PR #9, opened 2026-02-18
Container Policy Discussion Document

MGudgin:
This .md is the result of some thinking around container policy for Windows, Linix, macOS and a discussion with GHCP about the same.


alexsnitkovskiy:
Chromium also uses DisallowWin32kSystemCalls mitigation for some of their processes (I think storage and render). It uses Job with UILimits (on top of just using Job for resourcing management) to prevent certain UI operations when it cannot employ DisallowWin32kSystemCalls

MGudgin:
Superseded

PR #20, opened 2026-03-02
Forward the environment provided from the CLI.

shschaefer:
The CLI needs to be able to pass information through the environment.

In a follow on PR, we need to scrub this environment to ensure that it is not being used to pass sensitive information or otherwise hijack the container.


SteveSandersonMS (approved):
Makes sense - thanks!

PR #26, opened 2026-03-04
Grant AppContainer access to NUL device before process creation

vriveras:
Electron/Chromium opens \Device\Null internally during initialization, but the default DACL on the NUL device does not include an AppContainer ACE. This causes the sandboxed process to fail with exit code 1 in non-permissive mode.

**Add NullDeviceAccessGuard RAII class that:**
- Enables SeRestorePrivilege to obtain WRITE_DAC on the NUL device
- Adds a temporary ACE granting the AppContainer SID read/write access
- Restores the original DACL when the guard goes out of scope

### DIFF
--- a/wxc_common/AppContainerScriptRunner.cpp
+++ b/wxc_common/AppContainerScriptRunner.cpp
@@ -2,6 +2,7 @@
 
 #include <windows.h>
 
+#include <aclapi.h>
 #include <sddl.h>
 #include <userenv.h>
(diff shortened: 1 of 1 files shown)


shschaefer:
Thanks for this.  We think the correct solution to this is to resolve this problem from within the OS.  We have checked in code there.  Would prefer to not merge this and abandon the PR.

PR #28, opened 2026-03-13
Container Specific Policy - First Cut

MGudgin:
Naive container specific policy for Windows Sandbox and Hyperlight. Based off initial appContainer policies, with app container specifics removed, and general items removed if trivially enforceable by the container.


huzaifa-d:
Should be closed since the schema has evolved a lot since posting.

PR #29, opened 2026-03-13
Port wxc from C++ to Rust

MGudgin:
Rewrite the entire wxc (Windows Container Executor) codebase from C++ to
Rust, replacing the Visual Studio / vcxproj build system with a Cargo
workspace. Update documentation, build scripts, and SDK/CLI tooling to
reflect the new Rust-based build.

## New Rust crate structure (src/)

- **wxc**: CLI entry point (wxc-exec binary) using clap for argument
  parsing. Supports config via file path, --config flag, or
  --config-base64. Includes --delete mode for container profile cleanup.
- **wxc_common**: Shared library containing all core modules:
  - appcontainer: AppContainer profile creation and sandboxed script
    execution via Win32 Security Isolation APIs
  - config_parser: JSON configuration loading from file or base64
  - models: Data models (CodexRequest, ScriptResponse, NetworkPolicy,
    FilesystemPolicy, etc.) with serde serialization
  - error: Unified WxcError enum using thiserror
  - filesystem_bfs: BFS-based filesystem permission manager for
    read-write and read-only path policies
  - network_firewall: Windows Firewall rule management via COM/INetFwPolicy2
    for network allow/block enforcement
  - process_util: Process creation, pipe I/O, and output capture utilities
  - script_runner: Script execution orchestration
  - string_util: Wide-string (PCWSTR/HSTRING) conversion helpers
  - logger: Logging with debug and event-log modes
  - validator: Script request validation
- **wxc_test_driver**: Test harness binary (wxc-test-driver) for
  integration testing

## Deleted C++ sources

Remove all original C++ implementation files, headers, PCH files, and
Visual Studio project/filter files for wxc, wxc_common, wxc_test_driver,
and wxc_tests. Remove leftover C++ build artifacts: wxc.sln,
vcpkg.json, .clang-format, Directory.Build.props, build_all.bat,
build_debug.bat, and build_tests.bat.

## Updated CI pipeline for Rust

Replace the C++ build steps (MSBuild, vcpkg) in .github/workflows/build.yml
with Rust toolchain setup (rustup), cargo build, and cargo test steps.
Use the pre-installed rustup on GitHub-hosted runners.

## Updated docs and tooling for Rust migration

- Readme.md: Replace VS 2022/C++/vcpkg prerequisites with Rust toolchain
  and cargo build instructions
- build.bat: Replace msbuild wxc.sln with cargo build --release
- sdk/README.md: Update build instructions to reference cargo
- cli/ARCHITECTURE.md: Update C++ reference to Rust
- cli/src/cli.ts: Update default wxc-exec.exe path from x64/Debug to
  src/target/debug
- cli/example.ts: Same path update
- sdk/src/platform.ts: Update findWxcExecutable() to search Rust build
  output paths (src/target/{release,debug}) instead of VS layout
  (bin/{x64,ARM64})

## Other changes

- .gitignore: Add **/target/ for Rust build artifacts
- examples/08_pwsh.json: New PowerShell AppContainer example config
- src/expert-code-review-results.md: Code review findings from the port

Key dependencies: windows 0.58 crate (Win32 API bindings), serde/serde_json,
clap 4, thiserror, anyhow, base64.

Co-authored-by: shschaefer <stuart@theschaefers.com>


sdk/src/platform.ts
```diff
@@ -118,15 +118,11 @@ export function getPlatformSupport(): PlatformSupport {
...
-  // Detect current architecture
-  const arch = os.arch(); // 'x64' or 'arm64'
-  const platformDir = arch === 'arm64' ? 'ARM64' : 'x64';
-  
   const possiblePaths = [
-    // In SDK bin directory (deployed by build)
-    path.join(__dirname, '..', 'bin', platformDir, 'wxc-exec.exe'),
-    // Relative to project when used as a node module
-    path.join(__dirname, '..', 'node_modules', '@shschaefer', 'wxc-sdk', 'bin', platformDir, 'wxc-exec.exe'),
+    // Rust release build output
```
shschaefer, line 122: We need to adjust this.  It is not quite the same as what was there before.  Also, how are you handling platform builds with Rust?

src/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[workspace]
+members = ["wxc", "wxc_common", "wxc_test_driver"]
+resolver = "2"
```
bbonaby, line 3: latest resolver is v3, so we should update this in all our .tomls

src/wxc/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "wxc"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "wxc-exec"
+path = "src/main.rs"
+
+[dependencies]
+wxc_common = { path = "../wxc_common" }
```
bbonaby, line 11: We should make this  

```toml
wxc_common.workspace = true
```

Then add wxc_common to the workspace as a dependency. It's cleaner this way as the workspace will be manage the dependency.

src/wxc/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "wxc"
+version = "0.1.0"
+edition = "2021"
```
bbonaby, line 4: The edition can come from the workspace .toml file. Also not sure if we'll publish wxc itself to crates.io but if so we should consider this:

```toml
[package]
name = "wxc"
version = "0.1.0"
edition.workspace = true
license = "MIT"
keywords = ["windows", "containers"]
repository = "https://github.com/microsoft/mxc"
categories = ["os"]
description = "Windows sandboxed code execution runtime using AppContainer to safely run untrusted code."
```

Feel free to change the description. I'm not sure if this is what we want but it's my attempt at a description.

.github/workflows/build.yml
bbonaby, line 1: We should probably be building the Rust components for both x64 and ARM64 windows targets.

Also, most Rust repositories include two standard checks in their CI pipelines:

[rustfmt](https://doc.rust-lang.org/style-guide/):  enforces consistent code formatting

[clippy](https://doc.rust-lang.org/clippy/) : a linter that catches common mistakes and non-idiomatic patterns

These tools are widely used across the Rust ecosystem and are typically enforced in CI to ensure code quality and consistency. We might as well start now while we're early.

I also suggest splitting the SDK build from the wxc build. The SDK is currently Node/TypeScript, while wxc is a Rust binary crate, and the SDK could evolve to other languages in the future. Separating them will keep the pipeline easier to maintain and reason about.

I think this file should look something this, which should cover formatting, linting, and tests across the supported architectures for rust:

```yml
name: Build

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]

jobs:

  # Check formatting and linting for the WXC Rust code
  wxc-exec-lint:
    name: WXC Exec Lint
    runs-on: windows-latest

    defaults:
      run:
        working-directory: src

    steps:
      - uses: actions/checkout@v4

      - name: Setup Rust toolchain
        run: |
          rustup update stable
          rustup component add rustfmt clippy

      - name: Check formatting
        id: fmt
        continue-on-error: true
        run: cargo fmt --all -- --check

      - name: Run clippy
        run: cargo clippy --workspace --all-targets -- -D warnings

      - name: Fail if formatting check failed
        if: steps.fmt.outcome == 'failure'
        run: exit 1


  # Build and run Rust tests on x64 and ARM64
  wxc-exec-test:
    name: WXC Exec Rust tests
    strategy:
      matrix:
        include:
          - target: x86_64-pc-windows-msvc
            runner: windows-2025
          - target: aarch64-pc-windows-msvc
            runner: windows-11-arm

    runs-on: ${{ matrix.runner }}

    defaults:
      run:
        working-directory: src

    steps:
      - uses: actions/checkout@v4

      - name: Setup Rust toolchain
        run: |
          rustup update stable
          rustup target add ${{ matrix.target }}

      - name: Run tests
        run: cargo test --workspace


  # Build and package the TypeScript SDK
  wxc-typescript-sdk:
    name: WXC TypeScript SDK
    runs-on: windows-latest

    defaults:
      run:
        working-directory: sdk

    steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-node@v4
        with:
          node-version: 20

      - name: Install dependencies
        run: npm ci

      - name: Build
        run: npm run build

      - name: Pack npm package
        run: npm pack

      - uses: actions/upload-artifact@v4
        with:
          name: sdk-npm-package
          path: sdk/*.tgz
```

If you take my change above. You can run the same checks locally before pushing:

Format Rust code automatically: `cargo fmt --all`
Check formatting (same as CI): cargo fmt --all -- --check
Run the linter: `cargo clippy --workspace --all-targets -- -D warnings`
Run tests: `cargo test --workspace`
If needed, install them via cmdline: `rustup component add rustfmt clippy`

I found that installing the [GitHub CLI](https://cli.github.com/) enables GitHub Copilot to retrieve build pipeline failure information and quickly fix, commit, and push changes if you tell it to. Note: GitHub CLI is different from the GitHub Copilot CLI.

src/wxc/src/main.rs
```diff
@@ -0,0 +1,139 @@
+use std::fmt::Write;
```
bbonaby, line 1: FYI all rust files should have the following copyright header:

```rust
```
See: https://docs.opensource.microsoft.com/releasing/general-guidance/copyright-headers/

src/wxc_common/src/filesystem_bfs.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        Ok(combined)
+    }
+}
+
+/// Returns `false` for `C:\` (no inheritance), `true` for all other paths.
+fn test_for_root_path(path: &str) -> bool {
+    path != "C:\\"
+}
```
bbonaby, line 173: I assume we'd want the drive windows is installed on right?

```rust
use std::env;

// Returns the drive windows is installed on  + ":\" e.g "C:\" or "D:\"
fn windows_install_drive() -> Result<String, WxcError> {
    env::var("SystemRoot")
        .map_err(|e| WxcError::FilesystemPolicy(format!("SystemRoot not set: {e}")))?
        .get(..2)
        .map(|d| format!("{d}\\"))
        .ok_or_else(|| WxcError::FilesystemPolicy("SystemRoot has no drive prefix".into()))
}

fn test_for_root_path(path: &str) -> Result<bool, WxcError> {
    let root = windows_install_drive()?;
    Ok(!path.eq_ignore_ascii_case(&root))
}

```

Then you can use it like this in the functions above:

```rust
let inherit = test_for_root_path(path)?;
```
Then in the tests below since it's test code you can just do 
```rust
 assert!(test_for_root_path("some string").unwrap());
```

src/wxc_common/src/appcontainer.rs
```diff
@@ -0,0 +1,459 @@
...
+};
+use crate::script_runner::{get_timeout_milliseconds, ScriptRunner};
+use crate::string_util;
+
+// Attribute list constants (not always exported by the windows crate)
+const PROC_THREAD_ATTRIBUTE_SECURITY_CAPABILITIES: usize = 0x0002_0009;
+const PROC_THREAD_ATTRIBUTE_HANDLE_LIST: usize = 0x0002_0002;
+const PROC_THREAD_ATTRIBUTE_ALL_APPLICATION_PACKAGES_POLICY: usize = 0x0002_000F;
+const PROCESS_CREATION_ALL_APPLICATION_PACKAGES_OPT_OUT: u32 = 1;
+const EXTENDED_STARTUPINFO_PRESENT: PROCESS_CREATION_FLAGS = PROCESS_CREATION_FLAGS(0x0008_0000);
```
bbonaby, line 34: So
- PROC_THREAD_ATTRIBUTE_SECURITY_CAPABILITIES
- PROC_THREAD_ATTRIBUTE_HANDLE_LIST
- PROC_THREAD_ATTRIBUTE_ALL_APPLICATION_PACKAGES_POLICY
- EXTENDED_STARTUPINFO_PRESENT

Can all be imported by adding them to your import list from your `use windows::Win32::System::Threading::{}` on line 9.

For   `PROCESS_CREATION_ALL_APPLICATION_PACKAGES_OPT_OUT` we'll need to add `"Win32_System_WindowsProgramming",` to the windows crate feature list in the workspaces `Cargo.toml` file. And then in this file you can import it like:

```rust
use windows::Win32::System::WindowsProgramming::{
    PROCESS_CREATION_ALL_APPLICATION_PACKAGES_OPT_OUT,
};
```

src/wxc_common/src/string_util.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+/// Convert a UTF-16 slice to a UTF-8 String, stopping at the first null terminator if present.
+pub fn from_wide(wide: &[u16]) -> String {
+    let len = wide.iter().position(|&c| c == 0).unwrap_or(wide.len());
+    OsString::from_wide(&wide[..len])
+        .to_string_lossy()
+        .into_owned()
+}
+
```
bbonaby, line 20: Although Rust doesn't have a built in `wstring` a lot of crates use the [widestring](https://crates.io/crates/widestring) crate which makes it easier to work with wstrings or in this case for `CreateAppContainerProfile` or other APIS that take a `PCWSTR`.  

Alternatively, the `windows-rs` HSTRING struct can also be passed as a `PCWSTR`. It has a trait that does the conversion implicitly. Although you'd be creating a winrt string which has some a little overhead. My suggestion would be to just use the widestring crate which has functions to convert from a rust string to a wide string.

```rust
let title = "Hello from Rust";

// Convert to null-terminated UTF-16 for Win32 APIs
let title_wide = U16CString::from_str(title).expect("valid string");
let message_wide = U16CString::from_str("Rust String Literal").expect("valid string");

unsafe {
    MessageBoxW(
        None,
        PCWSTR(message_wide.as_ptr()),
        PCWSTR(title_wide.as_ptr()),
        MB_OK | MB_ICONINFORMATION,
    );
}


```

src/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[workspace.dependencies]
+windows = { version = "0.58", features = [
```
bbonaby, line 6: probably want to use the latest version which is `"0.62"` https://crates.io/crates/windows

src/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+    "Win32_System_IO",
+    "Win32_NetworkManagement_WindowsFirewall",
+    "Win32_Networking_WinSock",
+    "Win32_Security_Credentials",
+    "Win32_Security_Authorization",
+    "Win32_System_Memory",
+    "Win32_System_Diagnostics_Debug",
+    "Win32_Security_Isolation",
+] }
+windows-core = "0.58"
```
bbonaby, line 23: `windows-core = "0.62"`

src/wxc_test_driver/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "wxc_test_driver"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "wxc-test-driver"
+path = "src/main.rs"
```
bbonaby, line 8: This `[[bin]]` entry isn't necessary, Cargo already treats `src/main.rs` as the binary by default. 

As for the name, you can just use `wxc-test-driver` as the name of the crate itself on line 2.

src/wxc_common/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "wxc_common"
+version = "0.1.0"
+edition = "2021"
```
bbonaby, line 4:
```toml
[package]
name = "wxc_common"
version = "0.1.0"
edition.workspace = true
license = "MIT"
keywords = ["windows", "containers"]
repository = "https://github.com/microsoft/mxc"
categories = ["os"]
description = "Helper library for the wxc binary crate."
```

FYI if we'll be publishing the `wxc` binary crate to `crates.io`, we also have to publish the `wxc_common` crate as well since all crates on `crates.io` can only depend on published crates.

src/wxc_common/src/string_util.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Convert a SID pointer to its string representation (e.g. "S-1-5-...").
+/// Returns `default_value` on failure.
+///
+/// # Safety
+/// The caller must ensure `sid` points to a valid SID structure.
+pub unsafe fn sid_to_string(sid: *mut std::ffi::c_void, default_value: &str) -> String {
+    let mut string_sid = windows_core::PWSTR::null();
```
bbonaby, line 27: probably want 
```Rust
if sid.is_null() {
    return default_value.to_string();
}
```
above this line.

src/wxc/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+fn delete_app_container_profile(name: &str, logger: &mut Logger) -> bool {
+    // Clear BFS policy first
+    let mut bfs = FileSystemBfsManager::new(name.to_string());
+    bfs.remove_configuration(logger);
+
+    // Delete the AppContainer profile
+    let wide_name: Vec<u16> = name.encode_utf16().chain(std::iter::once(0)).collect();
```
bbonaby, line 65: see my comments in the `string_utils` module about using the [widestring](https://crates.io/crates/widestring) crate  for widestrings in rust.

src/wxc_common/src/process_util.rs
```diff
@@ -0,0 +1,203 @@
...
+            &mut group_sid_count,
+            &mut capability_sids,
+            &mut capability_sid_count,
+        )
+        .map_err(|_| {
+            WxcError::Process(format!(
+                "DeriveCapabilitySidsFromName({}) failed",
+                name
+            ))
+        })?;
```
bbonaby, line 174: FYI: The windows-rs crate's wrapper  already calls `GetLastError` internally on error. It converts it to an `HRESULT`, then returns it as an `windows::core::Error` in the Err variant of the Result. So, in your `map_err` you should do this:
```rust
.map_err(|e| WxcError::Process(format!(
    "DeriveCapabilitySidsFromName({name}) failed: {e}"
)))?;
```

src/wxc/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+    let _ = writeln!(logger, "Script timeout: {}", request.script_timeout);
+    let _ = writeln!(
+        logger,
+        "Container name: {}",
+        request.policy.app_container_name
+    );
+}
+
+fn display_script_results(response: &ScriptResponse, logger: &mut Logger) {
+    let _ = writeln!(logger, "Exit code: {}", response.exit_code);
```
bbonaby, line 53: I assume it's written like this because we don't care about logging failures? On the off chance we did care about them. We can return a `fmt::Result` and use the `?` operator which acts like our short circuit `RETURN_IF_FAILED` macros in c++, so something like:

```rust
fn log_request(request: &CodexRequest, logger: &mut Logger) -> fmt::Result {
   writeln!(logger, "Script code length: {}", request.script_code.len())?;
   // other writeln statements below
}
```

src/wxc_common/src/process_util.rs
```diff
@@ -0,0 +1,203 @@
...
+                WxcError::Process("Failed to set handle information on pipe".into())
+            })?;
+    }
+
+    Ok((OwnedHandle::new(h_read), OwnedHandle::new(h_write)))
+}
+
+/// Remove a specific Python "Failed to find real location of ..." line from stderr.
+pub fn suppress_python_location_error(stderr: &mut String) {
+    let needle = "Failed to find real location of ";
```
bbonaby, line 138: Hmm, I'm not sure if this error is localized based on the system or if it's always in english.

sdk/src/platform.ts
bbonaby: For differing `x64` and `arm64` windows builds in Rust we'd need to do something like this  (for debug no --release):
``` powershell
cargo build --release --target x86_64-pc-windows-msvc
cargo build --release --target aarch64-pc-windows-msvc
```
which produces:
```powershell
target/x86_64-pc-windows-msvc/release/wxc.exe
target/aarch64-pc-windows-msvc/release/wxc.exe
```
So the paths in this PR would need to change. Note:  VS does not support rust only VS Code. The equivalent of the cpp post build event [here in VS](https://github.com/microsoft/mxc/blob/d49afb88b03c2ec743d7c2c7a2f7b9d5442e9944/wxc/wxc.vcxproj#L103) would be to create a `tasks.json` file in the `.vscode/` folder. In there we would use powershell to run the cargo commands and copy the exes to where ever we wanted. 

Going a bit further though, as we think about distribution. I think we should leverage npm's optional dependency feature as it will automatically install the dependency package based on the hosts OS and CPU.

We can do something like this:

1. publish `wxc-exec-win-x64` to npm: (prebuilt wxc-exec.exe)
   1. Package.json :
   ```json
    {
      "name": "@microsoft/wxc-exec-win-x64",
      "os": ["win32"],
      "cpu": ["x64"],
      "files": ["wxc-exec.exe"]
     }
   ```
1. publish `wxc-exec-win-arm64` to npm: (prebuilt wxc-exec.exe)
   1. Package.json :
   ```json
    {
      "name": "@microsoft/wxc-exec-win-arm64",
      "os": ["win32"],
      "cpu": ["arm64"],
      "files": ["wxc-exec.exe"]
     }
   ```
1. publish  `mxc-sdk` to npm
   1. Package.json :
   ```json
    {
      "name": "@microsoft/wxc-sdk",
      "optionalDependencies": {
        "@microsoft/wxc-win-x64": "0.1.x",
        "@microsoft/wxc-win-arm64": "0.1.x"
       }
    }
   ```

Then the SDK will just need to do something like this to get the path to the executable:
```ts
import os from "os"

export function getWxcBinaryPath(): string {
  const platform = os.platform();
  const arch = os.arch();

  if (platform !== 'win32') {
    throw new Error(`Unsupported OS: ${platform}. WXC currently supports Windows only.`);
  }

  let packageName: string;
  if (arch === 'x64') {
    packageName = '@microsoft/wxc-exec-win-x64';
  } else if (arch === 'arm64') {
    packageName = '@microsoft/wxc-exec-win-arm64';
  } else {
    throw new Error(`Unsupported architecture: ${arch}`);
  }

  try {
    return require.resolve(`${packageName}/wxc-exec.exe`);
  } catch {
    throw new Error(
      `Failed to find wxc-exec binary. The platform package "${packageName}" may not be installed. `
    );
  }
}
```

src/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+thiserror = "2"
+anyhow = "1"
+base64 = "0.22"
+clap = { version = "4", features = ["derive"] }
```
bbonaby, line 29: we should add `wxc_common = { path = "wxc_common" }` here as well and probably sort this list

sdk/src/platform.ts
bbonaby, line 48: I think we'll need to update this before release

src/wxc/Cargo.toml
shschaefer: Please do not add AppContainer to descriptions.  This all needs to be scrubbed out...

.github/workflows/build.yml
shschaefer: Agree, we should add clippy and rustfmt (only if we add a custom format, some of the standard rules are rough).

src/wxc/src/main.rs
shschaefer: Can we do this as a commit/PR, rather than comment?

sdk/src/platform.ts
shschaefer: I have asked Anis to get me updated information.  Will take care of this in a separate PR.

bbonaby (approved):
Spoke to Stuart, we'll check this in and I'll do my comments as a PR. Thanks Martin

sdk/src/platform.ts
MGudgin: I've updated build.bat to build arm64/x64 release/debug. I'll tackle the package approach in a subsequent PR

src/wxc/src/main.rs
MGudgin: https://github.com/microsoft/mxc/pull/35

PR #44, opened 2026-03-23
Add WSL container configuration schema, routing

SohamDas2021:
Extends MXC's configuration schema and backend routing to support Linux containers via the WSL Container SDK (WSLC SDK). Also updates the design document with resolved design questions and naming corrections.

   - Add Wslc variant to ContainmentBackend enum
   - Add ContainerConfig struct with fields: target_os, image, cpu_count, memory_mb, gpu, storage_path, port_mappings
   - Add PortMapping struct for host↔container port forwarding
   - Extend JSON config parser to accept "containment": "wslc" and parse the new "container" section
   - Add placeholder match arm in main.rs (exits with error until Phase 3 implements WSLContainerRunner)
   - Doc changes

 No impact on existing behavior. AppContainer and Sandbox paths are unchanged. All existing tests pass.

### DIFF
--- a/src/wxc_common/src/config_parser.rs
+++ b/src/wxc_common/src/config_parser.rs
@@ -90,6 +90,10 @@ struct RawContainerConfig {
     storage_path: Option<String>,
     #[serde(rename = "portMappings")]
     port_mappings: Option<Vec<RawPortMapping>>,
+}
+
+#[derive(Deserialize, Default)]
+#[serde(default)]
 struct RawLxc {
     #[serde(rename = "containerName")]
     container_name: Option<String>,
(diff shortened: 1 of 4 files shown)


src/wxc_common/src/config_parser.rs
```diff
@@ -65,6 +65,33 @@ struct RawSandbox {
...
+    windows_port: Option<u16>,
+    #[serde(rename = "containerPort")]
+    container_port: Option<u16>,
+    protocol: Option<String>,
+}
+
+#[derive(Deserialize, Default)]
+#[serde(default)]
+struct RawContainerConfig {
+    #[serde(rename = "targetOs")]
```
shschaefer, line 81: Does WSLC call this targetOs or distro?

shschaefer (approved):
I am a bit concerned about the overlapping configuration and policy constructs.  Let's take some time this week to rationalize.

src/wxc_common/src/config_parser.rs
SohamDas2021: WSLC operates on image names, not OS/distro. targetOs here is an MXC-level routing hint so the SDK layer knows which backend/execution env to use.

SohamDas2021:
> I am a bit concerned about the overlapping configuration and policy constructs. Let's take some time this week to rationalize.



> I am a bit concerned about the overlapping configuration and policy constructs. Let's take some time this week to rationalize.

Yeah, the issue is for WSLC, we need both ContainerConfig for runtime settings and ContainerPolicy for security policy. But ContainerPolicy carries AppContainer-specific fields (app_container_name, capabilities, least_privilege_mode) that WSLC ignores. 

I will check this in and follow up after we discuss. Can we do something like this?

   SharedPolicy- paths, network -used by all backends
   AppContainerConfig- name, capabilities -AppContainer only
   SandboxConfig- daemon pipe -Sandbox only
   ContainerConfig- image, cpu, memory -WSLC only

shschaefer (approved):

PR #46, opened 2026-03-25
[AlexSn & AlexSn_AlterEgo 🐺] Add UI policy schema spec for Tessera co…

alexsnitkovskiy:
…ntainer configuration

   Define the "ui" section of WXC container config with 7 declarative
   fields (disable, dataChannels, isolation, desktopSystemControl,
   systemSettings, ime, injection) that map developer intent to
   JOB_OBJECT_UILIMIT_* flags and Win32k system call disable mitigation.
   Default-deny posture: "ui": {} = total lockdown.

### DIFF
--- a/docs/UIPolicy_Schema.md
+++ b/docs/UIPolicy_Schema.md
@@ -25,6 +25,18 @@ Developers declare *what the process is allowed to do* — Tessera translates th
 
 ---
 
+## Versioning and Compatibility
+
+Default-deny is a strong security posture, but it raises a compatibility question: what happens when a new field is introduced to the schema? Any config that does not explicitly set the new field will have it denied — which is the right behavior for security, but could silently break an application that was working fine under an older version of the policy.
+
+**Within a version**, default-deny applies fully — omitted fields are denied, and new fields introduced in a minor revision of the same version are denied for all existing configs.
+
+**Across major versions**, the behavior is different: applications pinned to an older version continue to run under the policy semantics of that version. New fields introduced in a later major version are not applied to configs that declare an older version — preserving the behavior the developer explicitly designed for. This allows the schema to evolve without silently changing the security posture of deployed applications.
+
+This makes version a compatibility contract: a config that says `"version": 1` will always behave as a v1 config, regardless of what the current schema version is.
+
+---
+
 ## Config Structure
 
 The `"ui"` section is a sibling of `"appContainer"`, `"filesystem"`, `"network"`, and `"sandbox"` in the WXC config. All fields shown explicitly for illustration — in practice, omitted fields default to their most restrictive value:
--- a/docs/UIPolicy_Schema.md
+++ b/docs/UIPolicy_Schema.md
@@ -1,0 +1,315 @@
+# Tessera UI Policy Schema — `"ui"` Section
+
+> **Status:** Draft — for review
+> **Location:** WXC container configuration JSON
+> **Reference:** [UIContainer.md](https://dev.azure.com/microsoft/OS/_git/os.2020?path=/onecoreuap/windows/core/ntuser/kernel/docs/UIContainer.md&version=GBofficial/ge_current_directwinpd_sf2)
+
+---
+
+## Overview
+
+The `"ui"` section of the WXC container configuration controls how a contained process interacts with the Windows GUI subsystem. It maps developer intent to the underlying enforcement mechanisms:
+
+- **Job Object UI Restrictions** (`NtUserSetJobUILimits`)
+- **Process Mitigation: Win32k System Call Disable** (`PROCESS_MITIGATION_SYSTEM_CALL_DISABLE_POLICY`)
+
+Developers declare *what the process is allowed to do* — Tessera translates that into the correct kernel flags and mitigations.
+
+### Design Principles
+
+1. **Default-deny** — All omitted fields default to the most restrictive value. `"ui": {}` = total lockdown.
+2. **Forward compatible** — New fields added to the schema are automatically denied by existing configs, requiring no updates. Default-deny ensures new capabilities are opt-in from day one.
+3. **Flat structure** — No nesting, no profiles, no precedence rules.
+4. **Intent over mechanism** — The schema describes permissions, not kernel APIs.
+5. **Opt-in** — Developers explicitly enable only what the process needs.
+
+---
+
+## Config Structure
+
+The `"ui"` section is a sibling of `"appContainer"`, `"filesystem"`, `"network"`, and `"sandbox"` in the WXC config. All fields shown explicitly for illustration — in practice, omitted fields default to their most restrictive value:
+
+```json
+{
+  "script": "myapp.exe",
+  "containment": "appcontainer",
+  "ui": {
+    "disable": false,
+    "dataChannels": "none",
+    "isolation": "container",
+    "desktopSystemControl": false,
+    "systemSettings": "none",
+    "ime": false,
+    "injection": false
+  }
+}
+```
+
+> **Note:** The `"ui"` section is only meaningful when `"containment"` is `"appcontainer"`. When `"containment"` is `"sandbox"`, the sandbox VM provides its own isolation — the `"ui"` section is ignored and a warning is emitted.
+
+---
+
+## Field Reference
+
+### `disable`
+
+| | |
+|---|---|
+| **Type** | `boolean` |
+| **Default** | `true` |
+| **Description** | Kill switch. Completely disables access to the GUI subsystem. The process cannot create windows, use GDI, or make any `NtUser*`/`NtGdi*` system calls. Atom table is also isolated. |
+| **Enforcement** | Process mitigation: `DisallowWin32kSystemCalls` (`PROC_THREAD_ATTRIBUTE_MITIGATION_POLICY`) + Job UI limit: `JOB_OBJECT_UILIMIT_GLOBALATOMS` (0x0020) |
+
+**When `true`:**
+- All other `"ui"` fields are ignored — the process has no GUI surface.
+
+**When `false`:**
+- The process retains GUI capability — other fields control what it can do.
+
+---
+
+### `dataChannels`
+
+| | |
+|---|---|
+| **Type** | `enum` |
+| **Default** | `"none"` |
+| **Values** | `"read"`, `"write"`, `"all"`, `"none"` |
+| **Description** | Controls access to data exchange channels between the contained process and the rest of the system. |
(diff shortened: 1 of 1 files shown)


docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+The `"ui"` section of the WXC container configuration controls how a contained process interacts with the Windows GUI subsystem. It maps developer intent to the underlying enforcement mechanisms:
+
+- **Job Object UI Restrictions** (`NtUserSetJobUILimits`)
+- **Process Mitigation: Win32k System Call Disable** (`PROCESS_MITIGATION_SYSTEM_CALL_DISABLE_POLICY`)
+
+Developers declare *what the process is allowed to do* — Tessera translates that into the correct kernel flags and mitigations.
+
+### Design Principles
+
+1. **Default-deny** — All omitted fields default to the most restrictive value. `"ui": {}` = total lockdown.
```
shschaefer, line 20: Though I agree with the sentiment, the challenge here is that we have to become clear about the impact of this policy.  if ui::disable is default true, will this not break a lot of things by default?

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+2. **Forward compatible** — New fields added to the schema are automatically denied by existing configs, requiring no updates. Default-deny ensures new capabilities are opt-in from day one.
```
shschaefer, line 21: This is the challenge.  When you add a new capability that is default deny, you will break existing policies.

shschaefer: Unless we handle versioning properly ;)

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+
+**When `true`:**
+- All other `"ui"` fields are ignored — the process has no GUI surface.
+
+**When `false`:**
+- The process retains GUI capability — other fields control what it can do.
+
+---
+
+### `dataChannels`
```
jeffbromberger: Do we imagine other things in this category besides clipboards? The dataChannel name was not obvious to me. If we think there are going to be other types of channels then should we have a "clipboard" element under dataChannels and have read/write/etc applied to clipboard directly? Or, if we think clipboard is the only likely channel should we name it more directly? If we imagine there being some other type of dataChannel possible then is it likely a user would want only a single value for all dataChannels? I.e. would they not want per-channel granularity on the access?

docs/UIPolicy_Schema.md
alexsnitkovskiy: Yes it may break many GUI apps if this is not explicitly overridden. However my strong believe when it comes to GUI subsystem, you have to by default do deny. This should be a mind set and policy should reflect it

docs/UIPolicy_Schema.md
alexsnitkovskiy: I want default to be deny. And I am optimizing into it, since not only we do not have all restriction we talked about in the code yet, we most def. will discover new once as we go.

So yes, if there is a versioning  - great. There should be versioning no matter what, and this would not be specific to the UI portion right. As such I do not think needs to be covered here?

docs/UIPolicy_Schema.md
alexsnitkovskiy: I am happy to rename this to clipboard.

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+| **Values** | `"desktop"`, `"handles"`, `"atoms"`, `"container"` |
+| **Description** | Determines whether the process can see and interact with UI objects (windows, menus, atoms) belonging to other processes. |
+| **Enforcement** | `JOB_OBJECT_UILIMIT_HANDLES` (0x0001), `JOB_OBJECT_UILIMIT_GLOBALATOMS` (0x0020) |
+
+| Value | Flags Set | Handle Access | Atom Table | Use Case |
+|-------|-----------|--------------|-----------|----------|
+| `"desktop"` | *(none)* | All handles on the desktop | Global (session-wide) | Process needs to interact with other windows |
+| `"handles"` | `UILIMIT_HANDLES` | Same-job handles only | Global (session-wide) | Handle isolation without atom isolation |
+| `"atoms"` | `UILIMIT_GLOBALATOMS` | All handles on the desktop | Per-job (isolated) | Atom isolation without handle restriction |
+| `"container"` | `UILIMIT_HANDLES` + `UILIMIT_GLOBALATOMS` | Same-job handles only | Per-job (isolated) | Sandboxed — process only sees its own UI objects |
```
jeffbromberger: I'm not clear on this one. It says same job handles, but then says _process_ only sees its own objects. If there are 2 processes in the same contained job, can they not see each other's handles?

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+
+---
+
+### `injection`
+
+| | |
+|---|---|
+| **Type** | `boolean` |
+| **Default** | `false` |
+| **Description** | Controls whether the process can inject synthetic input into other processes. |
```
jeffbromberger: Similarly here I'm not clear if this blocks between two processes in the same job, or only between processes in separate jobs.

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+
+The process has zero UI surface. Win32k attack surface eliminated.
+
+```json
+"ui": {}
+```
+
+Or equivalently:
+```json
+"ui": {
```
jeffbromberger, line 250: With a minimal config as this are we still taking steps to block, for example, system shutdown? A system service (at least AFAIK) can shutdown or restart the system without any UI or without running in an interactive session (provided it has SE_SHUTDOWN_NAME). Above we tied restart to desktopSystemControl but can we enforce that through this ui-centric configuration mechanism?

docs/UIPolicy_Schema.md
alexsnitkovskiy: Yes 2 processes in the same Job can see each othe handles.

The line 97 needs to be corrected "belonging to processes outside of the Job"

docs/UIPolicy_Schema.md
alexsnitkovskiy: It is blocked completely.

alexsnitkovskiy: The comment here is correct

docs/UIPolicy_Schema.md
alexsnitkovskiy: This configuration is specific for the GUI Subsystem, hence UI.  If there are API to exit windows / restart that do not flow via Win32 GUI Subsystem, they are not covered here, and should be considered somewhere else.

docs/UIPolicy_Schema.md
jsidewhite: Yeah breaking existing policies doesn't feel like it beholds the "forward compatible" promise.

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+}
+```
+
+> **Note:** The `"ui"` section is only meaningful when `"containment"` is `"appcontainer"`. When `"containment"` is `"sandbox"`, the sandbox VM provides its own isolation — the `"ui"` section is ignored and a warning is emitted.
+
+---
+
+## Field Reference
+
+### `disable`
```
jsidewhite, line 66: Shouldn't it be *enable*? i.e. `enable=true`**

disable is already the default

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+| **Default** | `"container"` |
+| **Values** | `"desktop"`, `"handles"`, `"atoms"`, `"container"` |
+| **Description** | • Handle isolation — restricts the process from seeing or interacting with USER handles (e.g. windows, menus) owned by processes outside the job. <br>• Atom table isolation — gives the job its own private atom table, preventing processes in the job from accessing the global (Window Station-based) atom table. |
+| **Enforcement** | `JOB_OBJECT_UILIMIT_HANDLES` (0x0001), `JOB_OBJECT_UILIMIT_GLOBALATOMS` (0x0020) |
+
+| Value | Flags Set | Handle Access | Atom Table | Use Case |
+|-------|-----------|--------------|-----------|----------|
+| `"desktop"` | *(none)* | All handles on the desktop | Global (Window Station) | Process needs to interact with other windows |
+| `"handles"` | `UILIMIT_HANDLES` | Same-job handles only | Global (Window Station) | Handle isolation without atom isolation |
+| `"atoms"` | `UILIMIT_GLOBALATOMS` | All handles on the desktop | Per-job (private) | Atom isolation without handle restriction |
```
jsidewhite, line 116: This is confusing.  Saying "atoms" gives you `handle` access.  Wouldn't that be like in `clipboard` saying `write` gives you `read` access?

jsidewhite: isolation is like a rejectlist, whereas the others are a allowlist.

docs/UIPolicy_Schema.md
```diff
@@ -0,0 +1,315 @@
...
+
+---
+
+### `systemSettings`
+
+| | |
+|---|---|
+| **Type** | `enum` |
+| **Default** | `"none"` |
+| **Values** | `"all"`, `"parameters"`, `"display"`, `"none"` |
```
jsidewhite, line 153: Allowing "all" means that you'll silently get new capabilities in future if more are added?

docs/UIPolicy_Schema.md
alexsnitkovskiy: Well as we talked we would not break them.
We want deny-by-default...and we would need to have proper versioning story

So for version X, it is deny-by-default

docs/UIPolicy_Schema.md
alexsnitkovskiy: Discussed during meeting. Let me know if you want to change

docs/UIPolicy_Schema.md
alexsnitkovskiy: We discussed this, it is a bit UI specific... and I gave examples with apps, where permutations used.

docs/UIPolicy_Schema.md
alexsnitkovskiy: Yes, but then in this case, this one is final.
This is truly legacy part

docs/UIPolicy_Schema.md
alexsnitkovskiy: I added versioning section

shschaefer (approved):

PR #49, opened 2026-03-26
[Config json format consolidation] #2 Add JSON schema, cross-field validation, and migrate examples to new config format

SohamDas2021:
Step 2 of multi-step consolidation process.

Adds schemas/mxc-config.v1.schema.json for editor validation and tooling. Adds platform/containment compatibility checks and schema version validation in config_parser.rs. Rewrites docs/schema.md to document the new format. Migrates all 15 example configs to use process.commandLine, containerId, lifecycle, and platform fields. All existing configs still accepted via dual-read fallback.

----------------------------------------------
**PR 1** ✅ Merged
   - `version`, `containerId`, `process` section with dual-read fallback. `Vm`/`MicroVm` enum variants.
   Manual `Default` for `CodexRequest`.

   **PR 2** 🔄 In review -> This PR
   - `schemas/mxc-config.v1.schema.json`. Schema version validation. `idleTimeoutMs` dual-read. `containerId` →
  `app_container_name` mapping. Migrate 15 examples + 16 test configs. Update `docs/schema.md`.

   **PR 3**
   - Add `LifecycleConfig` struct. Parse `lifecycle.destroyOnExit`/`preservePolicy` with dual-read from legacy
  fields. `wslc` alias for `container` key. Update `sdk/src/types.ts`, `cli/README.md`, `cli/ARCHITECTURE.md`,
  `sdk/README.md`.

   **PR 4**
   - Drop `script`, `workingDirectory`, top-level `timeout`, `container` alias, `appContainer.name`,
  `lxc.containerName`, `lxc.destroyOnExit`, `clearPolicyOnExit`, `removeRulesOnExit` from parser. Bump
  `SUPPORTED_MAJOR_VERSION` to 2.
----------------------------------------------


docs/schema.md
```diff
@@ -1,56 +1,106 @@
...
-    "timeout": 30000,                    // Optional: timeout in milliseconds (default is no timeout)
-    "appContainer": {                    // Optional: AppContainer-specific settings
-      "name": "CLI",                       // AppContainer profile name
-      "leastPrivilegeMode": false,         // Enable LPAC mode
-      "learningMode": false,               // Enable learning mode tracing, if available
-      "capabilities": [                    // Windows capabilities to grant
-        "internetClient",
-        "privateNetworkClientServer"
-      ]
+    "version": "1",                        // Schema version (current: "1")
```
shschaefer: Versions must use SemVer.  This should be 0.1.0-alpha.

schemas/mxc-config.v1.schema.json
```diff
@@ -0,0 +1,259 @@
...
+      "type": "string",
+      "enum": ["appcontainer", "sandbox", "vm", "wslc", "lxc", "microvm"],
+      "default": "appcontainer",
+      "description": "Containment backend to use for execution."
+    },
+    "platform": {
+      "type": "string",
+      "enum": ["linux", "windows"],
+      "default": "windows",
+      "description": "Host platform where wxc-exec runs. Must be compatible with the selected containment backend (e.g., AppContainer/Sandbox/WSLC require 'windows', LXC requires 'linux')."
```
bbonaby: thinking-out-loud: do we want to say `runtime executable` versus `wxc-exec`? But let's not focus too much on that for now.

shschaefer (approved):

shschaefer (approved):

schemas/mxc-config.v1.schema.json
```diff
@@ -0,0 +1,181 @@
+{
```
bbonaby, line 1: I like the direction here with `appContainer` and `lxc` etc already being scoped into their own objects. I wonder if we should lean into that more consistently and make backend-specific configuration fully structured around the selected containment.

Right now, we have backend-specific containers:
```json
"appContainer": { ... },
"lxc": { ... }
```
But there isn't a strong relationship between them. e.g
```json
"containment": "appcontainer" | "lxc" | 
```
configs can technically include multiple backend sections at once and nothing enforces that only the relevant backend config is present. E.g Folks looking at the schema might think they can mix and match containment backends.

My suggestion would be to see if we can formalize this pattern, so each backend owns its own schema and is conditionally validated.
Some like:

```json
{
  "type": "object",
  "required": [
    "containment",
    "backends"
  ],
  "properties": {
    "containment": {
      "type": "string",
      "enum": [
        "appcontainer",
        "lxc"
      ]
    },
    "backends": {
      "type": "object",
      "properties": {
        "appcontainer": {
          "$ref": "#/$defs/appContainerConfig"
        },
        "lxc": {
          "$ref": "#/$defs/lxcConfig"
        }
      },
      "additionalProperties": false
    }
  },
  "allOf": [
    {
      "if": {
        "properties": {
          "containment": {
            "const": "appcontainer"
          }
        }
      },
      "then": {
        "properties": {
          "backends": {
            "required": [
              "appcontainer"
            ],
            "maxProperties": 1
          }
        }
      }
    },
    {
      "if": {
        "properties": {
          "containment": {
            "const": "lxc"
          }
        }
      },
      "then": {
        "properties": {
          "backends": {
            "required": [
              "lxc"
            ],
            "maxProperties": 1
          }
        }
      }
    }
  ]
}
```

So then this would be **valid**:
```json
{
  "containment": "lxc",
  "backends": {
    "lxc": {
      "distribution": "alpine"
    }
  }
}
```
and these **will not be valid**
```json
{
  "containment": "lxc",
  "backends": {
    "appcontainer": {}
  }
}
```

```json
{
  "containment": "lxc",
  "backends": {
    "lxc": {},
    "appcontainer": {}
  }
}
```

```json
{
  "containment": "appcontainer",
  "backends": {}
}
```

Just a suggestion though, I thought it would be cleaner this way. This also allows what goes into `Networking`, `Filesystem` to be scoped inside the backends schema rather than at the top level. E.g today currently all of those are top level objects and not all their properties will be supported for every backend. Having each backend have their own schema imo gives us clear separation in the future. But this is all just a suggestion, I'll leave it up to you.

bbonaby (approved):

PR #59, opened 2026-03-26
Fix CLI build typescript build

bbonaby:
PR fixes the CLI build. Basically, it just needed to use the new way of passing in the script stuff that Soham added. I'll add a GitHub issue on myself so the test scripts can also be ran on in the pipelines too so we catch this eariler.  

Another issue it fixes is that our parser will only accept version 1 in the json. So as a quick fix, since I know that will be updated in the future, I've updated all the example and test script json files to use version 1 so they can be ran locally without the version error.

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -69,13 +69,14 @@ program
       const policy: WxcConfiguration = JSON.parse(content);
 
       // Basic validation
-      if (!policy.script) {
-        console.error('Invalid configuration: missing script.code');
+      const commandLine = policy.process?.commandLine;
+      if (!commandLine) {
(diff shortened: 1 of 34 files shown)


shschaefer (changes requested):
Please fix the parser instead of the scripts :(

bbonaby:
Closing in favor of Soham's. Confirmed with Soham that we'll update his to use https://crates.io/crates/semver for the [validate_schema_version](https://github.com/microsoft/mxc/blob/d50850278682c30f73b8a9c0ebf3daf608b6c5c5/src/wxc_common/src/config_parser.rs#L260) function.

PR #62, opened 2026-03-26
[Config json format consolidation] #3.5 Fix CLI TypeScript build error after SDK type changes

SohamDas2021:
WxcConfiguration.script became optional in PR3, breaking the CLI build. Resolves command line from process.commandLine (new) or script (legacy), working directory from process.cwd or workingDirectory, and container name from containerId with fallback to backend-specific fields.


shschaefer (approved):
This is intermediate.  Expecting consolidation to single field.

PR #78, opened 2026-03-27
[LXC] Fix LXC container execution, Alpine version, and veth discovery

SohamDas2021:
- Use lxc-attach instead of lxc-execute for script execution. The runner starts the container with lxc-start (for veth discovery and iptables setup), then lxc-execute fails because the container is already running. lxc-attach correctly attaches to the running container.

- Update default Alpine release from 3.19 to 3.20. Alpine 3.19 has been removed from the LXC image server, causing container creation to fail. Updated in models.rs, all LXC JSON configs, and docs.

- Fix veth interface discovery. lxc-info -i only returns IP addresses, not interface names. Changed to parse the Link: line from lxc-info (without -i) which includes the host-side veth name.

Fixes #71 #72 #73

### DIFF
--- a/src/lxc_common/src/lxc_runner.rs
+++ b/src/lxc_common/src/lxc_runner.rs
@@ -45,6 +45,14 @@ impl LxcScriptRunner {
 
     /// Core execution logic.
     fn run_internal(&self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
+        // Validate required LXC fields
+        if self.config.distribution.is_empty() || self.config.release.is_empty() {
+            return ScriptResponse::error(
+                "LXC distribution and release are required \
+                 (e.g., \"distribution\": \"alpine\", \"release\": \"3.20\")",
+            );
+        }
+
         let container_name = self.resolve_container_name();
         let _ = writeln!(logger, "Container name: {}", container_name);
         let _ = writeln!(
--- a/src/lxc_common/src/lxc_runner.rs
+++ b/src/lxc_common/src/lxc_runner.rs
@@ -117,11 +117,9 @@ impl LxcScriptRunner {
         }
 
         // Execute the script using lxc-attach (container is already running)
+        // TODO: Thread request.script_timeout through to attach_run for timeout enforcement.
         let _ = writeln!(logger, "Executing script inside container...");
-        let result = container.attach_run(
-            &request.script_code,
-            &request.working_directory,
-        );
+        let result = container.attach_run(&request.script_code, &request.working_directory);
 
         let response = match result {
             Ok((exit_code, stdout, stderr)) => ScriptResponse {
--- a/src/lxc_common/src/lxc_runner.rs
+++ b/src/lxc_common/src/lxc_runner.rs
@@ -116,12 +116,11 @@ impl LxcScriptRunner {
             }
         }
 
-        // Execute the script
+        // Execute the script using lxc-attach (container is already running)
         let _ = writeln!(logger, "Executing script inside container...");
-        let result = container.exec(
+        let result = container.attach_run(
             &request.script_code,
             &request.working_directory,
-            request.script_timeout,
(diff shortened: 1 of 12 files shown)


src/wxc_common/src/config_parser.rs
```diff
@@ -359,7 +359,7 @@ fn convert_raw_config(raw: RawConfig, logger: &mut Logger) -> Result<CodexReques
         let raw_lxc = raw.lxc.unwrap_or_default();
         LxcConfig {
             distribution: raw_lxc.distribution.unwrap_or_else(|| "alpine".to_string()),
-            release: raw_lxc.release.unwrap_or_else(|| "3.19".to_string()),
+            release: raw_lxc.release.unwrap_or_else(|| "3.20".to_string()),
```
bbonaby: To be honest I think we shouldn't have a default distro. We should just error and tell the user to provide one. If we keep a default then we'd be on the hook to keep updating it this like we do today. 

We can keep the .20 changes in the .md and .json files but remove these defaults and add an error message that they need to provide a distro and release.  You'd probably want to remove any mention from the .md of us using a default too if it's in there as well.

src/wxc_common/src/models.rs
```diff
@@ -49,15 +49,15 @@ impl Default for SandboxConfig {
...
+    /// Distribution release version (e.g., "3.20", "24.04").
     pub release: String,
 }
 
 impl Default for LxcConfig {
     fn default() -> Self {
         Self {
             distribution: "alpine".to_string(),
-            release: "3.19".to_string(),
+            release: "3.20".to_string(),
```
bbonaby: see comment above about defaults

src/lxc_common/src/lxc_runner.rs
```diff
@@ -116,13 +116,10 @@ impl LxcScriptRunner {
             }
         }
 
-        // Execute the script
+        // Execute the script using lxc-attach (container is already running)
+        // TODO: Thread request.script_timeout through to attach_run for timeout enforcement.
```
bbonaby, line 128: should we make the TODO an issue?

src/lxc_common/src/network_iptables.rs
```diff
@@ -44,35 +44,31 @@ impl NetworkIptablesManager {
...
-                        return Some(name);
-                    }
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        // Parse the "Link:" line from lxc-info output
+        for line in stdout.lines() {
+            let trimmed = line.trim();
+            if let Some(link_name) = trimmed.strip_prefix("Link:") {
+                let veth = link_name.trim();
+                if veth.starts_with("veth") {
```
bbonaby, line 70: Good call on removing the need to iterate through `/sys/class/net` 👍 :
```bash
bbonaby@bbonaby-dev-2:/mnt/z/mxc/test_scripts$ lxc-info --name mycontainer
Name:           mycontainer
State:          RUNNING
PID:            46102
IP:             10.0.3.63
Link:           veth1000_8oQy
 TX bytes:      1.40 KiB
 RX bytes:      1.54 KiB
 Total bytes:   2.94 KiB
bbonaby@bbonaby-dev-2:/mnt/z/mxc/test_scripts$ cd /sys/class/net
bbonaby@bbonaby-dev-2:/sys/class/net$ ls
eth0  lo  lxcbr0  veth1000_8oQy  vethMXlpew
```

src/wxc_common/src/config_parser.rs
SohamDas2021: yes, I agree, will remove.

src/lxc_common/src/lxc_runner.rs
SohamDas2021: https://github.com/microsoft/mxc/issues/84

bbonaby (approved):

PR #80, opened 2026-03-27
Align schema versioning doc with semver spec

SohamDas2021:
- Explain pre-release (-alpha) means schema is not yet stable
- Fix: new optional fields are Minor bumps, not Patch (per semver)
- Fix: migration fallback PRs are Minor bumps, not Patch
- Note that pre-release labels are ignored for version comparison
- Reference semver §4 for 0.x initial development semantics

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -71,7 +71,13 @@ MXC config files include an optional `version` field using
 this to detect incompatible configs and provide clear upgrade guidance. If
 `version` is absent, the config is assumed compatible with the current version.
 
-The parser compares the config's major.minor against its supported version:
+Versions with a pre-release suffix (e.g., `-alpha`) indicate the schema is not
+yet stable — breaking changes may occur in any release. Once the schema is
+stable, version `1.0.0` (no suffix) will be released. After `1.0.0`, breaking
+changes require a major version bump per semver.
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -85,17 +91,18 @@ The parser compares the config's major.minor against its supported version:
...
 | Remove a field / breaking change | **Major** (0.x → 1.0.0) | Dropping legacy fields |
 
-**Rule of thumb:** Follow [semver](https://semver.org/). While in `0.x`,
-minor bumps may include breaking changes. Once `1.0.0` is reached, breaking
+**Rule of thumb:** Follow [semver](https://semver.org/). While in `0.x` (initial
+development), any release may include breaking changes per
+[semver §4](https://semver.org/#spec-item-4). Once `1.0.0` is reached, breaking
 changes require a major bump.
 
 #### Migration process for breaking changes
```
bbonaby, line 103: It might be worth adding an explicit deprecation step here so users know what our migration expectations are. For example, we could produce a deprecation warning about the removal of functionality in a minor release, before removing it in the next major release. That would give downstream consumers time to adapt I think.

bbonaby (approved):

PR #82, opened 2026-03-27
Windows Sandbox: Fix stdout relay, multi-exec, boot reliability, Python discovery, E2E tests, and docs

huzaifa-d:
## Summary

Brings the Windows Sandbox backend from a partially-working prototype to a fully functional, tested, and documented state. Fixes 4 critical bugs, adds multi-exec VM reuse, and introduces the first E2E test suite for the sandbox backend.

### Bug Fixes

**1. stdout/stderr silently discarded** (`d0e9a2b`)
- Root cause: output was dropped at 3 layers - `tcp_bridge.rs` assigned to `_stdout`/`_stderr` (unused), `pipe_server.rs` RESULT format had no output fields, `sandbox_runner.rs` hardcoded empty strings
- Fix: extended RESULT protocol to `RESULT <code> <stdout-b64> <stderr-b64> <error>`, base64-encoded to handle binary output

**2. cmd.exe quoting breaks scripts with double quotes** (`5d1aa87`)
- Root cause: Rust's `Command::arg()` backslash-escapes quotes but `cmd.exe` doesn't understand that syntax
- Fix: use `cmd.raw_arg(script_code)` to pass script text literally
- Also fixed stale `sandbox_running` state preventing retry after rendezvous timeout

**3. ~60% sandbox boot failure rate** (`7a437ff`)
- Root causes: zombie Hyper-V processes surviving taskkill, insufficient 3s teardown cooldown, vGPU failures under nested Hyper-V
- Fix: poll up to 30s for process exit, 5s CmService cooldown, disable vGPU in .wsb config, 3 retry attempts with exponential backoff (0s, 10s, 20s)

**4. Single execution per sandbox lifetime** (`fe3e5e0`)
- Root cause: agent `take()`s stdin/stdout/stderr TCP streams on first EXEC - `None` for subsequent runs
- Fix: added `StreamsReady` protocol message for coordinated data stream reconnection. Agent signals readiness before accept (TCP backlog queues daemon's connects), daemon waits for signal then connects 3 new streams. Residual control buffer passed to handle coalesced TCP reads.
- Also fixed Python discovery finding Windows Store stub (`WindowsApps\python.exe`) instead of real Python
- Also added `PYTHONDONTWRITEBYTECODE=1` and `PYTHONNOUSERSITE=1` to bootstrap.cmd for read-only mount compatibility

### Documentation (`d86d8dd`, `bd5a9ba`)
- Added comprehensive `docs/sandbox-integration-plan.md` covering architecture, protocol, security model, issues found with fixes, debugging guide, and E2E test documentation
- Added `Sandbox-Handoff.md` with detailed session notes for future contributors

### E2E Test Suite (`6737f69`)
- 6 new test configs: echo, PowerShell, stderr relay, exit code 42, timeout, PowerShell env inspection
- `run_sandbox_tests.ps1` runner: starts daemon, runs 10 tests (6 basic + 1 timeout + 3 multi-exec on same VM), decodes base64 output, validates results, reports pass/fail
- Manual-only (requires Hyper-V + Windows Sandbox feature - cannot run in GitHub CI)
- All 10 tests pass in ~1:40

## Files Changed (20 files, +1553/-116)

| Area | Files | What |
|------|-------|------|
| Protocol | `sandbox_protocol.rs` | `StreamsReady` variant + roundtrip test |
| Agent | `executor.rs`, `listener.rs`, `main.rs` | Multi-exec loop, `accept_data_connections()`, listener passthrough |
| Daemon | `pipe_server.rs`, `tcp_bridge.rs`, `main.rs` | RESULT format, reconnect, retry logic, `guest_addr` state |
| VM mgmt | `sandbox_vm.rs` | Python discovery fix, bootstrap env vars, vGPU disable, teardown polling |
| Client | `sandbox_runner.rs` | Base64 RESULT parsing with legacy fallback |
| Configs | `basic_sandbox.json`, `sandbox_custom_timeout.json` | `-S -B` Python flags |
| Tests | 6 new configs + `run_sandbox_tests.ps1` | E2E test suite |
| Docs | `sandbox-integration-plan.md`, `Sandbox-Handoff.md` | Full architecture docs |

## Testing

- **156 unit tests pass** (including new `roundtrip_streams_ready`)
- **10/10 E2E tests pass** - echo, Python, PowerShell (x2), stderr, exit code, timeout, multi-exec (x3)
- E2E verified on Windows 11 Insider build 26570 (known sandbox regression - boot may be flaky)

## Known Issues (Future Work)

1. **Lock held across reconnect await** - `pipe_server.rs` holds `DaemonState` mutex during `reconnect_data_streams()` (up to 60s). Should take connection out of state before awaiting.
2. **Stale `sandbox_running` on reconnect failure** - `guest_connection` cleared but `sandbox_running` stays true, causing next request to wait 120s for a rendezvous file that won't appear.
3. **Agent hangs if daemon dies during reconnect** - `accept_data_connections()` has no timeout.
4. **No size limit on base64 decode** - could cause memory exhaustion with very large output.
5. **Buffered output** - stdout/stderr are captured and returned after completion, not streamed live (unlike AppContainer's console inheritance).


docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+| `wxc-exec.exe` | `wxc` | Host | CLI entry point, dispatches to SandboxScriptRunner |
+| `wxc-sandbox-daemon.exe` | `wxc_sandbox_daemon` | Host | Manages sandbox VM lifecycle, bridges IPC to TCP |
+| `wxc-sandbox-agent.exe` | `wxc_sandbox_agent` | Inside sandbox VM | Accepts commands, runs scripts, bridges stdio |
+
+### IPC Between wxc-exec and Daemon
+
+The daemon listens on a **localhost TCP port** derived deterministically from a pipe name (default `wxc-sandbox`):
+
+```rust
+fn pipe_name_to_port(name: &str) -> u16 {
```
bbonaby: issue: hmm, I'm not sure this will work as good as it sounds. Windows is also providing other processes port numbers while this is going on, so it's possible that this function chooses a port that is already taken. I think it might be better to ask windows for a port number by attempting to bind to port `0`. Then you can retrieve the port windows provides back to you.

I see that this is code that is already in the codebase and outside this PR, but if you can update this that would be appreciated. Feel free to do it in a follow up.

Sandbox-Handoff.md
```diff
@@ -0,0 +1,470 @@
+# MXC Windows Sandbox E2E Test — Agent Handoff Prompt
```
bbonaby: I think you can delete this file

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+| `src/wxc_common/src/config_parser.rs` | Parses JSON → `CodexRequest` with `SandboxConfig` |
+| `src/wxc_sandbox_daemon/src/main.rs` | Daemon entry point, IPC server, idle watchdog |
+| `src/wxc_sandbox_daemon/src/pipe_server.rs` | TCP IPC server, EXEC handling, retry logic |
+| `src/wxc_sandbox_daemon/src/sandbox_vm.rs` | .wsb generation, Python discovery, VM launch/teardown |
+| `src/wxc_sandbox_daemon/src/rendezvous.rs` | Polls rendezvous.txt for agent address |
+| `src/wxc_sandbox_daemon/src/tcp_bridge.rs` | 4-channel TCP bridge, execute_on_guest, reconnect |
+| `src/wxc_sandbox_agent/src/main.rs` | Agent entry point |
+| `src/wxc_sandbox_agent/src/listener.rs` | TCP listener, rendezvous writer, guest IP discovery |
+| `src/wxc_sandbox_agent/src/executor.rs` | Command loop, child process spawning, stdio bridging |
+| `src/wxc_sandbox_agent/src/firewall.rs` | Guest firewall lockdown via netsh |
```
bbonaby: question: Is the firewall part of this is working?

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Known Limitations
+
+1. **IPC uses TCP, not named pipes** — Port conflicts possible if another process occupies the derived port
+2. **Single language mapped** — Only Python is mapped from host; Node.js would need similar treatment
```
bbonaby: question: wonder if this will be an issue, as there isn't a way for the user of mxc to know only python is supported for sandboxes.

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+3. **Windows Insider regression** — Builds 26100+ have confirmed sandbox boot failures (zombie VM processes)
+4. **Cold boot time** — First sandbox boot takes 15-60s; subsequent boots ~15-30s
+5. **No filesystem/network policy forwarding** — The `filesystem` and `network` config sections are ignored for sandbox containment; isolation relies on the VM boundary and agent firewall
```
bbonaby: question: So does this mean the only thing that we care about for sandbox is passing the script in and then getting the result?

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
+# Windows Sandbox Backend
```
bbonaby: suggestion: We probably don't need the info in this doc in the repo but a high level design of what's happening would be good..

src/wxc_common/src/sandbox_runner.rs
```diff
@@ -132,25 +133,50 @@ impl SandboxScriptRunner {
...
+                    standard_out: stdout_text,
+                    standard_err: stderr_text.clone(),
+                    error_message: if error_msg.is_empty() {
+                        stderr_text
+                    } else {
+                        error_msg
+                    },
                 }
             } else {
+                // Fallback: legacy format "RESULT <exit-code> <error-message>"
```
bbonaby: question: will this work with your new changes? I figured this code path wouldn't work anymore and we'd just send an error.

src/wxc_sandbox_daemon/src/pipe_server.rs
```diff
@@ -127,34 +169,85 @@ async fn ensure_sandbox_ready(state: &Arc<Mutex<DaemonState>>) -> Result<()> {
...
-    rendezvous::cleanup(&rendezvous_dir).await?;
+    let python_dir = sandbox_vm::find_host_python()
+        .context("Python is required on the host for sandbox execution")?;
+    eprintln!("[daemon] host Python found at {:?}", python_dir);
+
+    let temp_dir = std::env::temp_dir().join("wxc-sandbox-config");
+    std::fs::create_dir_all(&temp_dir).context("create .wsb config dir")?;
+
+    let max_attempts = 3;
+    let backoff_secs = [0, 10, 20]; // backoff between retries
```
bbonaby: question: are these back off times enough? Not really sure myself tbh

src/wxc_sandbox_daemon/src/sandbox_vm.rs
```diff
@@ -66,6 +67,30 @@ pub fn find_host_python() -> Result<PathBuf> {
     )
 }
 
+/// Returns true if `dir` contains a real Python installation (not a
+/// Windows Store stub).  The WindowsApps stub passes `exists()` but
+/// fails when invoked, so we run `python --version` to verify.
+fn is_real_python(dir: &Path) -> bool {
```
bbonaby, line 94: thought: I wonder what happens if the script uses  python apis that aren't supported in the version installed in the sandbox. I guess we'd return the error back  via stderr, then they would do another script to get the version installed in the sandbox and then update the script to use supported apis.

src/wxc_sandbox_daemon/src/sandbox_vm.rs
```diff
@@ -164,18 +194,68 @@ pub async fn launch(wsb_path: &Path) -> Result<()> {
...
+/// not propagated.
 pub async fn teardown() {
     eprintln!("[daemon] tearing down sandbox");
 
-    // WindowsSandbox.exe hosts the UI; killing it closes the sandbox.
-    let _ = Command::new("taskkill")
-        .args(["/F", "/IM", "WindowsSandbox.exe"])
-        .stdout(std::process::Stdio::null())
+    // Kill the sandbox UI and session processes.
+    for process_name in [
```
bbonaby, line 238: question: if someone had other sandbox processes running (made without mxc), could this kill their other sandbox processes?

src/wxc_sandbox_daemon/src/tcp_bridge.rs
```diff
@@ -91,15 +91,17 @@ async fn wait_for_ready(control: &mut TcpStream, timeout: std::time::Duration) -
...
+/// beyond the EXIT frame (e.g. a StreamsReady that arrived in the same read).
 pub async fn execute_on_guest(
     conn: &mut GuestConnection,
     exec_id: &str,
     script_code: &str,
     working_directory: &str,
     timeout_ms: u32,
     host_stdin: &[u8],
-) -> Result<(i32, String)> {
+) -> Result<(i32, String, Vec<u8>, Vec<u8>, Vec<u8>)> {
```
bbonaby: suggestion: tbh it got confusing trying to find out where `control_residual` was coming from and how it's used. My suggestion would be to use a struct instead of a tuple since all of these are related.

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+4. Daemon receives `StreamsReady`, connects 3 new TCP streams (stdin, stdout, stderr) to the agent
+5. Next `EXEC` request reuses the existing sandbox VM with fresh data streams
+
+### Multi-Exec Security Considerations
+
+Multi-exec reuses the same sandbox VM across script executions for performance. This has implications:
+
+**What IS preserved between executions:**
+- VM boundary (separate OS instance) — unchanged
+- Firewall lockdown (host IP only) — applied once at agent startup, persists
```
bbonaby: question: are these applied and then removed between scripts? I assume we only ever have 1 sandbox instance running right?

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+8. Daemon sends `RESULT <code> <stdout-b64> <stderr-b64> <error>\n` to wxc-exec
+
+### Multi-Execution (Same Sandbox)
+
+After the first execution completes:
+
+1. Agent sends `Exit` on control channel
+2. Agent sends `StreamsReady` on control channel (signals it's listening for new connections)
+3. Daemon receives `Exit`, returns result to wxc-exec client
+4. Daemon receives `StreamsReady`, connects 3 new TCP streams (stdin, stdout, stderr) to the agent
```
bbonaby: thought: These tcp connections aren't encrypted so I guess anyone can snoop what's being read?  Since we don't have control over what is sent back by the scripts, I wonder if it's a security issue. (script sends back PII or token info for some website).

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+- Registry modifications
+- Background processes spawned by a previous script
+- Scheduled tasks or services installed by a previous script
+
+**Constraint:** Multi-exec assumes all scripts in a session come from the **same trust domain**. Do NOT reuse a sandbox VM across different callers or trust boundaries.
+
+**Recommended future safeguards:**
+- Kill orphan processes between executions (enumerate and terminate non-system processes)
+- Clean temporary directories between runs
+- Consider an optional `freshVM: true` config flag for callers who need per-execution isolation
```
bbonaby: thought: we should really think about this some more in the future. I could be completely off base here but if we think about unwanted data exfiltration, could these lead to that?

docs/sandbox-integration-plan.md
```diff
@@ -0,0 +1,402 @@
...
+
+Kills sandbox processes in order:
+1. `WindowsSandbox.exe`, `WindowsSandboxServer`, `WindowsSandboxRemoteSession`
+2. Polls for `WindowsSandbox*` and `vmmemWindowsSandbox` to exit (up to 30s)
+3. Waits 5s additional for Hyper-V backend / VHDX release
+
+## Security Model
+
+- **VM isolation**: Scripts run inside a separate Windows instance — full OS boundary
+- **Firewall lockdown**: After the agent accepts host connections, it blocks all other network traffic via `netsh advfirewall` rules scoped to the sandbox
```
bbonaby: question: I thought settings stuff using  `netsh advfirewall`  requires elevation?

test_scripts/run_sandbox_tests.ps1
```diff
@@ -0,0 +1,186 @@
...
+}
+
+# Preflight: check Windows Sandbox is available
+$sandboxFeature = dism /online /get-featureinfo /featurename:Containers-DisposableClientVM 2>&1 |
+    Select-String "State"
+if ($sandboxFeature -notmatch "Enabled") {
+    Write-Host "ERROR: Windows Sandbox feature is not enabled." -ForegroundColor Red
+    Write-Host "Run: dism /online /enable-feature /featurename:Containers-DisposableClientVM /all" -ForegroundColor Yellow
+    exit 1
+}
```
bbonaby, line 47: suggestion: I think we should do something similar in the wxc_common crate, we should check if the feature is available first before attempting to go through with the work required for the daemon and agent, then send the user an error message.

docs/sandbox-integration-plan.md
huzaifa-d: This is for future work. This PR is to only get this to a working state.

docs/sandbox-integration-plan.md
huzaifa-d: Good point, but this PR is to only get this to a working state.

docs/sandbox-integration-plan.md
huzaifa-d: I put it on purpose, but I will take your suggestion and split this in two files.

src/wxc_common/src/sandbox_runner.rs
huzaifa-d: Removed completely.

src/wxc_sandbox_daemon/src/pipe_server.rs
huzaifa-d: I just used these for testing. Will change once we have actual performance numbers.

src/wxc_sandbox_daemon/src/sandbox_vm.rs
huzaifa-d: Yeah, right now it would be to just return the error.

src/wxc_sandbox_daemon/src/tcp_bridge.rs
huzaifa-d: Good suggestion; implemented.

test_scripts/run_sandbox_tests.ps1
huzaifa-d: Good idea; implemented

docs/sandbox-integration-plan.md
huzaifa-d: Yes, src/wxc_sandbox_agent/src/firewall.rs applies _netsh advfirewall_ rules after the agent accepts the 4 host connections. It deletes all existing rules, allows only the host IP on the agent's listen port, and sets default policy to block-all.

docs/sandbox-integration-plan.md
huzaifa-d: Yes, that is the current model. Will probably change in the future.

docs/sandbox-integration-plan.md
huzaifa-d: No, they are persistent through the lifetime. Do you see a concern?

docs/sandbox-integration-plan.md
huzaifa-d: Yeah, I will be clarifying the scope and changing this accordingly later.

docs/sandbox-integration-plan.md
huzaifa-d: You are right but on the sandbox, the agent runs as WDAGUtilityAccount which has local admin privilege.

src/wxc_sandbox_daemon/src/sandbox_vm.rs
huzaifa-d: Added a comment to investigate a better approach.

docs/sandbox-integration-plan.md
huzaifa-d: Added a comment to follow up and check if this is a concern.

docs/sandbox-integration-plan.md
bbonaby: mostly because each invocation of wxc-exec for sandbox could have different networking configurations, so if they persist you'd have to make sure you remove all before the next one comes along.

bbonaby (approved):
Let's make GitHub issues for all the TODOs so we don't lose track of things (Combine them if necessary). When Martins back I'm sure he'd be interested in taking a look at them. I think my biggest issues are around security of the multi execution philosophy and us destroying all sandboxes (even ones not launched by us) on clean up.

We probably need to think about how we're going to end-to-end test all these backends in a CI pipeline, but that's for the future.

MGudgin:
I'm back and have started looking at the PR.

Gudge

huzaifa-d:
Completing the PR, and will be following up with issues/concerns @MGudgin may have in a separate PR.

PR #83, opened 2026-03-28
[LXC] Fix LXC denied file path masking and add run-all test script

SohamDas2021:
- Fix denied file paths: Container start aborted when deniedPaths contained a file (e.g., /etc/shadow) because
  the tmpfs mount used create=dir which can't overlay a file. Now detects file vs directory in the container rootfs- uses /dev/null bind mount for files, tmpfs for directories.
   - Add run_lxc_all_tests.sh: Runs all 6 LXC test configs (3 via test scripts + 3 examples) and reports pass/fail
  summary.

Fixes #75

### DIFF
--- a/src/lxc_common/src/filesystem_mounts.rs
+++ b/src/lxc_common/src/filesystem_mounts.rs
@@ -78,10 +78,8 @@ pub fn configure_filesystem_mounts(
 
         // Determine if the path is a file or directory inside the container rootfs
         // so we use the correct LXC `create=` type for the mount entry.
-        let rootfs_base = container
-            .config_path()
-            .map(|p| format!("{}/{}/rootfs", p, container.name()))
-            .unwrap_or_else(|| format!("/var/lib/lxc/{}/rootfs", container.name()));
+        let lxc_base = container.config_path().unwrap_or("/var/lib/lxc");
+        let rootfs_base = format!("{}/{}/rootfs", lxc_base, container.name());
         let full_path = format!("{}/{}", rootfs_base, container_path);
 
         // Use /dev/null bind mount for files, tmpfs for directories
--- a/src/lxc_common/src/filesystem_mounts.rs
+++ b/src/lxc_common/src/filesystem_mounts.rs
@@ -75,8 +75,27 @@ pub fn configure_filesystem_mounts(
     for host_path in &policy.denied_paths {
         validate_path(host_path)?;
         let container_path = host_path.trim_start_matches('/');
-        let mount_entry = format!("tmpfs {} tmpfs ro,size=0,create=dir 0 0", container_path);
-        logger.log_line(&format!("Masking denied path: /{}", container_path));
+
+        // Determine if the path is a file or directory inside the container rootfs
+        // so we use the correct LXC `create=` type for the mount entry.
+        let rootfs_base = container
+            .config_path()
+            .map(|p| format!("{}/{}/rootfs", p, container.name()))
+            .unwrap_or_else(|| format!("/var/lib/lxc/{}/rootfs", container.name()));
+        let full_path = format!("{}/{}", rootfs_base, container_path);
+
+        // Use /dev/null bind mount for files, tmpfs for directories
+        let is_file = std::path::Path::new(&full_path).is_file();
+        let mount_entry = if is_file {
+            format!("/dev/null {} none bind,ro,create=file 0 0", container_path)
+        } else {
+            format!("tmpfs {} tmpfs ro,size=0,create=dir 0 0", container_path)
+        };
+        let create_type = if is_file { "file" } else { "dir" };
+        logger.log_line(&format!(
+            "Masking denied path: /{} ({})",
+            container_path, create_type
+        ));
         container.set_config_item("lxc.mount.entry", &mount_entry)?;
     }
 
--- a/test_scripts/run_lxc_all_tests.sh
+++ b/test_scripts/run_lxc_all_tests.sh
@@ -7,6 +7,17 @@ PASSED=0
 FAILED=0
 FAILURES=""
 
+# Check for Windows line endings in test scripts
+check_line_endings() {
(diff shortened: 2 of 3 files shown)


test_scripts/run_lxc_all_tests.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
bbonaby, line 1: When I run this i get:
```bash
./run_lxc_all_tests.sh
-bash: ./run_lxc_all_tests.sh: cannot execute: required file not found
```

test_scripts/run_lxc_all_tests.sh
```diff
@@ -0,0 +1,60 @@
...
+        echo "FAIL: $name"
+        FAILED=$((FAILED + 1))
+        FAILURES="$FAILURES\n  - $name"
+    fi
+    echo ""
+}
+
+run_exec "LXC Hello World (example)" "$REPO_DIR/examples/11_lxc_hello_world.json"
+run_exec "LXC Filesystem Access (example)" "$REPO_DIR/examples/12_lxc_filesystem_access.json"
+run_exec "LXC Network Restricted (example)" "$REPO_DIR/examples/13_lxc_network_restricted.json"
```
bbonaby: We'll need to be careful about running the example jsons as is. Some examples might require some initial steps to set up.

src/lxc_common/src/filesystem_mounts.rs
```diff
@@ -75,8 +75,27 @@ pub fn configure_filesystem_mounts(
...
         let container_path = host_path.trim_start_matches('/');
-        let mount_entry = format!("tmpfs {} tmpfs ro,size=0,create=dir 0 0", container_path);
-        logger.log_line(&format!("Masking denied path: /{}", container_path));
+
+        // Determine if the path is a file or directory inside the container rootfs
+        // so we use the correct LXC `create=` type for the mount entry.
+        let rootfs_base = container
+            .config_path()
+            .map(|p| format!("{}/{}/rootfs", p, container.name()))
+            .unwrap_or_else(|| format!("/var/lib/lxc/{}/rootfs", container.name()));
```
bbonaby: should we just error out instead of doing the `or else` to attempt to get this full path? Seems like if the first isn't available that means something is wrong right?

test_scripts/run_lxc_all_tests.sh
SohamDas2021: I had hit this, this is most likely because of windows line endings issue (\r\n). The shebang #!/bin/bash\r makes Linux look for /bin/bash\r which doesn't exist. 

I had replaced them to run it.

test_scripts/run_lxc_all_tests.sh
SohamDas2021: I am thinking if we should only run the test scripts and not the examples.

bbonaby: yea I'm cool with only tests scripts.

test_scripts/run_lxc_all_tests.sh
bbonaby: Oh I see, what I ended up doing to get it to work was to add

```yml
[*.sh]
end_of_line = lf
```
to the editorconfig file at the root of the repo. This should fix it long turn I think, since it'll make sure they're all using LF when checked in. Also is this expected currently? I know you're fixing things:
### Without sudo/root
```bash
bbonaby@bbonaby-dev-2:/mnt/z/mxc/test_scripts$ ./run_lxc_all_tests.sh
=== Basic LXC ===
Running basic LXC test...
Hello from LXC
Linux CLI-LXC-Hello-World 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 Linux
Basic LXC test complete.
PASS: Basic LXC

=== LXC Filesystem ===
Running LXC filesystem test...
Readonly dir: /tmp/tmp.bMDTcbi5c9
Readwrite dir: /tmp/tmp.pGaHYxLTib
Failed to configure filesystem: Failed to set config item lxc.mount.entry: No such file or directory (os error 2)FAIL: LXC Filesystem

=== LXC Network ===
Running LXC network test...
Network policy error: iptables -N MXC-CLI-LXC-Filesystem-T failed: iptables v1.8.10 (nf_tables): Could not fetch rule set generation id: Permission denied (you must be root)
FAIL: LXC Network

================================
Results: 1 passed, 2 failed
Failures:
  - LXC Filesystem
  - LXC Network
```

### With sudo/root
```bash
bbonaby@bbonaby-dev-2:/mnt/z/mxc/test_scripts$ sudo ./run_lxc_all_tests.sh
[sudo] password for bbonaby:
=== Basic LXC ===
Running basic LXC test...
Hello from LXC
Linux CLI-LXC-Hello-World 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 Linux
Basic LXC test complete.
PASS: Basic LXC

=== LXC Filesystem ===
Running LXC filesystem test...
Readonly dir: /tmp/tmp.ECAkKFeZXZ
Readwrite dir: /tmp/tmp.e1smNohDor
test content
LXC filesystem test complete.
PASS: LXC Filesystem

=== LXC Network ===
Running LXC network test...
wget: bad address 'api.github.com'
FAIL: LXC Network

================================
Results: 2 passed, 1 failed
Failures:
```
I'm using wsl ubuntu distro

SohamDas2021: LXC container management inherently requires root privileges (unless unprivileged LXC is specifically configured, which most setups don't do). I have that in lxc-backend.md, may be I should just add an error message in the script. 

1 failure is expected- https://github.com/microsoft/mxc/issues/76

bbonaby (approved):

PR #87, opened 2026-03-31
Update network schema and validation

SohamDas2021:
- Adds validation in config_parser.rs to reject network.allowedHosts and network.blockedHosts for non-LXC containment backends. These fields are only meaningful for LXC (iptables enforcement). AppContainer uses network.proxy instead. Network proxy is already only supported with the 'appcontainer' containment backend
- Updates schemas/mxc-config.schema.json to include allowedHosts/blockedHosts with descriptions clarifying they are LXC-only.
- Removes allowedHosts/blockedHosts from AppContainer example and test configs (03, 04, 11 examples + 4 test configs).
- Removes redundant docs/examples.md (duplicated content from examples/ directory and docs/schema.md).
- Updates docs/schema.md network section to show backend-specific field comments.

Fixes #79


docs/schema.md
```diff
@@ -35,7 +35,9 @@ you add `"$schema": "./schemas/mxc-config.schema.json"` to your config file.
     "network": {
         "defaultPolicy": "block",          // "allow" or "block"
         "enforcementMode": "firewall",     // "capabilities", "firewall", or "both"
-        "proxy": { "localhost": 8080 }     // Localhost proxy port (appcontainer only)
+        "allowedHosts": ["api.github.com"],// LXC only: hostnames/IPs to allow
```
shschaefer, line 38: Please do not comment this like this.

shschaefer (changes requested):

PR #90, opened 2026-03-31
[SDK/CLI] Fix SDK payload format, add SandboxPolicy versioning, and test infrastructure

bbonaby:

The SDK's `buildSandboxPayload()` still emitted the deprecated `script` field, but schema 0.4.0-alpha requires `process.commandLine`. Every `spawnSandbox()` call failed with "Request error". Additionally, `SandboxPolicy` had no version field and there were no tests.

### What changed

**SDK**
- Renamed types: `WxcConfiguration` -> `ContainerConfig`, dropped `Container` prefix from sub-types
- `buildSandboxPayload()` emits `process.commandLine` instead of `script`
- Removed deprecated fields (`script`, `workingDirectory`, `timeout`) from `ContainerConfig`
- Added required `version` field to `SandboxPolicy` with semver validation via the `semver` package
- Exported `findWxcExecutable` (was missing from index.ts)
- Added 8 unit tests using `node:test`

**CLI**
- `run-sdk` rewritten to take `<script>` + `--policy`/`--policy-file` (mirrors `spawnSandbox(script, policy)` API)
- `run` auto-discovers wxc-exec/lxc-exec, removed `--wxc-path`
- Fixed `--config-base64` flag mismatch and base64 arg passing
- Added 3 integration test for app container usage

**CI**
- Added `npm test` for SDK (unit tests only since the CLI integration tests wxc integration and wxc requires a specific version of windows)

### How was it tested?

- 8 SDK unit tests: payload format, capabilities mapping, version validation
- 3 CLI integration tests (local): AppContainer command execution
- Manual: filesystem, proxy, and direct SDK `spawnSandbox()` calls

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -1,37 +1,50 @@
-#!/usr/bin/env node
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
 
 import { Command } from 'commander';
-import { WxcExecutor } from './wxc-executor';
+import { ContainerExecutor } from './wxc-executor';
 import {
   spawnSandbox,
   getPlatformSupport,
   SandboxPolicy,
-  getAvailableToolsPolicy
 } from '@microsoft/mxc-sdk';
 
 import * as fs from 'fs';
 import * as path from 'path';
-import { WxcConfiguration } from '@microsoft/mxc-sdk/dist/types';
+import { ContainerConfig } from '@microsoft/mxc-sdk/dist/types';
 
 const program = new Command();
 
 program
   .name('wxc-cli')
-  .description('CLI for invoking the WXC (Windows eXecution Container)')
+  .description('CLI test driver for the MXC SDK and container backends')
   .version('0.1.0');
 
 program
   .command('run')
-  .description('Run Python code with WXC sandbox')
-  .argument('<config>', 'Path to JSON configuration file or base64-encoded config')
+  .description('Run a config for a specific container backend')
+  .argument('<config>', 'Path to ContainerConfig JSON file or base64-encoded config')
   .option('--config-base64', 'Treat config argument as base64-encoded JSON')
   .option('--debug', 'Enable debug output')
-  .option('--wxc-path <path>', 'Path to wxc-exec.exe executable', path.join(__dirname, '..', '..', 'src', 'target', 'debug', 'wxc-exec.exe'))
-  .action(async (config: string, options: { base64?: boolean; debug?: boolean; wxcPath: string }) => {
+  .action(async (config: string, options: { configBase64?: boolean; debug?: boolean }) => {
     try {
-      const executor = new WxcExecutor(options.wxcPath);
+      const platform = require('os').platform();
+      let execPath: string | null;
+      if (platform === 'linux') {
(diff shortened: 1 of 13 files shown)


sdk/src/sandbox.ts
```diff
@@ -26,66 +49,62 @@ export function buildSandboxPayload(
...
+        },
+    };
+
+    if (platform === 'linux') {
+        config.containment = 'lxc';
+        config.lxc = {
+            containerName: name,
+            // Default Linux distro since SandboxPolicy doesn't expose this
+            distribution: 'alpine',
+            release: '3.23',
```
SohamDas2021, line 78: is this the latest? I remember using 3.20 in our jsons.

cli/src/cli.ts
```diff
@@ -109,98 +123,41 @@ program
...
-        if (!policy.filesystem) {
-          policy.filesystem = {};
-        }
-        policy.filesystem.readwritePaths = [
-          ...(policy.filesystem.readwritePaths ?? []),
-          ...toolsPolicy.readwritePaths,
-        ];
+
+program
+  .command('run-sdk')
```
SohamDas2021, line 128: Do we need to update cli/README.md?

bbonaby: Will publish a follow up which updates the readme

sdk/src/sandbox.ts
bbonaby: yea 3.23 is the latest currently as of december 2025: https://images.linuxcontainers.org/images/alpine/

sdk/src/sandbox.ts
```diff
@@ -26,66 +49,62 @@ export function buildSandboxPayload(
...
+    if (policy.network?.allowOutbound) {
+        capabilities.push("internetClient");
+    }
+    if (policy.network?.allowLocalNetwork) {
+        capabilities.push("privateNetworkClientServer");
+    }
+
+    config.appContainer = {
+        name: name,
+        leastPrivilege: false,
```
bbonaby, line 103: We'll need  to plumb this through the `SandboxPolicy` in the future or remove it. I know it's appcontainer specific right now, but I can see someone wanting what ever container is to be created from the policy to have the least amount of privileges possible in general. So seems like a good fit for `SandboxPolicy`

sdk/src/sandbox.ts
```diff
@@ -26,66 +49,62 @@ export function buildSandboxPayload(
...
+        process: {
+            commandLine: script,
+            cwd: workingDirectory,
+        },
+        containerId: name,
+        filesystem: {
+            readwritePaths: policy.filesystem?.readwritePaths,
+            readonlyPaths: policy.filesystem?.readonlyPaths,
+            deniedPaths: policy.filesystem?.deniedPaths,
+            clearPolicyOnExit: true,
```
bbonaby, line 68: Soham had `lifecycle` in his versioning.md spec [here](https://github.com/microsoft/mxc/blob/main/docs/versioning.md) when he talked about things that the `SandboxPolicy` dictates. Then I asked him to remove it since we don't support it today. I think what he had makes a lot of sense and that the `SandboxPolicy` should be updated in the future to have a `lifecycle/lifetime`  object. Seems super general for all kinds of containers too.

shschaefer (approved):

PR #94, opened 2026-04-01
Add a versioning doc

SohamDas2021:


docs/versioning.md
```diff
@@ -0,0 +1,285 @@
+# MXC Versioning Design
+
+## Core Concepts
+
+### Policy = Intent
+
+The policy (filesystem, network, lifecycle) expresses **what** the user wants — "block network, allow these paths, destroy on exit." It does not specify how the OS enforces it, nor which container type to use.
```
bbonaby: suggestion: I don't think we have `lifecycle` in the policy right now but it does make sense to have. For now we can remove that part.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+- **Patch** (x.y.Z) — bug fixes only
+- **Minor** (x.Y.0) — new features, backward compatible
+- **Major** (X.0.0) — breaking changes
+
+## Two Independent Version Streams
+
+| Stream | Owner | What it versions | Format |
+|---|---|---|---|
+| **MXC Policy/Config** | MXC team | SandboxPolicy + config format | Semver (e.g., 0.4.0-alpha) |
+| **OS Tech Language** | OS team | What the OS can enforce (sandbox spec) | Version info (e.g., 1.4.5, 2.0.0) |
```
bbonaby: suggestion: let's let Jeff  add this section in when he adds MXC <--> OS integration. So that code goes in the same time as this part.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+MXC is responsible for mapping policy intent onto the OS tech language version.
+They are **not** lock-stepped.
+
+## Schema Shipping Model
+
+```
+mxc/schemas/
+├── schema.0.3.0-alpha.json      (shipped — historical)
+├── schema.0.4.0-alpha.json      (shipped — current stable)
+└── mxc-config.schema.json       (current — includes experimental section)
```
bbonaby: suggestion: we probably could do `stable/`  and `dev/` folders, where the historical and current stable one go in `stable/` while `mxc-config.schema.json` says in `dev/`. just so we have a clean separation.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+```
+
+**Rules:**
+- **Generic section** (top) — shipped, stable, supported. Always executed.
+- **Experimental section** (bottom) — only executed when the experimental flag
+  is enabled (see below).
+- **Promotion:** When an experimental feature is ready to ship, move it from
+  `experimental` to the generic section and bump the minor version.
```
bbonaby: suggestion: we should probably say something about what is expected inside the  `experimental` section. Something a long the lines of Devs can put any keys or nested keys they want in there and that they would be responsible for adding the parsing code for their experimental feature. As long as their experimental code doesn't break what is shipped then they're free to iterate as much as they want.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Experimental Flag
+
+The experimental flag must be supported at every layer of the stack:
+
+**1. `wxc-exec.exe` / `lxc-exec` (Rust binaries):**
+```bash
+wxc-exec.exe --experimental config.json
+lxc-exec --experimental config.json
```
bbonaby: might be quicker to switch `--experimental` and `config.json` around since I _think_ we take the config without an flag as the first argument

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+);
+```
+
+## Error Contract
+
+MXC ↔ OS needs a defined error contract:
+- Which feature failed
+- Whether it's a version mismatch or runtime unavailability (e.g., Hyper-V off)
+- What the user should do (upgrade OS, enable feature, etc.)
+- Security policy is deterministic — no relaxation, no fuzzy fallback
```
bbonaby, line 259: suggestion: let's remove the OS related parts + negotiation out until jeff's integration stuff is in. Then we'll have the full picture. Folks might come here and be confused about the OS stuff that isn't already in.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+We are introducing a `version` field in the SandboxPolicy. This version must match the MXC config JSON version — they are the same version, tied 1:1.
+
+```typescript
+// sdk/src/types.ts
+SandboxPolicy {
+  version: "0.4.0-alpha",       // must match MXC config schema version
+  filesystem: { ... },           // policy (intent)
+  network: { ... },              // policy (intent)
+}
+```
```
bbonaby, line 26: question: Just so we're on the same page about the 1:1. 

Say we were on v0.5.0 and previous version was v0.4.0. We package all the schemas for a major version into the npm package. In this case lets say major version 0.

Then, if a consumer uses 0.4.0 as the SandboxPolicy version, then we create the WxcConfiguration object that uses the 0.4.0 schema.  

if 0.5.0 was used in the SandboxPolicys version field, then we create a WxcConfiguration object with 0.5.0 schema .

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+  debug: false
+});
+```
+
+The SDK passes `--experimental` to the underlying binary when this option is set.
+
+**3. CLI (`mxc-cli`):**
+```bash
+npm start run config.json --experimental
+```
```
bbonaby: suggestion: we can remove this part. I'd think of the cli as an implementation detail, as it's really just a test driver project for using the sdk.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+stable section, configs that still reference it under `experimental` will receive
+an error: "feature X has moved to the stable section." The parser will not
+silently fall back — explicit migration is required.
+
+## Open Questions
+
+1. **Experimental + stable interaction:** Can experimental features modify or
+   override the behavior of stable features? Or must they be fully isolated?
+   (e.g., experimental compartments changing how `network.defaultPolicy` is
+   enforced)
```
bbonaby: thought: since we're saying that the experimental flag is only for development then it is probably fine to say that experimental features should be fully isolated. If we change to allowing it in the stable schema then that's a different story.

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+   discover and target them?
+
+3. **Security of the experimental flag:** Should `--experimental` require
+   additional privilege or be restricted to debug builds? A malicious caller could
+   pass `--experimental` to enable a feature that weakens the sandbox boundary.
+
+4. **Telemetry / diagnostics:** When something fails with `--experimental`
+   enabled, how do we distinguish between experimental feature bugs and stable
+   feature regressions? Should experimental execution be tagged differently in
+   logs?
```
bbonaby: thought: well since we're saying it's only for development then technically only us devs will see the failure right? So, I think this is fine for now if the experimental feature is fully isolated.

docs/versioning.md
SohamDas2021: Clarified.

docs/versioning.md
jsidewhite: >experimental [](http://example.com/codeflow?start=53&length=12)

is there "experimental" anything in the schema?

docs/versioning.md
jsidewhite: yeah good point - it probably needs at least a "feature" level node

"experimental" : [
    "feature": {
            "name": "exp feat 1"
     },
    "feature": {
            "name": "exp feat 2"
     },
]

so we can surface the names - but otherwise, freeform inside of "feature"

?

docs/versioning.md
```diff
@@ -0,0 +1,285 @@
...
+
+The experimental flag must be supported at every layer of the stack:
+
+**1. `wxc-exec.exe` / `lxc-exec` (Rust binaries):**
+```bash
+wxc-exec.exe --experimental config.json
+lxc-exec --experimental config.json
+```
+
+When `--experimental` is passed:
```
jsidewhite, line 117: `--experimental <feat name>`

We have to allow experimental features to be on/off 1 at a time, not all or nothing.

docs/versioning.md
jsidewhite: i don't agree with this - experimental features must be shippable to anthropic/nemoclaw/whatever

docs/versioning.md
jsidewhite: ... what - cli must parse without order bias

docs/versioning.md
```diff
@@ -0,0 +1,275 @@
...
+a production feature.
+
+**Global flag:** The `--experimental` flag is a single global toggle. When enabled,
+all experimental features in the config are active. There is no per-feature
+enable/disable mechanism — simplicity over granularity.
+
+**Isolation:** Experimental features must be fully isolated from stable features.
+They must not modify or override the behavior of stable features. Since the
+experimental flag is a development tool, this isolation keeps the stable code path
+deterministic regardless of what experiments are running.
```
jsidewhite: why impose this req?  I don't think you can reason over this.

what if i turn on "AppContainer **as Medium IL**" with an experimental flag?  am i breaking the stable appcontainer feature?

docs/versioning.md
```diff
@@ -0,0 +1,275 @@
...
+}
+```
+
+**Promotion process:** When an experimental feature is ready to ship:
+1. Move the field from `RawExperimental` to the stable section of `RawConfig`
+2. Move the field from `ExperimentalConfig` to `CodexRequest`
+3. Remove the `if request.experimental_enabled` guard — the feature now always
+   applies
+4. Update the schema: move from `experimental` section to generic section
+5. Bump the minor version
```
jsidewhite, line 180: so actually just realized that this isn't strictly how it works.  there's no single VERSION.txt file that you open and bump.

for the set of *major* versions that you still care about where feature is **additive*, you add to a new *minor* version schema update to those *major* versions.

for the set of *major* versions that you still care about where feature is **breaking*, don't touch them, but instead add new *major* version schema.

for the set of *major* versions that you don't care about do nothing.

so you might have many schemas in dev, not just one.

    schemas/dev/1.vnext.json
    schemas/dev/2.vnext.json
    schemas/dev/3.vnext.json

docs/versioning.md
```diff
@@ -0,0 +1,275 @@
...
+We are introducing a `version` field in the SandboxPolicy. This version must match the MXC config JSON version — they are the same version, tied 1:1.
+
+When a consumer specifies a SandboxPolicy version (e.g., `0.4.0`), MXC creates
+the corresponding `WxcConfiguration` using the `0.4.0` schema. If a different
+version is specified (e.g., `0.5.0`), MXC uses the `0.5.0` schema. All schemas
+for a major version are packaged together in the SDK.
+
+```typescript
+// sdk/src/types.ts
+SandboxPolicy {
```
jsidewhite, line 21: >SandboxPolicy [](http://example.com/codeflow?start=0&length=13)

might be worth noting that this struct is subject to violent evolution

jsidewhite (approved):
:shipit:

PR #96, opened 2026-04-01
Config schema restructuring to support versioning

SohamDas2021:


docs/schema.md
```diff
@@ -2,9 +2,9 @@
 ## Configuration Schema
 
 MXC uses a JSON configuration file. The formal schema is at
-[`schemas/mxc-config.schema.json`](../schemas/mxc-config.schema.json) —
+[`schemas/dev/mxc-config.schema.json`](../schemas/dev/mxc-config.schema.json) —
 editors that support JSON Schema will provide autocomplete and validation when
-you add `"$schema": "./schemas/mxc-config.schema.json"` to your config file.
+you add `"$schema": "./schemas/dev/mxc-config.schema.json"` to your config file.
```
bbonaby, line 8: future:  I guess we can decide later whether we want to point to the dev version in the docs or the stable version

bbonaby (dismissed):

jsidewhite:
so is the idea we have a "prep for release" script (or manual instructions), which copies mxc-config.schema.json to schemas/stable/schema.0.5.0-alpha.json right before we release.  then the npm packing pipeline grabs all schemas in the stable dir and puts them in the package?

docs/schema.md
```diff
@@ -2,9 +2,9 @@
...
```
jsidewhite, line 7: >mxc-config.schema.json [](http://example.com/codeflow?start=35&length=22)

pasting comment from previous pr here,

so actually just realized that this isn't strictly how it works.  there's no single VERSION.txt file that you open and bump.

for the set of **major** versions that you still care about where feature is **additive**, you add to a new *minor* version schema update to those *major* versions.

for the set of **major** versions that you still care about where feature is **breaking**, don't touch them, but instead add new *major* version schema.

for the set of **major** versions that you don't care about do nothing.

so you might have many schemas in dev, not just one.
```
schemas/dev/1.vnext.json
schemas/dev/2.vnext.json
schemas/dev/3.vnext.json
```

SohamDas2021:
> so is the idea we have a "prep for release" script (or manual instructions), which copies mxc-config.schema.json to schemas/stable/schema.0.5.0-alpha.json right before we release. then the npm packing pipeline grabs all schemas in the stable dir and puts them in the package?



> so is the idea we have a "prep for release" script (or manual instructions), which copies mxc-config.schema.json to schemas/stable/schema.0.5.0-alpha.json right before we release. then the npm packing pipeline grabs all schemas in the stable dir and puts them in the package?

yes, that is correct.

docs/schema.md
SohamDas2021: I have captured this in the open questions in the versioning doc (https://github.com/microsoft/mxc/pull/94)

jsidewhite (approved):
:shipit:

PR #97, opened 2026-04-01
[NanVix] Adding initial NanVix support

huzaifa-d:
## NanVix Backend for MXC

Adds NanVix as a containment backend in wxc-exec.exe. When the JSON config specifies `"containment": "nanvix"`, the binary routes to a new `NanVixScriptRunner` that spawns `nanvixd.exe` (the NanVix microkernel VM daemon), pipes the script via stdin, and relays stdout/stderr directly to the parent process.

### What Changed

- `models.rs` Added `NanVix` variant to `ContainmentBackend` enum (replaces `MicroVm`)
- `config_parser.rs` Added `"nanvix"` containment parsing
- `nanvix_runner.rs` **NEW** `NanVixScriptRunner` implementing `ScriptRunner` trait
- `lib.rs` Added `pub mod nanvix_runner` (Windows-gated with `#[cfg(windows)]`)
- `main.rs` Added `ContainmentBackend::NanVix` dispatch arm
- `types.ts` Updated SDK types: `microvm` â†’ `nanvix`
- `schema.md` / `schema.json` Updated docs and JSON schema
- `nanvix_hello.json` Example test configuration

### NanVix Runner Design

- **I/O model**: stdin piped (script code), stdout/stderr inherited (relayed to parent)
- **Auto-discovery**: `nanvixd.exe`, `kernel.elf`, `python.elf`, `cpython-ramfs.img` discovered next to the running executable
- **Error handling**: Internal `NanVixError` enum with 4 variants (Preflight, Platform, Runtime, Timeout)
- **Watchdog**: Condvar-based timer thread with duplicated Windows HANDLE for safe process termination (avoids PID reuse)
- **Policy validation**: Rejects filesystem paths, network policies, proxy, and working directory (NanVix has its own isolation)
- **Handle leak fix**: Duplicated process HANDLE is closed on both normal-exit and timeout paths via shared closure

### Commits

1. **First Implementation** (`e380f3f`) Core NanVix backend: runner, config parsing, models, SDK types, schema updates
2. **Close Handle Fix** (`a9f39a5`) P0 fix: handle leak on normal-exit path, P1 proxy policy check, P2 python_home delimiter validation, 4 new tests
3. **Fix clippy lint** (`203cc52`) Add `Default` impl for `NanVixScriptRunner`, fix test config JSON format

### Tests

167 workspace tests pass. Zero clippy warnings. E2E test verified with actual NanVix VM producing correct output (`Hello from NanVix! sum=100`).

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -59,7 +59,7 @@ you add `"$schema": "./schemas/dev/mxc-config.schema.json"` to your config file.
 | `"wslc"` | Linux containers via the WSL Container SDK |
 | `"lxc"` | Native LXC container isolation |
 | `"vm"` | VM-based isolation (not yet implemented) |
-| `"microvm"` | MicroVM-based isolation (not yet implemented) |
+| `"nanvix"` | NanVix microkernel VM isolation via WHP |
 
 Only the backend section matching the selected `containment` value is used;
 other backend sections are ignored.
--- a/src/wxc_common/src/nanvix_runner.rs
+++ b/src/wxc_common/src/nanvix_runner.rs
@@ -9,8 +9,14 @@
 //! ## I/O model
 //!
 //! - **stdin**: runner writes script code, then closes (EOF triggers `exec()`)
-//! - **stdout**: relayed directly to parent process (not captured)
-//! - **stderr**: relayed directly to parent process (kernel traces)
+//! - **stdout**: relayed directly to parent process via `Stdio::inherit()` (not captured)
+//! - **stderr**: relayed directly to parent process via `Stdio::inherit()` (kernel traces)
+//!
+//! **Note for SDK consumers:** Because stdout/stderr are inherited (not captured),
+//! `ScriptResponse.standard_out` and `standard_err` are always empty strings for
+//! the NanVix backend. Output is streamed directly to the parent's console/pipes.
+//! Programmatic consumers that need captured output should redirect wxc-exec's
+//! stdout/stderr at the process level.
 //!
 //! ## Exit codes
 //!
@@ -149,6 +155,8 @@ impl NanVixScriptRunner {
 impl ScriptRunner for NanVixScriptRunner {
     fn run(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
         // -- Policy validation ---------------------------------------------------
+        // Reject unsupported policies — NanVix provides its own isolation model.
+        // Fail-closed: any policy setting not meaningful for NanVix is rejected.
         if !request.policy.readwrite_paths.is_empty()
             || !request.policy.readonly_paths.is_empty()
             || !request.policy.denied_paths.is_empty()
@@ -171,12 +179,32 @@ impl ScriptRunner for NanVixScriptRunner {
             )
             .to_response();
         }
+        if request.policy.network_proxy.is_enabled() {
+            return NanVixError::Preflight(
+                "network proxy is not supported by the NanVix backend \
+                 — NanVix has no network stack"
+                    .to_string(),
+            )
+            .to_response();
+        }
         if !request.working_directory.is_empty() {
             return NanVixError::Preflight(
                 "workingDirectory is not supported by the NanVix backend \
                  -- guest has its own filesystem namespace"
                     .to_string(),
             )
+            .to_response();
+        }
+
+        // Validate PYTHON_HOME doesn't contain NanVix delimiters that could
+        // corrupt the guest argument string (';' separates argv from env vars,
+        // spaces separate argv entries).
+        if PYTHON_HOME.contains(';') || PYTHON_HOME.contains(' ') {
+            return NanVixError::Preflight(format!(
+                "PYTHON_HOME '{}' contains invalid characters (';' or space) \
+                 — these are NanVix argument delimiters",
+                PYTHON_HOME
+            ))
             .to_response();
         }
 
@@ -285,17 +313,29 @@ impl ScriptRunner for NanVixScriptRunner {
                 let result = cvar.wait_timeout(cancelled, duration).unwrap();
                 cancelled = result.0;
 
+                // Always close the duplicated handle to avoid leaks.
+                let close_handle = |handle_raw: usize| {
+                    use windows::Win32::Foundation::{CloseHandle, HANDLE};
+                    let handle = HANDLE(handle_raw as *mut std::ffi::c_void);
+                    let _ = unsafe { CloseHandle(handle) };
+                };
+
                 if *cancelled {
+                    // Process already exited — close the handle and return.
+                    if let Some(handle_raw) = process_handle_raw {
+                        close_handle(handle_raw);
+                    }
                     return;
                 }
 
+                // Timeout elapsed and process is still running — kill it.
                 if let Some(handle_raw) = process_handle_raw {
-                    use windows::Win32::Foundation::{CloseHandle, HANDLE};
+                    use windows::Win32::Foundation::HANDLE;
                     use windows::Win32::System::Threading::TerminateProcess;
 
                     let handle = HANDLE(handle_raw as *mut std::ffi::c_void);
                     let kill_result = unsafe { TerminateProcess(handle, 1) };
-                    let _ = unsafe { CloseHandle(handle) };
+                    close_handle(handle_raw);
 
                     if kill_result.is_ok() {
                         timed_out_clone.store(true, Ordering::SeqCst);
@@ -489,5 +529,47 @@ mod tests {
             !resp.error_message.contains("workingDirectory"),
             "default request should not trigger workingDirectory rejection"
         );
+    }
+
+    #[test]
+    fn policy_rejects_network_proxy() {
+        let mut runner = NanVixScriptRunner::new();
+        let request = CodexRequest {
+            policy: ContainerPolicy {
+                network_proxy: crate::models::ProxyConfig {
+                    address: Some(crate::models::ProxyAddress::new(
+                        "127.0.0.1".to_string(),
+                        8080,
+                        true,
+                    )),
+                    builtin_test_server: false,
+                },
+                ..Default::default()
+            },
+            ..Default::default()
+        };
+        let mut logger = Logger::new(Mode::Buffer);
+        let resp = runner.run(&request, &mut logger);
+        assert_eq!(resp.exit_code, -1);
+        assert!(resp.error_message.contains("network proxy"));
+    }
+
+    #[test]
+    fn python_home_constant_has_no_delimiters() {
+        // PYTHON_HOME must not contain ';' or spaces — these are NanVix
+        // argument delimiters that would corrupt the guest arg string.
+        assert!(!PYTHON_HOME.contains(';'), "PYTHON_HOME contains ';'");
+        assert!(!PYTHON_HOME.contains(' '), "PYTHON_HOME contains space");
+    }
+
+    #[test]
+    fn guest_args_format_is_correct() {
+        let expected = "-S -B -c exec(__import__('sys').stdin.read());PYTHONHOME=/sysroot";
+        let actual = format!(
+            "-S -B -c exec(__import__('sys').stdin.read());PYTHONHOME={}",
+            PYTHON_HOME
+        );
+        assert_eq!(actual, expected);
+        assert!(!"exec(__import__('sys').stdin.read())".contains(' '));
     }
 }
--- a/src/wxc_common/src/nanvix_runner.rs
+++ b/src/wxc_common/src/nanvix_runner.rs
@@ -30,10 +30,11 @@
 use std::fmt::Write;
 use std::io::Write as IoWrite;
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
+use std::process::{Child, Command, Stdio};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Condvar, Mutex};
 use std::thread;
+use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 
 use crate::logger::Logger;
@@ -54,6 +55,13 @@ const RAMFS_IMAGE: &str = "cpython-ramfs.img";
 const BOOT_TIMEOUT_MS: u64 = 60_000;
 /// Generic error exit code returned to host callers.
 const ERROR_EXIT_CODE: i32 = -1;
+const ERR_FILESYSTEM_POLICY: &str =
+    "filesystem policy is not supported by the NanVix backend -- guest has a read-only ramfs";
+const ERR_NETWORK_POLICY: &str =
+    "network policy is not supported by the NanVix backend -- NanVix has no network stack";
+const ERR_PROXY_POLICY: &str =
+    "network proxy is not supported by the NanVix backend -- NanVix has no network stack";
+const ERR_WORKDIR: &str = "workingDirectory is not supported by the NanVix backend -- guest has its own filesystem namespace";
 
 // -- NanVix error classification ---------------------------------------------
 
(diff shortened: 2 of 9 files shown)


docs/schema.md
```diff
@@ -59,7 +59,7 @@ you add `"$schema": "./schemas/mxc-config.schema.json"` to your config file.
 | `"wslc"` | Linux containers via the WSL Container SDK |
 | `"lxc"` | Native LXC container isolation |
 | `"vm"` | VM-based isolation (not yet implemented) |
-| `"microvm"` | MicroVM-based isolation (not yet implemented) |
```
devigned, line 62: Perhaps, microVM or some non-specific product'ish name would be better understood by a user.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+
+    /// Compute total timeout: boot grace + script timeout.
+    fn total_timeout_ms(script_timeout: u32) -> u64 {
+        let script_ms = get_timeout_milliseconds(script_timeout) as u64;
+        BOOT_TIMEOUT_MS.saturating_add(script_ms)
+    }
+}
+
+impl ScriptRunner for NanVixScriptRunner {
+    fn run(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
```
bbonaby: suggestion: I want to say this for the other runners too but I think we should start breaking these function bodies up. This way these functions don't because unwieldly. (CreateProcess PTSD 🥲).

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+        }
+        if !request.working_directory.is_empty() {
+            return NanVixError::Preflight(
+                "workingDirectory is not supported by the NanVix backend \
+                 -- guest has its own filesystem namespace"
+                    .to_string(),
+            )
+            .to_response();
+        }
+
```
bbonaby: thought: I think this is actually interesting, so basically the only thing `nanvix/microVM` cares about is the script that it's gonna run?  I wonder if just surfacing a warning that these will be ignored is enough rather than hard erroring. Then we can document that nanvix doesn't support these.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+// -- NanVix error classification ---------------------------------------------
+
+/// Classifies NanVix runner errors for structured error handling.
+#[derive(Debug)]
+enum NanVixError {
+    /// Missing binaries, invalid paths, unsupported policies.
+    Preflight(String),
+    /// WHP unavailable, spawn failure.
+    Platform(String),
```
bbonaby: question: WHP == Windows Hypervisor Platform? Also what's the difference between a spawn failure and missing binaries in the preflight context? Just trying to understand the distinction between the preflight and platform.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+impl NanVixScriptRunner {
+    pub fn new() -> Self {
+        Self { _private: () }
+    }
+
+    /// Resolve and validate all required paths next to the running executable.
+    fn resolve_paths(&self) -> Result<(PathBuf, PathBuf, PathBuf, PathBuf), NanVixError> {
+        let dir = exe_dir()?;
+
+        let nanvixd = dir.join("nanvixd.exe");
```
bbonaby: question: is this a future executable? I don't see where it's located in the PR. If `exe_dir()` gets the current directory of `wxc-exec.exe` then I'd expect to see a script that copies `nanvixd.exe` there. I don't see one currently so that's why I ask.

That said, we'd want to be able to get dependencies via crates.io or through another distribution channel. Is your thinking that we'd check in the binary itself?

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+        if !nanvixd.exists() {
+            return Err(NanVixError::Preflight(format!(
+                "nanvixd.exe not found in {:?}",
+                dir
+            )));
+        }
+
+        let ramfs = dir.join("cpython-ramfs.img");
```
bbonaby: question: same as above, where's the `.img` coming from.

test_configs/nanvix_hello.json
```diff
@@ -0,0 +1,7 @@
+{
+    "process": {
+        "commandLine": "x = 42\ny = 58\nprint('Hello from NanVix! sum=%d' % (x + y))",
```
bbonaby, line 3: question: is it expected that nanvix will only run python? or can it run other scripting languages?

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+                 -- guest has its own filesystem namespace"
+                    .to_string(),
+            )
+            .to_response();
+        }
+
+        // Validate PYTHON_HOME doesn't contain NanVix delimiters that could
+        // corrupt the guest argument string (';' separates argv from env vars,
+        // spaces separate argv entries).
+        if PYTHON_HOME.contains(';') || PYTHON_HOME.contains(' ') {
```
bbonaby: question: don't we control what `PYTHON_HOME` contains when we initialized it at the top of the file? wondering if we should just move this comment up there and tell folks not to add these characters.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+use std::thread;
+use std::time::Duration;
+
+use crate::logger::Logger;
+use crate::models::{CodexRequest, NetworkPolicy, ScriptResponse};
+use crate::script_runner::{get_timeout_milliseconds, ScriptRunner};
+
+const PYTHON_BINARY: &str = "python.elf";
+const PYTHON_HOME: &str = "/sysroot";
+const BOOT_TIMEOUT_MS: u64 = 60_000;
```
ppenna, line 55: Could you add comments explaining these constants?

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+                total_ms, BOOT_TIMEOUT_MS, script_timeout_ms
+            ),
+        }
+    }
+}
+
+impl NanVixError {
+    fn to_response(&self) -> ScriptResponse {
+        ScriptResponse {
+            exit_code: -1,
```
ppenna: Should this be a file-scope constant?

src/wxc_common/src/nanvix_runner.rs
ppenna: `nanvixd.exe` should be a file-scope constant

src/wxc_common/src/nanvix_runner.rs
ppenna: `cpython-ramfs.img` should be afile scope constant.

src/wxc_common/src/nanvix_runner.rs
ppenna: I would breakup this unction input multiple functions to make code easier to read.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+            .arg(&bin_dir)
+            .arg("-ramfs")
+            .arg(&ramfs_path)
+            .arg("--")
+            .arg(&python_path)
+            .arg(&guest_args)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
+            .spawn()
```
ppenna: You may consume these from `nanvix` create from `github.com`

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+            let process_handle_raw: Option<usize> = if dup_ok.is_ok() {
+                Some(dup_handle.0 as usize)
+            } else {
+                None
+            };
+
+            Some(thread::spawn(move || {
+                let (lock, cvar) = &*cancel_pair_clone;
+                let mut cancelled = lock.lock().unwrap();
+                let result = cvar.wait_timeout(cancelled, duration).unwrap();
```
ppenna: We probably don't want to call `unwrap()` on production code. Instead, fail with an error message.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,579 @@
...
+            let dup_ok = unsafe {
+                DuplicateHandle(
+                    GetCurrentProcess(),
+                    HANDLE(raw),
+                    GetCurrentProcess(),
+                    &mut dup_handle,
+                    0,
+                    false,
+                    DUPLICATE_SAME_ACCESS,
+                )
```
ppenna: We should document `SAFETY` contracts.

src/wxc_common/src/nanvix_runner.rs
ppenna: +1

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: I will change this in the other PR.

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: Good point.

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: Thought about it; Hard error it is

test_configs/nanvix_hello.json
huzaifa-d: For this PR, I just want to make sure about Python

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: I have another PR that will add a build.rs

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: I have another PR that will add a build.rs

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: Preflight = before we try to spawn (paths, config validation). Platform = spawn attempt fails (WHP not available, binary missing at runtime)
Added comments.

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: Good catch. Old code from earlier implementation. Will remove.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+        }
+
+        Ok(())
+    }
+
+    fn build_guest_args() -> String {
+        format!(
+            "-S -B -c exec(__import__('sys').stdin.read());PYTHONHOME={}",
+            PYTHON_HOME
+        )
```
ppenna, line 289: We used to have a comment detailing why this is required. Should we put it back?

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+        let _ = writeln!(logger, "NanVix: python={:?}", python_path);
+
+        // Build the guest argument string.
+        // The ';' is NanVix's separator between argv and environment variables:
+        //   everything before ';' -> kernel splits on spaces -> argv[]
+        //   everything after ';'  -> kernel sets as env vars
+        // -S: skip site.py, -B: no .pyc writing
+        // -c exec(...): reads all of stdin and executes it (no interactive >>> prompts)
+        // Note: exec(__import__('sys').stdin.read()) has NO spaces, so it survives
+        //       NanVix's space-splitting in build_string_table().
```
ppenna: Actually, we should move this inside the function now.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+
+                use windows::Win32::Foundation::HANDLE;
+                use windows::Win32::System::Threading::TerminateProcess;
+
+                let handle = HANDLE(process_handle_raw as *mut std::ffi::c_void);
+                // SAFETY: `handle` is a valid duplicated process handle owned by
+                // this thread, and passing exit code 1 is valid for termination.
+                let _ = unsafe { TerminateProcess(handle, 1) };
+                close_handle(process_handle_raw);
+            }))
```
ppenna: We can encapsulate this in a function as well to improve redability.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+                )
+            };
+            if let Err(e) = dup_ok {
+                let err = NanVixError::Platform(format!(
+                    "failed to duplicate {} process handle: {}",
+                    NANVIXD_BINARY, e
+                ));
+                let _ = writeln!(logger, "{}", err);
+                let _ = child.kill();
+                let _ = child.wait();
```
ppenna: We may want to capture errors and log them for improved failure diagnostics.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+            (&nanvixd_path, &bin_dir, &ramfs_path, &python_path),
+            &guest_args,
+            &request.script_code,
+        ) {
+            Ok(c) => c,
+            Err(e) => {
+                let _ = writeln!(logger, "{}", e);
+                return e.to_response();
+            }
+        };
```
ppenna: If we fail beyond this point, we need to make sure to cleanup the process. For a safer programming pattern, we may want to have RAII semantics here: https://github.com/nanvix/nanvix/blob/dev/src/utils/nanvix-test/src/nanvixd/mod.rs

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+
+        // Signal watchdog to stop
+        {
+            let (lock, cvar) = &*cancel_pair;
+            let mut cancelled = lock.lock().unwrap_or_else(|e| e.into_inner());
+            *cancelled = true;
+            cvar.notify_one();
+        }
+        if let Some(t) = watchdog {
+            let _ = t.join();
```
ppenna: Log errors instead of silencing them.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+
+        let was_timed_out = timed_out.load(Ordering::SeqCst);
+
+        if was_timed_out {
+            let err = NanVixError::Timeout {
+                script_timeout_ms: request.script_timeout,
+                total_ms: timeout_ms,
+            };
+            let _ = writeln!(logger, "{}", err);
+            return err.to_response();
```
ppenna, line 461: No kill if we timeout?

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,622 @@
...
+        }
+
+        match exit_status {
+            Ok(status) => {
+                let exit_code = status.code().unwrap_or(ERROR_EXIT_CODE);
```
ppenna, line 466: Should we have an special exit code for this condition?

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: Added back; dropped during refactor.

src/wxc_common/src/nanvix_runner.rs
huzaifa-d: Added logging.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -0,0 +1,705 @@
...
+    // always sees it as true after child.wait() returns from a kill.
+    timed_out.store(true, Ordering::SeqCst);
+
+    use windows::Win32::Foundation::HANDLE;
+    use windows::Win32::System::Threading::TerminateProcess;
+
+    let handle = HANDLE(process_handle_raw as *mut std::ffi::c_void);
+    // SAFETY: `handle` is a valid duplicated process handle owned by
+    // this thread, and passing exit code 1 is valid for termination.
+    let _ = unsafe { TerminateProcess(handle, 1) };
```
ppenna, line 174: We should we have a file-scope constant for this hardcode value.

ppenna (approved):
LGTM left a minor suggestion.

ppenna:
@huzaifa-d This should wait for PR #108 to be merged first, as per offline discussion with @shschaefer .

SohamDas2021:
Also we should make this feature "experimental"- contained by the experimental flag, once my experimental infra PR is checked in- https://github.com/microsoft/mxc/pull/108. We should file an issue for this.

huzaifa-d:
Merging. Will be moved to experimental; tracked via #113

PR #100, opened 2026-04-01
[LXC] Fix DNS resolution failure in LXC

SohamDas2021:
Fixes #76 . 

When defaultPolicy: "block" with allowedHosts set, DNS resolution failed inside the container because the script executed before the container's network stack (DHCP/DNS) had fully initialized. After lxc-start, the container's DHCP lease and /etc/resolv.conf take ~3-4 seconds to initialize. The runner was executing the script immediately, causing wget: bad address errors even though iptables rules correctly allowed DNS (port 53).

Fix
Added wait_for_network() which polls lxc-info -iH every 250ms until the container reports an IP address (up to 10s timeout). Only runs when network features are in use (firewall enforcement mode or allowed/blocked hosts configured), so
non-networked containers have zero overhead.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Force LF line endings for shell scripts (prevents CRLF issues in WSL/Linux)
+*.sh text eol=lf
--- a/src/lxc_common/src/lxc_runner.rs
+++ b/src/lxc_common/src/lxc_runner.rs
@@ -10,7 +10,9 @@ use std::thread;
 use std::time::{Duration, Instant};
 
 use wxc_common::logger::Logger;
-use wxc_common::models::{CodexRequest, LifecycleConfig, LxcConfig, NetworkEnforcementMode, ScriptResponse};
+use wxc_common::models::{
+    CodexRequest, LifecycleConfig, LxcConfig, NetworkEnforcementMode, ScriptResponse,
+};
 use wxc_common::script_runner::ScriptRunner;
 use wxc_common::validator::validate_request;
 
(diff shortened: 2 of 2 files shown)


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Force LF line endings for shell scripts (prevents CRLF issues in WSL/Linux)
+*.sh text eol=lf
```
bbonaby, line 2: ❤️

src/lxc_common/src/lxc_runner.rs
```diff
@@ -43,6 +47,46 @@ impl LxcScriptRunner {
         }
     }
 
+    /// Wait for the container's network stack to initialize.
+    /// Polls `lxc-info` until the container has an IP address or the timeout is reached.
+    fn wait_for_network(container_name: &str, timeout: Duration, logger: &mut Logger) -> bool {
+        let start = Instant::now();
+        let poll_interval = Duration::from_millis(250);
```
bbonaby: suggestion: based on what you said in the PR description, if it takes a couple seconds then might be good to increase the poll duration to 500ms or a 1. I was also thinking about whether we should make this an async function since we're gonna be waiting for 10 seconds in the worse case  but I think for now it's probably fine.

SohamDas2021: It seems the LXC runner is inherently sequential, so I would skip making it async for now. Bumping the interval to 500ms.

bbonaby (approved):

PR #101, opened 2026-04-01
Add experimental feature guide and update versioning doc

SohamDas2021:
Adds a step-by-step developer guide for adding experimental features to MXC, and updates versioning.md to use the new object-based experimental section design.

### DIFF
--- a/docs/add-a-new-experimental-feature.md
+++ b/docs/add-a-new-experimental-feature.md
@@ -157,18 +157,37 @@ Add tests to verify:
 - Missing optional fields use defaults
 - Unknown fields under `experimental` are ignored (forward compatibility)
 
+Also ensure that `convert_raw_config()` populates `CodexRequest.experimental`:
+
+```rust
+Ok(CodexRequest {
+    // ... existing fields ...
+    experimental,   // ← include the parsed experimental config
+})
+```
+
 ## Step 4: Implement the feature in the runner
 
 > The `--experimental` CLI flag and `experimental_enabled` field on
 > `CodexRequest` already exist from when `compartments` was added. No changes
 > needed in `main.rs`.
 
+The full flow is:
+
+```
+main.rs: cli.experimental → request.experimental_enabled = true
+main.rs: runner.run(&request, &mut logger)
+  → runner checks request.experimental_enabled
+    → reads request.experimental.gpu_isolation
+      → applies the feature
+```
+
 In the appropriate runner (`appcontainer.rs`, `lxc_runner.rs`, etc.), guard
 your feature behind `experimental_enabled`:
 
 ```rust
 fn run(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
-    // ... normal execution ...
+    // ... normal execution (filesystem, network, etc.) ...
 
     if request.experimental_enabled {
         // existing experimental feature
@@ -182,7 +201,7 @@ fn run(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse
         }
     }
 
-    // ... continue normal execution ...
+    // ... execute the script ...
 }
 ```
 
@@ -220,6 +239,14 @@ wxc-exec.exe test_configs/experimental_gpu_isolation.json --experimental --debug
 # Without flag — experimental section silently ignored, normal execution
 wxc-exec.exe test_configs/experimental_gpu_isolation.json --debug
 ```
+
+Verify three things:
+1. **With `--experimental`:** debug output shows your feature was applied
+   (e.g., "Applying GPU isolation: device 0, 1024MB limit")
+2. **Without `--experimental`:** no trace of your feature in the output,
+   process executes normally
+3. **Stable features unaffected:** filesystem, network, and other policies
+   still work exactly as before in both modes
 
 ## Step 6: Update the SDK (if needed)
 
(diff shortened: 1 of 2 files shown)


docs/add-a-new-experimental-feature.md
```diff
@@ -0,0 +1,287 @@
...
+## Overview
+
+Adding an experimental feature touches these files:
+
+| File | What to change |
+|------|----------------|
+| `schemas/dev/mxc-config.schema.json` | Add `gpuIsolation` as a feature under `experimental` |
+| `src/wxc_common/src/models.rs` | Add `GpuIsolationConfig` struct, add field to `ExperimentalConfig` |
+| `src/wxc_common/src/config_parser.rs` | Add `gpuIsolation` field to `RawExperimental` |
+| Runner (`appcontainer.rs` or `lxc_runner.rs`) | Feature logic, guarded behind `experimental_enabled` |
```
bbonaby, line 27: thought: reading it out like this makes me realize we should add a new refactor issue. appcontainer.rs vs lxc_runner.rs etc. I'll add one.

docs/add-a-new-experimental-feature.md
```diff
@@ -0,0 +1,287 @@
...
+    }
+    if let Some(g) = raw_exp.gpu_isolation {
+        experimental.gpu_isolation = Some(GpuIsolationConfig {
+            device_index: g.device_index,
+            memory_limit_mb: g.memory_limit_mb,
+            allow_cuda: g.allow_cuda.unwrap_or(false),
+        });
+    }
+}
+```
```
bbonaby: thought: we're gonna want to tell folks to make their own new functions. e.g something like this
```rust
 if let Some(g) = raw_exp.gpu_isolation {
        experimental.gpu_isolation = Some(GpuIsolationConfig::new(some_parsed_struct));
    }
```
then they don't have to touch this function again. Only their `new::()` when they add/remove things from the config.

docs/add-a-new-experimental-feature.md
```diff
@@ -0,0 +1,287 @@
...
+    // ... existing fields ...
+    experimental,   // ← include the parsed experimental config
+})
+```
+
+## Step 4: Implement the feature in the runner
+
+> The `--experimental` CLI flag and `experimental_enabled` field on
+> `CodexRequest` already exist from when `compartments` was added. No changes
+> needed in `main.rs`.
```
bbonaby, line 183: question: I guess we'll add the initial framework code to make this all a reality after this PR right?

docs/add-a-new-experimental-feature.md
```diff
@@ -0,0 +1,287 @@
...
+
+When your experimental feature is ready to ship:
+
+1. Move the property from `experimental` to a top-level property in the schema
+   (e.g., `experimental.gpuIsolation` → `gpuIsolation`)
+2. Move the struct from `ExperimentalConfig` to `CodexRequest`
+3. Move the field from `RawExperimental` to `RawConfig`
+4. Remove the `if request.experimental_enabled` guard
+5. Bump the minor version
+6. Add a parser error for configs still referencing the feature under
```
bbonaby, line 285: thought: forever? or should we say until next release then we remove the error and we'd fallback into an error like `X is not an available experimental feature` or something like that?

docs/versioning.md
```diff
@@ -71,36 +71,35 @@ The current config schema has two sections:
...
-Each experimental feature has a `name` (used for identification and future
-per-feature gating — see Open Questions) and a freeform `config` object that
-the feature developer defines. The `name` must be unique across all experimental
-features. Today, the `--experimental` flag is a global toggle that enables all
+Each experimental feature is a typed property under `experimental` — the same
+pattern as stable features (`filesystem`, `network`) under the top-level
+config. This gives editors full autocomplete and validation for experimental
+configs. Today, the `--experimental` flag is a global toggle that enables all
 experimental features; per-feature gating (e.g., `--experimental compartments`)
 is under consideration.
```
bbonaby, line 94: issue: it just hit me that we're talking about one flag to turn on all experimental features but we never spoke about what to do if there are multiple backends that are experimental. I guess this goes for stable too. We should make sure we only ever allow one backend per parsed config right? e.g if we parsed the json and there are more than one backends we should hard error. Perhaps to show our intent in the schema we can do something similar to what we did here https://github.com/microsoft/mxc/blob/eb9f25193d1e29431ca25caa444fcd866eae8b74/schemas/dev/mxc-config.schema.json#L127 . This is more for the future, not this PR.

suggestion: would be that we state somewhere that we only support one backend per configuration regardless of stable vs experimental.

docs/add-a-new-experimental-feature.md
SohamDas2021: Yes. I had a chat with Anis on this, I will be working on the experimental infra tomorrow. Anis wanted this doc to be in today.

So, I am expecting an iteration to this doc if we find missing links during implementation.

docs/add-a-new-experimental-feature.md
SohamDas2021: Yes, I agree, the error should persist for at least one release cycle (so users have time to migrate), then it could be relaxed to the standard "unknown field" behavior.

docs/versioning.md
SohamDas2021: Yes, I agree with you. We should hard block jsons with multiple backends, I will create an issue on this to address later.

SohamDas2021: https://github.com/microsoft/mxc/issues/104

bbonaby (approved):

PR #107, opened 2026-04-02
[SDK/CLI] Remove export of findWxcExecutable and findLxcExecutable functions from SDK 

bbonaby:
Addresses part of #105.

SDK consumers should interact through `spawnSandbox()` which handles executable discovery internally. The `findWxcExecutable` and `findLxcExecutable` functions are implementation details that should not be part of the public API surface.

Changes

 - SDK: Remove `findWxcExecutable` and `findLxcExecutable` from `sdk/src/index.ts` exports (functions remain available internally for sandbox.ts)
 - CLI: Delete the legacy run command which was the only external consumer of these functions. Functional tests use the `run-sdk` command which goes through the public `spawnSandbox()` function of the SDK. This was only used to test passing wxc-exec configs directly to wxc-exec, which is not needed as `run-sdk` uses public sdk functions we eventually do this under the hood. 
 - See `run-sdk` here: https://github.com/microsoft/mxc/blob/93ccc2a420ce8b92a5aa62d098a8676b89b83db8/cli/src/cli.ts#L127

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/107)

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -22,52 +22,6 @@ program
   .version('0.1.0');
 
 program
-  .command('run')
-  .description('Run a config for a specific container backend')
-  .argument('<config>', 'Path to ContainerConfig JSON file or base64-encoded config')
-  .option('--config-base64', 'Treat config argument as base64-encoded JSON')
-  .option('--debug', 'Enable debug output')
-  .action(async (config: string, options: { configBase64?: boolean; debug?: boolean }) => {
-    try {
-      const platform = require('os').platform();
-      let execPath: string | null;
-      if (platform === 'linux') {
-        const { findLxcExecutable } = require('@microsoft/mxc-sdk');
-        execPath = findLxcExecutable();
-      } else {
-        const { findWxcExecutable } = require('@microsoft/mxc-sdk');
-        execPath = findWxcExecutable();
-      }
-      if (!execPath) {
-        console.error('Error: Executable not found. Ensure wxc-exec or lxc-exec is built.');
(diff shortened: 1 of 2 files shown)


cli/src/cli.ts
```diff
@@ -21,52 +21,6 @@ program
   .description('CLI test driver for the MXC SDK and container backends')
   .version('0.1.0');
 
-program
-  .command('run')
-  .description('Run a config for a specific container backend')
-  .argument('<config>', 'Path to ContainerConfig JSON file or base64-encoded config')
-  .option('--config-base64', 'Treat config argument as base64-encoded JSON')
-  .option('--debug', 'Enable debug output')
-  .action(async (config: string, options: { configBase64?: boolean; debug?: boolean }) => {
```
bbonaby, line 30: Note: Spoke to Stuart, this is a legacy bootstrap script from before we went all in on `SandboxPolicy` as the SDK’s primary interface. In this function we use the exported `findwxcExecutable` and `findlxcExecutable` from the sdk on lines `36` and `39` . These functions are intended to remain internal to the sdk itself, this entire function needed to be removed as the sdk will no longer export them. 

Interactions with wxc/lxc from the cli project should go through the run-sdk script below, which invokes them via the supported SDK pathways. So the removal of this does nothing, except help with #105 where I intend to remove other legacy stuff.

SohamDas2021 (approved):

PR #110, opened 2026-04-02
Add BaseContainer runner for AppContainer re-platform

jsidewhite:
Route AppContainer containment through `Experimental_CreateProcessInSandbox` API based on schema_version. Configs with version >= 0.5 use the new BaseContainer code-path; older or missing versions fall back to the legacy inline AppContainer implementation. An `--experimental` CLI flag forces the BaseContainer path regardless of version.

## Changes
- **Add BaseContainerRunner** (`base_container_runner.rs`): builds a FlatBuffer `SandboxSpec` from the container policy and calls `Experimental_CreateProcessInSandbox` via `processmodel.dll`. Includes `internetClient` auto-add parity with the legacy runner. Handles `WAIT_FAILED` from `WaitForSingleObject` by returning an error with `GetLastError()` detail.
- **Add SandboxSpec FlatBuffer schema** and generated Rust bindings
- **Update wxc-exec** `main.rs` with version-based runner dispatch and `--experimental` flag (replaces former `--tessera` flag)
- **Simplify BaseContainer identity string** (drop MSIX-style naming)
- **Move `SandboxSpec.fbs`** to `external/windows-sdk/` to clarify it is an ingested Windows SDK dependency
- **Update `basic_appcontainer.json`** test config with file output and sleep

###### Microsoft Reviewers: <a href="https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/110">Open in CodeFlow</a>


src/wxc/src/main.rs
```diff
@@ -40,6 +41,10 @@ struct Cli {
     /// Container name (required with --delete)
     #[arg(long = "containername")]
     containername: Option<String>,
+
+    /// Force the Tessera (CreateProcessInSandbox) code-path regardless of schema version
```
jsidewhite: >Tessera [](http://example.com/codeflow?start=18&length=7)

I guess we probably want to remove all traces of this name, but "sandbox" means something else in this repo :P

thoughts? #Resolved

src/wxc_common/src/sandbox_spec_generated.rs
```diff
@@ -0,0 +1,413 @@
+// automatically generated by the FlatBuffers compiler, do not modify
```
jsidewhite: @bbonaby do we check this in or generate it with flatc.exe during the build? #Pending

src/wxc/src/main.rs
SohamDas2021: Yes, I agree with you. If we are calling developer intent as "sandbox policy", this adds to the naming confusion :(.

SohamDas2021:
High-level question- so we would continue to have the app container path in MXC to support older Windows builds? #ByDesign

src/wxc_common/src/tessera_runner.rs
```diff
@@ -0,0 +1,337 @@
...
+
+        let _ = writeln!(logger, "Tessera: process created (PID: {})", pi.dwProcessId);
+
+        // 5. Wait for the child process to exit.
+        let timeout_ms = get_timeout_milliseconds(request.script_timeout);
+        let mut exit_code: u32 = u32::MAX;
+
+        unsafe {
+            let wait_result = WaitForSingleObject(pi.hProcess, timeout_ms);
+            if wait_result == WAIT_TIMEOUT {
```
SohamDas2021: do we need to check for other wait_results like WAIT_FAILED?

src/wxc_common/src/base_container_runner.rs
```diff
@@ -0,0 +1,337 @@
...
+        // 5. Wait for the child process to exit.
+        let timeout_ms = get_timeout_milliseconds(request.script_timeout);
+        let mut exit_code: u32 = u32::MAX;
+
+        unsafe {
+            let wait_result = WaitForSingleObject(pi.hProcess, timeout_ms);
+            if wait_result == WAIT_TIMEOUT {
+                let _ = writeln!(logger, "Tessera: process timed out, terminating...");
+                let _ = TerminateProcess(pi.hProcess, u32::MAX);
+                let _ = WaitForSingleObject(pi.hProcess, 5000);
```
SohamDas2021, line 269: we are ignoring the wait_result here? #Resolved

src/wxc/src/main.rs
jsidewhite: @copilot can you please update all mentions of "Tessera" in this PR to be replaced with "BaseContainer"

src/wxc_common/src/tessera_runner.rs
jsidewhite: @copilot notice how appcontainer.rs checks for other failures.  can we just copy what they did?

```rs
        let timeout_ms = get_timeout_milliseconds(request.script_timeout);

        let wait_result = unsafe { WaitForSingleObject(process_handle.get(), timeout_ms) };

        match wait_result {
            WAIT_OBJECT_0 => {}
            WAIT_TIMEOUT => unsafe {
                let _ = TerminateProcess(process_handle.get(), u32::MAX);
                let _ = WaitForSingleObject(process_handle.get(), u32::MAX);
            },
            WAIT_FAILED => {
                let err = unsafe { GetLastError() };
                return Err(WxcError::Process(format!(
                    "WaitForSingleObject failed: {:?}",
                    err
                )));
            }
            other => {
                return Err(WxcError::Process(format!(
                    "WaitForSingleObject returned unexpected value: {}",
                    other.0
                )));
            }
        }
```

jsidewhite:
(in theory yes, but) I don't think we need to answer that here and now - it's a discussion we can have with our crew and customers based on reach and functionality.

---
In reply to: [4181729193](https://github.com/microsoft/mxc/pull/110#issuecomment-4181729193) [](http://example.com/codeflow?ancestors=4181729193)

external/windows-sdk/BaseContainerSpecification.fbs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
```
bbonaby: issue: we'd still have to use the open source copyright header #Pending

src/wxc/src/main.rs
```diff
@@ -68,6 +73,26 @@ fn display_script_results(response: &ScriptResponse, logger: &mut Logger) {
...
+    if parts.len() < 2 {
+        return false;
+    }
+    let Ok(major) = parts[0].parse::<u32>() else {
+        return false;
+    };
+    let Ok(minor) = parts[1].parse::<u32>() else {
+        return false;
+    };
+    major > min_major || (major == min_major && minor >= min_minor)
```
bbonaby: suggestion: we can use https://crates.io/crates/semver. If you ever change how this works it'll come in handy i feel #WontFix

src/wxc/src/main.rs
```diff
@@ -139,9 +164,39 @@ fn main() {
...
+                        "(empty)"
+                    } else {
+                        &request.schema_version
+                    }
+                );
+                Box::new(AppContainerScriptRunner::new())
+            }
+        }
         ContainmentBackend::Sandbox => Box::new(SandboxScriptRunner::new(&request.sandbox_config)),
         ContainmentBackend::Wslc => {
```
bbonaby: suggestion: we should move the if else into a separate function. That said, I wonder if it is worth it to actually have a factory provide this main function with the correct runner. Then main can just say "Hey factory here is the containment type and info hand me back the correct runner to use". then it also doesn't have to worry about what is experimental or not. #WontFix

src/wxc_common/src/sandbox_spec_generated.rs
bbonaby: My takeaway from using FlatBuffers in Rust is that we should check in the generated code. This improves IntelliSense (no build required) when pulling down repo for the first time. Also removes the need for a separate generation step.

Note: the current Rust output appears to be the legacy format (single file). I believe the newer version generates a module per fbs and then a mod per type with `--rust-module-root-file`.

My suggestion would be to try out:
`--rust --gen-object-api --force-empty --no-prefix --rust-module-root-file --gen-all` 

We can place the generated code in its own crate/folder, it'll have a mod.rs file that exports SandboxSpec which would get generated as:

```rust
#[derive(Default, Clone, Debug)]
pub struct SandboxSpecT {
    pub version: String,
    pub app_container: bool,
    pub integrity_level: u32,
    pub ui_restrictions: u32,
    pub least_privilege: bool,
    pub capabilities: Option<String>,
    pub fs_read_write: Option<Vec<String>>,
    pub fs_read_only: Option<Vec<String>>,
}
```
then it can be used like : 

```rust
use flatbuffers::FlatBufferBuilder;

let spec = SandboxSpecT {
    version: "1.0".to_string(),
    app_container: true,
    integrity_level: 0x1000,
    ui_restrictions: 0,
    least_privilege: true,
    capabilities: Some("internetClient,registryRead".to_string()),
    fs_read_write: Some(vec!["C:\\temp".to_string()]),
    fs_read_only: Some(vec!["C:\\windows".to_string()]),
};

let mut builder = FlatBufferBuilder::new();

let offset = spec.pack(&mut builder);
builder.finish(offset, Some(b"SBOX"));

let bytes = builder.finished_data();
```

bbonaby: Oh also we probably should write a readme somewhere that has the version of the flatbuffer compiler you used and flags you used to generate the rust code. That would help future folks when they want to update the spec.

src/Cargo.toml
```diff
@@ -40,4 +40,5 @@ tokio = { version = "1", features = ["full"] }
 uuid = { version = "1", features = ["v4"] }
 libc = "0.2"
 nix = { version = "0.29", features = ["mount", "sched", "signal", "net", "process", "user"] }
-lxc_common = { path = "lxc_common" }
\ No newline at end of file
+lxc_common = { path = "lxc_common" }
+flatbuffers = "24"
```
bbonaby: suggestion: if I recall when I looked last month https://crates.io/crates/flatbuffers had some rust updates in the new version that came out in december. We can just update this to 25. #Pending

src/wxc_common/src/base_container_runner.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        finish_sandbox_spec_buffer(&mut builder, spec);
+        builder.finished_data().to_vec()
+    }
+
+    /// Load `processmodel.dll` and resolve the `Experimental_CreateProcessInSandbox` export.
+    fn load_api() -> Result<PfnCreateProcessInSandbox, String> {
+        let dll_name = string_util::to_wide("processmodel.dll");
+
+        unsafe {
```
bbonaby, line 147: thought: Not for your PR but we've been throwing around unsafe  mostly to use Windows apis, but we should probably start adding safety comments. https://std-dev-guide.rust-lang.org/policy/safety-comments.html#safety-comments. I am definitely guilty of not adding them. I'll have to add a github issue to do a pass through the repo. #Resolved

src/wxc_common/src/sandbox_spec_generated.rs
jsidewhite: lol i don't have flatc.exe - the llm just "generated" the code from the .fbs file O.o.

src/wxc/src/main.rs
jsidewhite: fully agree - but removing the version check for now since it's not snapping for release (i.e. we don't know the final version for enabling it, technically)

src/wxc/src/main.rs
jsidewhite: mmm - going to let Soham decide this infra - nuked it from my PR.

bbonaby (approved):

PR #115, opened 2026-04-06
[Nanvix] Added install script and tests

huzaifa-d:
## NanVix: Build-time binary download + E2E test suite

A new `nanvix_binaries` crate with a `build.rs` that **automatically downloads, extracts, verifies, and copies** NanVix binaries during `cargo build`. The entire flow is gated behind a Cargo feature flag — the default build is completely unaffected.

### Usage

```bash
# Default build — no change, no download, no network
cargo build

# Build with NanVix binaries (downloads from GitHub releases)
cargo build --features microvm

# Or via build.bat
build.bat --with-microvm
```

### Design

**Zero Rust build-dependencies.** Instead of pulling in HTTP clients, zip libraries, or crypto crates, the build script uses Windows built-in tools:

| Tool | Ships with | Replaces |
|------|-----------|----------|
| `curl.exe` | Windows 10 1803+ | HTTP client (minreq/ureq) |
| `tar.exe` | Windows 10 1803+ | Zip extraction (zip crate) |
| `certutil` | All Windows | SHA256 verification (sha2 crate) |

**Reproducible builds.** Versions are pinned in `versions.json` with exact release tags and asset names. SHA256 checksums in `checksums.json` are verified on every build (catches corrupted cache, not just post-download).

**Two GitHub repos.** Binaries come from `nanvix/nanvix` (nanvixd.exe, kernel.elf) and `nanvix/cpython` (python.elf, cpython-ramfs.img). Download URLs are constructed deterministically — no GitHub API calls needed.

### Files

| File | Purpose |
|------|---------|
| `src/nanvix_binaries/build.rs` | Download, extract, verify (477 lines) |
| `src/nanvix_binaries/versions.json` | Pinned release tags + exact asset names |
| `src/nanvix_binaries/checksums.json` | SHA256 hashes for all 4 binaries |
| `src/nanvix_binaries/src/lib.rs` | Exposes `NANVIX_BIN_DIR` for build-time use |
| `src/wxc/build.rs` | Copies binaries next to `wxc-exec.exe` |
| `build.bat` | `--with-microvm` flag → `--features microvm` |

### E2E Test Suite

7 test configs + a WHP-aware PowerShell test runner:

Run with: `powershell -ExecutionPolicy Bypass -File test_scripts\run_nanvix_tests.ps1`

The test runner detects WHP availability and skips gracefully (exit 0) on machines without it.

### Testing

- ✅ `cargo build` — default build unaffected (zero impact)
- ✅ `cargo build --features microvm` — downloads 4 binaries, checksums verified
- ✅ `cargo test -p wxc_common` — 173 tests pass
- ✅ E2E: `wxc-exec.exe --debug nanvix_hello.json` → `Hello from NanVix! sum=100`, exit 0


src/nanvix_binaries/versions.json
ppenna, line 1: We have nightly releases shipping for Nanvix and Cpython. Should we have a workflow that updates them?

src/wxc/build.rs
```diff
@@ -0,0 +1,78 @@
...
+    // Always emit rerun-if-changed so Cargo doesn't re-run unnecessarily.
+    println!("cargo:rerun-if-changed=build.rs");
+}
+
+#[cfg(all(windows, feature = "microvm"))]
+fn copy_nanvix_binaries() {
+    use std::fs;
+    use std::path::Path;
+
+    let nanvix_bin_dir = match std::env::var("DEP_NANVIX_BINARIES_BIN_DIR") {
```
ppenna, line 19: We may want to have hardcode values as constants at file scope-level.

src/wxc/build.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+    };
+
+    let src_dir = Path::new(&nanvix_bin_dir);
+    let binaries = [
+        "nanvixd.exe",
+        "kernel.elf",
+        "python.elf",
+        "cpython-ramfs.img",
+    ];
```
ppenna: We should consume these from a single source of truth for easy maintainability.

src/nanvix_binaries/build.rs
```diff
@@ -0,0 +1,504 @@
...
+    println!("cargo:rustc-env=NANVIX_BIN_DIR={}", bin_dir.display());
+    println!("cargo:BIN_DIR={}", bin_dir.display());
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=versions.json");
+    println!("cargo:rerun-if-changed=checksums.json");
+    println!("cargo:rerun-if-env-changed=GITHUB_TOKEN");
+    println!("cargo:rerun-if-env-changed=GH_TOKEN");
+}
+
+// -- Manual JSON parsing (versions.json + checksums.json) --------------------
```
ppenna: Can't we use a create for that?

ppenna (commented):
I left some questions/suggestions.

src/nanvix_binaries/versions.json
huzaifa-d: Yes, I will be updating in a later PR; this is the first iteration.

src/wxc/build.rs
huzaifa-d: Same comment as above.

src/wxc/build.rs
huzaifa-d: What would you suggest?

src/nanvix_binaries/build.rs
huzaifa-d: Yes, we can but it adds 10 transitive crates. Since this is my first time on a Rust codebase, I thought this would be better. What do you suggest?

src/wxc/build.rs
ppenna: Add another create ie `nanvix-common` that holds all these common constants and values, and consume values from it.

src/nanvix_binaries/build.rs
ppenna: Use `serde` + `serde_json`, which is widely used by the Rust community and well-maintained.

src/nanvix_binaries/build.rs
ppenna, line 1: Even in build scripts, prefer returning results to surface errors while doing some cleanup. For instance, by removing artifacts that were downloaded and may leave the file system in a dangling state.

src/wxc/build.rs
ppenna, line 1: Same comment as previous `build.rs` file

test_scripts/run_nanvix_tests.ps1
ppenna, line 1: Should we hook this in the ci, as follow-up task?

huzaifa-d: I thought of this too, but the CI environment doesn't have WHP

ppenna: Should we also open an issue to enable that?

src/nanvix_binaries/build.rs
huzaifa-d: Understood.

ppenna (approved):
LGTM

PR #117, opened 2026-04-06
[MXC/WSL] Setup for MXC to invoke WSLC SDK APIs

SohamDas2021:
Copies WSLC SDK artifacts to external/wslc-sdk. Adds the wslc_common crate with Rust FFI bindings to the WSLC SDK, laying the foundation for WSL Container support in MXC
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/117)


external/wslc-sdk/README.md
```diff
@@ -0,0 +1,40 @@
...
+**Source:** Microsoft internal self-host package (`Microsoft.WSL.Containers.2.8.1.nupkg`)
+
+## Contents
+
+- `include/wslcsdk.h` — C API header
+- `runtimes/win-x64/wslcsdk.lib` — Static import library (x64)
+- `runtimes/win-x64/wslcsdk.dll` — Runtime DLL (x64)
+- `runtimes/win-arm64/wslcsdk.lib` — Static import library (ARM64)
+- `runtimes/win-arm64/wslcsdk.dll` — Runtime DLL (ARM64)
+
```
bbonaby, line 11: issue: lets not add the binaries to the repo itself. My suggestion would be to update the `build.rs` file you added to pull down `Microsoft.WSL.Containers.2.8.1.nupkg` from a nuget feed. Our public nuget feed is https://dev.azure.com/shine-oss/mxc/_artifacts/feed/MxcDependencies. You should have permission to add it. However it's truly meant to stay internal then for now this is what I'd do:

1. add the nupkg to external/wslc-sdk.
2. use a build.rs file to unzip it copy contents to temp cargo build folder
3. use bindgen on the .h file see: https://rust-lang.github.io/rust-bindgen/introduction.html to generate the rust bindings.
4. check in the bindings file like you're doing now.  (now you don't have to check in the .h file at all)

Those might seem like extra steps but this will help when WSL Containers do go public. you'd just add it to the feed, then update step 2 to use `reqwest` to download it from the feed, then unzip and the rest stays the same.

bbonaby: That said, you're welcome to push back if you believe before Build you can get this set up.

src/wslc_common/build.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Build script for wslc_common — links against the WSLC SDK.
```
bbonaby, line 4: thought: took me way too long to realize this was a new crate and not `wxc_common`

src/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["wxc", "wxc_common", "wxc_test_driver", "wxc_test_proxy", "wxc_sandbox_guest", "wxc_sandbox_daemon", "wxc_winhttp_proxy_shim", "lxc", "lxc_common", "generated/base_container_specification"]
+members = ["wxc", "wxc_common", "wxc_test_driver", "wxc_test_proxy", "wxc_sandbox_agent", "wxc_sandbox_daemon", "wxc_winhttp_proxy_shim", "lxc", "lxc_common", "wslc_common", "generated/base_container_specification"]
```
bbonaby: question: sandbox name change intentional?

Also since you're in this file can you format the members so each are on a new line please. I would greatly appreciate it!

SohamDas2021: > question: sandbox name change intentional?
> 
> Also since you're in this file can you format the members so each are on a new line please. I would greatly appreciate it!

Fixed in local, forgot to push.

external/wslc-sdk/README.md
SohamDas2021: I agree on the part of not checking in binaries. Will update.

Regarding using bindgen, I intentionally wrote a curated subset (30 of 54 functions). This keeps the bindings lean and aligned with the runner's actual needs. We can reconsider when the WSLC API stabilizes and the surface area grows.

external/wslc-sdk/README.md
```diff
@@ -0,0 +1,39 @@
...
+  (`wslcsdk.dll`) for x64 and ARM64.
+
+## Usage
+
+The `wslc_common` crate's `build.rs` extracts the `.nupkg` (which is a zip
+file) into the cargo build output directory at compile time, then links
+against `wslcsdk.lib`.
+
+At runtime, `wslcsdk.dll` must be available on the system — it is installed
+by the WSLC SDK MSI and is **not** bundled with MXC.
```
bbonaby, line 19: thought: when you add the code lets do a check  to confirm wslc is available then error with a error message. I assume you already planned to do this but just wanted to make sure we're on the same page.

external/wslc-sdk/README.md
bbonaby: yea I guess the issue would be that with bindgen we know that it's used by _alot_ of projects so the generated code has been mulled over for years for the best performance and interoperability. I'd say the benefits of it out way the negatives of the extra 24 functions. Would also help with maintenance too if wslc adds more functionality. Then we don't have to handwrite them, we just call bindgen as usual. But It's your call, we can revisit in the future when we see how this all plays out.

src/wxc/Cargo.toml
```diff
@@ -15,10 +15,12 @@ windows.workspace = true
 
 [target.'cfg(windows)'.dependencies]
 nanvix_binaries = { path = "../nanvix_binaries", optional = true }
+wslc_common = { workspace = true, optional = true }
```
bbonaby, line 18: is you want we could do what Huzaifa did with nanvix `[target.'cfg(windows)'.dependencies]`

src/Cargo.toml
```diff
@@ -1,5 +1,17 @@
 [workspace]
-members = ["wxc", "wxc_common", "wxc_test_driver", "wxc_test_proxy", "wxc_windows_sandbox_guest", "wxc_windows_sandbox_daemon", "wxc_winhttp_proxy_shim", "lxc", "lxc_common", "generated/base_container_specification"]
+members = [
```
bbonaby, line 2: ❤️

bbonaby (approved):

external/wslc-sdk/README.md
SohamDas2021: https://github.com/microsoft/mxc/issues/142 to track

PR #122, opened 2026-04-07
[MXC/WSL] Add WSL Container execution backend with e2e validation

SohamDas2021:
This PR adds the WSL Container runner, policy mapping, schema wiring, and the end-to-end testing.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/122)


src/wslc_common/src/policy_mapping.rs
```diff
@@ -0,0 +1,402 @@
...
+        // Block specific hosts
+        for host in blocked_hosts {
+            rules.push(format!("iptables -A OUTPUT -d {} -j DROP", host));
+        }
+    }
+
+    if rules.is_empty() {
+        None
+    } else {
+        Some(rules.join(" && "))
```
bbonaby, line 223: question: what does the `&&` do exactly? A comment might be needed for maintenance purposes.

src/wslc_common/src/policy_mapping.rs
```diff
@@ -0,0 +1,402 @@
...
+/// Returns true if per-host network filtering is needed (requires iptables
+/// exec after container start and `Privileged` flag).
+///
+/// Only returns true when the host list can refine the selected default policy:
+/// - `Block` → only `allowed_hosts` matter (allowlist)
+/// - `Allow` → only `blocked_hosts` matter (blocklist)
+pub fn needs_host_filtering(
+    is_default_block: bool,
+    allowed_hosts: &[String],
+    blocked_hosts: &[String],
```
bbonaby, line 139: question: I believe it's more idomatic rust to pass in `&[&str]`since this function doesn't need to care about whether it's an owned string or a string slice. There are some other functions in this file too.

src/wslc_common/src/policy_mapping.rs
```diff
@@ -0,0 +1,402 @@
...
+) -> Option<String> {
+    if allowed_hosts.is_empty() && blocked_hosts.is_empty() {
+        return None;
+    }
+
+    let mut rules = Vec::new();
+
+    if is_default_block && !allowed_hosts.is_empty() {
+        // Allow loopback and established connections first
+        rules.push("iptables -A OUTPUT -o lo -j ACCEPT".to_string());
```
bbonaby, line 191: question: For wslc is root level access needed to perform these actions?

src/wslc_common/src/policy_mapping.rs
```diff
@@ -0,0 +1,402 @@
...
+        }
+    }
+
+    (mounts, warnings)
+}
+
+/// Map the network default policy to a WSLC networking mode.
+///
+/// When `allowedHosts` or `blockedHosts` are present, networking must be
+/// `Bridged` (so the container has connectivity), and per-host filtering
```
bbonaby, line 116: question: Is Bridged the right word here? If so are we talking about this kind of bridge? https://wiki.archlinux.org/title/Network_bridge

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+unsafe extern "system" fn exit_callback(_exit_code: i32, context: *mut c_void) {
+    if context.is_null() {
+        return;
+    }
+    let ctx = &*(context as *const IoContext);
+    if let Ok(mut exited) = ctx.exited.0.lock() {
+        *exited = true;
+        ctx.exited.1.notify_all();
+    }
+}
```
bbonaby, line 101: issue: We should talk putting safety tags on these unsafe functions: https://std-dev-guide.rust-lang.org/policy/safety-comments.html. These ones are especially dangerous since we're deferencing the context and then taking a reference to the deferenced memory. I assume context is guaranteed to have the memory alignment of IoContext always? 

I was really interested in `ctx.exited.0.lock()` and searched up what could go wrong since we have a notify all underneath. Turns out if the lock is ever poisoned somehow we'd end up skipping the notify_all call, meaning who ever is waiting will wait forever. Although here I don't see any reason why the thread would get poisoned, (nowhere to panic before exited is updated). If you want you can update it to something like:

```rust
let mut exited = match ctx.exited.0.lock() {
    Ok(guard) => guard,
    Err(poisoned) => poisoned.into_inner(),
};
```

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+    pub fn new(config: &WslcConfig) -> Self {
+        Self {
+            config: config.clone(),
+        }
+    }
+
+    /// Free a PWSTR returned by the SDK (allocated with CoTaskMemAlloc).
+    unsafe fn free_error_message(ptr: PWSTR) {
+        if !ptr.is_null() {
+            windows::Win32::System::Com::CoTaskMemFree(Some(ptr as *const c_void));
```
bbonaby: question: are there no smart pointers in windows-rs crate for this? if not I think we should add some in a common file.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+        if ptr.is_null() {
+            return String::new();
+        }
+        let mut len = 0;
+        while *ptr.add(len) != 0 {
+            len += 1;
+        }
+        let msg = String::from_utf16_lossy(std::slice::from_raw_parts(ptr, len));
+        Self::free_error_message(ptr);
+        msg
```
bbonaby: issue: I think we should use the string_utils module for this so everyone benefits. Might need to create a method that uses the wide-string crate though. Also technically you should check if the windows-rs crate already has something for this conversion which includes the freeing operation by way of a drop trait.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+        let session_name: Vec<u16> = format!("{}\0", request.container_id)
+            .encode_utf16()
+            .collect();
+        let storage_path_str = self.config.storage_path.clone().unwrap_or_else(|| {
+            std::env::temp_dir()
+                .join("mxc-wslc-sessions")
+                .to_string_lossy()
+                .to_string()
+        });
+        let storage_path_wide: Vec<u16> =
```
bbonaby, line 174: issue: can use string_util module for this

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+        if hr != S_OK {
+            return ScriptResponse::error(&Self::format_error(
+                "WslcInitSessionSettings failed",
+                hr,
+                "",
+            ));
+        }
+
+        if let Some(cpu) = self.config.cpu_count {
+            let _ = WslcSetSessionSettingsCpuCount(&mut session_settings, cpu);
```
bbonaby: issue: we're swallowing the hresult for this one and all the rest throughout this function

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+}
+
+impl ScriptRunner for WSLContainerRunner {
+    fn run(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
+        unsafe { self.run_internal(request, logger) }
+    }
+}
+
+impl WSLContainerRunner {
+    unsafe fn run_internal(&self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
```
bbonaby, line 133: issue: This function is pretty long, I think we should split it up. It was hard keeping track of all the steps.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,595 @@
...
+        let callbacks = WslcProcessCallbacks {
+            on_stdout: Some(io_callback),
+            on_stderr: Some(io_callback),
+            on_exit: None,
+        };
+        let _ = WslcSetProcessSettingsCallbacks(
+            &mut process_settings,
+            &callbacks,
+            &*io_ctx as *const IoContext as *mut c_void,
+        );
```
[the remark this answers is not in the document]
bbonaby: This sounds like a righteous find

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+        let _ = WslcSetProcessSettingsCmdLine(&mut process_settings, argv.as_ptr(), argv.len());
+
+        // Set environment variables
+        if !request.env.is_empty() {
+            let env_cstrings: Vec<Vec<u8>> = request
+                .env
+                .iter()
+                .map(|e| format!("{}\0", e).into_bytes())
+                .collect();
+            let env_ptrs: Vec<PCSTR> = env_cstrings.iter().map(|e| e.as_ptr() as PCSTR).collect();
```
bbonaby, line 353: question: is there no better way to do this using the windows-rs crate? or wide-string crate? Might be good for you to do an inventory of your string usage to see what can be consolidated to  helper functions in the string_util module.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+        if hr != S_OK {
+            return ScriptResponse::error(&Self::format_error(
+                "WslcInitContainerSettings failed",
+                hr,
+                "",
+            ));
+        }
+
+        // -- Step 7: Apply policy mapping --
+        // TODO: Port mappings (WslcConfig.port_mappings) are parsed but not yet applied.
```
bbonaby, line 393: issue: we should add github issues for our TODOs so we all can keep track of things that need to be fixed

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+            &request.policy.readwrite_paths,
+            &request.policy.readonly_paths,
+        );
+        for w in &warnings {
+            let _ = writeln!(logger, "[WSLC] Warning: {}", w);
+        }
+
+        if !mounts.is_empty() {
+            // Build WslcContainerVolume array — keep owned data alive.
+            let wide_paths: Vec<(Vec<u16>, Vec<u8>)> = mounts
```
bbonaby, line 406: issue: another call for wide-string in our string_utils module

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,651 @@
...
+        let _ = WslcSetContainerSettingsNetworkingMode(&mut container_settings, net_mode);
+        let _ = writeln!(logger, "[WSLC] Networking mode: {:?}", net_mode);
+
+        // Build iptables rules (if per-host filtering is needed)
+        let iptables_cmd = policy_mapping::build_iptables_rules(
+            &request.policy.allowed_hosts,
+            &request.policy.blocked_hosts,
+            is_default_block,
+        );
+
```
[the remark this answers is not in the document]
bbonaby: question: Is this a righteous change @SohamDas2021

src/wxc/src/main.rs
```diff
@@ -159,8 +159,28 @@ fn main() {
...
+                    .unwrap_or_default();
+                Box::new(wslc_common::wsl_container_runner::WSLContainerRunner::new(
+                    &wslc_config,
+                ))
+            }
+            #[cfg(not(feature = "wslc"))]
+            {
+                eprintln!("Error: WSLC backend not compiled. Rebuild with --features wslc.");
+                process::exit(1);
+            }
```
bbonaby, line 215: question: What's the rationale to compile by feature. So we can have targeted builds?  Just curious since we'd really only build everything since we need all runners compiled into the final binary. I can see the case for compile on Windows v Linux, just not seeing the case for individual feature compile flags the more I think about it.

src/wxc_common/src/config_parser.rs
```diff
@@ -529,9 +502,36 @@ fn convert_raw_config(raw: RawConfig, logger: &mut Logger) -> Result<CodexReques
...
+                    .into_iter()
+                    .map(|m| PortMapping {
+                        windows_port: m.windows_port.unwrap_or(0),
+                        container_port: m.container_port.unwrap_or(0),
+                        protocol: m.protocol.unwrap_or_else(|| "tcp".to_string()),
+                    })
+                    .collect();
+            }
+            config
+        });
```
bbonaby, line 601: note: we should really refactor the config_parser. Right now it's like a super object that's doing more than parsing and knows too much all the backends.

schemas/dev/mxc-config.schema.0.5.0-dev.json
```diff
@@ -217,6 +217,42 @@
...
+          "properties": {
+            "targetOs": {
+              "type": "string",
+              "default": "linux",
+              "description": "Target OS for the container. Currently only 'linux' is supported."
+            },
+            "image": {
+              "type": "string",
+              "default": "alpine:latest",
+              "description": "Container image name. Images should be pre-pulled externally; the runner will attempt an auto-pull if not found (temporary behavior — to be moved to a setup script).",
```
bbonaby: suggestion: we probably don't want this description in the schema itself. Seems too implementation heavy.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+            );
+            if hr != S_OK {
+                let msg = Self::take_error_message(err_msg);
+                return ScriptResponse::error(&Self::format_error(
+                    "Failed to exec iptables rules",
+                    hr,
+                    &msg,
+                ));
+            }
+            let ipt_guard = WslcProcessGuard::from_raw(ipt_process);
```
bbonaby, line 550: question: At this point it's not possible for ipt_process to still be null right?

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+            return ScriptResponse::error(&format!("COM initialization failed: {:?}", com_hr));
+        }
+        let _ = writeln!(logger, "[WSLC] COM initialized");
+
+        // -- Step 1: Prerequisites check --
+        let mut can_run: BOOL = 0;
+        let mut missing = WslcComponentFlags::None;
+        let hr = WslcCanRun(&mut can_run, &mut missing);
+        if hr != S_OK {
+            return ScriptResponse::error(&Self::format_error("WslcCanRun failed", hr, ""));
```
bbonaby: suggestion: for these errors in this file is there a way we can give `ScriptResponse` a function for formatting errors e.g `ScriptResponse::error_fmt`. Just seems like this should be a general thing

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,595 @@
...
+            let msg = Self::take_error_message(err_msg);
+            return ScriptResponse::error(&Self::format_error(
+                "WslcCreateSession failed",
+                hr,
+                &msg,
+            ));
+        }
+        let session_guard = WslcSessionGuard::from_raw(session);
+        let _ = writeln!(logger, "[WSLC] Session created");
+
```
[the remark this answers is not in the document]
bbonaby: This sounds like a righteous change to me

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+        let io_ctx_exited = Arc::clone(&io_ctx.exited);
+        let callbacks = WslcProcessCallbacks {
+            on_stdout: Some(io_callback),
+            on_stderr: Some(io_callback),
+            on_exit: Some(exit_callback),
+        };
+        let _ = WslcSetProcessSettingsCallbacks(
+            &mut process_settings,
+            &callbacks,
+            &*io_ctx as *const IoContext as *mut c_void,
```
bbonaby: thought: is it possible for the callbacks to fire after this function ends? If `io_ctx` is used as the context for these functions, wouldn't it be dropped after this call and could potentially point to dangling memory/garbage?

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,652 @@
...
+
+        if !stdout.is_empty() {
+            let _ = writeln!(logger, "[WSLC] Captured {} bytes stdout", stdout.len());
+        }
+        if !stderr.is_empty() {
+            let _ = writeln!(logger, "[WSLC] Captured {} bytes stderr", stderr.len());
+        }
+
+        // -- Step 14: Cleanup --
+        if request.lifecycle.destroy_on_exit {
```
bbonaby, line 644: thought: we should be thinking about some RAII wrapper for all this.

src/wxc_common/src/config_parser.rs
SohamDas2021: Yes, I have https://github.com/microsoft/mxc/issues/88 filed to take care of this later.

src/wslc_common/src/policy_mapping.rs
SohamDas2021: Callers pass &request.policy.allowed_hosts which is &Vec<String>, and that coerces to &[String] but not
&[&str], keep this to avoid extra conversions. Same applies to the parameters in build_iptables_rules and build_volume_mounts

src/wslc_common/src/policy_mapping.rs
SohamDas2021: These rules are exec'd inside the container after start via WslcCreateContainerProcess. The container must have the Privileged flag set (grants root + NET_ADMIN capability) for iptables to work. Added comment to clarify.

src/wslc_common/src/policy_mapping.rs
SohamDas2021: Clarified in comment.

src/wslc_common/src/policy_mapping.rs
SohamDas2021: Bridged is the WSLC SDK's enum name (WSLC_CONTAINER_NETWORKING_MODE_BRIDGED), and it provides NAT networking through the WSL2 VM, not a true L2 network bridge. I have clarified in comment.

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: Thanks for raising this. I will update.

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: I am not seeing any in windows-rs crate, creating our own in wxc_common/src/string_util.rs

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: So, we do have two waits, in sequence:

 // Step 12
 let _ = WslcGetProcessExitEvent(process_guard.as_raw(), &mut exit_event);
 if !exit_event.is_null() {
     WaitForSingleObject(HANDLE(exit_event), u32::MAX);  // blocks until process exits
 }

 // After Step 12
 let (lock, cvar) = &*io_ctx_exited;
 let mut exited = lock.lock().unwrap_or_else(|e| e.into_inner());
 if !*exited {
     cvar.wait_timeout(exited, Duration::from_secs(30));  // blocks until exit_callback fires
 }

Only after both waits complete do we read the buffers and return. Since the SDK guarantees "once exit callback fires, no more I/O callbacks will be called," the io_ctx pointer is safe for the entire callback lifetime.



But your concern is valid when there is an early return in between the container starting and the waiting. I will go ahead and fix this.

src/wxc/src/main.rs
SohamDas2021: wslc_common's build.rs links against wslcsdk.lib from the WSLC SDK NuGet package. Without the feature flag, CI builds would fail on machines that don't have the SDK installed.

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: Ack, this would be a big change, filed https://github.com/microsoft/mxc/issues/167 to track and follow up.

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: Will be addressed in https://github.com/microsoft/mxc/issues/167

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,612 @@
...
+        let mut flags = WslcContainerFlags::None;
+        if request.lifecycle.destroy_on_exit {
+            flags = flags | WslcContainerFlags::AutoRemove;
+        }
+        if self.config.gpu {
+            flags = flags | WslcContainerFlags::EnableGpu;
+        }
+        if has_host_rules {
+            // Privileged needed for iptables inside the container
+            flags = flags | WslcContainerFlags::Privileged;
```
huzaifa-d, line 471: If only iptables need this, should this privilege be held throughout?

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,612 @@
...
+                windows::Win32::System::Threading::WaitForSingleObject(
+                    windows::Win32::Foundation::HANDLE(ipt_exit_event),
+                    30_000, // 30s timeout for iptables
+                );
+            }
+
+            let mut ipt_exit_code: i32 = -1;
+            let _ = WslcGetProcessExitCode(ipt_guard.as_raw(), &mut ipt_exit_code);
+            if ipt_exit_code != 0 {
+                let _ = writeln!(
```
huzaifa-d: Shouldn't it error out instead of continuing?

src/wxc_common/src/lib.rs
```diff
@@ -7,7 +7,6 @@ pub mod encoding;
 pub mod error;
 pub mod logger;
 pub mod models;
-#[cfg(target_os = "windows")]
```
huzaifa-d, line 10: On purpose?

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,612 @@
...
+
+        // -- Step 6: Container settings --
+        let image_cstr = format!("{}\0", image_name);
+        let mut container_settings = std::mem::zeroed::<WslcContainerSettings>();
+        let hr = WslcInitContainerSettings(
+            image_cstr.as_bytes().as_ptr() as PCSTR,
+            &mut container_settings,
+        );
+        if hr != S_OK {
+            return sdk_error("WslcInitContainerSettings failed", hr, "");
```
huzaifa-d, line 389: No Arc::from_raw here? And all early returns before line 599?

huzaifa-d (dismissed):

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: This is a valid point. The WSLC SDK sets flags at container creation time via WslcSetContainerSettingsFlags, currently there's no API to downgrade privileges after start. So, the flag is immutable once the container is created. I will follow up with the WSLC SDK team to confirm.

src/wxc_common/src/lib.rs
SohamDas2021: Thanks for the catch, this was the one you added in https://github.com/microsoft/mxc/pull/144/changes. I will remove this change.

src/wxc_common/src/string_util.rs
```diff
@@ -55,6 +55,66 @@ pub fn base64_encode(input: &[u8]) -> String {
     STANDARD.encode(input)
 }
 
+/// RAII wrapper for a PWSTR allocated with `CoTaskMemAlloc`.
+/// Automatically frees the memory with `CoTaskMemFree` on drop.
+/// Used for error messages and other strings returned by COM/Win32 APIs.
+pub struct CoTaskMemPWSTR {
```
bbonaby, line 61: thought: there seems to be lots of examples on GitHub on how to do this. I should have pointed to https://github.com/agashlin/comedy-rs/blob/e065ab6aa0cec8f2eb63fdf3f64b67327f80418d/src/com.rs#L447. That's an example of how we can make it generic. Up to you whether you want to update to use generics though

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,679 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! WSL Container runner — implements `ScriptRunner` for the WSLC SDK backend.
+//!
+//! Orchestrates the full lifecycle:
+//! `WslcCanRun → Session → Image check → Process settings → Container → Start →
```
bbonaby, line 7: thought: in discussion with the agent session folks they've brought up a good point that we might need a better story around provisioning and deprovisioning versus handling everything in one shot. Especially for backends that need setups that can be long.

Basically the question of whether having the executors handle creation -> execution -> destruction. Versus just focusing on attempting to run a workload in a backend.

This is something for the future though, just thinking out loud.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,679 @@
...
+    }
+}
+
+/// Callback invoked by the WSLC SDK for stdout/stderr data.
+///
+/// # Safety
+/// `context` must be a valid pointer to an `Arc<IoContext>` that outlives all
+/// callbacks. This is guaranteed because `io_ctx` is kept alive in `run_internal`
+/// until after the exit event wait completes and all callbacks have finished.
+/// The `IoCtxRawGuard` ensures the Arc reference is reclaimed on all exit paths.
```
bbonaby: question: is this an old comment? I don't see `IoCtxRawGuard` usage in here

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,679 @@
...
+                env_ptrs.as_ptr(),
+                env_ptrs.len(),
+            );
+            if hr != S_OK {
+                return sdk_error("WslcSetProcessSettingsEnvVariables failed", hr, "");
+            }
+        }
+
+        // Set working directory
+        let _cwd_cstr; // kept alive for SDK pointer
```
bbonaby, line 365: question: is this comment still accurate?

src/wslc_common/src/policy_mapping.rs
```diff
@@ -0,0 +1,471 @@
...
+    } else {
+        !blocked_hosts.is_empty()
+    }
+}
+
+/// Validate that a host string is safe for use in an iptables command.
+/// Accepts hostnames (a-z, 0-9, dots, hyphens) and IPv4/IPv6 addresses
+/// (digits, dots, colons, brackets, slash for CIDR).
+/// Rejects empty strings and anything containing shell metacharacters.
+fn is_valid_host(host: &str) -> bool {
```
bbonaby, line 152: note: we have the url crate dependency in the repo now, not sure if this has a validate method but might be useful.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -0,0 +1,679 @@
...
+            drop(err_msg);
+        }
+
+        let _ = WslcTerminateSession(session_guard.as_raw());
+        let _ = writeln!(logger, "[WSLC] Cleanup complete");
+
+        // IoCtxRawGuard drops here, reclaiming the Arc reference.
+        // All callbacks are guaranteed complete after the exit event wait.
+
+        // RAII guards will call Release on drop.
```
bbonaby, line 669: thought: just to test can you confirm (print statements would do) if all are dropped like we expect? Looking at the Arc usage above I believe they should but just wanted to confirm

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: Yes, will fix

src/wslc_common/src/policy_mapping.rs
SohamDas2021: url crate parses full URLs, not bare hostnames/IPs, can be an overkill here.

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: Yes, still accurate. _cwd_cstr is declared outside the if block so it lives until end of function. The comment is correct, it holds the string alive so the raw pointer passed to the SDK remains valid.

src/wslc_common/src/wsl_container_runner.rs
SohamDas2021: Added prints and confirmed.

bbonaby (approved):

PR #123, opened 2026-04-07
Add MicroVM E2E test workflow for GitHub Actions

huzaifa-d:
Adds a new GitHub Actions workflow (\microvm-e2e.yml\) that:

1. **Builds** \wxc-exec.exe\ with \cargo build --features microvm\ (auto-downloads Nanvix/CPython binaries)
2. **Excludes** build output from Windows Defender (prevents quarantine of unsigned exes)
3. **Verifies** all 5 required binaries are present
4. **Checks WHP** availability — gracefully skips if unavailable
5. **Runs** the full 7-test E2E suite (\
un_microvm_tests.ps1\)
6. **Uploads logs** on failure

**Triggers:** PR to main, push to main, manual dispatch. Runs on \windows-latest\ with 30min timeout.

**Note:** WHP is available on \windows-latest\ runners per Nanvix project's proven CI pattern.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/123)

### DIFF
--- a/.github/workflows/microvm-e2e.yml
+++ b/.github/workflows/microvm-e2e.yml
@@ -86,3 +86,31 @@ jobs:
           path: |
             logs/
             **/*.log
+
+      - name: Print performance summary
+        if: steps.whp-check.outputs.whp_available == 'true' && !cancelled()
+        shell: pwsh
+        run: |
+          $jsonPath = Join-Path $env:GITHUB_WORKSPACE "microvm-perf-results.json"
+          if (Test-Path $jsonPath) {
+            $data = Get-Content $jsonPath -Raw | ConvertFrom-Json
+            Write-Host "`n=== MicroVM Performance Summary ==="
+            Write-Host "Commit: $($data.commit)"
+            Write-Host "Timestamp: $($data.timestamp)"
+            Write-Host ""
+            Write-Host ("{0,-35} {1,10} {2,8}" -f "Test", "Time (ms)", "Status")
+            Write-Host ("{0,-35} {1,10} {2,8}" -f "----", "---------", "------")
+            foreach ($r in $data.results) {
+              Write-Host ("{0,-35} {1,10} {2,8}" -f $r.description, $r.wall_time_ms, $r.status)
+            }
+          } else {
+            Write-Host "No performance results found."
+          }
+
+      - name: Upload performance results
+        if: steps.whp-check.outputs.whp_available == 'true' && !cancelled()
+        uses: actions/upload-artifact@v4
+        with:
+          name: microvm-perf-results-${{ github.event.pull_request.number || github.run_number }}
+          retention-days: 30
+          path: microvm-perf-results.json
--- a/.github/workflows/microvm-e2e.yml
+++ b/.github/workflows/microvm-e2e.yml
@@ -72,19 +72,6 @@ jobs:
           npm install
           npm run build
 
-      - name: Sync SDK build into CLI
-        shell: pwsh
-        run: |
-          # Ensure CLI uses the freshly-built SDK (npm may have copied a stale version)
-          $sdkDist = Join-Path $env:GITHUB_WORKSPACE "sdk\dist"
-          $cliSdkDist = Join-Path $env:GITHUB_WORKSPACE "cli\node_modules\@microsoft\mxc-sdk\dist"
-          if (Test-Path $cliSdkDist) {
-            Copy-Item "$sdkDist\*" $cliSdkDist -Recurse -Force
-            Write-Host "Synced SDK dist into CLI node_modules"
-          } else {
-            Write-Host "::warning::CLI SDK dist not found at $cliSdkDist"
-          }
-
       - name: Diagnose hypervisor environment
         shell: pwsh
         run: |
--- a/.github/workflows/microvm-e2e.yml
+++ b/.github/workflows/microvm-e2e.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Build CLI
         working-directory: cli
         run: |
-          npm ci
+          npm install
           npm run build
 
       - name: Diagnose hypervisor environment
--- a/.github/workflows/microvm-e2e.yml
+++ b/.github/workflows/microvm-e2e.yml
@@ -67,7 +67,15 @@ jobs:
         run: |
           $wxcExe = Join-Path $env:GITHUB_WORKSPACE "src\target\x86_64-pc-windows-msvc\debug\wxc-exec.exe"
           $configDir = Join-Path $env:GITHUB_WORKSPACE "test_configs"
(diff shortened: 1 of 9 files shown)


src/wxc_common/src/nanvix_runner.rs
```diff
@@ -249,21 +249,8 @@ impl NanVixScriptRunner {
...
-            || !request.policy.blocked_hosts.is_empty()
-            || request.policy.default_network_policy != NetworkPolicy::Allow
-        {
-            return Err(NanVixError::Preflight(ERR_NETWORK_POLICY.to_string()));
-        }
-        if request.policy.network_proxy.is_enabled() {
-            return Err(NanVixError::Preflight(ERR_PROXY_POLICY.to_string()));
-        }
+        // Filesystem policies are silently ignored — the guest has a read-only ramfs.
+        // Network and proxy policies are also ignored — no network stack in guest.
```
ppenna, line 253: If we have logging system in place, we could log that as debug level.

sdk/src/sandbox.ts
```diff
@@ -70,12 +72,31 @@ export function buildSandboxPayload(
...
+        }
+        return config;
+    }
+
+    // Default path: include filesystem with clearPolicyOnExit
+    config.filesystem = {
+        readwritePaths: policy.filesystem?.readwritePaths,
+        readonlyPaths: policy.filesystem?.readonlyPaths,
+        deniedPaths: policy.filesystem?.deniedPaths,
+        clearPolicyOnExit: true,
```
bbonaby, line 99: issue: I don't see the benefit of splitting this like how it's done here. If your if statement above with the lengths turns out false then the values are still set to null. Same as if the original

```typescript
 config.filesystem = {
        readwritePaths: policy.filesystem?.readwritePaths,
        readonlyPaths: policy.filesystem?.readonlyPaths,
        deniedPaths: policy.filesystem?.deniedPaths,
        clearPolicyOnExit: true,
```

was kept. Am I missing something here?

test_scripts/run_microvm_tests.ps1
```diff
@@ -25,35 +25,14 @@
 
 param(
     [string]$WxcExePath = "..\src\target\debug\wxc-exec.exe",
-    [string]$ConfigDir = "..\test_configs"
+    [string]$ConfigDir = "..\test_configs",
```
bbonaby, line 28: issue: I'd like us to move away from individual scripts for all our tests. For npm we only need a single script that runs `npm test` for cli.

src/wxc_common/src/nanvix_runner.rs
```diff
@@ -249,21 +249,8 @@ impl NanVixScriptRunner {
...
-        if !request.policy.allowed_hosts.is_empty()
-            || !request.policy.blocked_hosts.is_empty()
-            || request.policy.default_network_policy != NetworkPolicy::Allow
-        {
-            return Err(NanVixError::Preflight(ERR_NETWORK_POLICY.to_string()));
-        }
-        if request.policy.network_proxy.is_enabled() {
-            return Err(NanVixError::Preflight(ERR_PROXY_POLICY.to_string()));
-        }
+        // Filesystem policies are silently ignored — the guest has a read-only ramfs.
```
bbonaby, line 252: issue: I still believe we should never silently do anything. We should warn the user that these policies are not supported by nanvix. Think about it this way. If we silently ignore how would GHCP learn that the values it passes to us are never used?

sdk/src/sandbox.ts
```diff
@@ -280,5 +269,123 @@ export function spawnSandboxAsync(
...
+ * const result = await execSandbox(
+ *   "print('Hello from sandbox!')",
+ *   { version: '0.4.0-alpha' },
+ *   { containment: 'microvm', experimental: true }
+ * );
+ * console.log(result.stdout);    // "Hello from sandbox!"
+ * console.log(result.exitCode);  // 0
+ * ```
+ */
+export function execSandbox(
```
bbonaby, line 351: issue: I think this might be the second or third PR you have out with the same information in it. Would you be able to consolidate these PRs into a single one please? I think I've read the same files multiple times now 😔.

That said, I still think this function doesn't need to exist, and a flag should be passed to spawnSandbox instead.

cli/src/cli.ts
```diff
@@ -17,11 +19,46 @@ import { ContainerConfig } from '@microsoft/mxc-sdk/dist/types';
...
-  .name('wxc-cli')
-  .description('CLI test driver for the MXC SDK and container backends')
-  .version('0.1.0');
-
+program
+  .name('wxc-cli')
+  .description('CLI test driver for the MXC SDK and container backends')
+  .version('0.1.0');
+
+program
```
bbonaby, line 27: issue: no need to update this function. We should delete it honestly, since run-sdk will go through the end to end, not just pass json to wxc/lxc exec

.github/workflows/microvm-e2e.yml
```diff
@@ -0,0 +1,157 @@
...
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src
+
+      - name: Build with MicroVM support
+        working-directory: src
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: cargo build --features microvm --target x86_64-pc-windows-msvc
```
bbonaby, line 37: issue: no need to rebuild nanvix again if build.yml builds it already. You can just download the artifact from the pipeline

.github/workflows/microvm-e2e.yml
```diff
@@ -0,0 +1,157 @@
...
+
+      - name: Exclude build output from Windows Defender
+        shell: pwsh
+        run: |
+          $binDir = Join-Path $env:GITHUB_WORKSPACE "src\target\x86_64-pc-windows-msvc\debug"
+          Add-MpPreference -ExclusionPath $binDir
+          Write-Host "Added Defender exclusion for $binDir"
+
+      - name: Verify MicroVM binaries
```
bbonaby, line 46: issue: If someone updated where the Rust code is outputted to I don't think they'd know to update this file. I think we'd need a general README.md file in `src/` with a nanvix section that tells them this.

.github/workflows/microvm-e2e.yml
```diff
@@ -0,0 +1,157 @@
...
+        working-directory: sdk
+        run: |
+          npm ci --ignore-scripts
+          npm run build
+
+      - name: Build CLI
+        working-directory: cli
+        run: |
+          npm install
+          npm run build
```
bbonaby, line 73: issue: no need to do this since all of this is done in build.yml

huzaifa-d:
Old PR. Merged #132

PR #125, opened 2026-04-07
[SDK] Fix: propagate SandboxPolicy.version to ContainerConfig output in SDK

bbonaby:
buildSandboxPayload validated policy.version but never set it on the returned ContainerConfig, causing MXC to receive a versionless config.

- Added version: `policy.version` to the config object in `buildSandboxPayload`
- Made `ContainerConfig.version` a required field (compile-time guard)
- Added unit test for version propagation and empty version rejection
- Added integration tests for missing version, invalid version, and basic appcontainer launch with valid version
- Added npm test step to CLI job in build workflow. NOTE: unfortunately some tests are skipped because they can't run in the pipeline e.g no bfscfg.exe, appcontainer seemly not being able to register the PATH variable :(. But there are some that do run, which is still helpful to confirming sdk functionality is works. 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/125)

### DIFF
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -56,6 +56,7 @@ export function buildSandboxPayload(
     const name = containerName ?? generateRandomContainerName();
 
     const config: ContainerConfig = {
+        version: policy.version,
         process: {
             commandLine: script,
             cwd: workingDirectory,
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -24,14 +24,22 @@ function validatePolicyVersion(version: string): void {
 
     const parsed = semverParse(version);
     if (!parsed) {
-        throw new Error(`Invalid policy version '${version}': must be valid semver (e.g., '0.4.0' or '0.4.0-alpha')`);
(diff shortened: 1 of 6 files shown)


sdk/src/sandbox.ts
```diff
@@ -56,6 +56,7 @@ export function buildSandboxPayload(
     const name = containerName ?? generateRandomContainerName();
 
     const config: ContainerConfig = {
+        version: policy.version,
```
SohamDas2021, line 67: Where are we validating the policy version? I thought we are adding minversion checks?

bbonaby: Ah yes, forgot about our conversation. Will be in next commit. Look for updates to https://github.com/microsoft/mxc/blob/5fdd12a97ea98fb878c8609fde1520431e7f1922/sdk/src/sandbox.ts#L20

SohamDas2021 (approved):

SohamDas2021 (approved):

PR #128, opened 2026-04-08
Add SandboxPolicy spec v1

bbonaby:
Wanted to formalize things now that we've been making a lot of updates. There are still lots to do but I wanted to make sure we all align on what our developer story is for both the folks developing MXC  and folks consuming MXC. 

Note: In this there are changes that would have to be made to make this a reality. If we align on the docs in this then I can create Github Issues for this. 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/128)


docs/sandbox-policy/v1/policy.md
```diff
@@ -0,0 +1,819 @@
...
+
+### 6.1 Environment → Backend Selection
+
+The SDK maps `request.environment.isolation` to a containment backend per OS:
+
+| `request.environment.isolation` | Windows | Linux |
+|-------------|---------|-------|
+| `"process"` (default) | BaseProcessContainer | LXC |
+| `"container"` | WSLC (future) | Docker (future) |
+| `"microvm"` | Hyperlight/NanVix (future) | microVM (future) |
```
ppenna: Please reword to `Nanvix` to match the spelling of the project's name.

docs/sandbox-policy/v1/policy.md
```diff
@@ -0,0 +1,819 @@
...
+2. **No UI policy surface.** Contributors adding UI containment (clipboard, window isolation, input injection) don't
+know where their work fits: SandboxRequest? Config? Both? The UIPolicy_Schema draft is Windows-specific (Job Object UI
+limits, atom tables) but there's no cross-platform abstraction above it.
+
+3. **Unclear boundary between Policy and Config.**
+Contributors repeatedly ask: "Should I update SandboxPolicy or
+Config? How do I know?" There's no decision framework.
+
+4. **New backends are coming fast.** Disposable VMs, microVMs
+(Hyperlight/NanVix), WSLC, and macOS are all planned. Each
```
ppenna: Please reword to `Nanvix` to match the spelling of the project's name.

docs/sandbox-policy/v1/policy.md
```diff
@@ -0,0 +1,819 @@
...
+  filesystem?: { ... };                   // readwritePaths, readonlyPaths, deniedPaths, tempDir
+  network?: { ... };                      // policy, allowedHosts, blockedHosts,
+                                          // proxy, enforcementMode
+  ui?: { ... };                           // allowWindows, clipboard,
+                                          // allowInputInjection,
+                                          // isolation (Win), ime (Win),
+                                          // desktopSystemControl (Win),
+                                          // systemSettings (Win)
+  resources?: { ... };                    // maxMemoryMB, maxCpus
+  leastPrivilege?: boolean;               // Least privilege mode (Win)
```
MGudgin: Let's consider putting platform specific policy in a nested structure (e.g. "extensions" : { "platform": "windows", "integrityLevel", "low" }

MGudgin: And I also worry that integrityLevel is very specific to process based containment and is leaking mechanism somewhat

docs/sandbox-policy/v1/policy.md
```diff
@@ -0,0 +1,819 @@
...
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│ LAYER 1: SDK + SandboxRequest (policy + environment)        │
+│                                                             │
+│ Users: GitHub Copilot, Claude Code, OpenAI, etc.            │
+│                                                             │
+│ SandboxRequest (TypeScript): developer-facing intent        │
+│   policy:      security intent (filesystem, network, ui)    │
+│   environment: runtime selection (isolation, linux distro)  │
```
MGudgin: I worry that this layering/compartmentalization is mainly a post-hoc application of what we've already built. I'm not yet convinced of the split between policy and environment, or that all the things in them are in the right place. Similar concern around policy vs config. That said, I'm generally OK with us 'starting' here, and iterating to get to where we want to be.

docs/sandbox-policy/v1/policy.md
bbonaby: funnily enough my first iteration of all this when I was thinking about it I had an extensions model  like you mentioned. However the more I added the more it started to look like just like the config schema wrapper Stuart talked about yesterday. That's the struggle I guess. 

policy vs config schema for multiple backends versus MacOS just worrying about processes/apps. 

We could theoretically stick with the base policy that provides the filesystem + networking. Then have the user provide what container type they want and we give them a config with it's defaults. They can then update that config how they see fit and/our just call out `spawnSandbox()` api directly with it.

What are your thoughts on that for now?

bbonaby: I simplified this but ended up doing something similar to what I mentioned above. So we're sticking with the policy but allowing dev to get a config from it, to which they can edit themselves. No more policy + environment.

docs/sandbox-policy/v1/policy.md
bbonaby: Agreed. Removed the policy + environment split. Policy is now filesystem, network, UI intent, and timeout. Everything else moved to ContainerConfig. Users can get a backend-specific config via 
 createConfigFromPolicy(policy) and modify it before spawning. Windows-specific UI fields (isolation, ime, etc.) are now under appcontainer.ui in the config, not in policy.

MGudgin (approved):

PR #130, opened 2026-04-08
Rename appcontainer.rs to appcontainer_runner.rs

MGudgin:
Rename module to follow the _runner.rs naming convention for files containing runner classes. Update module declaration in lib.rs and import in main.rs accordingly.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/130)


bbonaby (approved):
I believe the future for this runner is for us to use the [base_container_runner](https://github.com/microsoft/mxc/blob/main/src/wxc_common/src/base_container_runner.rs) @jsidewhite recently added then remove this once all the functionality we need are in there.

PR #134, opened 2026-04-08
[MicroVM] Add SDK & CLI support

huzaifa-d:
Adds SDK/CLI support needed to invoke experimental containment backends (notably MicroVM), including a non-PTY execution path intended for CI/automation where accurate exit codes and separated stdout/stderr are required.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/134)

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -2,10 +2,10 @@
 // Licensed under the MIT License.
 
 
-import { Command, InvalidArgumentError } from 'commander';
+import { Command } from 'commander';
 import {
   spawnSandbox,
-  spawnSandboxWithPty,
+  spawnSandboxWithoutPty,
   getPlatformSupport,
   SandboxPolicy,
   ContainmentType,
@@ -90,14 +90,7 @@ program
   .option('--policy-file <path>', 'Path to a SandboxPolicy JSON file')
   .option('--cwd <path>', 'Working directory for the sandboxed process')
   .option('--container-name <name>', 'Name for the sandbox container')
-  .option('--containment <backend>', 'Override containment backend',
-    (value: string) => {
-      const valid = ['microvm'];
-      if (!valid.includes(value)) {
-        throw new InvalidArgumentError(`Must be one of: ${valid.join(', ')}`);
-      }
-      return value;
-    })
+  .option('--containment <backend>', 'Override containment backend')
   .option('--no-pty', 'Use child_process.spawn instead of node-pty (reliable exit codes)')
   .option('--debug', 'Enable debug output')
   .option('--experimental', 'Enable experimental features')
@@ -160,12 +153,12 @@ program
       };
 
       if (options.pty === false) {
-        // Non-PTY mode using spawnSandbox() (child_process.spawn).
+        // Non-PTY mode using spawnSandboxWithoutPty() (child_process.spawn).
         // This provides reliable exit code propagation and separate stdout/stderr
         // streams. Use this for CI, automation, or any scenario where correct exit
         // codes matter. The default PTY mode (node-pty/ConPTY) returns exit code -1
         // for all processes on Windows due to a known ConPTY limitation.
-        const child = spawnSandbox(scriptCommand, policy, spawnOptions, options.cwd, options.containerName, undefined, containment);
+        const child = spawnSandboxWithoutPty(scriptCommand, policy, spawnOptions, options.cwd, options.containerName, undefined, containment);
 
         child.stdout?.on('data', (data: Buffer) => {
           process.stdout.write(data);
@@ -183,7 +176,7 @@ program
       } else {
         // PTY mode: interactive terminal with colors/input
         console.log('Spawning sandboxed process using SDK...');
-        const ptyProcess = spawnSandboxWithPty(scriptCommand, policy, spawnOptions, options.cwd, options.containerName, undefined, containment);
+        const ptyProcess = spawnSandbox(scriptCommand, policy, spawnOptions, options.cwd, options.containerName, undefined, containment);
 
         ptyProcess.onData((data: string) => {
           process.stdout.write(data);
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -2,12 +2,13 @@
 // Licensed under the MIT License.
 
 
-import { Command } from 'commander';
-import { ContainerExecutor } from './wxc-executor';
+import { Command, InvalidArgumentError } from 'commander';
 import {
   spawnSandbox,
+  spawnSandboxWithPty,
   getPlatformSupport,
   SandboxPolicy,
+  ContainmentType,
   getAvailableToolsPolicy,
 } from '@microsoft/mxc-sdk';
 
@@ -81,7 +82,7 @@ program
 
 program
   .command('run-sdk')
-  .description('Simulate SDK usage: build a sandbox payload from a SandboxPolicy and spawn it')
+  .description('Run a workload in a sandbox using the MXC SDK')
   .option('--script <command>', 'Command line to execute')
   .option('--script-file <path>', 'Path to a script file (contents are read and passed as the command)')
   // Policy JSON should match the SandboxPolicy type defined in sdk/src/types.ts
@@ -89,9 +90,18 @@ program
   .option('--policy-file <path>', 'Path to a SandboxPolicy JSON file')
   .option('--cwd <path>', 'Working directory for the sandboxed process')
   .option('--container-name <name>', 'Name for the sandbox container')
+  .option('--containment <backend>', 'Override containment backend',
+    (value: string) => {
+      const valid = ['microvm'];
+      if (!valid.includes(value)) {
+        throw new InvalidArgumentError(`Must be one of: ${valid.join(', ')}`);
+      }
+      return value;
+    })
+  .option('--no-pty', 'Use child_process.spawn instead of node-pty (reliable exit codes)')
   .option('--debug', 'Enable debug output')
   .option('--experimental', 'Enable experimental features')
-  .action(async (options: { script?: string; scriptFile?: string; policy?: string; policyFile?: string; cwd?: string; containerName?: string; debug?: boolean; experimental?: boolean }) => {
+  .action(async (options: { script?: string; scriptFile?: string; policy?: string; policyFile?: string; cwd?: string; containerName?: string; containment?: string; pty?: boolean; debug?: boolean; experimental?: boolean }) => {
     try {
       let scriptCommand: string;
       if (options.script) {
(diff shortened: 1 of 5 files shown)


sdk/src/sandbox.ts
```diff
@@ -280,5 +269,123 @@ export function spawnSandboxAsync(
...
+    });
+
+    child.on('close', (code: number | null) => {
+      resolve({
+        stdout: Buffer.concat(stdoutChunks).toString(),
+        stderr: Buffer.concat(stderrChunks).toString(),
+        exitCode: code ?? -1,
+      });
+    });
+  });
```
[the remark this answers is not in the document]
huzaifa-d: Added tests.

bbonaby:
FYI: all code in here is the same as #141. If you look at that PR, you have it set to be merged with main, so it is showing all of the changes in this PR + new changes. I was going down the line of PRs so ended up reviewing it first. My suggestion would be to just close this one.

In the future if you want to do stacking you'll need to make sure each subsequent PR is set to be merged into the predecessor branch.

Also in the future we should make sure we have PRs with descriptions.

cli/src/cli.ts
```diff
@@ -1,97 +1,107 @@
...
   .option('--cwd <path>', 'Working directory for the sandboxed process')
   .option('--container-name <name>', 'Name for the sandbox container')
+  .option('--containment <backend>', 'Override containment backend',
+    (value: string) => {
+      const valid = ['microvm'];
+      if (!valid.includes(value)) {
+        throw new InvalidArgumentError(`Must be one of: ${valid.join(', ')}`);
+      }
+      return value;
+    })
```
bbonaby: issue: I believe this can be removed. The SDK itself will throw an error now if the containment backend you pass to it is not supported here:

https://github.com/microsoft/mxc/blob/45b3f00b9b71a9babaebf9fbf0c8fa50d54cf184/sdk/src/sandbox.ts#L214

We can keep the containment option. I'm specifically talking about the parameter validation. The test will fail as a result of the SDK doing the validation itself.

cli/src/cli.ts
```diff
@@ -142,47 +152,74 @@ program
...
+          process.exit(code ?? 1);
+        });
+        child.on('error', (err: Error) => {
+          console.error('Error:', err.message);
+          process.exit(1);
+        });
+      } else {
+        // PTY mode: interactive terminal with colors/input
+        console.log('Spawning sandboxed process using SDK...');
+        const ptyProcess = spawnSandboxWithPty(scriptCommand, policy, spawnOptions, options.cwd, options.containerName, undefined, containment);
```
bbonaby: issue: can we add whether to spawn with Pty as an option in spawnOptions? This way we keep the same spawnSandbox(). Then in spawnSandbox and spawnSandboxFromConfig we'd use the Pty option to know whether to use the pty version or the pipe version. Doing it this way GitHub code doesn't need to change the api they use.

sdk/src/sandbox.ts
```diff
@@ -167,11 +168,37 @@ export function createConfigFromPolicy(
...
         },
-        filesystem: {
-            readwritePaths: [...(policy.filesystem?.readwritePaths ?? [])],
-            readonlyPaths: [...(policy.filesystem?.readonlyPaths ?? [])],
-            deniedPaths: [...(policy.filesystem?.deniedPaths ?? [])],
-        },
+    };
+
+    // Microvm: minimal config with early return — no UI/network mapping needed
+    if (containment === 'microvm') {
```
bbonaby, line 205: issue: can we do the same thing that we do with process containers at the bottom of this function where we return the result of the function. See: https://github.com/microsoft/mxc/blob/45b3f00b9b71a9babaebf9fbf0c8fa50d54cf184/sdk/src/sandbox.ts#L207


I'd think we'd have something like:

```typescript
else if  if (containment === 'microvm') {
    return buildMicroVmConfig(config);
}
```

Then this logic just lives in there. The early return doesn't really save us much in the grand scheme of things. From a readability  (of the whole function) and maintenance standpoint I believe this option is better.

sdk/src/sandbox.ts
```diff
@@ -214,31 +241,33 @@ export function createConfigFromPolicy(
...
+ * @param containerName Optional container name; if not provided, a random name will be generated
+ * @param containment Optional containment backend type
+ * @returns The sandbox payload object
+ */
 export function buildSandboxPayload(
     script: string,
     policy: SandboxPolicy,
     workingDirectory?: string,
     containerName?: string,
+    containment?: ContainmentType,
```
bbonaby: suggestion: I believe you can just make "process" a default arg like:

```typescript
export function buildSandboxPayload(
    script: string,
    policy: SandboxPolicy,
    workingDirectory?: string,
    containerName?: string,
    containment?: ContainmentType = "process",
): ContainerConfig
```

then you don't have to have the `containment ?? "process"` below, we'd just pass in `containment`

sdk/src/sandbox.ts
```diff
@@ -282,14 +310,27 @@ function spawnWithConfig(
     throw new Error(`MXC is not supported on this platform: ${platformSupport.reason}`);
   }
 
+  // Validate containment against platform
+  if (config.containment) {
+    if (config.containment === 'microvm' && os.platform() !== 'win32') {
+      throw new Error('The microvm backend is only supported on Windows (requires WHP/Hyper-V).');
+    }
+    const experimentalBackends: string[] = ['microvm'];
```
bbonaby: issue: we should move this to types.ts like how we have this https://github.com/microsoft/mxc/blob/45b3f00b9b71a9babaebf9fbf0c8fa50d54cf184/sdk/src/types.ts#L39 but it would be a list like you have above, just located in types.ts for use everywhere in the sdk

sdk/src/sandbox.ts
```diff
@@ -317,52 +358,123 @@ function spawnWithConfig(
     args.push('--debug');
   }
 
-  if (options.experimental) {
+  if (options.experimental || config.containment === 'microvm') {
```
bbonaby: question: why is `config.containment === 'microvm'` needed? Do we need to add this to here https://github.com/microsoft/mxc/blob/45b3f00b9b71a9babaebf9fbf0c8fa50d54cf184/sdk/src/types.ts#L159

Might have to confirm with Soham, how he envisioned experimental features to be integrated.

sdk/src/sandbox.ts
```diff
@@ -317,52 +358,123 @@ function spawnWithConfig(
...
+ * const script = 'python -c "import sys; print(sys.version)"';
+ * const policy: SandboxPolicy = { version: '0.4.0-alpha' };
+ *
+ * const ptyProcess = spawnSandboxWithPty(script, policy);
+ * ptyProcess.onData((data) => console.log(data));
+ * ptyProcess.onExit((e) => console.log('Exit code:', e.exitCode));
  * ```
  */
-export function spawnSandbox(
+export function spawnSandboxWithPty(
```
bbonaby: issue: see my comments in the cli.ts, GitHub copilot cli currently uses `spawnSandbox` which uses pty by default. It would be better if we passed what kind of process interaction the caller wants via the SandboxSpawnOptions  with the default being to use `Pty`. I'd just keep in mind that we currently have 2 public entry points to spawn a sandbox and probably only need these two for now. `spawnSandbox()` which makes a process container only and `spawnSandboxFromConfig()` that allows the caller to specify the containment type they want, then update the produced config to fit their needs. How you augment them to support the piping  vs pty behavior is up to you but I feel like the easiest way would be through `SandboxSpawnOptions`.

cli/src/cli.ts
huzaifa-d: The return type is different. I need the child process. But I changed it to have a spawnSandboxWithoutPTY version.

sdk/src/sandbox.ts
```diff
@@ -315,8 +326,8 @@ function resolveExecutableAndArgs(
     if (config.containment === 'microvm' && os.platform() !== 'win32') {
       throw new Error('The microvm backend is only supported on Windows (requires WHP/Hyper-V).');
     }
-    const experimentalBackends: string[] = ['microvm'];
-    if (!experimentalBackends.includes(config.containment) &&
+    const experimentalList: readonly string[] = ExperimentalBackends;
```
bbonaby: suggestion: Now you can use `ExperimentalBackends` directly below

sdk/src/sandbox.ts
SohamDas2021: This is what I am doing for WSLC-  https://github.com/microsoft/mxc/pull/186/changes#diff-54eb8cc1bda7e2440c49a8a43ee30e2e1158079e40b22864468a8771ca47fa23:~:text=if%20(options.containment,%7D

if containment has value of a new experimental feature, we fail if the experimental flag is not there.

sdk/src/index.ts
```diff
@@ -29,22 +29,26 @@
...
+  findWxcExecutable,
 } from './platform';
 
 // Export sandbox spawning functions
 export {
   createConfigFromPolicy,
   spawnSandbox,
+  spawnSandboxWithoutPty,
   spawnSandboxFromConfig,
-  spawnSandboxAsync,
```
bbonaby, line 46: issue: Oh I missed this my first go around. We want to keep this public API.

sdk/src/index.ts
```diff
@@ -29,22 +29,26 @@
...
+  ContainmentType,
+  ExperimentalBackends,
   ContainerConfig,
   PlatformSupport,
 } from './types';
 
 // Export platform detection functions
 export {
   getPlatformSupport,
+  findWxcExecutable,
```
bbonaby: issue: We don't want to expose this api. I don't think we're using it directly in the cli in this PR so we can remove this.

sdk/src/sandbox.ts
```diff
@@ -317,38 +369,58 @@ function spawnWithConfig(
...
+  const { executablePath, args } = resolveExecutableAndArgs(config, options);
+
   const ptyOpts: pty.IPtyForkOptions = {
     name: "xterm-color",
     cols: 120,
     rows: 80,
-    cwd: workingDirectory || process.cwd(),
-    env: env
+    ...options.ptyOptions,
+    cwd: workingDirectory || options.ptyOptions?.cwd || process.cwd(),
```
bbonaby: suggestion: not sure if this is an issue, but to keep the previous behaviour should we swap the order of `process.cwd()` and   `options.ptyOptions?.cwd`

sdk/src/sandbox.ts
```diff
@@ -391,58 +514,3 @@ export function spawnSandboxFromConfig(
...
- *   version: '0.4.0-alpha',
- *   filesystem: { readwritePaths: ['/workspace'] },
- * };
- *
- * const result = await spawnSandboxAsync('echo hello', policy);
- * console.log('Output:', result.stdout);
- * console.log('Exit code:', result.exitCode);
- * ```
- */
-export function spawnSandboxAsync(
```
bbonaby, line 419: note: This is what I mentioned above, I think a previous change removed this in your commits and we'd want it back.

sdk/src/sandbox.ts
```diff
@@ -357,12 +435,63 @@ export function spawnSandbox(
...
+ * const child = spawnSandboxWithoutPty(
+ *   "print('Hello from sandbox!')",
+ *   { version: '0.4.0-alpha' },
+ *   { experimental: true }
+ * );
+ * child.stdout?.on('data', (data) => console.log(data.toString()));
+ * child.on('close', (code) => console.log('Exit code:', code));
+ * ```
+ */
+export function spawnSandboxWithoutPty(
```
bbonaby, line 472: note: For what it's worth we're taking the `SandboxPolicy` as an arg which is just filesystem, networking, ui and using that as is. No backend specific information like the `spawnSandboxFromConfig` function that allows folks to pass in a config object with  backend specific fields updated. 

Everything that can't be expressed in the SandboxPolicy, is just set to their default values in the config (which could be deny by default). 

That's why the `spawnSandbox()` function is only for process containers since it doesn't have back end specific options. That plus github copilot is only using it right now, and they're spinning up process containers.

All that to say if you think the right approach for this function is to do what `spawnSandboxFromConfig` does and take in a Config file we should do that since it will allow consumers to use this non Pty version for the backend of their choice

sdk/src/sandbox.ts
```diff
@@ -357,12 +434,63 @@ export function spawnSandbox(
   options: SandboxSpawnOptions = {},
   workingDirectory?: string,
   containerName?: string,
-  env?: { [key: string]: string | undefined }
+  env?: { [key: string]: string | undefined },
+  containment?: ContainmentType,
```
bbonaby, line 438: note: Just an FYI, not sure if this was missed in your last commit but this addition will allow someone to spawn a `microvm/nanvix vm` using `pty` if they pass in `microvm` to this function using the `containment` parameter you added to it. Circumventing the use of your new `spawnSandboxWithoutPty` api. 

Not sure if that was intended or not.

bbonaby (approved):

PR #137, opened 2026-04-08
refactor: standardize wide-string conversion via string_util::to_wide

MGudgin:
Replace manual encode_utf16().chain(once(0)).collect() patterns in appcontainer.rs and process_util.rs with calls to the existing string_util::to_wide() helper for consistency and reduced duplication.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/137)

### DIFF
--- a/src/wxc_common/src/appcontainer_runner.rs
+++ b/src/wxc_common/src/appcontainer_runner.rs
@@ -341,11 +341,7 @@ impl AppContainerScriptRunner {
         };
 
         // --- Build command line ---
-        let mut cmd_line_wide: Vec<u16> = request
-            .script_code
-            .encode_utf16()
-            .chain(std::iter::once(0))
-            .collect();
+        let mut cmd_line_wide = string_util::to_wide(&request.script_code);
 
         let working_dir_wide = string_util::to_wide(&request.working_directory);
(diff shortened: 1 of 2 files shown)


src/wxc_common/src/appcontainer_runner.rs
```diff
@@ -341,11 +341,7 @@ impl AppContainerScriptRunner {
         };
 
         // --- Build command line ---
-        let mut cmd_line_wide: Vec<u16> = request
-            .script_code
-            .encode_utf16()
-            .chain(std::iter::once(0))
-            .collect();
+        let mut cmd_line_wide = string_util::to_wide(&request.script_code);
```
bbonaby, line 344: In a follow up we should make the `string_util` functions use https://crates.io/crates/widestring under the hood. I'll make a GitHub issue when I can.

bbonaby (approved):

src/wxc_common/src/appcontainer_runner.rs
MGudgin: Was this what you had in mind https://github.com/microsoft/mxc/pull/140?

PR #141, opened 2026-04-09
[MicroVM] Add CLI E2E tests

huzaifa-d:
Adds CLI-based E2E tests for the MicroVM backend. Stacked on #134.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/141)

### DIFF
--- a/.github/workflows/microvm-e2e.yml
+++ b/.github/workflows/microvm-e2e.yml
@@ -95,3 +95,92 @@ jobs:
           name: microvm-perf-results-${{ github.event.pull_request.number || github.run_number }}
           retention-days: 30
           path: microvm-perf-results.json
+
+  microvm-cli-e2e:
+    name: CLI MicroVM
+    runs-on: windows-latest
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Rust toolchain
+        run: |
+          rustup update stable
+          rustup target add x86_64-pc-windows-msvc
+
+      - name: Cache Rust build artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src
+
+      - name: Build with MicroVM support
+        working-directory: src
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: cargo build --features microvm --target x86_64-pc-windows-msvc
+
+      - name: Exclude build output from Windows Defender
+        shell: pwsh
+        run: |
+          $binDir = Join-Path $env:GITHUB_WORKSPACE "src\target\x86_64-pc-windows-msvc\debug"
+          Add-MpPreference -ExclusionPath $binDir
+          Write-Host "Added Defender exclusion for $binDir"
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Build SDK
+        working-directory: sdk
+        run: |
+          npm ci
+          npm run build
+
+      - name: Build CLI
+        working-directory: cli
+        run: |
+          npm install
+          npm run build
+
+      - name: Diagnose hypervisor environment
+        shell: pwsh
+        run: ./scripts/ci/diagnose-whp.ps1
+
+      - name: Check Windows Hypervisor Platform
+        id: whp-check
+        shell: pwsh
+        run: ./scripts/ci/check-whp.ps1
+
+      - name: Run MicroVM CLI E2E Tests
+        if: steps.whp-check.outputs.whp_available == 'true'
+        shell: pwsh
+        working-directory: test_scripts
+        run: |
+          $cliJs = Join-Path $env:GITHUB_WORKSPACE "cli\dist\cli.js"
+          .\run_microvm_cli_tests.ps1 -CliPath $cliJs
+
+      - name: Upload logs on failure
+        if: failure() || cancelled()
+        uses: actions/upload-artifact@v6
+        with:
+          name: microvm-cli-e2e-logs-${{ github.event.pull_request.number || github.run_number }}
+          retention-days: 7
+          path: |
+            logs/
+            **/*.log
+
+      - name: Print performance summary
+        if: steps.whp-check.outputs.whp_available == 'true' && !cancelled()
+        shell: pwsh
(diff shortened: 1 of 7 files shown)


.github/workflows/microvm-e2e.yml
```diff
@@ -95,3 +95,92 @@ jobs:
...
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src
+
+      - name: Build with MicroVM support
+        working-directory: src
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: cargo build --features microvm --target x86_64-pc-windows-msvc
```
bbonaby, line 122: Issue: I think we should use the artifacts that are built from the build.yml action. No need to build again. Also are we sure that these tests don't need to run on arm64?

.github/workflows/microvm-e2e.yml
```diff
@@ -95,3 +95,92 @@ jobs:
           name: microvm-perf-results-${{ github.event.pull_request.number || github.run_number }}
           retention-days: 30
           path: microvm-perf-results.json
+
```
bbonaby, line 98: Thought: We need to add the nanvix build to the 1ES azure pipelines. See: https://github.com/microsoft/mxc/blob/main/.azure-pipelines/README.md. The azure pipeline is where the official build will happen. These GitHub workflows are just for pull requests but eventually my plan is to move them over to the 1ES pipelines. The 1ES pipeline has things like SBOM generation etc out the box. We will need to eventually start running our tests on machines that contain all tessera bits. I'll follow up on moving stuff to 1ES.

cli/src/cli.ts
```diff
@@ -81,17 +83,25 @@ program
 
 program
   .command('run-sdk')
-  .description('Simulate SDK usage: build a sandbox payload from a SandboxPolicy and spawn it')
+  .description('Run a sandboxed process from a SandboxPolicy using the MXC SDK')
```
bbonaby, line 86: issue: "Sandboxed process" would only be for the BaseProcessContainer/Appcontainer option today, whereas you're now adding hyperlight/nanvix  which technically isn't running in a sandbox. Just call it "run a workload in a sandbox using the MXC SDK"

sdk/src/sandbox.ts
```diff
@@ -70,12 +72,26 @@ export function buildSandboxPayload(
...
+        // Microvm uses its own ramfs; host filesystem policies are not supported.
+        if (containment === 'microvm') {
+            return config;
+        }
+        // Other backends (appcontainer, lxc, etc.) fall through to
+        // platform-specific setup below
+    }
+
+    // Default path: include filesystem with clearPolicyOnExit
+    config.filesystem = {
```
bbonaby, line 90: issue: We shouldn't silently ignore what the user passed in. I know you have a hard error in wxc-exec about the user providing  file paths to nanvix, so the better thing to do here is just have a microvm function. If platform is Windows we can check if there are policies nanvix doesn't support like filesystem/networking etc and then surface the error. What we have now will have the users thinking that either the sdk is broken because we don't pass along the policies to nanvix or worse that these policies are in affect for nanvix.

sdk/src/sandbox.ts
```diff
@@ -280,5 +265,138 @@ export function spawnSandboxAsync(
...
+): { executablePath: string; args: string[] } {
+  const platformSupport = getPlatformSupport();
+  if (!platformSupport.isSupported) {
+    throw new Error(`MXC is not supported on this platform: ${platformSupport.reason}`);
+  }
+
+  // Validate containment against available methods (skip experimental backends
+  // like microvm which aren't reported by platform detection)
+  if (options.containment) {
+    const experimental = ['microvm'];
```
bbonaby, line 290: suggestion: lets make a new folder called features under sdk/src and create a new typescript file called experimental. In there we can have a list of experimental features. Soham and I want to formalize how we handle experimental features in the sdk and adding this small part will help towards keeping things associated with how we handle experimental features all in one place. 

For example all of this could be a encapsulated in a helper function called `isExperimental` n that file. sandbox.ts doesn't need to have/know about what we do to handle experimental features.

sdk/src/sandbox.ts
```diff
@@ -168,66 +190,29 @@ export interface SandboxSpawnOptions {
...
-  };
-
-  const ptyProcess = pty.spawn(executablePath, args, ptyOpts);
-  return ptyProcess;
+export function spawnSandbox(
+  script: string,
+  policy: SandboxPolicy,
+  options: SandboxSpawnOptions = {},
+  workingDirectory?: string,
+  containerName?: string,
```
bbonaby, line 198: see: https://github.com/microsoft/mxc/blob/main/docs/sandbox-policy/v1/policy.md#two-api-paths the plan is not to plumb containment into `spawnSandbox`. We will offer a `createConfigFromPolicy` method that will take the containment method and return a typescript config object for that containment method without it's default values set. I have not made that change yet, but will be doing that soon please hold off on this PR until then if you can.

sdk/src/sandbox.ts
```diff
@@ -280,5 +265,138 @@ export function spawnSandboxAsync(
...
+ * const result = await execSandbox(
+ *   "print('Hello from sandbox!')",
+ *   { version: '0.4.0-alpha' },
+ *   { containment: 'microvm', experimental: true }
+ * );
+ * console.log(result.stdout);    // "Hello from sandbox!"
+ * console.log(result.exitCode);  // 0
+ * ```
+ */
+export function execSandbox(
```
bbonaby, line 360: thought: is this function needed? Why not allow the developer to pass in a `useNodePty` flag. when set we use node-pty and when not set we use the std/in/out/err pipes. does node-pty not work with nanvix? if so we need an error for that.

test_scripts/run_microvm_cli_tests.ps1
```diff
@@ -0,0 +1,222 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+.SYNOPSIS
```
bbonaby, line 5: issue: please add your tests directly to the cli project here: https://github.com/microsoft/mxc/blob/main/cli/src/cli.test.ts so they can be run with npm test all in one place.

huzaifa-d:
Abandoned since we will be removing the SDK CLI

PR #151, opened 2026-04-10
[LXC] Add CLI e2e tests for LXC

SohamDas2021:
Adds end-to-end tests that validate the full CLI → SDK → lxc-exec → LXC container path on Linux.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/151)


test_scripts/cli_e2e_tests/run_lxc_cli_tests.sh
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
```
bbonaby: issue: I think these tests can go directly in the cli project itself actually, we can make then only run if the platform is Linux. This way we don't need another script and then we can just run `npm test` and all the npm tests will run in the cli directory

SohamDas2021: The reason I avoided that is because the LXC tests need root (sudo) access. But I see your point, will move.

cli/src/cli.test.ts
```diff
@@ -195,3 +195,108 @@ describe.skip('SDK proxy end-to-end', () => {
     assert.ok(output.includes('Proxy policy active'), 'Expected proxy policy to be set up');
   });
 });
+
+// LXC end-to-end tests — only run on Linux with root privileges
+const isLinux = os.platform() === 'linux';
+const isRoot = isLinux && process.getuid?.() === 0;
```
bbonaby: suggestion: maybe better if we say `isLinuxRoot`

cli/src/cli.test.ts
```diff
@@ -37,7 +37,7 @@ function writeTempPolicy(dir: string, policy: object): string {
   return filePath;
 }
 
-describe('SDK end-to-end', () => {
+describe('SDK end-to-end', { skip: os.platform() !== 'win32' ? 'AppContainer tests require Windows' : undefined }, () => {
```
bbonaby: suggestion: we can probably say "Windows Process Container tests can only be ran on Windows". The rename these tests to `Windows Process Container end-to-end` or remove the end-to-end part.

bbonaby (approved):

bbonaby (approved):

PR #152, opened 2026-04-10
Fixing some discovered bugs.

shschaefer:
### [ISSUE 1] `tokio::time::Instant` subtraction panics when past deadline
- **Confidence:** Confirmed
- **Category:** Runtime panic / denial of service
- **Lines:** `wxc_sandbox_daemon/src/tcp_bridge.rs:86`, `wxc_sandbox_daemon/src/tcp_bridge.rs:285`

`tokio::time::Instant` subtraction (`deadline - Instant::now()`) panics if `now() > deadline`. The code computes `remaining` at the *top* of a loop — after a `read()` call that may return data just before the deadline expires. On the next loop iteration the deadline has been crossed and the subtraction panics, crashing the daemon's client-handling task.

### [ISSUE 2] Idle-timeout `process::exit(0)` bypasses sandbox VM teardown
- **Confidence:** Confirmed
- **Category:** Resource leak / VM leak
- **Lines:** `wxc_sandbox_daemon/src/main.rs:101`

The idle watchdog calls `std::process::exit(0)` to shut down the daemon. This terminates the process immediately, bypassing all async cleanup — including the sandbox VM teardown at lines 75–81 of `main()`. The Windows Sandbox VM (Hyper-V backed) continues running after the daemon exits, consuming CPU, memory, and a Hyper-V slot until manually killed.

### [ISSUE 3] `read_from_pipe` mixes byte-length and character-count for output cap
- **Confidence:** High-confidence
- **Category:** Incorrect truncation logic
- **Lines:** `wxc_common/src/process_util.rs:172-183`

`MAX_OUTPUT_CHARS` (line 19) is named and documented as a character limit (1 MB of characters). However, the remaining-budget computation at line 172 uses `result.len()` which returns the **byte** length of the String, not the character count:

### [ISSUE 4] `encode_message` truncates frame length for payloads > 4 GB
- **Confidence:** Needs-domain-check
- **Category:** Integer truncation / protocol corruption
- **Lines:** `wxc_common/src/sandbox_protocol.rs:72`

`encode_message` casts `json.len()` to `u32` without bounds checking:
```rust
let len = json.len() as u32;
```
If the serialized JSON exceeds `u32::MAX` bytes (~4 GB), `len` is silently truncated. The 4-byte frame header then advertises a shorter length than the actual payload. The decoder reads only `len` bytes of the payload, attempts JSON deserialization on the truncated data, and fails with a parse error — or worse, if the truncated data happens to be valid JSON, processes a corrupted message.


MGudgin (approved):

src/wxc_common/src/sandbox_protocol.rs
```diff
@@ -69,6 +70,11 @@ pub struct ExitNotification {
 /// Layout: `[len: u32 LE][json: len bytes]`
 pub fn encode_message(msg: &ControlMessage) -> Result<Vec<u8>, serde_json::Error> {
     let json = serde_json::to_vec(msg)?;
+    if json.len() > u32::MAX as usize {
```
bbonaby, line 73: oh good catch

bbonaby (approved):

huzaifa-d (approved):

PR #154, opened 2026-04-10
Add PowerShell directory discovery to SDK getAvailableToolsPolicy

MGudgin:
When PowerShell (pwsh.exe) is detected on PATH, getAvailableToolsPolicy()
now includes the extra filesystem paths that PowerShell needs to function
inside a sandboxed container:

- C:\ added to readonlyPaths so pwsh.exe can enumerate the drive root
  during startup.
- %USERPROFILE%\AppData\Roaming\Microsoft\Windows\PowerShell\PSReadLine
  added to readwritePaths so the PSReadLine module can persist command
  history.

This is implemented as a private getPowerShellPolicy() helper in
sdk/src/policy.ts that is called by getAvailableToolsPolicy(). Detection
checks whether any PATH directory contains pwsh.exe (machine-level
discovery, not command-line-level). Windows-only; returns empty policy on
other platforms.

Includes five SDK tests covering:
- C:\ and PSReadLine path injection when pwsh.exe is on PATH
- No injection when pwsh.exe is absent from PATH
- No injection on non-Windows platforms
- No PSReadLine path when USERPROFILE is unset

### DIFF
--- a/sdk/src/policy.test.ts
+++ b/sdk/src/policy.test.ts
@@ -1,0 +1,108 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import { describe, it, afterEach } from 'node:test';
+import assert from 'node:assert';
+import { getAvailableToolsPolicy } from './policy';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+describe('getAvailableToolsPolicy - PowerShell discovery', () => {
+    let originalPlatform: PropertyDescriptor | undefined;
+    let tmpDir: string | undefined;
+
+    const mockWindows = () => {
+        originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform');
+        Object.defineProperty(process, 'platform', { value: 'win32' });
+    };
+
+    const mockLinux = () => {
+        originalPlatform = Object.getOwnPropertyDescriptor(process, 'platform');
+        Object.defineProperty(process, 'platform', { value: 'linux' });
+    };
+
+    /** Create a temp directory containing a fake pwsh.exe and return its path. */
+    const createFakePwshDir = (): string => {
+        tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'mxc-test-'));
+        fs.writeFileSync(path.join(tmpDir, 'pwsh.exe'), '');
+        return tmpDir;
+    };
+
+    afterEach(() => {
+        if (originalPlatform) {
+            Object.defineProperty(process, 'platform', originalPlatform);
+            originalPlatform = undefined;
+        }
+        if (tmpDir) {
+            fs.rmSync(tmpDir, { recursive: true, force: true });
+            tmpDir = undefined;
+        }
+    });
+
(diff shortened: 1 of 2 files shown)


shschaefer (dismissed):
This code should be done in the SDK.  There is a method called getAvailableToolsPolicy().  Having MXC augment policy will be problematic when we go to validate, enforce policy, etc...

MGudgin:
> This code should be done in the SDK. There is a method called getAvailableToolsPolicy(). Having MXC augment policy will be problematic when we go to validate, enforce policy, etc...

PR updated, although I'm not convinced I like getAvailableToolsPolicy in general. It's not really 'least privilege' if we always add all the paths regardless of what the script is actually running.

sdk/src/policy.ts
```diff
@@ -185,13 +185,66 @@ function deduplicatePaths(paths: string[]): string[] {
...
+        } catch {
+            return false;
+        }
+    });
+
+    if (!pwshFound) {
+        return { readonlyPaths: [], readwritePaths: [] };
+    }
+
+    const readonlyPaths: string[] = ['C:\\'];
```
bbonaby: suggestion: I'd probably just do something like this:

```typescript
const systemDrive = process.env.SystemDrive || 'C:';
const systemRoot = systemDrive + '\\';
const readonlyPaths: string[] = [systemRoot ];
```

extremely likely that that environment variable will be set in all Windows machines. This way incase Windows is installed on the D drive or something we have that case covered. Then on the off chance it's not we can fallback on C:.

sdk/src/policy.ts
```diff
@@ -185,13 +185,66 @@ function deduplicatePaths(paths: string[]): string[] {
...
+    pathDirs: string[],
+    env: { [key: string]: string | undefined },
+): FilesystemPolicyResult {
+    if (os.platform() !== 'win32') {
+        return { readonlyPaths: [], readwritePaths: [] };
+    }
+
+    const pwshFound = pathDirs.some(dir => {
+        try {
+            return fs.existsSync(path.join(dir, 'pwsh.exe'));
```
bbonaby, line 217: question: To confirm `PowerShell 5.1` (powershell.exe) is all good right? Only issue is the undocked PowerShell e.g v6+?

sdk/src/policy.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    afterEach(() => {
+        restore();
+    });
+
+    it('should add C:\\ to readonlyPaths when pwsh.exe is on PATH', () => {
+        mockWindows();
+        // Use the real PowerShell directory from the current machine's PATH
+        const pwshDir = findPwshDir();
+        if (!pwshDir) {
```
bbonaby: suggestion: Since these are unit tests we can just mock having powershell in the path right?

sdk/src/policy.ts
MGudgin: I don't actually know the answer to this question.

asklar: neither is "all good". Modules folder needs to be available to both. However, not having the modules folder specified is only a problem after you circumvent the C:\ root ACL issue, which only affects pwsh, not Windows Powershell.

bbonaby (approved):

PR #161, opened 2026-04-13
SDK: implement Sandbox Policy Spec v1

bbonaby:
To update SDK to align closer to the Sandbox Policy spec I sent out last week. This will facilitate subsequent PRs that will plumb both UI and networking down into the `CreateProcessInSandbox` OS function.

### What changed?
- Added `createConfigFromPolicy()` as another primary API more advanced users can use. 
- Added SandboxPolicy fields: ui and networking allowedHosts/blockedHosts
- Add UiConfig, BaseProcessUiConfig, ContainmentType types to SDK
- Added `spawnSandboxFromConfig()` so users can pass in a config file based on our discussions
- Network policy filled centrally, backends add specifics

### How was this tested?
- Added unit tests
- Confirmed cli tests still run. Another change will be added with more test updates.

-----
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/161)

### DIFF
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -5,7 +5,7 @@ import * as pty from 'node-pty';
 import * as os from 'os';
 import { randomBytes } from "crypto";
 import { parse as semverParse } from 'semver';
-import { SandboxPolicy, ContainerConfig } from './types';
+import { SandboxPolicy, ContainerConfig, ContainmentType } from './types';
 import { findWxcExecutable, findLxcExecutable, getPlatformSupport } from './platform';
 
 const SUPPORTED_VERSION = '0.5.0-alpha';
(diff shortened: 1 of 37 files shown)


sdk/src/sandbox.ts
```diff
@@ -55,87 +55,169 @@ function validatePolicyVersion(version: string): void {
...
+ *
+ * @param policy - The sandbox policy expressing security intent
+ * @param containment - Containment backend type (default: "process")
+ * @param containerName - Optional container name; auto-generated if omitted
+ * @returns A fully populated ContainerConfig ready for modification or spawning
+ *
+ * @example
+ * ```typescript
+ * const policy: SandboxPolicy = {
+ *   version: '0.4.0-alpha',
```
SohamDas2021: is ui policy is a stable 0.4.0 feature? To me it is a promoted feature and we should use 0.5.0? Typically, 0.4.0 should also work for promoted features with experimental, but I recommend 0.5.0 to clear confusion.

sdk/src/sandbox.ts
```diff
@@ -55,87 +55,169 @@ function validatePolicyVersion(version: string): void {
...
+        distribution: 'alpine',
+        release: '3.23',
+        destroyOnExit: true,
+    };
+
+    return config;
+}
+
+/**
+ * Builds the Windows process container (BaseProcessContainer) portion of a ContainerConfig.
```
jsidewhite: >BaseProcessContainer [](http://example.com/codeflow?start=41&length=20)

I don't think this term is used anywhere else.  it's BaseContainer, right?

sdk/src/sandbox.ts
```diff
@@ -55,87 +55,169 @@ function validatePolicyVersion(version: string): void {
...
-            cwd: workingDirectory,
+            commandLine: '',
+            timeout: policy.timeoutMs ?? 0,
         },
-        containerId: name,
         filesystem: {
-            readwritePaths: policy.filesystem?.readwritePaths,
-            readonlyPaths: policy.filesystem?.readonlyPaths,
-            deniedPaths: policy.filesystem?.deniedPaths,
-            clearPolicyOnExit: true,
```
SohamDas2021, line 90: clearPolicyOnExit removed from filesystem config — is this intentional?

sdk/src/sandbox.ts
bbonaby: So I ended up using BaseProcessContainer because BaseContainer is ambiguous if MXC is supporting all kinds of containers not just process.

sdk/src/sandbox.ts
```diff
@@ -160,110 +242,150 @@ export interface SandboxSpawnOptions {
...
- *
- * const result = await spawnSandboxAsync(script, policy);
- * console.log('Output:', result.stdout);
- * console.log('Exit code:', result.exitCode);
- * ```
+/**
+ * Internal helper: resolves the executor binary path and spawns a PTY process.
+ */
+function spawnWithConfig(
+  script: string,
```
SohamDas2021: unused?

sdk/src/sandbox.ts
jsidewhite: then buildProcessBaseContainer

sdk/src/sandbox.ts
```diff
@@ -55,87 +55,169 @@ function validatePolicyVersion(version: string): void {
...
-        if (policy.network) {
-            config.network = {
-                defaultPolicy: policy.network.allowOutbound ? 'allow' : 'block',
-                enforcementMode: 'firewall',
-            };
-        }
-
-        return config;
+        config.network = {
+            defaultPolicy: policy.network.allowOutbound ? 'allow' : 'block',
```
SohamDas2021, line 190: Do we need to add validation here for allowOutbound? Throw an error if allowedHosts or blockedHosts are set but allowOutbound is block?

src/wxc_common/src/base_container_runner.rs
```diff
@@ -303,8 +306,34 @@ impl ScriptRunner for BaseContainerRunner {
     fn run(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
         let _ = writeln!(logger, "BaseContainer: building sandbox specification...");
 
+        // Launch builtin test proxy if requested (before building spec so we have the port).
+        let mut request = request.clone();
+        if request.policy.network_proxy.builtin_test_server {
```
jsidewhite, line 328: >builtin_test_server [](http://example.com/codeflow?start=40&length=19)

I see it's already in other places - but I think it's not the right move to have `builtin_test_server` in the schema.

sdk/src/sandbox.ts
```diff
@@ -55,87 +55,169 @@ function validatePolicyVersion(version: string): void {
...
         version: policy.version,
+        containerId,
+        lifecycle: {
+            destroyOnExit: true,
+            preservePolicy: false,
+        },
         process: {
-            commandLine: script,
-            cwd: workingDirectory,
+            commandLine: '',
```
SohamDas2021, line 162: What happens today if we pass an empty string as script to downlevel?

jsidewhite (approved):
:shipit:

jsidewhite:
>         },

Are we okay with keeping this in if it's not supported in OS?

---
Refers to: schemas/dev/mxc-config.schema.0.5.0-dev.json:187 in 864b9b3. [](commit_id = 864b9b3ea7680e10c7826b35471af2774aac4385, deletion_comment = False)

bbonaby:
>         },

You mean leastPriviledge part? 

---
In reply to: [4255984786](https://github.com/microsoft/mxc/pull/161#issuecomment-4255984786) [](http://example.com/codeflow?ancestors=4255984786)

---
Refers to: schemas/dev/mxc-config.schema.0.5.0-dev.json:187 in 864b9b3. [](commit_id = 864b9b3ea7680e10c7826b35471af2774aac4385, deletion_comment = False)

sdk/src/sandbox.ts
bbonaby: we would throw an error in the rust code but now I've made it so we throw an error earlier from the SDK

sdk/src/sandbox.ts
bbonaby: updated

sdk/src/sandbox.ts
bbonaby: The 0.4.0/0.5.0 schemas don't have clearPolicyOnExit in filesystem. They use lifecycle.preservePolicy and lifecycle.destroyOnExit instead. We never actually passed either lifecycle field from the SDK to wxc-exec; the Rust code just defaulted preservePolicy = false. 
 
For backward compat with GHCP, I've kept filesystem.clearPolicyOnExit in the SandboxPolicy input, but now it properly drives lifecycle.preservePolicy in the
 output config. We'll deprecate clearPolicyOnExit from filesystem in a future version and move it to the top-level policy.

sdk/src/sandbox.ts
bbonaby: updated

sdk/src/sandbox.ts
bbonaby: removed

sdk/src/sandbox.ts
bbonaby: updated

SohamDas2021 (approved):

src/wxc_common/src/base_container_runner.rs
bbonaby: spoke offline, in the future we'll look to remove this from the schema, and instead provide the proxy exe and instructions on how to use.

PR #163, opened 2026-04-14
Allow MXC to accept 0.5.0-alpha jsons

SohamDas2021:
We are allowing MXC to accept 0.5.0-alpha jsons. So, with this change, MXC accepts the following:
 - `0.4.0-alpha` without `--experimental` → stable features only
 - `0.5.0-alpha` without `--experimental` → stable + promoted features
 - `0.5.0-alpha` with `--experimental` → stable + promoted + experimental features
 - `0.4.0-alpha` with `--experimental` → stable + experimental features (also works)

### DIFF
--- a/src/wxc_common/src/config_parser.rs
+++ b/src/wxc_common/src/config_parser.rs
@@ -254,8 +254,8 @@ pub fn load_request(
 
 // ---------- Cross-field validation ----------
 
-/// Supported schema version (semver). Configs with a higher major.minor are rejected.
-const SUPPORTED_VERSION: (u32, u32) = (0, 5); // 0.5.x
+/// Maximum supported schema version (major.minor). Configs with a higher major.minor are rejected.
+const SUPPORTED_VERSION: &str = "<=0.5";
 
 /// Validate that the schema version (semver) is supported by this binary.
 /// Compares major.minor only — patch and pre-release labels are ignored.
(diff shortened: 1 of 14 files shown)


jsidewhite:
>     let parts: Vec<&str> = version_core.split('.').collect();

I thought the Rust SemVer crate was going to give us parsing functions (I know it's a simple parse, but i think @bbonaby suggested it, and we already depend on it i think?)

```rust
use semver::Version;

fn main() {
    let v1 = Version::parse("1.2.3").unwrap();
    let v2 = Version::parse("1.5.0").unwrap();

    if v2 > v1 {
        println!("upgrade!");
    }
}
```

```rust
use semver::{Version, VersionReq};

fn main() {
    let req = VersionReq::parse(">=1.2.3, <2.0.0").unwrap();
    let v = Version::parse("1.5.1").unwrap();

    println!("{}", req.matches(&v));
}
```

---
Refers to: src/wxc_common/src/config_parser.rs:269 in c41c495. [](commit_id = c41c495d9614235782b73c81de88c73dab9f0058, deletion_comment = False)

src/wxc_common/src/config_parser.rs
```diff
@@ -255,7 +255,7 @@ pub fn load_request(
 // ---------- Cross-field validation ----------
 
 /// Supported schema version (semver). Configs with a higher major.minor are rejected.
-const SUPPORTED_VERSION: (u32, u32) = (0, 4); // 0.4.x
+const SUPPORTED_VERSION: (u32, u32) = (0, 5); // 0.5.x
```
jsidewhite: >SUPPORTED_VERSION [](http://example.com/codeflow?start=6&length=17)

don't we need to handle 0.4.0 still, with VersionReq?

```
    let req = VersionReq::parse(">=1.2.3, <2.0.0").unwrap();
```

jsidewhite (approved):
:shipit:

src/wxc_common/src/config_parser.rs
SohamDas2021: The parser checks major.minor ≤ SUPPORTED, so 0.4.0 (minor 4 ≤ 5) will still be accepted.

jsidewhite: but we want to reject 0.3.0 and below, right?

SohamDas2021: Yes, while that is true currently, it may not be true when we work on 0.6.0? Are you saying we should reject any < 0.n-1.0?

SohamDas2021: Maybe we should enforce min supported is 0.4.0, guess that is what you are suggesting anyway.

SohamDas2021: Updated.

jsidewhite (approved):

PR #180, opened 2026-04-17
SDK: update network enforcement and allow binary path to be passed in as parameter

bbonaby:
### Why is this change being made?

- Sandboxes were unintentionally network-open when no `network` section was provided.
- Network enforcement also always used `'both'` mode (capabilities + firewall) for the windows process base container even though firewall rules aren't working yet.
- Electron apps for example that consumed the SDK as a dependency have no clean way to specify the wxc-exec binary location should they be packaged in a different location.

### What changed?

- **Default-deny network**: `createConfigFromPolicy` now sets `defaultPolicy: 'block'` when no network section is provided.
- **enforcementMode**: Defaults to `'capabilities'` Only uses `'both'` when `allowedHosts`/`blockedHosts` are specified.
- **`executablePath` option**: New `SandboxSpawnOptions.executablePath` lets consumers pass the binary path directly.

## How was the change tested?

- Tested via VM using an electron app that consumes the sdk where I ran through a scenario where a configs script attempts to go out to the web and there is no networking section to the config. Before this change it would go out to the web. After this change it would it should be blocked

-----
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/180)

### DIFF
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -3,6 +3,7 @@
 
 import * as pty from 'node-pty';
 import * as os from 'os';
+import * as fs from 'fs';
 import { randomBytes } from "crypto";
 import { parse as semverParse } from 'semver';
 import { SandboxPolicy, ContainerConfig, ContainmentType } from './types';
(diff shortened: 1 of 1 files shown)


sdk/src/sandbox.ts
```diff
@@ -268,7 +284,9 @@ function spawnWithConfig(
   const platform = os.platform();
   let executablePath: string | null;
 
-  if (platform === 'linux') {
+  if (options.executablePath) {
+    executablePath = options.executablePath;
```
SohamDas2021, line 294: Do we need to check if the path exists? findWxcExecutable() / findLxcExecutable() codepaths do check for that below.

SohamDas2021 (approved):

SohamDas2021 (approved):

PR #186, opened 2026-04-20
[WSLC] SDK integration

SohamDas2021:
Enables the SDK to generate config JSON for the WSLC (WSL Container) executor, completing the end-to-end path from SandboxPolicy to WSLC container execution.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/186)

### DIFF
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -325,13 +325,6 @@ export interface SandboxSpawnOptions {
   experimental?: boolean;
 
   /**
-   * Containment backend to use. Defaults to "process" (AppContainer on
-   * Windows, LXC on Linux). Set to "wslc" for Linux containers on Windows
-   * via the WSL Container SDK.
-   */
-  containment?: ContainmentType;
-
-  /**
    * Explicit path to the wxc-exec (or lxc-exec) binary.
    * When set, the SDK uses this path directly instead of searching.
    * Useful for packaged apps (e.g., Electron) where the binary
@@ -478,7 +471,7 @@ export function spawnSandbox(
   env?: { [key: string]: string | undefined },
   containment?: ContainmentType,
 ): pty.IPty {
-  const config = buildSandboxPayload(script, policy, workingDirectory, containerName, containment ?? options.containment);
+  const config = buildSandboxPayload(script, policy, workingDirectory, containerName, containment);
   return spawnWithConfig(config, options, workingDirectory, env);
 }
 
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -250,8 +250,13 @@ export function createConfigFromPolicy(
             throw new Error('Proxy configuration is not supported on Linux');
         }
 
-        if ((policy.network.allowedHosts?.length || policy.network.blockedHosts?.length) && !policy.network.allowOutbound) {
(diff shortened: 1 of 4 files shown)


sdk/src/sandbox.ts
```diff
@@ -250,6 +285,13 @@ export interface SandboxSpawnOptions {
    */
   experimental?: boolean;
 
+  /**
+   * Containment backend to use. Defaults to "process" (AppContainer on
+   * Windows, LXC on Linux). Set to "wslc" for Linux containers on Windows
+   * via the WSL Container SDK.
+   */
+  containment?: ContainmentType;
```
bbonaby: issue: I believe this one is in types.ts

sdk/src/sandbox.ts
```diff
@@ -359,7 +401,12 @@ export function spawnSandbox(
   containerName?: string,
   env?: { [key: string]: string | undefined }
 ): pty.IPty {
-  const config = buildSandboxPayload(script, policy, workingDirectory, containerName);
+  if (options.containment === 'wslc' && !options.experimental) {
+    throw new Error(
+      "WSLC containment requires experimental mode. Set 'experimental: true' in SandboxSpawnOptions."
+    );
+  }
```
bbonaby: issue: I believe we already have something like this in https://github.com/microsoft/mxc/blob/8fe637c86021670c8fc26d647a18c10eb89b6d68/sdk/src/sandbox.ts#L329 can you check if this already convers this case? You might need to add wslc to this list https://github.com/microsoft/mxc/blob/8fe637c86021670c8fc26d647a18c10eb89b6d68/sdk/src/types.ts#L45

sdk/src/sandbox.ts
```diff
@@ -389,6 +436,11 @@ export function spawnSandboxFromConfig(
   workingDirectory?: string,
   env?: { [key: string]: string | undefined }
 ): pty.IPty {
+  if (config.containment === 'wslc' && !options.experimental) {
```
bbonaby: note: see comment above about experimental backends list and function. I don't think these if statements are needed anymore

sdk/src/sandbox.ts
```diff
@@ -59,6 +59,31 @@ function validatePolicyVersion(version: string): void {
...
+ */
+function buildWslcContainerConfig(
+    config: ContainerConfig,
+    policy: SandboxPolicy,
+    containerId: string,
+): ContainerConfig {
+    config.containment = 'wslc';
+    config.containerId = containerId;
+
+    config.experimental = {
```
bbonaby, line 76: question: Can you check with Huzaifa if his microVM should have been added to this experimental json config as well? I don't see it here https://github.com/microsoft/mxc/blob/8fe637c86021670c8fc26d647a18c10eb89b6d68/sdk/src/types.ts#L165 

If so we might need to add it and a microvm config object

bbonaby:
fyi, you have merge conflicts now

sdk/src/sandbox.ts
SohamDas2021: Thanks for the catch. IMO, yes. I will sync with him.

SohamDas2021: Huzaifa will update this for microvm.

bbonaby (approved):

PR #205, opened 2026-04-23
[CLI/SDK] Add auto file logging for debug mode

huzaifa-d:
Adds opt-in diagnostic file logging across the MXC stack
Auto enabled with CLI : --debug.
And SDK : spawnSandbox('echo hi', policy, { debug: true });

-----
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/205)

### DIFF
--- a/sdk/src/logger.ts
+++ b/sdk/src/logger.ts
@@ -1,0 +1,42 @@
+// sdk/src/logger.ts
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+/**
+ * Appends timestamped log lines to a file.
+ * Emits console.warn if the file cannot be opened, then degrades to no-op.
+ */
+export class FileLogger {
+  private fd: number | null = null;
+
+  constructor(filePath: string) {
+    try {
+      const dir = path.dirname(filePath);
+      if (!fs.existsSync(dir)) fs.mkdirSync(dir, { recursive: true });
+      this.fd = fs.openSync(filePath, 'a');
+    } catch (err) {
+      console.warn(`[mxc-sdk] Could not open log file '${filePath}': ${err}`);
+      this.fd = null;
+    }
+  }
+
+  log(level: 'info' | 'warn' | 'error', message: string, data?: Record<string, unknown>): void {
+    if (this.fd === null) return;
+    try {
+      const ts = new Date().toISOString();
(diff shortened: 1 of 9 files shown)


sdk/src/sandbox.ts
```diff
@@ -424,7 +440,30 @@ function spawnWithConfig(
   workingDirectory?: string,
   env?: { [key: string]: string | undefined },
 ): pty.IPty {
-  const { executablePath, args } = resolveExecutableAndArgs(config, options);
+  let logger: FileLogger | undefined;
```
SohamDas2021, line 443: Not adding logs in other functions like spawnSandboxWithoutPty?

sdk/src/logger.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+  log(level: 'info' | 'warn' | 'error', message: string, data?: Record<string, unknown>): void {
+    if (this.fd === null) return;
+    try {
+      const ts = new Date().toISOString();
+      const suffix = data ? ' ' + JSON.stringify(data) : '';
+      const line = `[${ts}] ${level.toUpperCase()} ${message}${suffix}\n`;
+      fs.writeSync(this.fd, line);
```
SohamDas2021, line 32: blocking call? Shall we use buffering or an async fs.write?

src/wxc_common/src/logger.rs
```diff
@@ -13,21 +17,37 @@ pub enum Mode {
...
+        self.file = Some(file);
+        Ok(())
+    }
+
     pub fn log(&mut self, msg: &str) {
         match self.mode {
             Mode::Console => print!("{}", msg),
             Mode::Buffer => self.buffer.push_str(msg),
         }
+        if let Some(ref mut f) = self.file {
```
SohamDas2021, line 44: refactor into a helper function and call from log and log_line?

sdk/src/sandbox.ts
huzaifa-d: Good point. New function and was already working on logging. Added.

sdk/src/logger.ts
huzaifa-d: On purpose. Logging is only when debug is chosen and Async() adds more complexity. Keeping it simple for now.

SohamDas2021 (approved):

PR #209, opened 2026-04-24
[WSLC] Add usePty option to SandboxSpawnOptions and WSLC SDK E2E test

SohamDas2021:
Adds a usePty option to SandboxSpawnOptions so spawnSandboxFromConfig can use child_process.spawn (non-PTY) for reliable exit codes and separate stdout/stderr streams. Also adds the first SDK-level E2E test for WSLC, validating that WSLC-specific fields flow through end-to-end.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/209)

### DIFF
--- a/sdk/src/helper.ts
+++ b/sdk/src/helper.ts
@@ -2,10 +2,12 @@
 // Licensed under the MIT License.
 
 import * as os from 'os';
+import * as fs from 'fs';
 import * as path from 'path';
 import { randomBytes } from 'crypto';
 import { FileLogger } from './logger';
-import { ContainerConfig } from './types';
+import { ContainerConfig, ContainmentType, ExperimentalBackends, SandboxingMethod } from './types';
+import { findWxcExecutable, findLxcExecutable, getPlatformSupport } from './platform';
 import { SandboxSpawnOptions } from './sandbox';
 
 /**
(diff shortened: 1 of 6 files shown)


sdk/src/sandbox.ts
```diff
@@ -535,22 +542,48 @@ export function spawnSandboxWithoutPty(
...
+  env?: { [key: string]: string | undefined }
+): pty.IPty;
 export function spawnSandboxFromConfig(
   config: ContainerConfig,
   options: SandboxSpawnOptions = {},
   workingDirectory?: string,
   env?: { [key: string]: string | undefined }
-): pty.IPty {
+): pty.IPty | ChildProcess {
+  if (options.usePty === false) {
```
huzaifa-d, line 469: This will skip logging. Consider a spawnWithLogging(executablePath, args, spawnFn, options) helper.

SohamDas2021: Updated.

sdk/src/sandbox.ts
```diff
@@ -21,6 +21,37 @@ function makeLogFilePath(dir: string): string {
...
 }
 
+/**
+ * Sets up logging and resolves the executable path + args.
+ * Shared by both PTY and non-PTY spawn paths.
+ */
+function prepareSpawn(
+  config: ContainerConfig,
+  options: SandboxSpawnOptions,
+): { executablePath: string; args: string[]; logger?: FileLogger; logFile?: string; startTime: number } {
```
bbonaby: suggestion: I think this function should return a named object so it's clear  rather than an unnamed one. That said, I also think it might be time for us to start moving these kinds of functions out of the sandbox.ts file and into a helper.ts file. I'm realizing now that the public api surface is kind of buried below all our helpers. If you don't mind can you start us off my moving this function into a helper.ts file please?

sdk/src/sandbox.ts
```diff
@@ -588,22 +604,64 @@ export function spawnSandboxWithoutPty(
...
  * const ptyProcess = spawnSandboxFromConfig(config);
+ *
+ * // Non-PTY mode — returns ChildProcess with reliable exit codes:
+ * const child = spawnSandboxFromConfig(config, { usePty: false });
+ * child.stdout?.on('data', (data) => console.log(data.toString()));
  * ```
  */
+export function spawnSandboxFromConfig(
+  config: ContainerConfig,
+  options: SandboxSpawnOptions & { usePty: false },
```
bbonaby, line 453: question: is the ` & { usePty: false }` needed?  we have usePty in the spawnOptions already and we check it in the implementation.

sdk/src/sandbox.e2etest.ts
```diff
@@ -0,0 +1,79 @@
...
+import fs from 'node:fs';
+import path from 'node:path';
+import os from 'os';
+import { createConfigFromPolicy, spawnSandboxFromConfig } from './sandbox';
+import { SandboxPolicy } from './types';
+import { ChildProcess } from 'child_process';
+
+const isWslcAvailable = os.platform() === 'win32';
+
+describe('WSLC SDK E2E — createConfigFromPolicy → customize → spawn', {
```
bbonaby, line 27: note: the location of this file will change when my other one goes in

SohamDas2021: ack.

sdk/src/sandbox.ts
SohamDas2021: It tells the compiler when usePty is exactly false, return ChildProcess. Without it, both overloads would have the same signature and the caller wouldn't get type narrowing.

sdk/src/sandbox.ts
```diff
@@ -440,51 +438,29 @@ function spawnWithConfig(
...
-    logger.log('info', 'mxc.log.created', { logFile });
-  }
-
-  const startTime = Date.now();
-  logger?.log('info', 'mxc.spawn.start', {
-    platform: os.platform(),
-    arch: os.arch(),
-    containment: config.containment,
-  });
+  const { executablePath, args, logger, startTime } = prepareSpawn(config, options, resolveExecutableAndArgs);
```
bbonaby: question: just noticed this not under the try/catch. Should it be?

sdk/src/helper.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const startTime = Date.now();
+  logger?.log('info', 'mxc.spawn.start', {
+    platform: os.platform(),
+    arch: os.arch(),
+    containment: config.containment,
+  });
+
+  const { executablePath, args } = resolveExecutableAndArgs(config, options);
```
bbonaby: suggestion: to be honest, `resolveExecutableAndArgs` isn't  a publicly exported function either. I'd say let's just move that function into this helpers file then we don't have to take in the function as a parameter. What do you think?

bbonaby (approved):

PR #210, opened 2026-04-24
[WSLC] Add execution timeout enforcement for WSLC containers

SohamDas2021:
Implements process execution timeout for the WSLC backend. Previously, WaitForSingleObject waited indefinitely (u32::MAX) for the process to exit, meaning a hung or long-running process would block forever. Now the configured script_timeout is enforced- the container is killed after the specified duration.

Fixes #160 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/210)

### DIFF
--- a/src/wslc_common/src/wsl_container_runner.rs
+++ b/src/wslc_common/src/wsl_container_runner.rs
@@ -814,17 +814,45 @@ impl WSLContainerRunner {
         }
         let process_guard = WslcProcessGuard::from_raw(process, sdk.WslcReleaseProcess);
 
-        // -- Step 12: Wait for exit (callbacks capture I/O during execution) --
+        // -- Step 12: Wait for exit with timeout enforcement --
         let mut exit_event: HANDLE = ptr::null_mut();
         let hr = (sdk.WslcGetProcessExitEvent)(process_guard.as_raw(), &mut exit_event);
         if hr != S_OK {
             return sdk_error("WslcGetProcessExitEvent failed", hr, "");
         }
+
+        // Use configured timeout, or wait indefinitely if not set.
+        let wait_ms = if request.script_timeout > 0 {
+            request.script_timeout
+        } else {
+            u32::MAX
+        };
+
+        let mut timed_out = false;
         if !exit_event.is_null() {
-            windows::Win32::System::Threading::WaitForSingleObject(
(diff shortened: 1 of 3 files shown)


src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -814,17 +814,45 @@ impl WSLContainerRunner {
...
+        let mut timed_out = false;
         if !exit_event.is_null() {
-            windows::Win32::System::Threading::WaitForSingleObject(
+            let wait_result = windows::Win32::System::Threading::WaitForSingleObject(
                 windows::Win32::Foundation::HANDLE(exit_event),
-                u32::MAX,
+                wait_ms,
             );
+            // WAIT_TIMEOUT = 0x00000102 (258)
+            if wait_result.0 == 258 {
```
bbonaby, line 838: suggestion: I believe you might be able to use the WAIT_TIMEOUT value from the windows-rs crate instead of manually writing the number

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -814,17 +814,45 @@ impl WSLContainerRunner {
...
+                // Send SIGTERM with 2-second grace period; the SDK escalates
+                // to SIGKILL automatically if the process doesn't exit.
+                err_msg = CoTaskMemPWSTR::null();
+                let _ = (sdk.WslcStopContainer)(
+                    container_guard.as_raw(),
+                    WslcSignal::SigTerm,
+                    2,
+                    err_msg.as_mut_ptr(),
+                );
+                drop(err_msg);
```
bbonaby, line 854: suggestion: I don't think you need to explicitly drop. When this goes out of scope it should drop it automatically since you've implemented it.

src/wslc_common/src/wsl_container_runner.rs
```diff
@@ -900,10 +932,14 @@ impl WSLContainerRunner {
         // for the Release calls, and the process exits shortly after.
 
         ScriptResponse {
-            exit_code,
+            exit_code: if timed_out { -1 } else { exit_code },
```
bbonaby, line 935: note: we could consider returning the win32 or hresult value for timeout if you want rather than -1.

bbonaby (approved):

PR #211, opened 2026-04-24
Refactor SDK tests: dry-run support and test restructure

bbonaby:

We plan to remove the `CLI/` project via: #212 to make the repository easier to understand/navigate. One step in this direction is to move the integration tests from there directly into the SDK.

That said, some tests can't run in the pipelines because they require OS-level sandbox infrastructure such as BaseContainer API, LXC networking. These aren't all available on standard build runners. So we also need a way to test SDK -> wxc binary interaction without creating any sandboxes. This change helps alleviate this issue by adding "dry-run" tests where data flow goes through sdk -> wxc-exec and returns right before attempting to run the runners.

### What Changed?

- **`--dry-run` flag** added to `wxc-exec` and `lxc-exec` : validates that the executables are invoked.
- **`dryRun` option** added to SDK `SandboxSpawnOptions` : passes `--dry-run` to the native binary
- **`ScriptRunner` trait refactored**: `run()` is now a provided method with a dry-run guard; runners implement `execute()`, structurally enforcing dry-run for any new backend
- **Moved SDK tests** : Moved integration tests to sdk under `integration/` folder and `unit` folder for unit tests. The integration tests consumes the SDK npm package directly.
- **Dry-run smoke tests** added exercising all3 SDK spawn functions (`spawnSandbox`, `spawnSandboxAsync`, `spawnSandboxFromConfig`)
- **Azure Pipeline** : added `SDK_Integration_Tests` and `SDK_Unit_Tests` stages to `1ES.Build.yml`
- **Pipeline Environment variables** : added `MXC_SKIP_OS_BUILD_DEPENDENT_TESTS` and `MXC_SKIP_LXC_NETWORK_TESTS`  for tests the can't run in the pipeline.
- `test issues`: Added sdk/tests/integration/test_issues.md which has a list of current issues.

### How was it tested?

- All integration tests should run locally based on the platform unless `MXC_SKIP_OS_BUILD_DEPENDENT_TESTS` or ``MXC_SKIP_LXC_NETWORK_TESTS`` for lxc is set in the environment variables.
   - Note: We will need to figure out how we can unskip these in the pipeline in the future (different windows internal build pool maybe).
- Dry-run tests pass which should validate SDK <--> exec binary communication
- Confirmed via build in azure pipeline as well: https://microsoft.visualstudio.com/Dart/_build/results?buildId=145686041&view=results

fixes #202

-----
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/211)

### DIFF
--- a/sdk/src/platform.ts
+++ b/sdk/src/platform.ts
@@ -91,9 +91,9 @@ export function getPlatformSupport(): PlatformSupport {
 
   if (platform === 'linux') {
     // LXC is the only containment backend on Linux
-    support.availableMethods = ['lxc'];
     if (isLxcAvailable()) {
       support.isSupported = true;
+      support.availableMethods = ['lxc'];
     } else {
       support.reason = 'LXC is not installed or not available on this system';
     }
--- a/sdk/src/platform.ts
+++ b/sdk/src/platform.ts
@@ -90,13 +90,13 @@ export function getPlatformSupport(): PlatformSupport {
   }
 
   if (platform === 'linux') {
-    // Check if LXC is available
+    // LXC is the only containment backend on Linux
+    support.availableMethods = ['lxc'];
     if (isLxcAvailable()) {
       support.isSupported = true;
-      support.availableMethods = ['lxc'];
-      return support;
+    } else {
+      support.reason = 'LXC is not installed or not available on this system';
     }
-    support.reason = 'LXC is not installed or not available on this system';
     return support;
   }
 
--- a/sdk/src/policy.ts
+++ b/sdk/src/policy.ts
@@ -7,7 +7,7 @@
  */
 
 import * as fs from 'fs';
-import os from 'os';
+import * as os from 'os';
 import * as path from 'path';
 import { execSync } from 'child_process';
 import { randomBytes } from 'crypto';
--- a/sdk/src/policy.ts
+++ b/sdk/src/policy.ts
@@ -7,7 +7,7 @@
  */
 
 import * as fs from 'fs';
-import * as os from 'os';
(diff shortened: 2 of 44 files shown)


sdk/src/platform.ts
```diff
@@ -90,13 +90,13 @@ export function getPlatformSupport(): PlatformSupport {
   }
 
   if (platform === 'linux') {
-    // Check if LXC is available
+    // LXC is the only containment backend on Linux
+    support.availableMethods = ['lxc'];
```
SohamDas2021: >support.availableMethods = ['lxc']; [](http://example.com/codeflow?start=4&length=35)

Why are we doing this outside the if (isLxcAvailable()) block?

sdk/src/policy.ts
```diff
@@ -7,7 +7,7 @@
  */
 
 import * as fs from 'fs';
-import * as os from 'os';
+import os from 'os';
```
SohamDas2021: intended?

sdk/src/platform.ts
bbonaby: Updated, running azure pipeline to confirm we're still good.

sdk/src/policy.ts
bbonaby: removed  thanks

sdk/src/platform.ts
bbonaby: confirmed we're good now

SohamDas2021 (approved):
:shipit:

PR #214, opened 2026-04-24
Add registry-based per-executable config override for wxc-exec

bbonaby:
When diagnosing workloads running inside MXC, it is useful to override specific policies (filesystem, network, UI) for a particular executable without modifying the caller's config. This lets administrators iterate on policy tuning from the host machine using registry keys, without requiring changes to the application that invokes wxc-exec.

Note: this is temporary to allow folks to test today. We have plans to do this via a hooking mechanism in the future.

### What changed

- **New module** `registry_config.rs` in `wxc_common`: parses the command line from a `CodexRequest` to resolve the target executable name, checks `HKLM\SOFTWARE\Microsoft\MXC\Diagnostics\Exec\{exe_name}` for an override config path, and merges or replaces the original request
- **Documentation**: `docs/dev/registry-config-override.md` walkthrough with registry layout, examples, and testing instructions
- **Two override modes**: `OverrideConfig=1` fully replaces the request; default merge mode selectively overrides filesystem, network, and/or UI policies via per-policy `OverrideFilesystemPolicy`, `OverrideNetworkPolicy`, `OverrideUiPolicy` flags
- **Integration** in `wxc/src/main.rs`: `check_registry_override()` runs after `load_request()`, non-fatal on error
- **winreg dependency**: added `winreg = "0.55"` (Windows-only) for clean registry access
- **Test configs**: 3 PowerShell 5.1 configs under `test_configs/registry_override_configs/` (`ps51_baseline.json`, `ps51_override.json`, `ps51_override_restrictive_ui.json`)
- **Test script**: `test_scripts/registry_override/run_registry_override_tests.ps1` with 6 automated tests covering no-override, full override, merge mode, and UI-only override

### How was it tested?

- 11 unit tests in `registry_config.rs` covering command-line parsing edge cases, merge vs. full-override logic, and per-policy flag behavior (`cargo test -p wxc_common -- registry_config`)
- added 6 tests that test the override scenario on a Windows machine that supports MXC bits ( see `run_registry_override_tests.ps1`).

### DIFF
--- a/schemas/dev/mxc-config.schema.0.5.0-dev.json
+++ b/schemas/dev/mxc-config.schema.0.5.0-dev.json
@@ -155,12 +155,12 @@
 
     "ui": {
       "type": "object",
-      "description": "Cross-platform UI policy. Mapped from SandboxPolicy.ui by createConfigFromPolicy.",
+      "description": "UI policy for container.",
       "properties": {
         "disable": {
           "type": "boolean",
           "default": true,
-          "description": "Whether UI is disabled (no visible windows)."
+          "description": "Whether UI is disabled."
         },
         "clipboard": {
           "type": "string",
@@ -172,6 +172,27 @@
           "type": "boolean",
           "default": false,
           "description": "Whether input injection (keyboard/mouse) is allowed."
+        },
+        "isolation": {
+          "type": "string",
+          "enum": ["desktop", "handles", "atoms", "container"],
+          "default": "container",
+          "description": "UI isolation level for the desktop (Windows BaseProcessContainer only)."
+        },
+        "desktopSystemControl": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether desktop system control is allowed (Windows BaseProcessContainer only)."
+        },
+        "systemSettings": {
+          "type": "string",
+          "default": "none",
+          "description": "System settings access level (Windows BaseProcessContainer only)."
+        },
+        "ime": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether IME (Input Method Editor) is allowed (Windows BaseProcessContainer only)."
         }
       }
     },
@@ -190,33 +211,6 @@
           "items": { "type": "string" },
           "default": [],
           "description": "AppContainer capabilities (e.g., 'internetClient', 'registryRead')."
-        },
-        "ui": {
-          "type": "object",
-          "description": "BaseProcessContainer-specific UI settings (Windows only).",
-          "properties": {
-            "isolation": {
-              "type": "string",
-              "enum": ["desktop", "handles", "atoms", "container"],
-              "default": "container",
-              "description": "UI isolation level for the desktop."
-            },
-            "desktopSystemControl": {
-              "type": "boolean",
-              "default": false,
-              "description": "Whether desktop system control is allowed."
-            },
-            "systemSettings": {
-              "type": "string",
-              "default": "none",
-              "description": "System settings access level."
-            },
-            "ime": {
-              "type": "boolean",
-              "default": false,
-              "description": "Whether IME (Input Method Editor) is allowed."
-            }
-          }
         }
       }
     },
--- a/schemas/dev/mxc-config.schema.0.5.0-dev.json
+++ b/schemas/dev/mxc-config.schema.0.5.0-dev.json
@@ -155,12 +155,12 @@
 
     "ui": {
       "type": "object",
-      "description": "UI policy for container.",
+      "description": "Cross-platform UI policy. Mapped from SandboxPolicy.ui by createConfigFromPolicy.",
       "properties": {
         "disable": {
           "type": "boolean",
           "default": true,
-          "description": "Whether UI is disabled."
+          "description": "Whether UI is disabled (no visible windows)."
         },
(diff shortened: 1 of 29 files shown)


test_configs/base_container_ui_configs/02_clip_none_write_blocked.json
```diff
@@ -6,6 +6,16 @@
     "commandLine": "cmd.exe /c \"echo test | clip.exe && echo FAIL: wrote clipboard || echo PASS: clipboard write blocked\"",
     "timeout": 15000
   },
-  "appContainer": { "capabilities": [], "ui": { "isolation": "desktop", "desktopSystemControl": false, "systemSettings": "none", "ime": false } },
-  "ui": { "disable": false, "clipboard": "none", "injection": false }
-}
+  "appContainer": {
```
SohamDas2021: Please make sure all the new updated config jsons pass? Your description doesn't say if you have validated those. just making sure.

schemas/dev/mxc-config.schema.0.5.0-dev.json
```diff
@@ -172,6 +172,27 @@
           "type": "boolean",
           "default": false,
           "description": "Whether input injection (keyboard/mouse) is allowed."
+        },
```
SohamDas2021: I understand the purpose here, but once we ship. going forward changing non-experimental features in the schema would be a strict no, are we on the same page?

jsidewhite:
I don't think this PR is the right approach.  Slinging all the raw-tech options from the config.json schema into the `SandboxPolicy` ts struct in support of a dev-only debugging scenario is misguided and will confuse developers.

(Commenting for durability.)

schemas/dev/mxc-config.schema.0.5.0-dev.json
bbonaby: Spoke to Jeff and went a different direction. That said yea we're on the same page.

test_configs/base_container_ui_configs/02_clip_none_write_blocked.json
bbonaby: Spoke to jeff and went a different route

bbonaby:
> I don't think this PR is the right approach. Slinging all the raw-tech options from the config.json schema into the `SandboxPolicy` ts struct in support of a dev-only debugging scenario is misguided and will confuse developers.
> 
> (Commenting for durability.)

Spoke to Jeff and went a different direction. See reverted commit and now registry solution

src/wxc/src/main.rs
```diff
@@ -156,6 +157,18 @@ fn main() {
...
+    match check_registry_override(&mut request, &mut logger) {
+        Ok(true) => {
+            let _ = writeln!(logger, "Registry override applied");
+        }
+        Ok(false) => {} // No override found — continue normally.
+        Err(e) => {
+            let _ = writeln!(logger, "Registry override error: {}", e);
+            // Non-fatal: proceed with original config.
+        }
+    }
```
jsidewhite, line 170: I don't like this block of code - i.e. side-effects.

Took me a while to realize that check_registry_override is clobbering request.

Is it possible to return the new request instead of clobbering the existing one secrety in the "check_" api?

jsidewhite (approved):
:shipit:

shschaefer:
I do not think this is a good solution.  It only works on Windows.  It is disruptive to automated tool flows.

asklar:
there's already 2 ways to configure policy for the scenarios we've been talking about (config.json and governance layer), so I don't think we need a third.

jsidewhite:
Assume you're talking about config.json for copilot.exe - which is just one consumer, and not as expressive as the config layer - not really the vector we'd use for a debugging tool for MXC.  I'm not sure governance is the right tool either, but I'd love to understand more when we have a beat.

---
In reply to: [4329675416](https://github.com/microsoft/mxc/pull/214#issuecomment-4329675416) [](http://example.com/codeflow?ancestors=4329675416)

jsidewhite:
It'll be fine - this will satisfy the immediate need to debug/tweak his UI bobbits now, and tomorrow I'm going to convince Branden that we can keep all of his work almost as-is, but,
 
1. Move it into a separate policy injection .exe
1. Change wxc-exec.exe to expose a general hooking mechanism
1. Hooking mechanism in MXC will be invoked right before policy application to allow a transform of policy
1. Registry will then merely become an implementation artifact of this particular injection .exe tool - which can evolve if the interface is confrontational.

---
In reply to: [4320911786](https://github.com/microsoft/mxc/pull/214#issuecomment-4320911786) [](http://example.com/codeflow?ancestors=4320911786)

MGudgin:
I'm struggling to understand the design and direction here, why can't we just do all this with config.json for now? I'm also concerned about the security holes this introduces (e.g. caller of SDK thinks they are getting a particular policy/config but instead that gets overridden silently by something in the registry).

MGudgin (changes requested):
I do not think this is the correct approach to the stated problem. And it introduces security issues as noted in my other comment. Marking this as 'Request Changes' for now. Please do not merge.

bbonaby:
No point in keeping this up since we're not going through with this so I'm going to close it.

PR #217, opened 2026-04-25
[IsolationSession] Add experimental containment backend

adpa-ms:

-----

## Summary

Adds an experimental `isolation_session` containment backend that runs commands inside a Windows Isolation Session via the OS-side IsoEnvBroker Session API. Behind `--experimental`. Companion issue: #216. Full design in [`docs/isolation-session-plan.md`](docs/isolation-session-plan.md).

## What's in this PR

**Implementation:**
- New backend module `src/wxc_common/src/isolation_session_runner.rs` — two-layer (`IsolationSessionManager` + `IsolationSessionRunner`).
- New bindings crate `src/isolation_session_bindings/` — generated from a private WinMD; provenance tracked in `external/windows-sdk/isolation-environment-session/GENERATION_INFO.toml`.
- New generator tool at `tools/generate-isolation-session-bindings/` (separate workspace, not a member of `src/`).
- New JSON config: `"containment": "isolation_session"` with `experimental.isolation_session.configurationId`.
- New `isolation_session` Cargo feature on `wxc` (off by default).
- Dispatch in `wxc/src/main.rs` behind `--experimental`; `CoInitializeEx(COINIT_MULTITHREADED)` added at startup (benign for non-WinRT backends).

**Documentation and tests:**
- `docs/isolation-session-plan.md` — design doc covering problem, architecture, config, OS API dependency, bindings workflow, v0.1 scope, test plan, known issues, risks, prerequisites, examples, supported workloads.
- `test_configs/isolation_session_hello.json` + `isolation_session_exit42.json` — integration test configs.
- `test_scripts/run_isolation_session_tests.ps1` — runner that invokes both via `wxc-exec --experimental`, validates exit codes and output substrings, reports pass/fail.
- `.github/copilot-instructions.md` — added IsolationSession to the backends table, the integration tests list, and the key documentation list.

## Testing

**Automated** (covered by `cargo test --workspace`):
- All 287 workspace tests pass; no regressions vs. main.
- ~22 of those are backend-specific (config parsing, policy validation, options building, feature-unavailable). The feature-unavailable test is what runs in CI — it returns a clean error on machines without the IsoEnvBroker service.

**Integration / manual** (require a host with `Feature_IsoBrokerSessionApis`):
- `test_scripts/run_isolation_session_tests.ps1` runs both configs and validates output. Verified passing on a Hyper-V VM: agent identity (`<caller>-IEB-<NNN>`), env-var pass-through, working-directory pass-through, and exit-code propagation (42) all green.
- These tests must run interactively on the host — the IsoEnvBroker cohort check rejects network-logon tokens, so PSSession-driven invocations fail with `Access Denied`. Documented in the design doc.

Branch was rebased onto current `main` immediately before opening this PR; all tests above were re-run on the rebased branch.

## Scope notes

This PR is intentionally narrow — it introduces the experimental backend and nothing else. Explicitly deferred (each warrants its own issue when its conversation begins):
- TypeScript SDK exposure (the JSON/CLI entry point works today; SDK hookup is follow-up work).
- Stateful API where one provisioned agent + session hosts multiple exec calls. The `Manager` / `Runner` split exists to make this migration clean.
- Interactive ConPTY, control signals, console resize — fire-and-forget script execution doesn't need them.

Closes #216.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/217)


external/windows-sdk/isolation-environment-session/README.md
```diff
@@ -0,0 +1,41 @@
...
+- The WinMD file from an OS build (e.g., amd64chk flavor)
+- Rust toolchain (the generation tool builds from source)
+
+```sh
+cargo run --manifest-path tools/generate-isolation-session-bindings/Cargo.toml -- <path-to-winmd>
+```
+
+After regeneration:
+1. Review the generated `src/isolation_session_bindings/src/bindings.rs`
+2. Update `GENERATION_INFO.toml` manually — the generator does not touch it. Fields:
```
bbonaby, line 25: question:  So we we just need to wait  a couple months until the new APIs show up in the windows-rs crate right? we should make a GitHub issue to track this

src/isolation_session_bindings/src/bindings.rs
```diff
@@ -0,0 +1,2645 @@
+// Bindings generated by `windows-bindgen` 0.62.1
```
bbonaby, line 1: ❤️

src/isolation_session_bindings/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License.
+
+//! Generated WinRT bindings for the IsoEnvBroker Session API.
+//!
+//! This crate contains Rust projections generated from the
+//! `Windows.AI.IsolationEnvironment.Session` WinMD using `windows-bindgen`.
+//!
+//! To regenerate:
+//! ```sh
+//! cargo run --manifest-path tools/generate-isolation-session-bindings/Cargo.toml -- <winmd-path>
```
bbonaby, line 11: note: just incase I don't see it, just adding a note that we should make sure we add this step to the build script and pipelines

src/isolation_session_bindings/build.rs
```diff
@@ -0,0 +1,91 @@
...
+    // If the provenance file doesn't exist yet (e.g., first-time setup before
+    // generation has been run), skip the check.
+    if !info_path.exists() {
+        return;
+    }
+
+    let contents = std::fs::read_to_string(&info_path).unwrap_or_default();
+
+    // Extract the expected windows crate version from the TOML.
+    let expected = contents.lines().find_map(|line| {
```
bbonaby, line 27: suggestion: I believe we can use the semver crate from the workspace to do the version matching in this file. That might be nicer. I checked the windows-rs crate to see if bindgen has a version we can check as run time but it doesn't look like it 😢 so the lock file might be the only way for now.

Alternatively you could specify the version of the windows crate you want in this crates toml file instead of using the version in the workspace. Then put a comment that the version used there should match the version in your generation_info file.

src/wxc/src/main.rs
```diff
@@ -103,6 +104,15 @@ fn delete_app_container_profile(name: &str, logger: &mut Logger) -> bool {
 }
 
 fn main() {
+    // Initialize COM/WinRT for backends that use WinRT APIs (Isolation Session).
+    // COINIT_MULTITHREADED is benign for backends that don't use COM.
+    let _ = unsafe {
```
bbonaby, line 119: issue: We probably want to add a // Safety comment here and anywhere where unsafe functions are used just for completeness in this PR. See: https://std-dev-guide.rust-lang.org/policy/safety-comments.html#inside-unsafe-elements

test_scripts/run_isolation_session_tests.ps1
```diff
@@ -0,0 +1,154 @@
...
+
+$ErrorActionPreference = "Stop"
+$RepoRoot = Split-Path -Parent $PSScriptRoot
+
+if (-not $ConfigDir) {
+    $ConfigDir = Join-Path $RepoRoot "test_configs"
+}
+
+# Locate wxc-exec.exe — explicit path > target-specific release > default release.
+$Target = "x86_64-pc-windows-msvc"
```
bbonaby: issue: I believe we also need the [aarch64-pc-windows-msvc](https://doc.rust-lang.org/nightly/rustc/platform-support/windows-msvc.html) target as well for the folks on arm64 Windows devices.

tools/generate-isolation-session-bindings/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+//!
+//! The WinMD file is built from the OS repo (amd64chk flavor):
+//!   obj/<flavor>/onecoreuap/windows/core/isoenvbroker/src/published/objchk/amd64/windows.ai.isolationenvironment.winmd
+
+use std::env;
+use std::path::Path;
+use std::process;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
```
bbonaby: suggestion: we can use the cli crate for arg passing. checkout the wxc crate to see an  example

tools/generate-isolation-session-bindings/src/main.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Generates Rust bindings for the IsoEnvBroker Session API from a WinMD file.
```
bbonaby, line 4: question: any reason you decided to put this crate in a tools  folder? was it because you rather it not get built  in the same workspace as the other crates?

tools/generate-isolation-session-bindings/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+            // Walk up to find the cargo registry cache for windows-bindgen.
+            let candidate = dir.join("default").join("Windows.winmd");
+            if candidate.exists() {
+                return Some(candidate);
+            }
+            None
+        });
+
+    // Fallback: search the cargo registry directly.
+    let default_winmd = bindgen_manifest.unwrap_or_else(|| {
```
bbonaby: issue: I think we should just fail fast here. If the repo structure changes without updating this we just fail if we can't find the winmd so we're alerted right away

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+            if name.is_empty() {
+                None
+            } else {
+                Some((name, value))
+            }
+        })
+        .collect();
+
+    ProcessOptions {
+        process_path: r"C:\Windows\System32\cmd.exe".to_string(),
```
bbonaby: question: so always cmd? Would we ever want powershell 5.1? I ask since that's also installed on Windows out the box. Also I'm thinking we should use `std::env::var("SystemDrive")` to get the drive windows is installed on first and use that. If the environment variable isn't set, which is unlikely then we can fall back to C

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+/// This is called once from `IsolationSessionManager::new()`.
+pub(crate) fn check_service_available() -> Result<(), String> {
+    // Try the lightest possible call: RegisterClient with an empty registration id.
+    // If the WinRT activation factory is not registered (feature disabled),
+    // this fails with CLASS_E_CLASSNOTAVAILABLE or similar.
+    match IsolationSessionClient::RegisterClient(&windows_core::HSTRING::new()) {
+        Ok(_) => Ok(()),
+        Err(e) => {
+            let code = e.code().0 as u32;
+            // CLASS_E_CLASSNOTAVAILABLE (0x80040111) or REGDB_E_CLASSNOTREG (0x80040154)
```
bbonaby: suggestion: I believe you can find these two HRESULT in either Win32::Foundation or Win32::System::WinRT if you want to rely on the windows rust crate

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+/// Manages the IsoEnvBroker Session API lifecycle. Methods map 1:1 to the
+/// Session API steps.
+pub struct IsolationSessionManager {
+    registration_id: windows_core::HSTRING,
+    provision_id: windows_core::HSTRING,
+}
+
+impl IsolationSessionManager {
+    /// Activate the WinRT factory and verify the service is available.
+    pub fn new() -> Result<Self, String> {
```
bbonaby: suggestion: I believe we should have actual error variants for IsolationSessionManager. I think I need to do an inventory for the other runners but you can see how the nanvix runner does it here: https://github.com/microsoft/mxc/blob/a5fd45e935860869f88d4f2c40e82c773a3339b7/src/wxc_common/src/nanvix_runner.rs#L70.  You can then create a `from<T>`  impl for ScriptResponse as well so it can be auto converted to one.

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+            .map_err(|e| format!("get Status failed: {}", e))?;
+        if status != IsolationSessionProvisionStatus::Succeeded {
+            let ext = result.ExtendedError().unwrap_or_default();
+            return Err(format!(
+                "ProvisionAgentUserAsync status: {} (extended: {:#010x})",
+                status.0, ext.0
+            ));
+        }
+
+        let name = result
```
bbonaby, line 292: question: if they fails does the user stay provisioned forever?

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+            .map_err(|e| format!("StartSessionAsync: {}", e))?;
+
+        match status {
+            IsolationSessionOperationStatus::Succeeded
+            | IsolationSessionOperationStatus::SessionAlreadyStarted => Ok(()),
+            _ => Err(format!("StartSessionAsync returned status: {}", status.0)),
+        }
+    }
+
+    /// Step 3: Create an isolated process and capture its output.
```
bbonaby: question: this is creating a process inside the agent session right? If so should we update the comment to say that? Maybe it's just me but at first I thought this was creating a process inside the user session.

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+                .map_err(|e| format!("SetTimeoutMilliseconds: {}", e))?;
+        }
+
+        if !options.working_directory.is_empty() {
+            create_opts
+                .SetWorkingDirectory(&windows_core::HSTRING::from(&options.working_directory))
+                .map_err(|e| format!("SetWorkingDirectory: {}", e))?;
+        }
+
+        if !options.env_vars.is_empty() {
```
bbonaby, line 357: question: In CreateProcess on Windows environment variables need to be sorted in alphabetical order https://learn.microsoft.com/en-us/windows/win32/procthread/changing-environment-variables. Do we need to do that here?

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+            ));
+        }
+
+        let worker = result
+            .Process()
+            .map_err(|e| format!("get Process failed: {}", e))?;
+
+        // Read stdout/stderr via pipe handles.
+        let stdout = {
+            let h = worker
```
bbonaby: suggestion: how about out_handle and err_handle below for the variable names?

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+            return ScriptResponse::error(&format!("start_session failed: {}", e));
+        }
+
+        let result = match manager.create_process(&options) {
+            Ok(r) => r,
+            Err(e) => {
+                // Attempt cleanup even on failure.
+                let _ = manager.stop_session();
+                let _ = manager.deprovision_agent_user();
+                let _ = manager.unregister_client();
```
bbonaby, line 612: note: something to think about is whether the lifetime clean up  should be handled by the manager e.g tracking or some RAII type that represents the client side session.

src/wxc_common/src/models.rs
```diff
@@ -43,6 +46,27 @@ impl Default for WindowsSandboxConfig {
     }
 }
 
+/// Session configuration size for the Isolation Session backend.
+/// Maps to `IsolationConfigurationId` in the IsoEnvBroker Session API.
+#[derive(Debug, Default, Copy, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum IsolationSessionConfigurationId {
```
bbonaby, line 54: question: should we add comments on what each of these mean ?

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,749 @@
...
+    }
+}
+
+// -- IsolationSessionManager (lifecycle core) --------------------------------
+
+/// Manages the IsoEnvBroker Session API lifecycle. Methods map 1:1 to the
+/// Session API steps.
+pub struct IsolationSessionManager {
+    registration_id: windows_core::HSTRING,
+    provision_id: windows_core::HSTRING,
```
bbonaby, line 226: note: might be usual to have a comment on what each of these mean in practise. For example I might have missed it but I don't see who updates these from their defaults. Is it the winrt call to `ProvisionAgentUserAsync`?

.github/copilot-instructions.md
```diff
@@ -69,9 +69,10 @@ node --test dist/sandbox.test.js
 node --test dist/cli.test.js
 
 # Integration tests — run from repo root, requires built wxc-exec.exe
-test_scripts\run_test_configs.bat           # All test configs via wxc_test_driver
-test_scripts\run_basicac_test.bat           # Single AppContainer test
-test_scripts\run_lxc_all_tests.sh           # All LXC tests (Linux)
+test_scripts\run_test_configs.bat                # All test configs via wxc_test_driver
+test_scripts\run_basicac_test.bat                # Single AppContainer test
+test_scripts\run_isolation_session_tests.ps1     # IsolationSession E2E (requires IsoEnvBroker host)
```
bbonaby: note: You probably want to update the build.bat file too so any agent session stuff gets compiled and built. E.g the tools folder you're adding.  We'd also need to update the azure pipelines to make sure it's building the new generated isolated session stuff if needed. That can be handled in a separate PR though.

external/windows-sdk/isolation-environment-session/README.md
adpa-ms: I believe so, though "a couple of months" might be optimistic. I'm happy to make an issue tracking this.

src/isolation_session_bindings/src/lib.rs
adpa-ms: Actually I intentionally made the *generation* of the bindings a separate manual step that won't interfere with the rest of the MXC build. This currently needs a WinMD from internal OS builds and shouldn't be something MXC devs should have to do.

tools/generate-isolation-session-bindings/src/main.rs
adpa-ms: Yes, exactly. This shouldn't even need to be built on every build. Just when new bindings are needed because the OS updates APIs underneath us.

.github/copilot-instructions.md
adpa-ms: The binding generation shouldn't be happening build over build, the bindings themselves are committed to the source. However, I will take a look and confirm that everything else is set up appropriately.

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: Fair point, though this is also just "v0.1". The underlying APIs are currently being updated in the OS. I'll add one-line comments for now since they will very likely get obsoleted soon. You are right, these just use their defaults in the one-shot runner.

external/windows-sdk/isolation-environment-session/README.md
adpa-ms: Filed as #225 to track when the IsolationSession OS APIs land in windows-rs upstream

src/isolation_session_bindings/build.rs
adpa-ms: Done - added semver = "1" as a [build-dependencies] entry in isolation_session_bindings/Cargo.toml and replaced the starts_with comparison with a VersionReq caret match

tools/generate-isolation-session-bindings/src/main.rs
adpa-ms: Claude found a way to remove the lookup entirely that seems to work: using "default"

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: Fixed the system drive thing. For now just sticking to cmd though.

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: Done - introduced IsolationSessionError with three variants

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: Added some defensive code.

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: The underlying IsoEnvBroker OS API doesn't seem to care so I'm not bothering either.

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: I think I want to defer this to after whatever the "state-aware" design ends up being, we might not need to use the manager in this way.

.github/copilot-instructions.md
adpa-ms: Ok I think I've fixed it now

docs/isolation-session-plan.md
```diff
@@ -0,0 +1,395 @@
+# MXC IsolationSession Backend — Design Document
```
bbonaby: suggestion: oh forgot about this, but can we move this under a `isolation-session` folder inside `docs/`? When your other changes come in we'd probably want the md files all inside the same folder versus at the top level

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -0,0 +1,869 @@
...
+    "filesystem policy is not supported by the isolation session backend";
+pub(crate) const ERR_NETWORK_POLICY: &str =
+    "network policy is not supported by the isolation session backend";
+pub(crate) const ERR_PROXY_POLICY: &str =
+    "network proxy is not supported by the isolation session backend";
+
+/// Validates that the request does not contain policy fields unsupported by
+/// the isolation session backend. Returns `Ok(())` if valid, or a
+/// `Policy`-variant error on rejection.
+pub(crate) fn validate_policy(request: &CodexRequest) -> Result<(), IsolationSessionError> {
```
bbonaby, line 94: issue: I think you need to merge with `main`. I updated the script runner trait so validation is handled consistently across runners:
  - I added an internal `execute` trait function. This has the same body and return type that `run()` has, and runners now put their execution logic there. For you it should be as simple as just changing 551 from `run` to `execute`.
  - I kept `run` as the public interface for starting runner execution. `run` now has a default implementation that calls `validate_runner` first, then calls `execute` if validation succeeds. So now runners can implement `validate_runner` to define their validation logic.

The goal with that was to give all runners a unified input-validation flow before execution.

test_scripts/run_isolation_session_tests.ps1
```diff
@@ -0,0 +1,167 @@
...
+    'aarch64-pc-windows-msvc'
+}
+
+if ($WxcExePath) {
+    $WxcExec = $WxcExePath
+} else {
+    $CandidatePaths = @(
+        (Join-Path $RepoRoot "src\target\$HostTarget\release\wxc-exec.exe"),
+        (Join-Path $RepoRoot "src\target\$OtherTarget\release\wxc-exec.exe"),
+        (Join-Path $RepoRoot "src\target\release\wxc-exec.exe")
```
bbonaby: question: do we want to care about folks that built wxc using debug mode `src\target\debug\wxc-exec.exe`?

docs/isolation-session-plan.md
adpa-ms: Done — moved to `docs/isolation-session/initial-bringup-plan.md`. Filename pinned to "initial-bringup-plan" rather than just "plan" so future Phase 2 / state-aware docs get a clean adjacent slot. Updated the H1 inside the file to match ("Initial Bringup Plan") and the link in `.github/copilot-instructions.md`.

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: Done — rebased onto fresh `origin/main` (5 commits, including #211). `IsolationSessionRunner` now implements `validate_runner` (calls the existing `validate_policy` and maps the error via the `From<IsolationSessionError> for ScriptResponse` impl) and `execute`; the inline policy check at the top of the old `run` body is gone. Pre-push gates green in both feature variants (build/test/clippy/fmt with and without `--features isolation_session`), plus 2/2 manual VM integration tests.

test_scripts/run_isolation_session_tests.ps1
adpa-ms: Yes — appended three debug-flavored paths after the three release paths so a release build still wins on collision. Kept it parameter-free; the simple append matches the spirit of the closest peer scripts (`run_microvm_tests.ps1` defaults to `target/debug/`, `run_windows_sandbox_tests.ps1` supports both via `$Profile`).

bbonaby (approved):

bbonaby (approved):

PR #230, opened 2026-04-29
Add real-time diagnostic logging and shared diagnostic console

jsidewhite:
Launch mxc-diagnostic-console.exe to get real-time output from each layer of the MXC runtime stack.
1. SDK (.ts)
2. Runtime: wxc-exec
3. OS APIs
4. Kernel (via Learning Mode)

# Feature notes
- mxc-diagnostic-console.exe: shared named pipe server with ETW consumer for live inspection of MXC execution across processes
- wxc-exec.exe: emits structured logs (preamble, config, lifecycle phases) to the diagnostic pipe when enabled
- SDK: TypeScript diagnostic module connects to the pipe and logs sandbox spawn parameters
- Controlled via registry (ConsoleEnabled) or env (MXC_DIAG_CONSOLE=1)

image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/230)


jsidewhite:
@copilot resolve the merge conflicts in this pull request #Resolved

sdk/src/diagnostic.ts
```diff
@@ -0,0 +1,153 @@
...
+        return diagnosticEnabled;
+    }
+
+    // Environment variable takes precedence.
+    const envVal = process.env['MXC_DIAG_CONSOLE'];
+    if (envVal === '1' || envVal === 'true') {
+        diagnosticEnabled = true;
+        return true;
+    }
+    if (envVal === '0' || envVal === 'false') {
```
bbonaby: question: do we need one for null/undefined? and I think case might matter here  for the true and false strings #ByDesign

docs/diagnostics.md
```diff
@@ -0,0 +1,72 @@
...
+| **Runtime** | `wxc-exec.exe` (Rust) | Input config, parsed request, sandbox spec, process lifecycle, timing |
+| **OS** | Tessera ETW provider | Kernel-side sandbox creation and validation events |
+| **Internals** | Kernel-General ETW (learning mode) | Access checks that would have been denied, logged instead of blocked |
+
+All layers stream into a single `mxc-diagnostic-console.exe` window in real time.
+
+## Quick Start
+
+```powershell
+# Terminal 1: start the diagnostic console (run as admin for ETW)
```
bbonaby, line 17: question: does it really need admin for this? #ByDesign

docs/diagnostics.md
```diff
@@ -0,0 +1,72 @@
...
+A unified diagnostic view across every layer of the MXC stack:
+
+| Layer | Source | What you see |
+|-------|--------|--------------|
+| **SDK** | `mxc-sdk` (TypeScript) | SDK version, policy construction |
+| **Runtime** | `wxc-exec.exe` (Rust) | Input config, parsed request, sandbox spec, process lifecycle, timing |
+| **OS** | Tessera ETW provider | Kernel-side sandbox creation and validation events |
+| **Internals** | Kernel-General ETW (learning mode) | Access checks that would have been denied, logged instead of blocked |
+
+All layers stream into a single `mxc-diagnostic-console.exe` window in real time.
```
bbonaby, line 12: question:does it need to be a separate exectuable? #ByDesign

docs/diagnostics.md
```diff
@@ -0,0 +1,72 @@
...
+```powershell
+New-Item -Path "HKLM:\SOFTWARE\Microsoft\MXC\Diagnostics" -Force
+Set-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\MXC\Diagnostics" -Name "ConsoleEnabled" -Value 1 -Type DWord
+```
+
+## Configuration
+
+| Method | Setting | Description |
+|--------|---------|-------------|
+| Registry | `HKLM\...\MXC\Diagnostics\ConsoleEnabled` = 1 | Machine-wide, persistent |
```
bbonaby: question: should we remove registry part for now and just use environment variable? Note sure folks would be onboard for a registry addition again.  #ByDesign

docs/diagnostics.md
```diff
@@ -0,0 +1,72 @@
...
+| Env var | `MXC_DIAG_CONSOLE=1` | Per-session, no admin needed (takes precedence) |
+
+## What Gets Logged
+
+- Input JSON config and parsed `CodexRequest` (env values redacted, script truncated)
+- Sandbox spec details (size, UI flags, capabilities, filesystem/network policy)
+- Process lifecycle (command line, identity, child PID, exit code, elapsed time)
+- Phase markers for key execution stages
```
bbonaby: note: "phase markers" might be ambiguous with the new lifecycle apis. #Resolved

sdk/src/diagnostic.ts
```diff
@@ -0,0 +1,153 @@
...
+        diagnosticEnabled = false;
+        return false;
+    }
+
+    // On Windows, check registry (best-effort).
+    if (os.platform() === 'win32') {
+        try {
```
bbonaby: note: fyi there is a npm package that can be used for this without spinning up a new executable :https://www.npmjs.com/package/@vscode/windows-registry #WontFix

sdk/src/diagnostic.ts
```diff
@@ -0,0 +1,153 @@
...
+            const result = execSync(
+                'reg query "HKLM\\SOFTWARE\\Microsoft\\MXC\\Diagnostics" /v ConsoleEnabled 2>nul',
+                { encoding: 'utf-8', timeout: 1000 },
+            );
+            if (result.includes('0x1')) {
+                diagnosticEnabled = true;
+                return true;
+            }
+        } catch {
+            // Registry key doesn't exist or access denied.
```
bbonaby: question: would you want to log this? #Resolved

sdk/src/diagnostic.ts
```diff
@@ -0,0 +1,153 @@
...
+        });
+        socket.on('close', () => {
+            pipeConnected = false;
+            pipeSocket = null;
+        });
+        // Don't let the pipe keep the process alive.
+        socket.unref();
+
+        pipeSocket = socket;
+        pipeConnected = true;
```
bbonaby: question: in the connect handler we set this to true. If we set it to true here, is it possible that callers attempt to write to the pipe before the pipe is actually connected? #Resolved

sdk/src/diagnostic.ts
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * Messages are best-effort and non-blocking. If the console is not
+ * running, the message is silently dropped.
+ *
+ * Each message is sent as a newline-delimited JSON envelope so the
+ * console can split coalesced stream writes back into individual messages.
+ *
+ * @param message - The log message to send
+ */
+export function diagLog(message: string): void {
```
bbonaby, line 120: question: should this be an async function since it's writing to the pipe or is this a quick function? #WontFix

sdk/src/sandbox.ts
```diff
@@ -8,6 +8,34 @@ import { randomBytes } from "crypto";
...
+let sdkVersionLogged = false;
+
+/** Log the SDK version to the diagnostic console (once per process). */
+function diagLogVersion(): void {
+    if (!sdkVersionLogged) {
+        sdkVersionLogged = true;
+        const sdkPath = require.resolve('@microsoft/mxc-sdk/package.json').replace(/[/\\]package\.json$/, '');
+        diagLog(`mxc-sdk v${SDK_VERSION} (PID ${process.pid}) @ ${sdkPath}`);
+    }
+}
```
bbonaby: issue: so we're trying to make this file  as slim as possible since it's the file with the main public apis. So, we should move these functions to helpers.ts. That said I also think it would be easier to just have the `SDK_VERSION` be a  normal variable with a value, instead of attempting to look at the package.json file at runtime with a lambda. In the build pipeline we can make sure this version matches what's in package.json so the build fails if they're different. This way the sdk's runtime code doesn't take the perf hit. #Resolved

sdk/src/index.ts
```diff
@@ -48,18 +48,24 @@ export {
...
 export {
   FileLogger,
-} from './logger';
+} from './logger';
+
+// Export diagnostic logging
+export {
+  diagLog,
+  diagClose,
+} from './diagnostic';
```
bbonaby: question: do these actually need to be exposed as public apis to sdk consumers here? I'm just noticing  the `FileLogger` above is exposed too, which probably shouldn't be. #Resolved

src/wxc_common/src/diagnostic.rs
```diff
@@ -0,0 +1,351 @@
...
+        Self { console_enabled }
+    }
+
+    /// Check whether `ForceLearningMode` is enabled via registry.
+    ///
+    /// When `HKLM\SOFTWARE\Microsoft\MXC\Diagnostics\ForceLearningMode` is set
+    /// to DWORD 1, the `learningModeLogging` capability is injected into the
+    /// container policy regardless of what the config specifies.
+    pub fn force_learning_mode() -> bool {
+        let subkey_wide: Vec<u16> = REGISTRY_SUBKEY
```
bbonaby: issue: we have a `string_utils` module for dealing with wide strings #Resolved

src/wxc_common/src/diagnostic.rs
```diff
@@ -0,0 +1,351 @@
...
+            .encode_utf16()
+            .chain(std::iter::once(0))
+            .collect();
+
+        let mut hkey = HKEY::default();
+        let result = unsafe {
```
bbonaby: note: for this and other registry stuff in this file. the winreg crate is usually used when needing to interact with the registry https://crates.io/crates/winreg.  #Resolved

src/wxc_common/src/diagnostic.rs
```diff
@@ -0,0 +1,351 @@
...
+        val
+    }
+}
+
+/// Produce a redacted JSON representation of a `CodexRequest` suitable for diagnostic logging.
+///
+/// - Environment variable values are replaced with `<redacted>`.
+/// - `script_code` is truncated to [`SCRIPT_CODE_TRUNCATE_LEN`] characters.
+/// - `network_proxy` (which is `#[serde(skip)]`) is logged separately.
+pub fn redacted_request_json(request: &CodexRequest) -> String {
```
bbonaby, line 115: question: do we need to log the json at all? I assumed if its a tool for diagnostics the developer would already know what json was passed in. #WontFix

src/wxc_common/src/diagnostic.rs
```diff
@@ -0,0 +1,351 @@
...
+        Err(_) => return "unknown".to_string(),
+    };
+
+    let mut entry = PROCESSENTRY32W {
+        dwSize: std::mem::size_of::<PROCESSENTRY32W>() as u32,
+        ..Default::default()
+    };
+
+    let mut parent_pid = None;
+    unsafe {
```
bbonaby, line 199: issue: safety comments needed in this file #Resolved

src/wxc_common/src/diagnostic.rs
```diff
@@ -0,0 +1,351 @@
...
+                let mut len = buf.len() as u32;
+                let name = if QueryFullProcessImageNameW(
+                    handle,
+                    PROCESS_NAME_FORMAT(0),
+                    windows::core::PWSTR(buf.as_mut_ptr()),
+                    &mut len,
+                )
+                .is_ok()
+                {
+                    let full = String::from_utf16_lossy(&buf[..len as usize]);
```
bbonaby: issue: we'd want to use our `string_utils` module for this conversion #Resolved

src/wxc_common/src/logger.rs
```diff
@@ -36,6 +66,145 @@ impl Logger {
...
+                eprintln!(
+                    "[MXC Diagnostics] Hint: Start mxc-diagnostic-console.exe first, \
+                     or disable console logging."
+                );
+            }
+        }
+
+        /// Verify the pipe server process is running at High integrity level or above.
+        fn verify_server_integrity(pipe_file: &std::fs::File) -> Result<(), String> {
+            unsafe {
```
bbonaby: issue: it'd be preferable if each function call had the unsafe label and safety  comment rather than the whole function, otherwise we'd want to mark the function itself as usafe. #Resolved

src/wxc_common/src/logger.rs
```diff
@@ -36,6 +66,145 @@ impl Logger {
...
+                let integrity_rid = *windows::Win32::Security::GetSidSubAuthority(
+                    sid,
+                    (sub_authority_count - 1) as u32,
+                );
+
+                let _ = CloseHandle(token);
+                let _ = CloseHandle(process);
+
+                // SECURITY_MANDATORY_HIGH_RID = 0x3000
+                const SECURITY_MANDATORY_HIGH_RID: u32 = 0x3000;
```
bbonaby: note: you might be able to actually find this const in the windows-rs crate. #Resolved

src/mxc_diagnostic_console/src/etw.rs
```diff
@@ -0,0 +1,667 @@
...
+// Licensed under the MIT License.
+
+//! Real-time ETW consumer for MXC diagnostic providers.
+//!
+//! Listens for events from:
+//! - **Tessera** TraceLogging provider (all events)
+//! - **Microsoft-Windows-Kernel-General** (learning-mode AccessCheckLog only, event ID 0xe)
+//!
+//! Starts a trace session, enables the providers, and delivers decoded events
+//! to the diagnostic console's display channel. Requires administrator privileges.
```
bbonaby, line 11: issue: can this all be done via the rust tracelogging crate? https://crates.io/crates/tracelogging. Or if you want multi platform tracing you can use https://crates.io/crates/tracing-etw. Both crates are made by Microsoft. #WontFix

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+use windows::Win32::System::Registry::{
+    RegCloseKey, RegGetValueW, RegOpenKeyExW, HKEY, HKEY_LOCAL_MACHINE, KEY_READ, RRF_RT_DWORD,
+};
+use windows::Win32::System::Threading::{
+    GetCurrentProcess, OpenProcess, OpenProcessToken, PROCESS_QUERY_LIMITED_INFORMATION,
+};
+use windows_core::PCWSTR;
+
+/// Well-known pipe name shared between console and wxc-exec clients.
+const PIPE_NAME: &str = r"\\.\pipe\mxc-diagnostics";
```
bbonaby: question: If I understand this correctly to get etw working the user would need to run mcx_diagnostic_console elevated right? and this executable is the server for this pipe. I'm not extremely knowledgable about pipes and their vulnerabilities but is there anything to watch for with this elevated process + named pipe combination security wise?

Is the purpose for the named pipe really to get multi process real time logging in a console (I assume we save to a file later)? #WontFix

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+fn main() {
+    // Parse CLI args.
+    let args: Vec<String> = std::env::args().collect();
+    if args.iter().any(|a| a == "--help" || a == "-h") {
+        println!("Usage: mxc-diagnostic-console [--verbose | --minified]");
+        println!();
+        println!("Options:");
+        println!("  --verbose    Show all ETW event properties");
+        println!("  --minified   Show reduced properties for common events (default)");
+        return;
```
bbonaby: issue: we can use the cli rust crate to do this. Check out wxc's main.rs  #Resolved

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+    Message { pid: u32, text: String },
+    /// A client disconnected.
+    Disconnected { pid: u32 },
+    /// An ETW event from the Tessera provider.
+    EtwEvent { pid: u32, text: String },
+}
+
+/// Check whether the current process is running elevated (as admin).
+fn is_elevated() -> bool {
+    unsafe {
```
bbonaby, line 93: issue: safety comments  missing in this file #Resolved

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+    }
+
+    // Check registry.
+    let subkey: Vec<u16> = DIAG_REGISTRY_SUBKEY
+        .encode_utf16()
+        .chain(std::iter::once(0))
+        .collect();
+    let mut hkey = HKEY::default();
+    let status = unsafe {
+        RegOpenKeyExW(
```
bbonaby: issue: see my comments in diagnostics.rs but we can use the winreg crate for registry interaction and our string_utils module for wide strings #Resolved

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum DisplayMode {
+    /// Show all decoded properties.
+    Full = 0,
+    /// Show a reduced set of properties for common event types.
+    Minified = 1,
+}
+
+/// Global display mode, set once at startup from CLI args.
+static DISPLAY_MODE: AtomicU8 = AtomicU8::new(DisplayMode::Minified as u8);
```
bbonaby, line 67: note: for what it's worth since you only want to initialize this once you can use the call_once method on the Once struct to do this https://doc.rust-lang.org/std/sync/struct.Once.html. #WontFix

src/wxc/src/main.rs
```diff
@@ -161,8 +163,78 @@ fn main() {
...
+            .any(|c| c == "learningModeLogging")
+    {
+        request
+            .policy
+            .capabilities
+            .push("learningModeLogging".to_string());
+        true
+    } else {
+        false
+    };
```
bbonaby, line 254: question: since learning mode is only used for appcontainer/base container can this be done in that runner?  #WontFix

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+
+    // Create a security descriptor with a NULL DACL (allows all access).
+    let mut sd = SECURITY_DESCRIPTOR::default();
+    let psd = PSECURITY_DESCRIPTOR(std::ptr::addr_of_mut!(sd).cast());
+    unsafe {
+        InitializeSecurityDescriptor(psd, 1)
+            .map_err(|e| format!("InitializeSecurityDescriptor: {e}"))?;
+        SetSecurityDescriptorDacl(psd, true, None, false)
+            .map_err(|e| format!("SetSecurityDescriptorDacl: {e}"))?;
+    }
```
bbonaby, line 327: question: Is this needed if we can just pass `None` as the last argument to `CreateNamedPipeW` below to get the default security descriptor? #WontFix

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+}
+
+/// Get the ANSI color code and exe name for a PID.
+fn get_pid_info(map: &[(u32, usize, String)], pid: u32) -> (&'static str, &str) {
+    map.iter()
+        .find(|(p, _, _)| *p == pid)
+        .map(|(_, idx, exe)| (PID_COLORS[*idx], exe.as_str()))
+        .unwrap_or((PID_COLORS[0], "?"))
+}
```
bbonaby: suggestion: was somewhat hard to read so wonder if we should just put those 3 in a struct. #Resolved

src/mxc_diagnostic_console/src/main.rs
```diff
@@ -0,0 +1,679 @@
...
+/// Look up the executable name for a process ID. Returns just the filename.
+fn process_exe_name(pid: u32) -> String {
+    unsafe {
+        let handle = OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, false, pid);
+        let handle = match handle {
+            Ok(h) => h,
+            Err(_) => return format!("PID{pid}"),
+        };
+        let mut buf = [0u16; 260];
+        let len = GetModuleFileNameExW(Some(handle), None, &mut buf);
```
bbonaby, line 536: issue:  probably need to look though this file at all the win32 functions and check for error handling. E.g this one on line 627 we never do GetLastError() #Resolved

docs/diagnostics.md
jsidewhite: I can't think of a better way to collate multiple layers of the stack (SDK, wxc-exec runtime, OS etw events) into one place.

What were you thinking?

docs/diagnostics.md
jsidewhite: For ETW yes.  For listening security, I think so, but not 100% sure.

sdk/src/diagnostic.ts
jsidewhite: but then it would never progress to the

`if (os.platform() === 'win32')`

case, would it?

docs/diagnostics.md
jsidewhite: I think that was only in reference to overriding people's sandbox configs (well, and in reference to not cross plat, which i disagreed with).

IMO registry is fine as a gate for logging - but I invite specific critiques.

sdk/src/diagnostic.ts
jsidewhite: AI says,

● socket.write() on a Node.js net.Socket is non-blocking — it buffers internally and writes asynchronously under the hood. Returning a Promise
  would only add complexity for callers with no benefit since we don't await success (best-effort, fire-and-forget).

  Proposed action: No code change. The JSDoc already says "best-effort and non-blocking." The function is intentionally sync/fire-and-forget — 
  socket.write() doesn't block. This is a design choice, not a bug.

src/wxc_common/src/diagnostic.rs
jsidewhite: oooh my... this is soooo much nicer!

src/wxc_common/src/diagnostic.rs
jsidewhite: definitely yes.  it let's you paste the json as-is into the debugger hook. or make your own json config for a test scenario from a real world scenario.

otherwise you gotta reverse-craft your json from a CodexRequest.. so hard.

also just to eye'ing bugs in the translation layer... which we're seeing with the UILIMITs, e.g. IME, bit mask

src/mxc_diagnostic_console/src/etw.rs
jsidewhite: I would love this - but looks like crate is only for emission, not capture.

"The tracelogging crate is for producing events, not consuming them. The tracing-etw crate is also a
producer.

Neither tracelogging nor tracing-etw provides a real-time ETW consumer API. They're for emitting events. Consuming events (what this code
does) requires the low-level StartTrace/ProcessTrace pattern."

src/mxc_diagnostic_console/src/main.rs
jsidewhite: This is a design/security question, not a code change request. The answers:

 1. Yes, elevated for ETW. Pipe messages work without elevation.
 2. The pipe security is already addressed — FILE_FLAG_FIRST_PIPE_INSTANCE prevents squatting, client PIDs are resolved server-side, and 
clients verify the server is High integrity before sending.
 3. Yes, the pipe enables multi-process real-time logging. File saving could be added later.

Proposed action: No code change

src/mxc_diagnostic_console/src/main.rs
jsidewhite: ai says doesn't matter since the store happens exactly once anyway? (this isn't a general purpose lazy eval situation).  i'll do it though if you feel strongly


● This is about DISPLAY_MODE being set once at startup. The current AtomicU8 approach is fine — it's set once in main() and only read
  afterward. Once would be more semantically precise but adds complexity for no real benefit since the store happens once deterministically at
  startup.

src/mxc_diagnostic_console/src/main.rs
jsidewhite: oh man - i was nodding along and then tried it and it doesn't work.

turns out the difference is,

`a security descriptor containing a NULL DACL` (allows all access)
is different than
`a security descriptor pointer that is NULL` (based on creator's token)

sdk/src/diagnostic.ts
jsidewhite: i love it - but the AI told me not to (???)

```
● The native module failed to build (missing Spectre-mitigated libraries). Since @vscode/windows-registry requires node-gyp compilation, it's not a
  lightweight dependency — it brings in native build complexity.

  Revised proposal: Instead of @vscode/windows-registry, let's keep the current approach but clean it up. The reg query shell-out is wrapped in
  try/catch, has a 1-second timeout, and is only called once (cached). It's pragmatic and zero-dependency. We can leave a comment acknowledging the
  alternative.

  Alternatively, we could use PowerShell's Get-ItemProperty instead of reg.exe, but that's not really better.
```

src/wxc/src/main.rs
jsidewhite:
```
● This injects learningModeLogging into the request before it reaches any runner. Branden asks if it should be done in the runner instead.

  The logic here makes sense in main.rs because it modifies the CodexRequest before dispatch — the runner receives an already-configured
  request. Moving it to the runner would mean only the base container runner checks, which is correct since learning mode only applies there.
  However, the current approach is simpler — one place to inject, and if the request goes to a different runner, the extra capability is
  harmless.

  Proposed action: This is a design preference. Moving it into the runner adds coupling (runner needs to know about registry). The current
  approach keeps config injection in one place (main.rs). I'd leave it as-is.
```

honestly i'd prefer to keep it visible at a high-level since it's an injection. and all injections in place.  it's really not an implementation detail of the backends. imo.

src/mxc_diagnostic_console/src/main.rs
jsidewhite: still looks like gibberish

sdk/src/sandbox.ts
jsidewhite: took your suggestion in full - but then undid the hardcoded version ask - i just think it's another failure point.

jsidewhite:
@copilot resolve the merge conflict

jsidewhite:

bbonaby (approved):

PR #231, opened 2026-04-29
Add state-aware sandbox lifecycle API design proposal

adpa-ms:

-----

## Summary

Lands the design proposal for state-aware sandbox lifecycle APIs in MXC. The proposal introduces a 5-phase lifecycle (`provision`, `start`, `exec`, `stop`, `deprovision`), a new opt-in Rust trait (`StatefulSandboxBackend`), and five TypeScript SDK functions keyed by an opaque `SandboxId`. Code changes are deferred to follow-up PRs gated on this design's approval. Tracking issue: #232.

## What's in this PR

**Documentation:**
- `docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md` — 15-section reference design covering SDK, JSON wire contract, Rust layer architecture, plug-in guide, error model, failure semantics, graduation path, and v1 out-of-scope items (1429 lines).
- `docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md` — companion 2-pager for 30-minute walkthroughs (340 lines).

**Implementation:** None — this PR is design-only.

## How to review

For reviewers coming in cold, the recommended reading order:

1. **Start with the overview** (`mxc-state-aware-sandbox-api-overview.md`, ~340 lines) — designed for a 30-minute walkthrough. Covers the public surface and design decisions via tables and minimal prose.
2. **Reach for the reference** (`mxc-state-aware-sandbox-api.md`, ~1429 lines) for any section you want to dig deeper on; it has its own ToC.

Readers may stop after the overview and have enough context for the design meeting; the reference is for follow-up.

## Testing

**Automated:** N/A — documentation-only PR.

**Integration / manual:** N/A.

## Scope notes

This PR is intentionally narrow. Explicitly deferred (each warrants its own PR when implementation begins):
- Rust executor changes: dispatch routing for state-aware requests, the `StatefulSandboxBackend` trait, and the first state-aware backend (likely the IsolationSession backend on top of #217 once it merges).
- TypeScript SDK additions: five new functions, branded `SandboxId` type, typed per-phase `*Config` interfaces, `AbortSignal` cancellation, typed exception classes.
- Wire-format schema additions for `experimental.<backend>.<phase>`.

References #232.

-----
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/231)

### DIFF
--- a/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
+++ b/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
@@ -39,7 +39,7 @@ state-aware calls carry `sandboxId` only.
 
 | MXC layer | What's new | What's unchanged |
 |---|---|---|
-| TypeScript SDK (reference §6) | Five new functions: `provisionSandbox`, `startSandbox`, `execInSandbox` / `execInSandboxAsync`, `stopSandbox`, `deprovisionSandbox`. Branded `SandboxId` type. Per-phase typed `*Config` types per backend. `AbortSignal` cancellation. Typed exception classes. | `spawnSandbox` family preserved. `SandboxPolicy` reused as cross-cutting policy. `SandboxingMethod` extension reused. `*Config` naming convention reused. |
+| TypeScript SDK (reference §6) | Five new functions: `provisionSandbox`, `startSandbox`, `execInSandbox` / `execInSandboxAsync`, `stopSandbox`, `deprovisionSandbox`. Branded `SandboxId` type. Per-phase typed `*Config` types per backend. Per-phase typed `*Result` types per backend. `AbortSignal` cancellation. Typed exception classes. | `spawnSandbox` family preserved. `SandboxPolicy` reused as cross-cutting policy. `SandboxingMethod` extension reused. `*Config` naming convention reused. |
 | JSON wire format (reference §7) | Top-level `phase` discriminator. Top-level `sandboxId`. Per-phase nesting under `experimental.<backend>.<phase>`. Named envelope types as a TypeScript discriminated union. | One-shot configs (no `phase`) work unchanged. Cross-cutting `filesystem` / `network` / `ui` at top level for state-aware too — backends declare per-phase honor. |
 | Rust executor (reference §9) | Dispatch arm for state-aware. New `StatefulSandboxBackend` trait. Rust mirror of the wire envelope (private `Raw*` parser pattern). | `ScriptRunner` trait. Existing one-shot dispatch path. Existing backends unchanged. |
 | Error model (reference §8) | Closed enum of 12 codes. `MxcError` base + per-code subclasses. `details` open object. | Existing one-shot error paths preserved. |
@@ -50,10 +50,10 @@ state-aware calls carry `sandboxId` only.
 | Phase | Valid from state | Resulting state | Output |
 |---|---|---|---|
 | `provision` | (not provisioned) | provisioned | `sandboxId`, optional metadata |
-| `start` | provisioned | running | (success) |
+| `start` | provisioned | running | optional metadata |
 | `exec` | running | running | stdout, stderr, exit code |
-| `stop` | running | provisioned | (success) |
-| `deprovision` | provisioned | (not provisioned) | (success) |
+| `stop` | running | provisioned | optional metadata |
+| `deprovision` | provisioned | (not provisioned) | optional metadata |
 
 A backend that has no native equivalent for a phase implements it as a no-op. Every
 state-aware backend implements all five.
@@ -61,7 +61,7 @@ state-aware backend implements all five.
 ## TypeScript SDK
 
 Types used in the function signatures below. Per-phase `<Phase>SandboxOptions<C>` and
-`ProvisionResult<C>` are typed-generic over the chosen backend; full definitions are in
+`<Phase>Result<C>` are typed-generic over the chosen backend; full definitions are in
 [reference §6.1](./mxc-state-aware-sandbox-api.md#6-typescript-sdk).
 
 ```typescript
@@ -87,7 +87,7 @@ function startSandbox<C extends StateAwareSandboxingMethod>(
   containment: C,
   sandboxId: SandboxId,
   options?: StartSandboxOptions<C>,
-): Promise<void>;
+): Promise<StartResult<C>>;
 
 function execInSandbox<C extends StateAwareSandboxingMethod>(
   containment: C,
@@ -107,19 +107,21 @@ function stopSandbox<C extends StateAwareSandboxingMethod>(
   containment: C,
   sandboxId: SandboxId,
   options?: StopSandboxOptions<C>,
-): Promise<void>;
+): Promise<StopResult<C>>;
 
 function deprovisionSandbox<C extends StateAwareSandboxingMethod>(
   containment: C,
   sandboxId: SandboxId,
   options?: DeprovisionSandboxOptions<C>,
-): Promise<void>;
+): Promise<DeprovisionResult<C>>;
 ```
 
 Each phase's options carry `policy?: SandboxPolicy` (cross-cutting restrictions, shared
 with one-shot), `config?: <Phase>ConfigFor<C>` (per-phase backend config, typed), and
-`signal?: AbortSignal` (cancellation). `execInSandbox` returns an `IPty` for live
-streaming; `execInSandboxAsync` is a buffered convenience that resolves on exit.
+`signal?: AbortSignal` (cancellation). Each non-exec phase returns a typed
+`<Phase>Result<C>`: provision carries `sandboxId` plus optional metadata; start, stop,
+and deprovision carry optional metadata only. `execInSandbox` returns an `IPty` for
+live streaming; `execInSandboxAsync` is a buffered convenience that resolves on exit.
 Existing policy-discovery helpers (`getAvailableToolsPolicy`, etc.) compose into
 `SandboxPolicy` unchanged.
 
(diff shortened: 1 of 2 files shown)


docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,340 @@
...
+| JSON wire format (reference §7) | Top-level `phase` discriminator. Top-level `sandboxId`. Per-phase nesting under `experimental.<backend>.<phase>`. Named envelope types as a TypeScript discriminated union. | One-shot configs (no `phase`) work unchanged. Cross-cutting `filesystem` / `network` / `ui` at top level for state-aware too — backends declare per-phase honor. |
+| Rust executor (reference §9) | Dispatch arm for state-aware. New `StatefulSandboxBackend` trait. Rust mirror of the wire envelope (private `Raw*` parser pattern). | `ScriptRunner` trait. Existing one-shot dispatch path. Existing backends unchanged. |
+| Error model (reference §8) | Closed enum of 12 codes. `MxcError` base + per-code subclasses. `details` open object. | Existing one-shot error paths preserved. |
+| Plug-in surface (reference §11) | Implement `StatefulSandboxBackend`. Define typed per-phase `*Config` interfaces. Document the cross-cutting honor matrix. | Ephemeral-only backends require no changes. |
+
+## Lifecycle
+
+| Phase | Valid from state | Resulting state | Output |
+|---|---|---|---|
+| `provision` | (not provisioned) | provisioned | `sandboxId`, optional metadata |
```
jeffstall, line 53: We'll also have direct paths, from `provision` to `deprovision`, etc.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,340 @@
...
+function provisionSandbox<C extends StateAwareSandboxingMethod>(
+  containment: C,
+  options?: ProvisionSandboxOptions<C>,
+): Promise<ProvisionResult<C>>;
+
+function startSandbox<C extends StateAwareSandboxingMethod>(
+  containment: C,
+  sandboxId: SandboxId,
+  options?: StartSandboxOptions<C>,
+): Promise<void>;
```
jeffstall: I'm guessing that startSandbox should return a result of some kind?

adpa-ms: Fixed it.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
adpa-ms: Yes, this is just the general table.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+}
+
+interface NonProvisionStateAwareRequest {
+  phase: 'start' | 'exec' | 'stop' | 'deprovision';
+  sandboxId: SandboxId<StateAwareSandboxingMethod>;  // backend resolved from prefix
+  process?: ProcessConfig;            // exec only
+  filesystem?: FilesystemConfig;
+  network?: NetworkConfig;
+  ui?: UiConfig;
+  experimental?: ExperimentalStateAwareConfigs;
```
bbonaby, line 162: Note: something to think about, is in the future when they aren't experimental and are stable. We'd want `ProvisionStateAwareRequest`, `NonProvisionStateAwareRequest` and `OneShotRequest` to all have a field basically for `StableBackendConfigs` similare to what you have in `ExperimentalStateAwareConfigs`

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+    fn exec(&mut self, sandbox_id: &str, policy: Option<&SandboxPolicy>, config: Option<&Self::ExecConfig>, process: &ProcessConfig)
+        -> Result<ExecHandle, MxcError>;
+    fn stop(&mut self, sandbox_id: &str, policy: Option<&SandboxPolicy>, config: Option<&Self::StopConfig>)
+        -> Result<StopResult<Self::StopMetadata>, MxcError>;
+    fn deprovision(&mut self, sandbox_id: &str, policy: Option<&SandboxPolicy>, config: Option<&Self::DeprovisionConfig>)
+        -> Result<DeprovisionResult<Self::DeprovisionMetadata>, MxcError>;
+}
+```
+
+Backends declare per-phase config and metadata as associated types (use `()` for
```
bbonaby, line 220: note: for the implementation of the trait you can actually create default trait functions for all the methods. They will return `Ok(<some empty value>)`. Then the backend will only need to implement the trait functions is cares about.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+phases that don't need either). The dispatch layer deserialises configs into the typed
+shape before invoking the trait method; result structs (`StartResult<M>`,
+`StopResult<M>`, `DeprovisionResult<M>`) carry an optional metadata payload, parallel
+to `ProvisionResult<M>`. `SandboxPolicy` is the typed Rust mirror of the SDK type;
```
bbonaby, line 224: note: the `SandboxPolicy` is only an SDK concept and is just a subset of what's in the schema. The filesystem`, `networking` and `ui` fields etc are just regular json fields in the request parsed by the `config_parser` in the Rust code. All that to say we probably don't want to use the concept of policy in the rust layer and stick to configs.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+§5 covers id prefixes; reference §9 describes the Rust mirror struct and dispatch.
+
+## Worked example: IsolationSession
+
+Provision and exec — the two most distinctive shapes. Reference §7.4 has all five phases
+(provision, start, exec, stop, deprovision) end-to-end.
+
+#### Provision
+
+```typescript
```
bbonaby, line 240: note: I think this example might need to get updated based on the api changes and based on the reference .md

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+
+```json
+{ "result": { "sandboxId": "iso:reg-abc:prov-123", "metadata": { "agentUserName": "_iso_abc_123" } } }
+```
+
+#### Exec (buffered)
+
+```typescript
+// sandboxId from the provision example above.
+const r = await execInSandboxAsync(sandboxId, {
```
bbonaby, line 277: note: same with this example

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+- exit code: `0`
+
+The SDK constructs `{ stdout: "hello\n", stderr: "", exitCode: 0 }` from PTY events.
+
+The SDK auto-wraps backend-specific config under `experimental.<backend>.<phase>`.
+Cross-backend exec fields flow through top-level `process`. `SandboxPolicy` fields map to
+top-level `filesystem` / `network` / `ui` for state-aware (backend declares per-phase
+honor per reference §10.3).
+
+## Error codes
```
bbonaby, line 308: question: If I'm reading this right, this mean the SDK will need to check the json inside the std output and/or std error produced wxc-exec to know the result of one of the lifecycle functions right? E.g wxc-exec's main for example returns an integer, so the sdk will need to do some parsing to capture any structured output and then relay that to the caller via your Result methods.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -0,0 +1,357 @@
...
+3. Define typed `*Config` interfaces in `@microsoft/mxc-sdk` and slot into
+   `ExperimentalStateAwareConfigs`. If newly SDK-exposed, extend `SandboxingMethod` and
+   `StateAwareSandboxingMethod`.
+4. Register a variant in the `ContainmentBackend` enum, declare the backend's id prefix
+   alongside the variant (it serves as the dispatcher's routing key for non-provision
+   calls), and add a dispatch arm.
+5. Add `Raw*` intermediate structs in `config_parser.rs` for the backend's wire-format
+   block.
+6. Document policy-honor matrix, idempotence, concurrency, and error mapping in
+   `docs/<backend>-plan.md`.
```
bbonaby, line 343: suggestion: would prefer  `docs/<backend/feature>/design-plan.md`

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -0,0 +1,1481 @@
...
+The runtime value is a plain string; the brand exists at compile time only. The
+`__mxcBackend` phantom field carries the backend identity through the type system so
+non-provision SDK calls can infer their backend from the id without the caller restating
+it. The brand also prevents callers from accidentally passing other strings (a
+`containerId`, a path, a literal) where a `SandboxId` is expected.
+
+Persisting the id between calls is the caller's responsibility. The caller chooses the
+storage mechanism. MXC neither tracks the id after `provision` returns nor verifies its
+validity until the caller passes it back. If an id refers to a resource that no longer
+exists, the next call returns `error.code: "stale_id"` (§8); the caller decides whether
```
bbonaby, line 195: note: if we don't track or store this I don't see how we'd be able to know when to return this `error.code: "stale_id"`. I think it's ok to leave all of that to the sdk caller and we will just pass it down. It'd be on the underlying container system to know if it's stale or not and return an error, which the backend runner would then surface.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -0,0 +1,1481 @@
...
+  stderr: string;
+  exitCode: number;
+}
+
+// Per-phase options interfaces. All carry the cross-cutting `policy: SandboxPolicy`
+// (single source of truth, shared with one-shot), the per-phase backend `config`
+// (typed per backend), and an optional `signal` for cancellation.
+interface ProvisionSandboxOptions<C extends StateAwareSandboxingMethod> {
+  policy?: SandboxPolicy;
+  config?: ProvisionConfigFor<C>;
```
bbonaby, line 325: note: we'd really want these apis to take the config and not both policy + config, since the config is what in the end is passed to wxc-exec and lxc-exec. The config should already have info derived from policy inside it already. Also note that the "Policy" aspect of MXC is destined to change as Aegis comes a long, so we don't want to keep tying things to that "SandboxPolicy" object we have today. All that to say the config should be the currency for these. No issue with AbortSignal.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -0,0 +1,1481 @@
...
+  execInSandbox,
+  execInSandboxAsync,
+  stopSandbox,
+  deprovisionSandbox,
+  getAvailableToolsPolicy,
+  SandboxPolicy,
+} from '@microsoft/mxc-sdk';
+
+const tools = getAvailableToolsPolicy();
+const policy: SandboxPolicy = {
```
bbonaby, line 428: Note: see my comment about policy usage above for when you do your implementation. Can these be per backend provision Configs that are exposed to the user? This way only the supported objects and fields are available to the user. Then file system, networking and ui aren't always jammed in even if it's not supported for that backend.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -0,0 +1,1481 @@
...
+
+**Exec phase, dispatch succeeded**: the script's stdout/stderr stream live, matching
+ProcessContainer's existing behaviour. The executor's stdout/stderr inherit through to
+the SDK (PTY by default; piped via `child_process` in non-PTY mode). Process exit code is
+the script's exit code. No JSON envelope is emitted in this case — the SDK constructs
+`{ stdout, stderr, exitCode }` from PTY/process events, exactly as `spawnSandboxAsync`
+does.
+
+**Exec phase, dispatch failed**: the executor emits a JSON envelope on stdout (same shape
+as non-exec phases) and exits non-zero. No script ever runs, so stdout otherwise is empty
```
bbonaby, line 632: note: when --debug is passed to wxe-exec and lxc-exec we output data to stdout and stderr. So it's not always the case that they will be empty.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -0,0 +1,1481 @@
...
+```rust
+/// Dispatch outcome. Distinguishes structured-envelope responses (non-exec phases or
+/// dispatch failure) from exec success (where stdio has already streamed live through
+/// the relay).
+enum DispatchOutcome {
+    Envelope(ResponseEnvelope),
+    ExecCompleted { exit_code: i32 },
+}
+
+fn run(req: MxcRequest) -> Result<DispatchOutcome, MxcError> {
```
bbonaby, line 1061: note: for the current script runners we have a validate_common and validate_runner methods that are called in the run method so you'll have to make sure these get called in your implementation as well.

bbonaby (approved):

jeffstall (approved):
Thanks for organizing!

PR #263, opened 2026-05-06
Add ISV integration guidance to SDK README

denelon:
## 📖 Description

Add practical integration guidance to the SDK README for external consumers (ISVs integrating MXC into agentic applications). The npm readme is the first thing consumers read, and currently lacks coverage of common friction points.

## 🔗 References

Resolves https://github.com/microsoft/mxc/issues/262

## 🔍 Validation

Documentation-only change. Verified markdown renders correctly and code examples are syntactically valid TypeScript.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [x] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/263)


sdk/README.md
```diff
@@ -39,6 +39,7 @@ Then reference it from your project (e.g., via `npm link` or a relative path in
 
 **Requirements**:
 - **Windows**: Windows 11 build 26100+ with UBR ≥ 7965 (for builds 26100–26500)
+- **Windows**: The host process must run **elevated (Administrator)** to create AppContainer-based sandboxes
```
bbonaby: So this shouldn't be required to create app containers/process containers.  That said, right now the only thing that admin is required for as it relates to mxc is if the user used `schema/policy` version `0.4.0-alpha` and either used:
1. Firewall network enforcement: `enforcementMode: firewall` or `enforcementMode: both`. This is where the user can allow/deny hosts. 
Here is a snippet:

```json
"network": {
    "defaultPolicy": "allow",
    "enforcementMode": "both",
    "allowedHosts": [],
    "blockedHosts": ["1.1.1.1"]
  },
```
Unfortunately, only ip addresses can be allowed/denied currently, and even then only the be blockedHosts works as far as I know. 

2. **OR** `Proxy` configuration in the network part of their json. That is because the APIs used require elevation and are called in MXC directly. In this case the user would receive a UAC prompt.

Otherwise if `0.5.0-alpha+` policy/version is used then:
1.  `firewall enforcement` configuration is not implemented/not used
2. Proxy configuration is implemented and used but the API calls themselves are called within the OS itself. No UAC prompt. 

That said, for Windows process containment, if `0.5.0-alpha` is used then it would require a build of Windows that has our new `CreateProcessInSandbox` API (4D WIP+), otherwise we tell users to use `0.4.0-alpha` which uses APIs already in Windows prior to us adding `CreateProcessInSandbox`. I'll make a GitHub issue so this firewall elevation stuff doesn't fall through the cracks and gets update/documented

sdk/README.md
```diff
@@ -431,6 +432,120 @@ console.log('Output:', result.stdout);
...
+
+```typescript
+const policy: SandboxPolicy = {
+  version: '0.4.0-alpha',
+  filesystem: { readonlyPaths: tools.readonlyPaths },
+  timeoutMs: 30000, // 30 seconds
+};
+```
+
+## Recommended Policy for Agentic Workloads
```
bbonaby: note: One thing to keep in mind is that the `spawnSandbox` and `spawnSandboxAsync` APIs only spin up process containers based on the OS since process. 

If an agentic workload wants to use anything other than process containers, e.g wslc for example, they would use the more advanced `createConfigFromPolicy` to set any specific details for the container they want and then call `spawnSandboxFromConfig`.

sdk/README.md
```diff
@@ -431,6 +432,120 @@ console.log('Output:', result.stdout);
 console.log('Exit code:', result.exitCode);
 ```
 
+## Common Pitfalls
+
+### `spawnSandboxAsync` merges stdout and stderr
```
bbonaby: note: right now the only spawnSandbox* API that supports the ability to see the separated stdout and stderr streams is the `spawnSandboxFromConfig` API. That said, there isn't a reason why we couldn't augment both `spawnSandboxAsync` and `spawnSandbox` to do the same thing. Not for this docs PR but just wanted to mention that we could do it as a feature request.

sdk/README.md
```diff
@@ -431,6 +432,120 @@ console.log('Output:', result.stdout);
...
+
+```typescript
+import { createConfigFromPolicy, spawnSandboxFromConfig } from '@microsoft/mxc-sdk';
+
+const config = createConfigFromPolicy(policy);
+config.process.commandLine = 'python -c "print(\'hello\')"'; // ← Required!
+const pty = spawnSandboxFromConfig(config);
+```
+
+### Elevation failures on Windows
```
bbonaby: see my comment above about elevation. If `spawnSandbox` throws due to needing elevation other than from what I mentioned above then I think that's a bug on us.

sdk/README.md
```diff
@@ -431,6 +432,120 @@ console.log('Output:', result.stdout);
...
+
+If the host process is not elevated, `spawnSandbox` will throw. The error message may not immediately indicate a privilege issue. Always ensure your host process runs as Administrator during development and testing.
+
+### Timeouts default to no limit
+
+If you delegate work to a sandbox, set `timeoutMs` in your policy to prevent runaway processes:
+
+```typescript
+const policy: SandboxPolicy = {
+  version: '0.4.0-alpha',
```
asklar: we should point to the 0.5.0-alpha policy + the requirements (ui policy etc)

sdk/README.md
```diff
@@ -431,6 +432,120 @@ console.log('Output:', result.stdout);
...
+  SandboxPolicy,
+  getAvailableToolsPolicy,
+  getTemporaryFilesPolicy,
+} from '@microsoft/mxc-sdk';
+
+const tools = getAvailableToolsPolicy(process.env);
+const temp = getTemporaryFilesPolicy();
+
+const agentPolicy: SandboxPolicy = {
+  version: '0.4.0-alpha',
```
asklar: same here / others

bbonaby:

jsidewhite (approved):
:shipit:

denelon:

.github/copilot-instructions.md
```diff
@@ -107,7 +107,6 @@ The Rust workspace (`src/`) implements multiple sandboxing backends behind the `
...
-- **CLI** (`cli/`, `mxc-cli`) — thin Commander.js wrapper around the SDK. Depends on `@microsoft/mxc-sdk` via `file:../sdk`.
 
 The SDK auto-discovers native binaries by checking `sdk/bin/<target-triple>/` (npm-packaged) and `src/target/<target-triple>/{release,debug}/` (local dev). The `build.bat`/`build.sh` scripts copy binaries into the SDK bin directory.
```
richiemsft: There is a build-mac.sh as well

.github/copilot-instructions.md
richiemsft, line 1: Add mac section

richiemsft (approved):

PR #264, opened 2026-05-06
Set all UILIMIT bits when UI is disabled, improve debug output from sandbo…

jsidewhite:
…x spec

- Change: ui_restrictions_bitmask to now set UILIMIT_IME and friends when, e.g. base_process_ui.ime=false when ui.disable=true (previously the early-return skipped IME, which was confusing)
- Debug output now reads fields from the built FlatBuffer spec instead of the CodexRequest, verifying the actual bytes sent to the OS
- UILIMIT flags shown as allowed/blocked lists with emoji indicators
- SECTION: lines prefixed with a triangle for visibility

Output appears as,
image

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/264)

### DIFF
--- a/src/wxc_common/src/base_container_runner.rs
+++ b/src/wxc_common/src/base_container_runner.rs
@@ -22,6 +22,9 @@ use windows::Win32::System::Threading::{
 };
 use windows_core::PCWSTR;
 
+use crate::log_symbols::{
+    EMOJI_ALLOWED, EMOJI_BLOCKED, EMOJI_NEUTRAL, EMOJI_SECTION, EMOJI_WARNING,
+};
 use crate::logger::Logger;
 use crate::models::{
     CodexRequest, NetworkEnforcementMode, NetworkPolicy, ProxyAddress, ScriptResponse,
@@ -260,7 +263,7 @@ impl ScriptRunner for BaseContainerRunner {
     }
 
     fn execute(&mut self, request: &CodexRequest, logger: &mut Logger) -> ScriptResponse {
-        let _ = writeln!(logger, "\u{25B6} SECTION: Backend runner 'BaseContainer'");
+        let _ = writeln!(logger, "{EMOJI_SECTION} SECTION: Backend runner 'BaseContainer'");
 
         let run_start = std::time::Instant::now();
 
@@ -296,7 +299,7 @@ impl ScriptRunner for BaseContainerRunner {
             );
         }
 
-        let _ = writeln!(logger, "\u{25B6} SECTION: Build sandbox spec");
+        let _ = writeln!(logger, "{EMOJI_SECTION} SECTION: Build sandbox spec");
 
         // 1. Build the FlatBuffer sandbox spec from the request policy.
         let spec_bytes = Self::build_sandbox_spec(&request);
@@ -314,9 +317,9 @@ impl ScriptRunner for BaseContainerRunner {
             // Log Token
             let _ = writeln!(logger, "[token]");
             let integrity_emoji = if spec.integrity() == IntegrityLevel::system_default {
-                "\u{26AA}" // white circle (neutral)
+                EMOJI_NEUTRAL
             } else {
-                "\u{26A0}" // warning sign
+                EMOJI_WARNING
             };
             let _ = writeln!(
                 logger,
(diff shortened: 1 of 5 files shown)


src/wxc_common/src/base_container_runner.rs
```diff
@@ -382,60 +392,108 @@ impl ScriptRunner for BaseContainerRunner {
...
-            ui_restrictions & Self::UILIMIT_INJECTION != 0,
-        );
+        // Read back all fields from the built FlatBuffer spec for debug verification
+        if let Ok(spec) = sandbox_spec::base_container_layout::root_as_sandbox_spec(&spec_bytes) {
 
-        let _ = writeln!(logger, "SECTION: Load API");
+            // Log Token
+            let _ = writeln!(logger, "[token]");
+            let integrity_emoji = if spec.integrity() == IntegrityLevel::system_default {
+                "\u{26AA}" // ⚪ white circle (neutral)
```
bbonaby: suggestion: should we move these unicode characters to a contants file so they can be shared?

src/wxc_common/src/base_container_runner.rs
```diff
@@ -382,60 +392,108 @@ impl ScriptRunner for BaseContainerRunner {
...
+                    logger,
+                    "  capabilities:    {}", caps);
+            }
+
+            // Log network access
+            let _ = writeln!(logger, "[network]");
+            if let Some(np) = spec.network_policy() {
+                if let Some(proxy) = np.proxy() {
+                    if let Some(url) = proxy.url() {
+                        let _ = writeln!(logger, "  network_policy.proxy.url: {}", url);
```
bbonaby: suggestion: think we can invert some of these to get rid of the nesting?

src/wxc_common/src/base_container_runner.rs
```diff
@@ -382,60 +392,108 @@ impl ScriptRunner for BaseContainerRunner {
...
+                logger,
+                "  win32k_system_calls: {}",
+                if spec.disallow_win32k_system_calls() {
+                    "\u{274C} blocked"
+                } else {
+                    "\u{2705} allowed"
+                }
+            );
+            let r = spec.ui_restrictions();
+            let flags: &[(&str, u64)] = &[
```
bbonaby: suggestion: Should we just move all this pre-work for logging to it's own function in this file rather than have it in the execute function?

sdk/tests/integration/package-lock.json
```diff
@@ -19,12 +19,8 @@
     },
     "../..": {
       "name": "@microsoft/mxc-sdk",
-      "version": "0.1.6",
+      "version": "0.1.8",
```
bbonaby: suggestion: I believe the newest version is 0.2.0, once you merge with main can you make sure these two use 0.2.0 please? I missed updating both `sdk/package-lock.json` and `sdk/tests/integration/package-lock.json`

bbonaby: actually looks like they've been updated already. So once you merge with main they should go away. Resolving this

bbonaby:
looks like you just need to merge with main and fix the conflicts to get all green

jsidewhite:
@copilot please fix the merge conflicts - tyty.

jsidewhite:
@copilot please resolve merge conflicts.

MGudgin:
From looking at the diffs, it seems that this PR is no longer setting UI flags and is only about improved debug output?

src/wxc_common/src/base_container_runner.rs
jsidewhite: fixed

src/wxc_common/src/base_container_runner.rs
jsidewhite: fixed

src/wxc_common/src/base_container_runner.rs
jsidewhite: fixed

jsidewhite:
> From looking at the diffs, it seems that this PR is no longer setting UI flags and is only about improved debug output?

It's about output formatting, and the weird behavior which is that _regardless of whether the current debug output shows IME `true` or IME `false`_ in the CodexRequest...

```
    "base_process_ui": {
      "ime": true|false
    }
```

The IME would be `false` in the sandboxspec output...

```
UILIMIT flags: HANDLES={} READCLIP={} WRITECLIP={} SYSPARAM={} DISPLAY={} ATOMS={} DESKTOP={} EXIT={} IME=**FALSE**" INJECT={}
```

..because setting IME in sandboxspec was shortcircuited if user forgot to specify (ui->disable:false).

bbonaby (approved):

PR #255, opened 2026-05-05
[IsolationSession] Add state-aware lifecycle support to wxc-exec

adpa-ms:
- [x] I have signed the [Contributor License Agreement](https://opensource.microsoft.com/cla/).
- [x] This pull request is related to an issue.
- [x] If this PR changes build commands, project architecture, or key conventions, I have updated [`.github/copilot-instructions.md`](.github/copilot-instructions.md).

-----

## Summary

Closes #254. Adds state-aware lifecycle support to the IsolationSession backend in `wxc-exec`. Each lifecycle phase (provision, start, exec, stop, deprovision) becomes a separate `wxc-exec` invocation, threaded by a `sandboxId` returned from provision. Output streams live on exec (re-uses the one-shot `create_process` path); non-exec phases return a JSON envelope on stdout (`{"result":...}` or `{"error":...}`). The cross-backend Rust contract — a new `StatefulSandboxBackend` trait, dispatcher, parser additions, and error model — compiles unconditionally; IsolationSession's impl stays behind the existing `isolation_session` Cargo feature.

Full design: `docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md`. IsolationSession-specific Rust plan: `docs/isolation-session/state-aware-rust-initial-plan.md`.

## What's in this PR

**Cross-backend infrastructure (new modules in `wxc_common`):**
- `mxc_error` — `MxcError` + 12-variant `MxcErrorCode` + `ErrorEnvelope` + `ResponseEnvelope<T>` wire types.
- `id` — `mint_random_token` (4 bytes via `getrandom`, 8-hex format) + `parse_sandbox_id_prefix`. New `getrandom = "0.2"` workspace dep.
- `state_aware_backend` — `StatefulSandboxBackend` trait (5 phase methods + 5 validate hooks + per-phase `*Result<M>` types + `ExecHandle`); takes `&CodexRequest` + `sandbox_id` + `Option<C>` per design §9.2 (no per-phase wrapper structs, no Rust mirrors of `ProcessConfig` / `FilesystemConfig` / `NetworkConfig` / `UiConfig` — those names exist as TS SDK interfaces only).
- `state_aware_request` — `Phase` enum, `ParsedStateAwareRequest`, public `MxcRequest::OneShot/StateAware` discriminator.
[...]

PR #265, opened 2026-05-06
[IsolationSession] Add state-aware lifecycle support to TypeScript SDK

adpa-ms:
## 📖 Description

The SDK-side complement to #255. Adds five new functions plus the supporting types that callers use to drive a sandbox through `provision -> start -> exec -> stop -> deprovision` against IsolationSession (the only state-aware backend today; future backends opt in by extending `StateAwareSandboxingMethod` and adding an arm to `ConfigsForBackend`).

New public functions:

- `provisionSandbox(containment, config?, options?) -> Promise<ProvisionResult<C>>`
- `startSandbox(sandboxId, config?, options?) -> Promise<StartResult<C>>`
- `execInSandbox(sandboxId, config, options?) -> IPty` (streaming)
- `execInSandboxAsync(sandboxId, config, options?) -> Promise<ExecResult>` (buffered)
- `stopSandbox(sandboxId, config?, options?) -> Promise<StopResult<C>>`
- `deprovisionSandbox(sandboxId, config?, options?) -> Promise<DeprovisionResult<C>>`

Types layer:

- `SandboxId<C>` — branded TypeScript string parameterised by backend.
- Per-(backend, phase) `*Config` interfaces declaring only the fields valid at that phase, with cross-cutting fields (`filesystem` / `network` / `ui`) appearing only in phases the backend's policy honor matrix marks as `applied` — giving compile-time enforcement of the matrix on TypeScript callers.
- `*Result<C>` per phase with backend-typed `metadata` via `StateAwareMetadata` and `<Phase>MetadataFor<C>`.
- Closed `ErrorCode` union plus one typed `MxcError` subclass per code (12 total), mirroring the Rust `MxcErrorCode` enum, plus an `mxcErrorFromCode` helper for the SDK's wire-envelope parser.

Schema version bumps to `0.6.0-alpha`. The Rust parser was capped at `<=0.5` and was rejecting 0.6 envelopes from the SDK at parse time; the wxc-exec change in this PR raises the ceiling to `<=0.6` and adds parser coverage for the new accepted version.

`SandboxSpawnOptions` gains an optional `signal: AbortSignal` honored by the state-aware functions. Cancellation is best-effort: aborting kills the executor process and rejects the pending promise, but OS-side state may need a follow-up `deprovisionSandbox` to clean up.

Note: the design baked into this PR drops `SandboxPolicy` from the state-aware surface in favor of per-(backend, phase) Configs that absorb cross-cutting fields directly. The design docs in `docs/state-aware-lifecycle/` are revised in the first commit on this PR. The umbrella issue's older summary still references `SandboxPolicy` — that summary is design-stale relative to the design as it now ships.

## 🔗 References

Resolves #232 (umbrella issue tracking state-aware sandbox lifecycle APIs).

Rust-side state-aware support: #255.

Design: `docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md` and `docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md` (revised in the first commit on this PR).

## 🔍 Validation

Each commit was validated locally with the four MXC pre-push checks (build + Rust unit tests + clippy + fmt) plus `npm test` in `sdk/`. Final state at the tip of the branch:

- 141 SDK unit tests pass (61 pre-existing + 80 new). The new tests cover the `MxcError` class hierarchy and `mxcErrorFromCode`, the per-(backend, phase) Config matrix and `SandboxId<C>` brand at compile time via `// @ts-expect-error` patterns, and the state-aware function logic (envelope construction, response parsing, error mapping, AbortSignal cancellation, exec dispatch-vs-script-failure discrimination) using an injected fake spawn so unit tests do not depend on a real `wxc-exec` binary.
- Default-features and `--features isolation_session` Rust unit tests pass. The wxc-exec commit adds `schema_version_state_aware_06_accepted` and repoints the existing `schema_version_too_new_rejected` test at `0.7.0` as the new "too new" sentinel.
- Integration tests verified two ways. (1) On a Windows host without an IsolationSession runtime, the WinRT activation in `wxc-exec` returns `IsolationSessionError::ServiceUnavailable` which the dispatcher maps to wire `error.code: "backend_unavailable"`; the SDK throws `MxcBackendUnavailableError` and the new integration tests skip with a clear diagnostic. (2) Interactively on a Windows host with the IsolationSession runtime registered, the full provision -> start -> exec -> stop -> deprovision lifecycle passes end-to-end.

## Out of scope — deferred follow-ups

These items are intentionally not addressed in this PR. Each is a discrete concern that benefits from its own PR.

- IsolationSession detection in `sdk/src/platform.ts`. OS build-number alone is insufficient (IS depends on component dependencies beyond OS version); proper detection belongs in a follow-up, likely via a `wxc-exec --query-capabilities` subcommand the SDK consumes. The integration tests' try-and-skip pattern handles probing implicitly today, so this does not block the test suite.
- `'isolation_session'` in the consumer-facing `ContainmentType` plus a one-shot translation arm in `createConfigFromPolicy`. Today, callers wanting one-shot IS hand-build a `ContainerConfig` and call `spawnSandboxFromConfig`.
- Removing `SandboxPolicy` from the existing one-shot SDK paths. The state-aware surface added here does not use it; the one-shot surface still does and is left untouched in this PR.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Feature

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/265)

### DIFF
--- a/docs/sandbox-policy/v1/policy.md
+++ b/docs/sandbox-policy/v1/policy.md
@@ -55,11 +55,11 @@ Omitted policy fields = most restrictive permissions. Adding a field opts *in* t
 
 ```typescript
 // Fully locked down:
-spawnSandbox("script.sh", { version: "0.5.0-dev" });
+spawnSandbox("script.sh", { version: "0.5.0-alpha" });
 
 // Allow outbound network:
 spawnSandbox("script.sh", {
-  version: "0.5.0-dev",
+  version: "0.5.0-alpha",
   network: { allowOutbound: true },
 });
 ```
@@ -242,7 +242,7 @@ Execution timeout in milliseconds. Omitted = SDK default (no timeout).
 An empty policy is fully locked down:
 
 ```typescript
-spawnSandbox("script.sh", { version: "0.5.0-dev" });
+spawnSandbox("script.sh", { version: "0.5.0-alpha" });
 // No filesystem, no network, no UI, no input injection.
 ```
 
@@ -274,7 +274,7 @@ type ContainerConfig =
 
 ```json
 {
-  "version": "0.5.0-dev",
+  "version": "0.5.0-alpha",
   "containment": "process",
   "process": {
     "commandLine": "node agent.js",
@@ -324,7 +324,7 @@ cross-platform fields mapped from Policy.
 
 ```json
 {
-  "version": "0.5.0-dev",
+  "version": "0.5.0-alpha",
   "containment": "lxc",
   "process": {
     "commandLine": "bash run.sh",
@@ -406,7 +406,7 @@ Only `"process"` is end-to-end implemented today.
 
 ```typescript
 const policy: SandboxPolicy = {
-  version: "0.5.0-dev",
+  version: "0.5.0-alpha",
   filesystem: { readwritePaths: ["C:\\workspace"] },
   network: {},
   ui: {
@@ -560,7 +560,7 @@ thing they modify.
 ### "What happens if I omit all policy fields?"
 
 ```typescript
-spawnSandbox("script.sh", { version: "0.5.0-dev" });
+spawnSandbox("script.sh", { version: "0.5.0-alpha" });
 ```
 
 Produces a fully locked-down config: no filesystem access, no network, UI disabled, no input
--- a/docs/sandbox-policy/v1/policy.md
+++ b/docs/sandbox-policy/v1/policy.md
@@ -55,11 +55,11 @@ Omitted policy fields = most restrictive permissions. Adding a field opts *in* t
 
 ```typescript
 // Fully locked down:
-spawnSandbox("script.sh", { version: "0.5.0-alpha" });
+spawnSandbox("script.sh", { version: "0.5.0-dev" });
 
 // Allow outbound network:
 spawnSandbox("script.sh", {
-  version: "0.5.0-alpha",
+  version: "0.5.0-dev",
   network: { allowOutbound: true },
 });
 ```
@@ -242,7 +242,7 @@ Execution timeout in milliseconds. Omitted = SDK default (no timeout).
 An empty policy is fully locked down:
 
 ```typescript
-spawnSandbox("script.sh", { version: "0.5.0-alpha" });
+spawnSandbox("script.sh", { version: "0.5.0-dev" });
 // No filesystem, no network, no UI, no input injection.
 ```
 
@@ -274,7 +274,7 @@ type ContainerConfig =
 
 ```json
 {
-  "version": "0.5.0-alpha",
+  "version": "0.5.0-dev",
   "containment": "process",
   "process": {
     "commandLine": "node agent.js",
@@ -324,7 +324,7 @@ cross-platform fields mapped from Policy.
 
 ```json
 {
-  "version": "0.5.0-alpha",
+  "version": "0.5.0-dev",
   "containment": "lxc",
   "process": {
     "commandLine": "bash run.sh",
@@ -406,7 +406,7 @@ Only `"process"` is end-to-end implemented today.
 
 ```typescript
 const policy: SandboxPolicy = {
-  version: "0.5.0-alpha",
+  version: "0.5.0-dev",
   filesystem: { readwritePaths: ["C:\\workspace"] },
   network: {},
   ui: {
@@ -560,7 +560,7 @@ thing they modify.
 ### "What happens if I omit all policy fields?"
 
 ```typescript
-spawnSandbox("script.sh", { version: "0.5.0-alpha" });
+spawnSandbox("script.sh", { version: "0.5.0-dev" });
 ```
 
 Produces a fully locked-down config: no filesystem access, no network, UI disabled, no input
--- a/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
+++ b/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
@@ -4,10 +4,10 @@
 *Compiled 2026-04-28.*
 
 A state-aware sandbox API for MXC, surfaced alongside the existing one-shot
-`spawnSandbox*` family. Five lifecycle phases; opaque caller-owned `SandboxId`; typed
-per-phase backend config; cross-cutting `SandboxPolicy` shared with one-shot. Backends
-opt in by implementing a new `StatefulSandboxBackend` Rust trait. MXC retains no state
-between calls.
+`spawnSandbox*` family. Five lifecycle phases; opaque caller-owned `SandboxId`;
+per-(backend, phase) typed Config interfaces that absorb cross-cutting fields directly
+(no separate `SandboxPolicy` parameter). Backends opt in by implementing a new
+`StatefulSandboxBackend` Rust trait. MXC retains no state between calls.
 
 `spawnSandbox` is the composition of the five phases run end-to-end. State-aware exposes
 each phase individually.
@@ -20,12 +20,12 @@ function signatures throughout. One-line summaries; full definitions live in
 
 | Type | Where | Role |
 |---|---|---|
-| `SandboxPolicy` | `sdk/src/types.ts` | Cross-platform restriction policy: filesystem, network, UI, timeout. Reused as the cross-cutting policy on state-aware. |
 | `SandboxingMethod` | `sdk/src/types.ts` | String union of MXC backend names (`'appcontainer' \| 'windows_sandbox' \| 'lxc' \| 'wslc' \| 'vm' \| 'microvm' \| 'isolation_session'`). |
-| `ProcessConfig` | `sdk/src/types.ts` | Per-process settings: `commandLine`, `cwd`, `env`, `timeout`. Reused for state-aware exec. |
-| `FilesystemConfig`, `NetworkConfig`, `UiConfig` | `sdk/src/types.ts` | Wire-format restriction blocks. The `SandboxPolicy` → wire mapping is the existing `createConfigFromPolicy` logic. |
+| `ProcessConfig` | `sdk/src/types.ts` | Per-process settings: `commandLine`, `cwd`, `env`, `timeout`. Reused inside state-aware exec Configs. |
+| `FilesystemConfig`, `NetworkConfig`, `UiConfig` | `sdk/src/types.ts` | Wire-format-aligned cross-cutting interfaces. Reused inline as field types inside the per-(backend, phase) state-aware Configs. |
+| `SandboxSpawnOptions` | `sdk/src/sandbox.ts` | Existing options bag (debug, dryRun, logDir, executablePath, ptyOptions, usePty, experimental). State-aware reuses it as the third positional arg, extended with `signal?: AbortSignal` for cancellation. |
 | `pty.IPty` | `node-pty` package | Interactive PTY handle. Used as the streaming-exec return type, matching existing `spawnSandbox`. |
-| `getAvailableToolsPolicy`, `getUserProfilePolicy`, `getTemporaryFilesPolicy` | `sdk/src/policy.ts` | Filesystem-policy discovery helpers. Compose unchanged with state-aware via `SandboxPolicy`. |
+| `getAvailableToolsPolicy`, `getUserProfilePolicy`, `getTemporaryFilesPolicy` | `sdk/src/policy.ts` | Filesystem-policy discovery helpers. Produce `FilesystemPolicyResult` fragments that compose into a state-aware Config's `filesystem` field. |
 | `ContainmentBackend` (Rust) | `wxc_common::models` | Rust dispatch enum (one variant per backend). State-aware adds `IsolationSession` and future variants. |
 | ProcessContainer | MXC's existing AppContainer-based one-shot backend | Relevant context: ProcessContainer streams stdout/stderr live via PTY; state-aware exec preserves that streaming model. |
 
@@ -40,11 +40,11 @@ on the response, and neither shape carries `containerId`.
 
 | MXC layer | What's new | What's unchanged |
 |---|---|---|
-| TypeScript SDK (reference §6) | Five new functions: `provisionSandbox`, `startSandbox`, `execInSandbox` / `execInSandboxAsync`, `stopSandbox`, `deprovisionSandbox`. Branded `SandboxId<C>` type tagging ids by backend (`containment` named once at provision, inferred from the id thereafter). Per-phase typed `*Config` types per backend. Per-phase typed `*Result` types per backend. `AbortSignal` cancellation. Typed exception classes. | `spawnSandbox` family preserved. `SandboxPolicy` reused as cross-cutting policy. `SandboxingMethod` extension reused. `*Config` naming convention reused. |
+| TypeScript SDK (reference §6) | Five new functions: `provisionSandbox`, `startSandbox`, `execInSandbox` / `execInSandboxAsync`, `stopSandbox`, `deprovisionSandbox`. Branded `SandboxId<C>` type tagging ids by backend (`containment` named once at provision, inferred from the id thereafter). Per-(backend, phase) typed `*Config` interfaces (e.g. `IsolationSessionProvisionConfig`) that absorb cross-cutting fields directly — no separate policy parameter. Per-phase typed `*Result` types per backend. `AbortSignal` cancellation via the existing `SandboxSpawnOptions`. Typed exception classes. | `spawnSandbox` family preserved. `SandboxingMethod` extension reused. The wire-format-aligned `Process` / `Filesystem` / `Network` / `UiConfig` interfaces from `sdk/src/types.ts` are reused as field types inside state-aware Configs. `SandboxSpawnOptions` reused as the third-arg options bag (gains `signal?: AbortSignal`). `*Config` naming convention reused. |
 | JSON wire format (reference §7) | Top-level `phase` discriminator. Top-level `sandboxId`. `containment` carried on provision only; non-provision phases route via the `sandboxId` prefix. Per-phase nesting under `experimental.<backend>.<phase>`. Named envelope types as a TypeScript discriminated union. | One-shot configs (no `phase`) work unchanged. Cross-cutting `filesystem` / `network` / `ui` at top level for state-aware too — backends declare per-phase honor. |
 | Rust executor (reference §9) | Dispatch arm for state-aware. New `StatefulSandboxBackend` trait. Rust mirror of the wire envelope (private `Raw*` parser pattern). | `ScriptRunner` trait. Existing one-shot dispatch path. Existing backends unchanged. |
 | Error model (reference §8) | Closed enum of 12 codes. `MxcError` base + per-code subclasses. `details` open object. | Existing one-shot error paths preserved. |
-| Plug-in surface (reference §11) | Implement `StatefulSandboxBackend`. Define typed per-phase `*Config` interfaces. Register a backend-specific id prefix. Document the cross-cutting honor matrix. | Ephemeral-only backends require no changes. |
+| Plug-in surface (reference §11) | Implement `StatefulSandboxBackend`. Define typed per-(backend, phase) `*Config` interfaces. Declare the trait's `ID_PREFIX` and `BACKEND_KEY` consts. Document the cross-cutting honor matrix. | Ephemeral-only backends require no changes. |
 
 ## Lifecycle
 
@@ -61,8 +61,10 @@ Phases without native equivalents fall through to the trait's default no-op bodi
 
 ## TypeScript SDK
 
-Types used in the function signatures below. Per-phase `<Phase>SandboxOptions<C>` and
-`<Phase>Result<C>` are typed-generic over the chosen backend; full definitions are in
+Each phase takes a per-(backend, phase) Config — a single named interface that absorbs
+both the cross-cutting fields the backend honors at that phase and any backend-specific
+per-phase fields. `SandboxPolicy` does not appear in the state-aware surface. Full
+definitions and worked types are in
 [reference §6.1](./mxc-state-aware-sandbox-api.md#6-typescript-sdk).
 
 ```typescript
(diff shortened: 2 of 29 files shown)


test_scripts/README.md
```diff
@@ -36,6 +36,17 @@ These scripts are local helpers. Not every script is run by CI because several
 depend on local OS features such as Windows Sandbox, WHP, proxy setup, or stress
 test duration.
 
+### Deployment helpers (TShell `putd`)
```
bbonaby: note: probably better to use powershell remoting so non windows folks coming to the repo don't have to know what tshell is

sdk/tests/integration/isolation-session-state-aware.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  try {
+    await sdk.deprovisionSandbox(sandboxId, undefined, { experimental: true });
+  } catch (err) {
+    // Don't override the original test failure. Surface for debugging only.
+    console.error(`Cleanup deprovision failed for ${sandboxId}: ${err}`);
+  }
+}
+
+describe('IsolationSession state-aware lifecycle E2E', {
+  skip: !isWindows ? 'IsolationSession is Windows-only' : undefined,
```
bbonaby: suggestion: I suppose you could actually call a function here to check if Windows + iso session available. Then you can skip all if both are not true versus having to call `runOrSkipIfBackendUnavailable` in every function.

sdk/tests/integration/isolation-session-state-aware.test.ts
```diff
@@ -0,0 +1,111 @@
...
+}, () => {
+  it('full lifecycle: provision -> start -> exec -> stop -> deprovision', async (t) => {
```
bbonaby: suggestion: `it('runs full isolation session lifecycle to completion', async (t) => {`

docs/sandbox-policy/v1/policy.md
```diff
@@ -55,11 +55,11 @@ Omitted policy fields = most restrictive permissions. Adding a field opts *in* t
 
 ```typescript
```
bbonaby, line 56: issue: we'll probably not want to make changes to this file

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
```diff
@@ -195,6 +207,7 @@ SDK parsing rules.
 ```rust
 pub trait StatefulSandboxBackend {
     const ID_PREFIX: &'static str;
+    const BACKEND_KEY: &'static str;
```
bbonaby, line 210: question: Thinking about this more could these be enums? In rust you can use the [strum](https://crates.io/crates/strum) crate and then you can convert between the enum and string easily. Now you can deal with enums instead of strings directly. That said I think this is feedback for the previous PR right, not necessarily for this PR.

sdk/src/errors.ts
```diff
@@ -0,0 +1,144 @@
...
+    case 'stale_id': return new MxcStaleIdError(message, details);
+    case 'not_provisioned': return new MxcNotProvisionedError(message, details);
+    case 'not_started': return new MxcNotStartedError(message, details);
+    case 'already_started': return new MxcAlreadyStartedError(message, details);
+    case 'already_stopped': return new MxcAlreadyStoppedError(message, details);
+    case 'policy_validation': return new MxcPolicyValidationError(message, details);
+    case 'backend_error': return new MxcBackendError(message, details);
+    default: return new MxcError(code as ErrorCode, message, details);
+  }
+}
```
bbonaby, line 58: suggestion: interms of conciseness I think this could all be refactored into something like this. I'm wondering if we really need a class per error for our purposes when the error code would suffice:
```typescript
export const ErrorCodes = {
  MalformedRequest: 'malformed_request',
  UnsupportedContainment: 'unsupported_containment',
  UnsupportedPhase: 'unsupported_phase',
  BackendUnavailable: 'backend_unavailable',
  MalformedId: 'malformed_id',
  StaleId: 'stale_id',
  NotProvisioned: 'not_provisioned',
  NotStarted: 'not_started',
  AlreadyStarted: 'already_started',
  AlreadyStopped: 'already_stopped',
  PolicyValidation: 'policy_validation',
  BackendError: 'backend_error',
} as const;

export type ErrorCode = typeof ErrorCodes[keyof typeof ErrorCodes];

export class MxcError extends Error {
  readonly code: ErrorCode | string;
  readonly details?: Record<string, unknown>;

  constructor(code: ErrorCode | string, message: string, details?: Record<string, unknown>) {
    super(message);
    this.name = 'MxcError';
    this.code = code;
    this.details = details;
    Object.setPrototypeOf(this, new.target.prototype);
  }
}

export function mxcErrorFromCode(
  code: string,
  message: string,
  details?: Record<string, unknown>,
): MxcError {
  return new MxcError(code, message, details);
}
```

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+ * non-exec state-aware calls and the buffered `execInSandboxAsync`. Not
+ * exported from `index.ts` — production code uses the real `spawn`.
+ */
+export function _setSpawnImpl(fn: SpawnImpl): void {
+  spawnImpl = fn;
+}
+
+/** Test-only hook: restore the default `child_process.spawn`. */
+export function _resetSpawnImpl(): void {
+  spawnImpl = spawn as unknown as SpawnImpl;
```
bbonaby: issue: should probably keep test code in separate file so it isn't interleaved with production data code

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+  StopMetadataFor,
+  StopResult,
+} from './state-aware-types';
+
+const STATE_AWARE_VERSION = '0.6.0-alpha';
+
+// Wire-format cross-cutting fields that live at the envelope's top level.
+// Anything else on a per-(backend, phase) Config is backend-specific and is
+// nested under `experimental.<backend>.<phase>`.
+const CROSS_CUTTING_FIELDS = ['filesystem', 'network', 'ui', 'process'] as const;
```
bbonaby: suggestion: maybe we should call it common fields instead of cross cutting so it's meaning is more straight forward

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+
+// Mapping from a sandboxId's leading prefix segment to the wire-format
+// backend key. Extended as more state-aware backends opt in.
+const PREFIX_TO_BACKEND: Record<string, StateAwareSandboxingMethod> = {
+  iso: 'isolation_session',
```
bbonaby: suggestion: prefer to spell out the name or in my opinion we could move from strings to enums for the key in the Record

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+  if (colon < 0) {
+    throw mxcErrorFromCode('malformed_id', `sandboxId must carry a backend prefix: ${sandboxId}`);
+  }
+  const prefix = sandboxId.slice(0, colon);
+  const backend = PREFIX_TO_BACKEND[prefix];
+  if (!backend) {
+    throw mxcErrorFromCode('malformed_id', `sandboxId prefix '${prefix}' does not match a known state-aware backend`);
+  }
+  return backend;
+}
```
bbonaby: question: would all this be avoided if we used an enum instead? The string splicing I mean.

sdk/src/index.ts
```diff
@@ -58,3 +58,61 @@ export {
...
+  MxcMalformedIdError,
+  MxcStaleIdError,
+  MxcNotProvisionedError,
+  MxcNotStartedError,
+  MxcAlreadyStartedError,
+  MxcAlreadyStoppedError,
+  MxcPolicyValidationError,
+  MxcBackendError,
+  mxcErrorFromCode,
+} from './errors';
```
bbonaby: question: do we need to export all of the errors to sdk consumers?

sdk/src/index.ts
```diff
@@ -58,3 +58,61 @@ export {
...
+  ConfigsForBackend,
+  ProvisionConfigFor,
+  StartConfigFor,
+  ExecConfigFor,
+  StopConfigFor,
+  DeprovisionConfigFor,
+  StateAwareMetadata,
+  ProvisionMetadataFor,
+  StartMetadataFor,
+  StopMetadataFor,
```
bbonaby, line 89: note:  looking at it now it does seem weird to have the work `For` after these. Might be better to drop that

sdk/src/state-aware-types.ts
```diff
@@ -0,0 +1,158 @@
...
+
+export interface IsolationSessionStopConfig {
+  /** Schema version (semver). */
+  version?: string;
+}
+
+export interface IsolationSessionDeprovisionConfig {
+  /** Schema version (semver). */
+  version?: string;
+}
```
bbonaby, line 66: suggestion: perhaps these should all inherit from a base type that has a version, filesystem, network and ui. potentially the user might want to call exec multiple times with different settings for the  filesystem, network and ui. Doing it this way will also establish a pattern for future  cones as well.

sdk/tests/unit/state-aware.test.ts
```diff
@@ -0,0 +1,354 @@
...
+    ac.abort();
+    await assert.rejects(promise);
+    assert.ok(activeFake.killCount() >= 1, 'expected child.kill() to fire on abort');
+  });
+});
+
+describe('startSandbox / stopSandbox / deprovisionSandbox', () => {
+  afterEach(() => { _resetSpawnImpl(); });
+
+  it('startSandbox infers backend from sandboxId prefix and nests configurationId under experimental', async () => {
```
bbonaby: suggestion: so  usually the `it(`  should flow so it'd be something like `it('infers backend` rather than `it('startSandbox infers backend`. If it can't flow then I'd suggest having a describe block for each of the functions so the meaning of the tests are scoped.

sdk/tests/unit/state-aware.test.ts
```diff
@@ -0,0 +1,354 @@
...
+  stderr?: string;
+  exitCode?: number;
+  error?: Error;
+}
+
+function fakeSpawn(opts: FakeChildOpts): {
+  spawn: (cmd: string, args: string[], spawnOpts: unknown) => unknown;
+  captured: { cmd?: string; args?: string[]; envelope?: Record<string, unknown> };
+  killCount: () => number;
+} {
```
bbonaby: suggestion: might be good to move these to a common helper file. This seems like it's a good candidate for reuse.

sdk/tests/unit/state-aware-types.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    assert.ok(true);
+  });
+
+  it('runtime value is a string', () => {
+    const id = 'iso:abcd' as SandboxId<'isolation_session'>;
+    assert.strictEqual(typeof id, 'string');
+  });
+});
+
+describe('Per-(backend, phase) Configs', () => {
```
bbonaby: suggestion: perhaps the items in this describe block should should be in a iso-session-types.test.ts file, just to keep it scoped since these are all concerrned with the iso session configs. Also note that usually the `it(` block should flow. So, `  it('IsolationSessionProvisionConfig accepts cross-cutting fields per the policy honor matrix', () => {` would end up as `  it('accepts cross-cutting fields per the policy honor matrix', () => {` while the describe says what it is that will be tested.

sdk/tests/integration/isolation-session-state-aware.test.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+async function safeDeprovision(sandboxId: SandboxId<'isolation_session'>): Promise<void> {
+  try {
+    await deprovisionSandbox(sandboxId, undefined, { experimental: true });
+  } catch (err) {
+    // Don't override the original test failure. Surface for debugging only.
+    console.error(`Cleanup deprovision failed for ${sandboxId}: ${err}`);
+  }
+}
```
bbonaby: suggestion: I'd put all of this in a general helper file for reuse. Also for ` t.skip(`${label}: wxc-exec lacks the isolation_session feature; rebuild with --features isolation_session to run this test`);`. I think it would be better to error instead of skip with that same message.

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+  stdout: string;
+  stderr: string;
+  exitCode: number;
+}
+
+/**
+ * Spawns the executor with the given envelope, captures stdout/stderr,
+ * and resolves on close. Honors `options.signal` for cancellation.
+ */
+function spawnAndCollect(
```
bbonaby: suggestion: I'd just call this `spawn` and the return value `SpawnResult` tbh versus `CollectOutput`

sdk/src/state-aware-types.ts
```diff
@@ -0,0 +1,158 @@
...
+    ? NonNullable<StateAwareMetadata[C]>['start']
+    : undefined;
+export type StopMetadataFor<C extends StateAwareSandboxingMethod> =
+  'stop' extends keyof NonNullable<StateAwareMetadata[C]>
+    ? NonNullable<StateAwareMetadata[C]>['stop']
+    : undefined;
+export type DeprovisionMetadataFor<C extends StateAwareSandboxingMethod> =
+  'deprovision' extends keyof NonNullable<StateAwareMetadata[C]>
+    ? NonNullable<StateAwareMetadata[C]>['deprovision']
+    : undefined;
```
bbonaby: suggestion: these were kind of a tough read could we do something like:

```typescript
type MetadataForPhase<
  C extends StateAwareSandboxingMethod,
  Phase extends string,
> =
  Phase extends keyof NonNullable<StateAwareMetadata[C]>
    ? NonNullable<StateAwareMetadata[C]>[Phase]
    : undefined;

export type ProvisionMetadataFor<C extends StateAwareSandboxingMethod> =
  MetadataForPhase<C, 'provision'>;

export type StartMetadataFor<C extends StateAwareSandboxingMethod> =
  MetadataForPhase<C, 'start'>;

export type StopMetadataFor<C extends StateAwareSandboxingMethod> =
  MetadataForPhase<C, 'stop'>;

export type DeprovisionMetadataFor<C extends StateAwareSandboxingMethod> =
  MetadataForPhase<C, 'deprovision'>;
```

seems slightly more manageable.

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+ */
+export function buildStateAwareEnvelope(args: BuildEnvelopeArgs): Record<string, unknown> {
+  const { phase, backendKey, containment, sandboxId, config } = args;
+  const remaining: Record<string, unknown> = { ...(config ?? {}) };
+  const version = (typeof remaining.version === 'string' && remaining.version) || STATE_AWARE_VERSION;
+  delete remaining.version;
+
+  const envelope: Record<string, unknown> = { version, phase };
+  if (containment) envelope.containment = containment;
+  if (sandboxId) envelope.sandboxId = sandboxId;
```
bbonaby: issue: would prefer for maintenance purposes if these happen on different lines within curly brackets

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+/**
+ * Constructs the wire-format JSON-shaped envelope for a state-aware request
+ * from a per-(backend, phase) Config. Lifts cross-cutting fields
+ * (filesystem, network, ui, process) to envelope top-level; nests any
+ * remaining backend-specific fields under `experimental.<backend>.<phase>`.
+ * Exported for unit testing; consumers do not call this directly.
+ */
+export function buildStateAwareEnvelope(args: BuildEnvelopeArgs): Record<string, unknown> {
+  const { phase, backendKey, containment, sandboxId, config } = args;
+  const remaining: Record<string, unknown> = { ...(config ?? {}) };
```
bbonaby: note: it was a bit hard to see what was happening here and how it was used. Probably needs a comment like:
```typescript
// Copy config, then remove fields as they are lifted into the envelope.
// Anything left becomes experimental.<backend>.<phase>
```

That plus my suggestion would be to rename `remaining` to `backendSpecific`. That said would this function need to change when someone decides to move from out of experimental? I feel like I asked this question before although, not specifically related to this code so sorry in advance, juggling lots of stuff in my head 😅

sdk/src/state-aware.ts
```diff
@@ -0,0 +1,406 @@
...
+
+async function nonExecCall<T>(
+  envelope: Record<string, unknown>,
+  options: SandboxSpawnOptions,
+): Promise<T> {
+  const { stdout } = await spawnAndCollect(envelope, options);
+  return parseNonExecResponse<T>(stdout);
+}
+
+// --- Public API: the 5 functions ---
```
bbonaby: suggestion: Lets move the functions and types above this point to a `state-aware-helpers` file and keep this file with only the public apis so it stays small and what developers so non mxc developers don't get confused.

test_scripts/README.md
adpa-ms: I actually do use powershell remoting! I was attempting to match the other files in the repo that use putd for this PR. Happy to change it though.

adpa-ms: Swapped to PowerShell Remoting. The integration `node_modules` symlinks the SDK back into itself (`@microsoft/mxc-sdk` → SDK root → `tests/integration/node_modules` → ...), which makes `Copy-Item -ToSession -Recurse` recurse indefinitely; the script now stages a flat payload on the host (skipping the symlink, copying the SDK's published shape directly), zips it via `Compress-Archive`, copies the single zip via PSRemoting, and expands it on the VM. Other two `push_*` scripts stay on TShell for now — separate migration when someone wants to invest.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api-overview.md
adpa-ms: Agreed, but not for this PR. Noted for the pending Rust cleanup and refactor.

sdk/src/state-aware.ts
adpa-ms: "Cross-cutting" is the term used through the design doc for fields that apply across phases (and across backends) rather than being scoped to one per-(backend, phase) Config. "Common" loses the orthogonal meaning, so I'd rather keep the const aligned with the spec.

sdk/src/state-aware.ts
adpa-ms: Good call on the magic string. `'iso'` is the wire-format prefix — it has to match the Rust side's `IsolationSessionRunner::ID_PREFIX` for the dispatcher to route non-provision calls correctly, so we can't expand it to a longer name. Extracted it to a named `ISOLATION_SESSION_ID_PREFIX` const so the role is visible at the call site (each future backend declares its own).

sdk/src/state-aware.ts
adpa-ms: The string splicing in `backendForSandboxId` is wire-format parsing — the input is a string from the wire (`"iso:wxc-abcd1234..."`) and we have to extract the prefix segment to look it up. An enum representation of the lookup table would still need a string→enum step at the parse boundary, so it doesn't avoid the slicing. (The Rust side does the same thing in `parse_sandbox_id_prefix`.)

sdk/src/index.ts
adpa-ms: The `For` suffix is intentional across the generic resolver family — `<Phase>ConfigFor<C>`, `<Phase>MetadataFor<C>`, plus the underlying `ConfigsForBackend<C>` — to distinguish these from the concrete per-backend types (`IsolationSessionProvisionConfig`, `IsolationSessionProvisionMetadata`, etc.). Without the suffix, a reader seeing `ProvisionConfig<'isolation_session'>` could confuse it with the concrete shape; with the suffix, the indexed-resolver semantics are explicit. Dropping just from metadata would introduce asymmetry with config, and dropping from both loses that signal. Keeping as-is.

sdk/src/state-aware-types.ts
adpa-ms: The per-phase asymmetry is intentional. Each Config exposes only the cross-cutting fields the backend honors at that phase per the design doc's honor matrix, narrowed to what the runtime actually implements — for IsolationSession, `filesystem` at provision. A shared base would force every field at every phase, erasing both the TS enforcement and the implemented-only narrowing.

sdk/tests/unit/state-aware-types.test.ts
adpa-ms: All the tests in this file use `'isolation_session'` today because IsolationSession is the only state-aware backend, so a split right now leaves the "remaining" file (brand / conditional / result tests) still exercising IsolationSession concretely — not yet backend-agnostic in shape, just in name. I'd rather wait for the second backend to arrive: at that point the natural shape is `<new-backend>-types.test.ts` for its Configs alongside an actually-backend-agnostic `state-aware-types.test.ts`, and the existing IsolationSession describes either stay where they are or rename as appropriate.

sdk/tests/integration/isolation-session-state-aware.test.ts
adpa-ms: Keeping the skip — CI can't exercise this end-to-end today regardless of build config, because CI hosts don't have the IsolationSession runtime. The test gates out either way (at `unsupported_phase` without the feature, at `backend_unavailable` with it). I'll revisit if/when CI environments gain IsolationSession runtime support and we can actually run this end-to-end.

sdk/src/state-aware.ts
adpa-ms: I'd rather keep `spawnAndCollect` / `CollectedOutput`. The "And Collect" suffix distinguishes this from `child_process.spawn` — they have different semantics (Promise-buffered vs event-emitter streaming), and renaming ours to `spawn` would force aliasing the std-lib import everywhere it's referenced and risk a reader assuming our function is the std-lib one.

sdk/src/state-aware.ts
adpa-ms: State-aware is the long-term design — not awaiting graduation. But individual backends can be experimental, and IsolationSession is right now; that's what `experimental.<backend>.<phase>` encodes. When IsolationSession graduates, this function would need updates to handle the per-backend stable path. New cross-cutting fields just need an entry in `CROSS_CUTTING_FIELDS`.

bbonaby (approved):

PR #266, opened 2026-05-07
macos_sandbox implementation

richiemsft:
## 📖 Description
Adds the macOS sandbox backend process-level sandboxing on macOS using Apple's Seatbelt kernel sandbox via /usr/bin/sandbox-exec.

 1. Filesystem — read/write paths, read-only paths, denied paths.
 2. Network — default allow/block policy, allowedHosts (falls back to allow-all due to Seatbelt limitation), blockedHosts (rejected with error)
 3. UI — blocks WindowServer, pasteboard, and HID mach services when ui.disable=true
 4. Process execution — timeout enforcement, working directory, environment variables, stdin/stdout/stderr piping
 5. Experimental gating — requires --experimental flag, rejects inproc mode

## 🔗 References
- `docs/macos-sandbox-backend.md` — full backend documentation including environment setup, configuration, policy mapping, limitations
- 8 example configs (`examples/15–22_mac_*.json`) covering hello world, filesystem access/deny, network block/allow/restrict, combined restrictions, and Python execution

## 🔍 Validation
- Built with `./build-mac.sh --rust-only` on Apple Silicon (aarch64-apple-darwin)
- All 8 macOS example configs executed successfully with `mxc-exec-mac --debug`
- Profile builder unit tests pass on all platforms (`cargo test -p macos_sandbox_common`)
- Verified sandbox enforcement: filesystem deny, network block, and UI isolation all working

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/266)


sdk/src/types.ts
```diff
@@ -37,12 +37,12 @@ export interface LifecycleConfig {
  * - "process": BaseProcessContainer (Windows) / LXC (Linux) / macOS sandbox (macOS)
  * - "microvm": MicroVM/Nanvix backend (Windows only, experimental)
  */
-export type ContainmentType = "process" | "wslc" | "microvm";
+export type ContainmentType = "process" | "wslc" | "microvm" | "macos_sandbox";
```
bbonaby: issue: see comment above. That said, I assume you added macos_sandbox  here because  there is no way to currently state a process containment is experimental? That said it looks like Gudge is updating how this works in #268, so we probably should wait until that goes in. Note He's also changing the name from macos_sandbox back to seatbelt so if you have any issues with that you should comment on the PR.

sdk/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@microsoft/mxc-sdk",
-  "version": "0.2.0",
+  "version": "0.1.8",
```
bbonaby: issue: you'll need to merge with main

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! `SeatbeltScriptRunner` — executes scripts inside Apple's Seatbelt
+//! sandbox via `/usr/bin/sandbox-exec`.
+//!
+//! This is the Phase A "exec mode" implementation. It generates a
```
bbonaby: suggestion: we likely don't need this part "//! This is the Phase A "exec mode" implementation."

src/seatbelt_common/src/seatbelt_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+//! `SeatbeltScriptRunner` — executes scripts inside Apple's Seatbelt
+//! sandbox via `/usr/bin/sandbox-exec`.
+//!
+//! This is the Phase A "exec mode" implementation. It generates a
+//! TinyScheme profile from the [`CodexRequest`] using
+//! [`crate::profile_builder::build_profile`], writes it to a tempfile in
+//! `TMPDIR`, then spawns `sandbox-exec -f <profile> /bin/sh -c <script>`
+//! with the request's env and working directory.
+//!
+//! Compiled only on macOS — the rest of the workspace continues to build
```
bbonaby, line 12: note: we'll need to add a macOS stage to build this  in using our 1ES pipelines. I can take care of this for you before the next release.

src/seatbelt_common/src/seatbelt_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+/// release (deprecated in headers since 10.7 but still shipped through
+/// current versions).
+const SANDBOX_EXEC: &str = "/usr/bin/sandbox-exec";
+
+/// Default shell used to execute `script_code`. `/bin/sh` is guaranteed
+/// to exist and is on the SIP-protected path so it's always reachable
+/// from inside the sandbox.
+const DEFAULT_SHELL: &str = "/bin/sh";
+
+pub struct SeatbeltScriptRunner;
```
bbonaby, line 37: note: this is implementation code so it's up to you but, the file is called `macos_sandbox_runner.rs` but we're calling it `SeatbeltScriptRunner`. We should probably just stick with one

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+impl Default for SeatbeltScriptRunner {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
bbonaby: suggestion above `SeatbeltScriptRunner` on line 39 you can use the derive macro e.g 

```rust
[Derive(Default)]
pub struct SeatbeltScriptRunner;
```

and that will create a default implementation automictically at compile time.

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+        // Seatbelt cannot filter network by hostname — reject blockedHosts
+        // rather than silently allowing traffic the user expects to be denied.
+        if !request.policy.blocked_hosts.is_empty() {
+            return error_response(
+                "macOS Seatbelt does not support per-host network filtering. \
+                 'blockedHosts' cannot be enforced; remove it or use \
+                 defaultPolicy: \"block\" to deny all network."
+                    .to_string(),
+            );
+        }
```
bbonaby: issue: there is a `validate_runner` rate method that these would go into. We can overwrite it with an implementation for your new `SeatbeltScriptRunner` struct

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+            Ok(t) => t,
+            Err(e) => return error_response(format!("failed to create profile tempfile: {e}")),
+        };
+
+        if let Err(e) = tmp.write_all(profile.as_bytes()) {
+            return error_response(format!("failed to write profile: {e}"));
+        }
+        if let Err(e) = tmp.flush() {
+            return error_response(format!("failed to flush profile: {e}"));
+        }
```
bbonaby: suggestion: we can probably use better names for the one letter variables throught the file

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+        // 4. Drain stdout/stderr in background threads to avoid deadlock
+        //    if the child fills the OS pipe buffer (~64KB on macOS).
+        let stdout_handle = child
+            .stdout
+            .take()
+            .map(|r| std::thread::spawn(move || read_to_string(r)));
+        let stderr_handle = child
+            .stderr
+            .take()
+            .map(|r| std::thread::spawn(move || read_to_string(r)));
```
bbonaby: question: are we considering these reads to be best effort? `read_to_string` returns a Result<T,E>, so if there is an error we silently will continue

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+            }
+            Err(WaitError::Io(e)) => return error_response(format!("wait failed: {e}")),
+        };
+
+        let stdout = stdout_handle
+            .and_then(|h| h.join().ok())
+            .unwrap_or_default();
+        let stderr = stderr_handle
+            .and_then(|h| h.join().ok())
+            .unwrap_or_default();
```
bbonaby: suggestion: Since this is repeated code from the match block above we could have something like:

```rust
fn join_reader(
    name: &str, // name of stio e.g stdout or stderr, or you could use an enum
    handle: Option<std::thread::JoinHandle<std::io::Result<String>>>,
) -> Result<String, String> {
    match handle {
        Some(handle) => match handle.join() {
            Ok(Ok(output)) => Ok(output),
            Ok(Err(err)) => Err(format!("failed to read child {name}: {err}")),
            Err(_) => Err(format!("child {name} reader thread panicked")),
        },
        None => Ok(String::new()),
    }
}
```

and then use it like:

```rust
let stdout = join_reader("stdout", stdout_handle);
let stderr = join_reader("stderr", stderr_handle);
```

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+fn error_response(msg: String) -> ScriptResponse {
+    ScriptResponse {
+        exit_code: -1,
+        standard_out: String::new(),
+        standard_err: String::new(),
+        error_message: msg,
+    }
+}
+
+fn read_to_string<R: std::io::Read>(mut r: R) -> String {
```
bbonaby: note: pipe errors are swallowed here as `r.read_to_string` returns a ` std::io::Result<String>`

src/macos_sandbox_common/src/macos_sandbox_runner.rs
```diff
@@ -0,0 +1,278 @@
...
+    };
+
+    loop {
+        match child.try_wait() {
+            Ok(Some(status)) => return Ok(status),
+            Ok(None) => {
+                if Instant::now() >= deadline {
+                    return Err(WaitError::Timeout);
+                }
+                std::thread::sleep(Duration::from_millis(50));
```
bbonaby: suggestion: 50ms is 20 times a second, do we expect the average work to be done, to be quick?  could 250ms suffice? If not another thing to think about would be the need for us to validate that the timeout the developer gives us is not less than what we do here for each interval.

src/mxc_darwin/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    eprint!("{}", response.standard_err);
+    process::exit(response.exit_code);
+}
+
+#[cfg(not(target_os = "macos"))]
+fn run_macos_sandbox(_request: &CodexRequest, logger: &mut Logger) -> ! {
+    // Build-only stub on non-macOS hosts so cargo workspace builds in CI.
+    // Suppress the unused-import warnings for `Instant` etc.
+    let _ = Instant::now();
+    let _ = ScriptResponse::default();
```
bbonaby: note: both of these can actually be deleted since they're unused. You can delete the comment above too.

docs/macos-support/seatbelt-backend.md
```diff
@@ -0,0 +1,314 @@
...
+./build-mac.sh --debug
+
+# Rust binary only, skip TS SDK/CLI
+./build-mac.sh --rust-only
+```
+
+The script writes to `sdk/bin/<arch>/mxc-exec-mac` so the SDK's
+`findDarwinExecutable()` picks up the dev build automatically.
+
+### Codesigning and notarization
```
bbonaby, line 266: note: the signing would need to happen in our 1ES rust build pipeline template. I'll need to request our ESRP signing be updated to include the developer Id signing for macOS in the meantime. I should be able to set this up within a few days. (takes a couple days for the ESRP folks to approve requests). The steps below shouldn't be needed when using ESRP signing, it'll do all that for us.

src/seatbelt_common/src/profile_builder.rs
```diff
@@ -0,0 +1,446 @@
...
+    if !allow_outbound {
+        if policy.allowed_hosts.is_empty() {
+            // Pure deny — implicit from `(deny default)`.
+            out.push_str(";; --- network: default-deny (no allow-network rules emitted) ---\n");
+            return;
+        }
+        // defaultPolicy=block + allowedHosts = allowlist mode.
+        // Seatbelt limitation: only `*` or `localhost` are valid hosts in
+        // `(remote ...)` filters — per-hostname filtering is not supported.
+        // Fall back to allowing all outbound when allowedHosts is specified.
```
bbonaby, line 164: thought: we should probably document this, maybe in your .md file. That said, could a y firewall apis or network filtering executables be used to block layer 3 hosts? or addition of entries in `/etc/hosts `

sdk/src/sandbox.ts
```diff
@@ -278,11 +306,24 @@ export function createConfigFromPolicy(
         return buildWslcContainerConfig(config, policy, containerId);
     }
 
+    if (containment === 'macos_sandbox') {
```
bbonaby: issue: `process containment == macos_sandbox`.  So we actually don't need a `macos_sandbox` as a containment option. If you look at what we're doing for linux and windows when it comes to process containment, your `platform === 'darwin'` if statement below is likely all we need here, since seatbelt/mac sandbox is only for processes if I read the docs right.

docs/macos-support/seatbelt-backend.md
```diff
@@ -0,0 +1,314 @@
+# macOS Sandbox Container Backend
```
bbonaby, line 1: issue: lets move this to it's own folder under `docs/`  e.g `docs/macos_support` or something like that

src/seatbelt_common/src/seatbelt_runner.rs
richiemsft: I have a CI/CD task right after this.

src/seatbelt_common/src/seatbelt_runner.rs
richiemsft: Will follow up with Gudge, finalize the name and adapt to all parts in code.

docs/macos-support/seatbelt-backend.md
richiemsft: will leave it as is and remove the comment when we have it working.

src/seatbelt_common/src/seatbelt_runner.rs
richiemsft: moving away from macos_sandbox, bringing back seatbelt.

src/macos_sandbox_common/src/macos_sandbox_runner.rs
richiemsft: best effort, now added a log warning if there is an error.

src/macos_sandbox_common/src/macos_sandbox_runner.rs
richiemsft: nice catch, I was just testing very small, quick scripts. bumping to 500ms to be consistent with the other platforms.

src/seatbelt_common/src/profile_builder.rs
richiemsft: added documentation to the .md file. the only path that doesn't require root access is to leverage the network extension framework, but we need to be signed and leverage apple's entitlements. I added it as future improvement.

bbonaby:
hey @richiemsft I think ones you fix the merge conflicts we should be good.

richiemsft:
@bbonaby merged and everything worked locally, just pushed it, lets see what the pipeline says.

schemas/stable/mxc-config.schema.0.5.0-alpha.json
```diff
@@ -19,9 +19,9 @@
     },
     "containment": {
       "type": "string",
-      "enum": ["appcontainer", "windows_sandbox", "lxc", "microvm", "wslc", "macos_sandbox"],
+      "enum": ["appcontainer", "windows_sandbox", "lxc", "microvm", "wslc", "seatbelt"],
```
bbonaby: oh dang missed changes to this file. Are you able to revert the changes to this file? We moved 0.5.0 to stable eariler with Adibs changes. You can continue to iterate in the 0.6.0 schema. After Gudges changes macos_sandbox and seatbelt should both work in the rust code.

schemas/dev/mxc-config.schema.0.6.0-dev.json
```diff
@@ -321,12 +321,6 @@
           "type": "object",
           "description": "macOS sandbox backend (experimental). Used when containment is 'seatbelt'.",
           "properties": {
-            "mode": {
-              "type": "string",
-              "enum": ["exec", "inproc"],
-              "default": "exec",
-              "description": "How the sandbox profile is applied: 'exec' spawns /usr/bin/sandbox-exec; 'inproc' calls sandbox_init_with_parameters after fork (uses a private macOS API)."
-            },
```
bbonaby, line 329: accidental removal I think?

richiemsft: no, I removed the `mode` since we are only going to support the exec path for now

schemas/stable/mxc-config.schema.0.5.0-alpha.json
richiemsft: reverted the file

schemas/stable/mxc-config.schema.0.5.0-alpha.json
```diff
@@ -285,7 +285,7 @@
             },
             "imageTarPath": {
               "type": "string",
-              "description": "Path to a local tar file to import as the container image. Supports both rootfs tars (from `docker export`) and Docker image archives (from `docker save`). When set, the image is imported from this file instead of pulling from a registry. The format is auto-detected: archives containing `manifest.json` are treated as Docker image format, archives with Linux root directories (bin/, etc/, usr/) are treated as rootfs tars, and unrecognized formats are rejected.",
+              "description": "Path to a local tar file to import as the container image. Supports both rootfs tars (from `docker export`) and Docker image archives (from `docker save`). When set, this overrides `image`.",
```
bbonaby: note: looks like this wslc description was changed. But I won't hold up the PR for it since it's just the description. We can change it back in a follow up.

bbonaby (approved):

schemas/stable/mxc-config.schema.0.5.0-alpha.json
richiemsft: lets do it now. pushed the change. unsure why the revert kept this new string.

bbonaby (approved):

PR #268, opened 2026-05-07
Split SDK containment into intent vs. backend, and rename to match

MGudgin:
Split SDK containment into intent vs. backend, and rename to match

Introduce a two-tier containment naming scheme in the MXC SDK and align
the wire format, schema, runner dispatch, and docs with it. The wire
field accepts either layer; the binary owns intent resolution.

* `ContainmentType`: abstract intent. `'process' | 'vm' | 'microvm'`
  today. Names the *kind* of isolation the caller wants.
* `ContainmentBackend`: concrete runner. `'processcontainer' |
  'windows_sandbox' | 'wslc' | 'lxc' | 'microvm' | 'seatbelt' |
  'isolation_session'`. Names a specific implementation.

The deprecated `SandboxingMethod = ContainmentType | ContainmentBackend`
alias is retained for back-compat. A runtime `ContainmentTypes` list is
exported so callers can recognise abstract intents at run time without
re-deriving the union.

Intent resolution
-----------------

The Rust parser resolves abstract intents at run time, in both
`convert_raw_config_inner` and `parse_containment_str`:

* `process` resolves to `processcontainer` on Windows, `lxc` on Linux,
  and `seatbelt` on macOS.
* `vm` resolves to `windows_sandbox` on Windows; on other platforms it
  falls through to the placeholder `Vm` variant which the host binaries
  surface as a "VM backend not yet implemented" error.
* `microvm` keeps its existing single-backend mapping.

The SDK validator (`resolveExecutableAndArgs`) accepts abstract intents
without checking them against the host's concrete-backend list — the
binary is the source of truth at run time.

Wire-string and type-name renames
---------------------------------

Two PR-feedback renames target the abstraction layer only. The
underlying OS APIs and their Rust modules keep their original names.

* `appcontainer` → `processcontainer` (wire) /
  `appContainer` → `processContainer` (JSON sub-block).
  Reflects that the value names the Windows process-level abstraction;
  wxc-exec dispatches between the legacy AppContainer implementation
  (honors `capabilities`, `learningMode`, `leastPrivilege`) and the
  newer BaseContainer implementation (honors `ui`) at run time based on
  host OS and `--experimental`.
  Rust: `ContainmentBackend::AppContainer` → `ProcessContainer`,
  `RawAppContainer` → `RawProcessContainer`.
  TypeScript: `AppContainerConfig` → `ProcessContainerConfig`.

* `macos_sandbox` → `seatbelt` (wire). Internal types follow:
  `ContainmentBackend::MacosSandbox` → `Seatbelt`,
  `MacosSandboxMode` → `SeatbeltMode`,
  `MacosSandboxConfig` → `SeatbeltConfig`,
  `RawMacosSandbox` → `RawSeatbelt`,
  `experimental.macos_sandbox` field → `seatbelt`. Redundant
  `#[serde(rename)]` attributes are dropped where the field name now
  matches the wire form. Test names follow.

Backward compatibility
----------------------

The parser keeps the pre-0.6 wire vocabulary as dual-read aliases so
configs written against the stable schemas (`0.4.0-alpha`,
`0.5.0-alpha`) continue to parse unchanged:

* `containment: "appcontainer"` is accepted and routed to
  `ContainmentBackend::ProcessContainer`.
* `containment: "macos_sandbox"` is accepted and routed to
  `ContainmentBackend::Seatbelt`.
* JSON sub-block `appContainer` is accepted via `#[serde(alias)]` and
  parsed into the same `process_container` slot as `processContainer`.
* JSON sub-block `experimental.macos_sandbox` is accepted via
  `#[serde(alias)]` and parsed into the same `seatbelt` slot as
  `experimental.seatbelt`.

Type-level back-compat in the TypeScript SDK mirrors the wire-
level aliases:

* `ContainerConfig` exposes a `@deprecated appContainer?:
  ProcessContainerConfig` field next to `processContainer`, so
  callers who construct configs in code with the pre-0.6 name
  keep compiling. The SDK forwards configs to wxc-exec via
  `JSON.stringify` without transforming them, so the Rust-side
  serde alias handles the runtime mapping; the TS alias just
  restores the matching compile-time surface.

Each legacy wire-value encounter emits a one-line deprecation log via
the existing `Logger`, pointing callers at the canonical name. The
aliases are intended for the deprecation window; the 0.6.0-dev schema
documents only the canonical names, and a future schema bump may remove
the aliases.

Kept unchanged
--------------

* `wxc_common::appcontainer_runner` module and its
  `AppContainerScriptRunner` struct (implements the legacy AppContainer
  OS API specifically).
* `BaseContainerSpecification.fbs` `app_container` flatbuffer field
  (controlled by the OS-side schema).
* WinHTTP per-AppContainer proxy policy references (the Win32 API
  itself is named after AppContainer).
* JSDoc, comments, and prose describing AppContainer-the-OS-API
  behaviour.

Schema, docs, and configs
-------------------------

* Dev schema `0.6.0-dev` enum updated; default is `processcontainer`;
  sub-block renamed to `processContainer`. Stable schemas
  `0.4.0-alpha` and `0.5.0-alpha` are untouched (immutable release
  artifacts); the parser's dual-read aliases keep configs targeting
  them working without edits.
* `docs/schema.md` adds an "Abstract intents" table alongside the
  concrete-backends table and uses the new wire strings.
* `docs/examples.md`, `docs/UIPolicy_Schema.md`,
  `docs/windows-sandbox.md`, `docs/nanvix-integration-plan.md`,
  `docs/versioning.md`, `docs/sandbox-policy/v1/policy.md`, and the
  `docs/state-aware-lifecycle/` design docs all migrate to the new
  wire / type names. Prose that describes the underlying OS APIs is
  preserved.
* `examples/*.json` and `test_configs/*.json` updated to the new wire
  values and sub-block keys; appcontainer-named files renamed to their
  processcontainer equivalents (with the e2e and PowerShell harnesses
  updated to match).

Tests and validation
--------------------

* SDK validator gets unit-level coverage for the spawn-path so the
  abstract-intent regression cannot recur.
* New parser tests `process_containment_resolves_per_target` and
  `vm_containment_resolves_per_target` pin the per-target resolution.
* New parser tests
  `legacy_appcontainer_wire_value_aliases_processcontainer`,
  `legacy_macos_sandbox_wire_value_aliases_seatbelt`,
  `legacy_app_container_subblock_alias_accepted`, and
  `legacy_experimental_macos_sandbox_subblock_alias_accepted` pin the
  dual-read behaviour.
* The redundant `containment_vm_accepted` test is removed (now covered
  by the per-target test).

Validation (post-rebase onto main):

* `cargo build --workspace` clean
* `cargo test --workspace --lib` 383/383 passing
* `cargo fmt --all -- --check` clean
* `cargo clippy --workspace --lib --tests -- -D warnings` clean
* `npm run build` (sdk) clean
* `npm test` (sdk) 125/125 passing

### DIFF
--- a/sdk/src/types.ts
+++ b/sdk/src/types.ts
@@ -227,6 +227,16 @@ export interface ContainerConfig {
   process?: ProcessConfig;
   /** ProcessContainer configuration */
   processContainer?: ProcessContainerConfig;
+  /**
+   * Legacy alias of {@link processContainer}. Retained so callers
+   * migrating from pre-0.6 SDK versions can keep their existing code
+   * compiling; the native binary parses both names into the same slot
+   * via a serde alias.
+   *
+   * @deprecated Use {@link processContainer} instead. This alias may be
+   * removed in a future minor release.
+   */
+  appContainer?: ProcessContainerConfig;
   /** LXC container configuration (Linux only) */
   lxc?: LxcConfig;
   /** Filesystem access configuration */
--- a/sdk/src/types.ts
+++ b/sdk/src/types.ts
@@ -32,17 +32,59 @@ export interface LifecycleConfig {
 }
 
 /**
- * Containment type abstraction for createConfigFromPolicy.
- * Maps to platform-specific backends:
- * - "process": BaseProcessContainer (Windows) / LXC (Linux) / macOS sandbox (macOS)
- * - "microvm": MicroVM/Nanvix backend (Windows only, experimental)
+ * Abstract containment intent. Names the *kind* of isolation the caller
+ * wants. The native binary (or the SDK as a fallback) resolves this to a
+ * concrete {@link ContainmentBackend} at run time based on what the host
+ * supports.
+ *
+/**
+ * Abstract containment intent. Names the *kind* of isolation the caller
+ * wants; the native binary resolves it to a concrete
+ * {@link ContainmentBackend} per host capability.
+ *
+ * Today's intents:
+ * - "process": OS-native process-level isolation. Resolves to
+ *   `processcontainer` (Windows), `lxc` (Linux), or `seatbelt` (macOS).
+ * - "vm": full hardware-virtualised VM isolation. Resolves to
+ *   `windows_sandbox` on Windows; no concrete VM backend exists on other
+ *   platforms today.
+ * - "microvm": lightweight-VM isolation. Resolves to the current MicroVM
+ *   runner (Windows only, experimental); intended to expand as additional
+ *   microvm backends (e.g. NanVix) are added.
+ *
+ * Concrete-only backends (such as `"wslc"`) live on
+ * {@link ContainmentBackend} until there is a meaningful abstraction over
+ * multiple implementations of the same kind.
  */
-export type ContainmentType = "process" | "wslc" | "microvm";
+export type ContainmentType = "process" | "vm" | "microvm";
 
 /**
- * Containment backends that require the --experimental flag.
+ * Runtime list of {@link ContainmentType} values. Kept in sync with the
+ * `ContainmentType` union via the type annotation. Use this to recognise
+ * abstract intents at run time (the union itself only exists at compile
+ * time).
  */
-export const ExperimentalBackends: readonly ContainmentType[] = ['microvm', 'wslc'];
+export const ContainmentTypes: readonly ContainmentType[] = ['process', 'vm', 'microvm'];
+
+/**
+ * Concrete containment backend. Each value names a specific runner
+ * implementation in the native binary. Prefer a {@link ContainmentType}
+ * value unless you specifically need to force a particular backend.
+ */
+export type ContainmentBackend =
+  | 'processcontainer'
+  | 'windows_sandbox'
+  | 'wslc'
+  | 'lxc'
+  | 'microvm'
+  | 'seatbelt'
+  | 'isolation_session';
+
+/**
+ * Containment values (abstract intent or concrete backend) that require
+ * the `--experimental` flag.
+ */
+export const ExperimentalBackends: readonly (ContainmentType | ContainmentBackend)[] = ['microvm', 'wslc', 'seatbelt'];
 
 /**
  * Clipboard access policy levels
@@ -64,7 +106,7 @@ export interface UiConfig {
 
 /**
  * BaseProcess-specific UI configuration (Windows only).
- * Lives under appContainer.ui in ContainerConfig.
+ * Lives under processContainer.ui in ContainerConfig.
  */
 export interface BaseProcessUiConfig {
   /** UI isolation level for the desktop */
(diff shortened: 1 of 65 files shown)


shschaefer (approved):

schemas/dev/mxc-config.schema.0.5.0-dev.json
```diff
@@ -19,9 +19,9 @@
     },
     "containment": {
       "type": "string",
-      "enum": ["appcontainer", "windows_sandbox", "lxc", "microvm", "wslc", "macos_sandbox"],
-      "default": "appcontainer",
-      "description": "Containment backend to use for execution. Note: 'windows_sandbox', 'wslc', and 'macos_sandbox' are experimental and require the --experimental CLI flag."
+      "enum": ["process", "processcontainer", "windows_sandbox", "lxc", "microvm", "wslc", "seatbelt"],
```
bbonaby: issue: so this is a breaking change to the schema, so we'll want to move our current 0.5.0-alpha to the stable schema folder and then this file will be the new `0.6.0-dev schema`. That said Adib's current PR already does this, so we can wait for his changes to go in, then you can merge with main, remove changes to this files and put them in the `0.6.0-dev schema` file instead.

Edit: his PR just merged, so we can do it now.

MGudgin: I don't understand. I thought the point of the -dev schemas was that we accumulated changes there before promoting them to an immutable published version.

bbonaby: Sorry for the confusion but this is because we actually wanted to promote 0.5.0-dev to stable since we shipped the mxc npm package which GitHub is using but we missed moving it to stable before publishing. So, doing the clean 0.6.0-dev now will put us in a good place to keep iterating with new accumulated changes from here on out. Does that make sense?

MGudgin: Ah, yes. Thanks.

MGudgin: Fixed, I think (rebased onto Adib's changes)

bbonaby (approved):

sdk/src/types.ts
```diff
@@ -171,14 +219,14 @@ export interface ContainerConfig {
...
+  /** Containment intent (preferred) or concrete backend (override). */
+  containment?: ContainmentType | ContainmentBackend;
   /** Container lifecycle settings */
   lifecycle?: LifecycleConfig;
   /** Process execution settings (required) */
   process?: ProcessConfig;
-  /** AppContainer configuration */
-  appContainer?: AppContainerConfig;
+  /** ProcessContainer configuration */
+  processContainer?: ProcessContainerConfig;
```
bbonaby, line 229: issue: I think we just need to add `appContainer?: ProcessContainerConfig` here as well and do something like:

```typescript
+  /**
+   * Legacy alias of {@link processContainer}. Retained so callers
+   * migrating from pre-0.6 SDK versions can keep their existing code.
+   *
+   * @deprecated Use {@link processContainer} instead. This alias may be
+   * removed in a future minor release.
+   */
+  appContainer?: ProcessContainerConfig
```

MGudgin: Done in c7d3a8d. Added the field exactly as suggested. The SDK doesn't transform user-supplied configs along the spawn path — it just `JSON.stringify`s and forwards to wxc-exec — so the Rust-side `#[serde(alias = "appContainer")]` handles the runtime side, and this purely-additive TS field restores the matching compile-time surface.

bbonaby (approved):

PR #239, opened 2026-05-03
[IsolationSession] Migrate to in-proc IsoSessionOps API

adpa-ms:
- [x] I have signed the [Contributor License Agreement](https://opensource.microsoft.com/cla/).
- [x] This pull request is related to an issue.
- [x] If this PR changes build commands, project architecture, or key conventions, I have updated [`.github/copilot-instructions.md`](.github/copilot-instructions.md).

-----

## Summary

Migrates the `isolation_session` backend from the cross-proc
`Windows.AI.IsolationEnvironment.Session` API (introduced in #217) to
the in-proc `Windows.AI.IsolationSession` `IsoSessionOps` API surface
(loaded from `IsoSessionApp.dll`). Closes #238.

The migration is enabled by OS branch
`build/ge_current_directwinpd_uxip/26611.260501-1700`, the first build
where the in-proc IDL exposes process I/O handles, environment variables,
separate stderr, and exit-code retrieval — the gaps that pinned the
original implementation to the cross-proc surface.

Companion schema change: renames `IsolationSessionConfigurationId::CommandLine`
to `Composable` (matching the in-proc IDL's `IsoSessionConfigId.Composable`)
and changes the default from `Small` to `Composable`. JSON value
`"commandline"` becomes `"composable"`. No backwards-compat layer — the
field lives under `experimental.*` and the SDK does not expose it.

## What's in this PR

Two commits structured for review:

**Commit 1 — `[IsolationSession] Retarget bindings infrastructure at Windows.AI.IsolationSession`**
- Renames `external/windows-sdk/isolation-environment-session/` to `isolation-session/`.
- Switches the bindings generation tool's `--filter` from `Windows.AI.IsolationEnvironment.Session` to `Windows.AI.IsolationSession`.
- Updates path strings in `tools/generate-isolation-session-bindings/src/main.rs`, `src/isolation_session_bindings/build.rs`, `src/isolation_session_bindings/src/lib.rs`, and the README inside the renamed directory.
- The generated `bindings.rs`, the runner that imports it, and `GENERATION_INFO.toml` are deliberately untouched here so this commit builds cleanly with the existing cross-proc surface.

**Commit 2 — `[IsolationSession] Migrate to in-proc IsoSessionOps API and Composable default`**
- Regenerates `src/isolation_session_bindings/src/bindings.rs` against the new namespace using the WinMD from the OS branch above; refreshes `GENERATION_INFO.toml`.
[...]

PR #270, opened 2026-05-08
[IsolationSession] Honor readwritePaths and readonlyPaths at provision

adpa-ms:
## 📖 Description

Adds filesystem-policy support to the IsolationSession backend in `wxc-exec`. Caller-supplied `policy.filesystem.readwritePaths` and `readonlyPaths` are honored at provision via the OS-side `Windows.AI.IsolationSession.IsoSessionOps::ShareFolderBatchAsync` API: each path receives an inheritable ACE granting the agent user read+write or read-only access, applied to the full subtree. `policy.filesystem.deniedPaths` is rejected because the underlying API has no Deny-ACE primitive. Filesystem policy is bound to the agent user's lifecycle and immutable post-provision; non-provision phases (start / exec / stop / deprovision) reject every filesystem field with `policy_validation`.

The implementation centres on a new `IsolationSessionManager::share_folders(rw, ro)` wrapper that aggregates per-path failures into a single `Lifecycle` error. The state-aware `provision` path and the one-shot `IsolationSessionRunner::execute` call it after `provision_agent_user`, with defensive cleanup on failure. Validate hooks split into `validate_provision_policy` (accepts rw/ro, rejects `deniedPaths`) and `validate_post_provision_policy` (rejects every filesystem field); the obsolete `validate_policy` is removed.

The IsolationSession bindings are regenerated against OS build `26613.260505-1700`. Single source-name change: `IsoSessionFolderSharingAccessLevel::Denied` → `Remove` (integer 2 unchanged, ABI-compatible). The `Remove` semantics are `REVOKE_ACCESS` (undo a prior grant), not Deny ACE — hence the rejection of `deniedPaths`.

Bundles a small TTY/cwd fix in flight on the same branch: initial console size is synced into the inner ConPTY at process create (PSReadLine prompt no longer drifts in interactive shells), and the one-shot test runner creates the cwd before invoking the hello fixture so the inner agent's `chdir` succeeds.

## 🔗 References

Resolves #269.

Rust-side state-aware support this builds on: #255. Original IsolationSession bringup: #217. Cross-proc-to-in-proc migration: #239.

Per-backend Rust plan with the updated honor matrix: `docs/isolation-session/state-aware-rust-initial-plan.md`.

## 🔍 Validation

Each commit on the branch was validated locally with the four MXC pre-push checks (`cargo fmt --check`, `cargo build --workspace`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`) under both the default features and `--features wxc/isolation_session` configurations. Zero failures and zero lint warnings across both flag states for every commit.

VM integration tests run on a Hyper-V VM at OS build `26613.260505-1700` with `Composable` config (release build). Both runners are invoked interactively from the VM desktop, since the OS-side calling-process identity check rejects PSSession-driven invocations — same constraint as #217 / #239 / #255.

- `run_isolation_session_state_aware_tests.ps1`: **31 / 31** PASS. Lifecycle A (17 tests, 6 of them new for per-phase fs-policy rejection plus a control test that proves the main no-fs sandbox cannot reach the locked-down test directory). New 9-test Lifecycle B covers the full filesystem-policy lifecycle on a separate sandbox: provision rw+ro → start → write-to-rw with host-side file verification → read-from-rw → read-from-ro → write-to-ro denied with host-absent verification → read-from-restricted denied (proves grants are path-specific, not blanket on the parent) → stop → deprovision. Lifecycle C Medium (5) added separately for `IsolationSessionConfigId::Medium` end-to-end.
- `run_isolation_session_tests.ps1`: **7 / 7** PASS. Six baseline plus one new filesystem test that provisions with a `readwritePaths` grant on a locked-down host directory and asserts the agent reads the marker.

Both runners now lock down their host test directories so the agent user has no inherited access via the default Authenticated Users / Users ACEs — proves `share_folders` is what enables access in Lifecycle B and the one-shot filesystem test. Both runners hardened with try-catch around test bodies and `String.Contains()` for substring checks (replaced `-notmatch`, which had a brittle array-return edge case in PowerShell).

## Out of scope — deferred follow-ups

Each item is a discrete concern that benefits from its own PR.

- **Active deny support (`deniedPaths` honored).** Requires an OS-side IDL primitive that doesn't exist today (Deny ACE installation, or equivalent). Pending OS-team interest. Rejecting `deniedPaths` for now is consistent with how every other unsupported policy field is handled and is easy to loosen later.
- **Mid-lifecycle policy mutation.** Allowing `share_folders` to fire on a started sandbox would require either MXC-side persistent state (sandboxId → applied policy) or an OS-side query API. Neither exists today, so MXC keeps filesystem policy immutable post-provision. To change folders, deprovision and reprovision.
- **Server-side cleanup of granted ACEs on deprovision.** The OS-side wrapper has a documented TODO to track grants and clean them up on agent-user deletion. Until that lands, granted ACEs persist on host folders after deprovision as orphan SID-references. The agent SIDs are LSA-managed and unique per provision, so the SID can never be re-activated — the leftovers are operational clutter rather than a security concern. MXC does not work around this client-side.
- **TypeScript SDK exposure.** Filesystem-policy support in the SDK lands alongside the broader state-aware SDK work in a separate PR.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Feature

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/270)


jeffstall (approved):
Thanks for wiring this up! We'll add Deny support in the next round.

PR #275, opened 2026-05-11
fix(lxc): resolve unprivileged lxcpath instead of hard-coding /var/lib/lxc

caarlos0:

-----

## Summary

Fixes the LXC unprivileged-mode failure reported in #274:

```
Error: Failed to configure filesystem: Failed to set config item lxc.mount.entry: No such file or directory (os error 2)
```

### Root cause

`LxcContainer::config_file_path()` hard-coded `/var/lib/lxc`, but every `lxc-*` shell-out (`lxc-create`, `lxc-info`, `lxc-start`, ...) was invoked **without `-P`**. As a non-root user, liblxc auto-routes the container to `~/.local/share/lxc/<name>/config` (per `XDG_DATA_HOME` / `HOME`); our wrapper then tried to append `lxc.mount.entry` to `/var/lib/lxc/<name>/config`, which doesn't exist → `ENOENT`.

The same hard-coded path bug also affected the denied-paths rootfs probe in `filesystem_mounts.rs`, where `is_file()` would always return false for unprivileged users and silently mask files with `tmpfs` instead of `/dev/null`.

### Fix

- New `resolve_default_lxcpath()` helper that mimics liblxc's resolution: `LXC_PATH` → `/var/lib/lxc` for root → `$XDG_DATA_HOME/lxc` → `$HOME/.local/share/lxc` → `/var/lib/lxc` last-resort fallback. Extracted into a parameterized `resolve_lxcpath_with_env(get_env, geteuid)` so every branch is covered by deterministic unit tests.
- `LxcContainer` now stores the resolved `lxc_path` (always populated, no `Option`) and passes `-P <resolved>` to **every** `lxc-*` invocation, so behavior is identical regardless of how the binary is launched (cron, systemd unit with non-default `HOME`, etc.).
- `filesystem_mounts.rs` denied-path rootfs probe now uses the resolved `container.lxc_path()` instead of the hard-coded fallback.
- `set_config_item`'s error now includes the key, value, and target config-file path, addressing the diagnostic gap also called out in #274.

### Tests

Added 12 unit tests in `lxc_common::lxc_bindings::tests` covering:

- All `resolve_lxcpath_with_env` branches (`LXC_PATH` precedence even as root, root default, XDG_DATA_HOME wins over HOME, HOME fallback, trailing-slash normalization, empty-env handling, last-resort fallback when neither env var is set).
- `LxcContainer::new(..., None)` always yields a non-empty resolved path.
- `LxcContainer::new(..., Some(...))` honors explicit overrides and `config_file_path()` derives from the resolved path.
- `set_config_item` error message contains key, value, and the container config path.

`cargo test -p lxc_common`: **25 passed** (12 new). `cargo build -p lxc` and `cargo clippy -p lxc_common --no-deps -- -D warnings` clean.

Closes #274
Refs https://github.com/github/copilot-agent-runtime/issues/164

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/275)

### DIFF
--- a/src/lxc_common/src/lxc_bindings.rs
+++ b/src/lxc_common/src/lxc_bindings.rs
@@ -51,11 +51,20 @@ where
 /// See [`resolve_lxcpath_with_env`] for the exact algorithm. This wrapper
 /// reads the real environment and effective UID.
 pub fn resolve_default_lxcpath() -> String {
+    // `geteuid` only exists on Unix; on other targets the function is never
+    // invoked in production (lxc-exec is Linux-only) but the crate still
+    // needs to compile workspace-wide, so fall back to a non-root EUID.
+    #[cfg(unix)]
     // SAFETY: `geteuid` is a thread-safe, side-effect-free libc call.
-    resolve_lxcpath_with_env(
-        |k| std::env::var(k).ok(),
-        || unsafe { libc::geteuid() } as u32,
-    )
+    fn current_euid() -> u32 {
+        unsafe { libc::geteuid() as u32 }
+    }
+    #[cfg(not(unix))]
+    fn current_euid() -> u32 {
+        1
+    }
+
+    resolve_lxcpath_with_env(|k| std::env::var(k).ok(), current_euid)
 }
 
 /// Safe wrapper around an LXC container.
--- a/src/lxc_common/src/lxc_bindings.rs
+++ b/src/lxc_common/src/lxc_bindings.rs
@@ -7,18 +7,79 @@
 //! function pointer fields. This module provides an RAII `LxcContainer` wrapper
 //! that calls the appropriate function pointers and handles cleanup.
 
+/// Resolve the default LXC storage path the way liblxc does.
+///
+/// Replicates the algorithm liblxc applies when no explicit `-P <lxcpath>` is
+/// provided to its CLI tools, using the supplied environment lookup and
+/// effective-uid hooks. Extracted into a free function so unit tests can
+/// exercise every branch deterministically.
+///
+/// Resolution order:
+///  1. `LXC_PATH` env var (if non-empty).
+///  2. `/var/lib/lxc` when running as root (EUID 0).
+///  3. `$XDG_DATA_HOME/lxc` if `XDG_DATA_HOME` is set and non-empty.
+///  4. `$HOME/.local/share/lxc` if `HOME` is set and non-empty.
+///  5. `/var/lib/lxc` as a last-resort fallback.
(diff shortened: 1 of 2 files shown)


src/lxc_common/src/lxc_bindings.rs
```diff
@@ -7,18 +7,88 @@
...
+
+/// Resolve the default LXC storage path for the current process.
+///
+/// See [`resolve_lxcpath_with_env`] for the exact algorithm. This wrapper
+/// reads the real environment and effective UID.
+pub fn resolve_default_lxcpath() -> String {
+    // `geteuid` only exists on Unix; on other targets the function is never
+    // invoked in production (lxc-exec is Linux-only) but the crate still
+    // needs to compile workspace-wide, so fall back to a non-root EUID.
+    #[cfg(unix)]
```
bbonaby, line 57: suggestion: could use `#[cfg(target_os = "linux")]` for linux only

src/lxc_common/src/lxc_bindings.rs
```diff
@@ -7,18 +7,88 @@
...
+    // SAFETY: `geteuid` is a thread-safe, side-effect-free libc call.
+    fn current_euid() -> u32 {
+        unsafe { libc::geteuid() as u32 }
+    }
+    #[cfg(not(unix))]
```
bbonaby, line 62: suggestion: could use `#[cfg(not(target_os = "linux"))]`

src/lxc_common/src/lxc_bindings.rs
```diff
@@ -27,68 +97,69 @@ impl LxcContainer {
...
-        self.config_path.as_deref()
+    /// Get the resolved LXC storage path (the "lxcpath") used by this handle.
+    pub fn lxc_path(&self) -> &str {
+        &self.lxc_path
+    }
+
+    /// Build a `Command` for an `lxc-*` tool with `-P <lxc_path> -n <name>`
+    /// already populated. Centralizes the argv prefix so we can't accidentally
+    /// drop `-P` again (see #274).
+    fn lxc_command(&self, tool: &str) -> std::process::Command {
```
bbonaby, line 108: praise: good call

bbonaby (commented):
thanks!

bbonaby (approved):

caarlos0:
Thanks @bbonaby - do you have plans for next release?

bbonaby:
Hey @caarlos0 so we'll likely publish our new 0.2.0 npm package either tomorrow or Wednesday. Will keep you posted.

PR #280, opened 2026-05-11
[IsolationSession] Add Entra cloud-agent support to TypeScript SDK

adpa-ms:
## 📖 Description

- Adds `IsolationSessionUserConfig` (class with `wamToken` redacted under `util.inspect`) and `user?` field on `IsolationSessionProvisionConfig` / `IsolationSessionStartConfig`. SDK callers can now opt into Entra-backed sandboxes.
- Extends the 0.6.0-dev schema under `experimental.isolation_session.{provision,start}`. Additive within the experimental section — no version bump.
- Adds `docs/isolation-session/state-aware-typescript-initial-plan.md` — SDK companion to the existing Rust plan. The cross-backend state-aware design doc stays focused on lifecycle design; IsolationSession-specific Entra surface lives in the IsolationSession-specific doc.
- Renames `StateAwareSandboxingMethod` → `StateAwareContainmentBackend` (separate commit) to align with the new naming convention from #268. Done now while state-aware is still in 0.6.0-dev.

## 🔗 References

Closes #279

## 🔍 Validation

Full re-validation pass after rebase onto `origin/main` and rename cleanup:

- SDK unit tests: 131/131 pass.
- Rust workspace tests: 378/378 substantive pass.
- Rust with `isolation_session` feature: 392/392 pass.
- `build.bat --x64` and `build.bat --x64 --with-isolation-session`: both complete.
- Integration suite with `MXC_SKIP_OS_BUILD_DEPENDENT_TESTS=1`: 17 pass / 12 skip.
- IsolationSession state-aware E2E on Hyper-V VM (manual interactive console — the OS-side caller check requires the call to originate from an interactive session, so PSRemoting cannot drive provision): 5/5 pass.
- Entra validation matrix (UPN match, malformed bundle, sandbox/user inconsistencies) covered by Rust-side `test_scripts/run_isolation_session_state_aware_tests.ps1` Lifecycle D. SDK-level integration coverage for the Entra path isn't feasible without WAM credentials available to CI.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Feature


docs/isolation-session/state-aware-typescript-initial-plan.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Stop, Deprovision
+
+Each Config carries only `version?`. Neither phase returns metadata.
+
+## `IsolationSessionUserConfig`
+
+```typescript
+export class IsolationSessionUserConfig {
+  readonly upn: string;
```
bbonaby, line 90: question (non-blocking): so this is just the username of a logged on user? or is it deeper than just that?

docs/isolation-session/state-aware-typescript-initial-plan.md
```diff
@@ -0,0 +1,193 @@
...
+const skipReason = os.platform() !== 'win32'
+  ? 'IsolationSession is Windows-only'
+  : await probeStateAwareRuntime('isolation_session');
+
+describe('IsolationSession state-aware lifecycle E2E', { skip: skipReason }, () => {
+  it('runs full lifecycle', async () => { /* ... */ });
+});
+```
+
+The Entra path can't be exercised in CI without WAM credentials; the Rust-side
```
bbonaby, line 184: question (non-blocking): do you know what they're doing in the OS to test this path for iso session? Just curious if they have a test account for this kind of thing in the windows test labs

docs/isolation-session/state-aware-typescript-initial-plan.md
```diff
@@ -0,0 +1,193 @@
...
+phase. See the [Rust plan](state-aware-rust-initial-plan.md) for the full Rust-side
+contract (including fields not yet exposed via the SDK).
+
+| Phase | Config | Metadata |
+|---|---|---|
+| provision | `IsolationSessionProvisionConfig` | `IsolationSessionProvisionMetadata` |
+| start | `IsolationSessionStartConfig` | none |
+| exec | `IsolationSessionExecConfig` | n/a (exec returns an exit code, not metadata) |
+| stop | `IsolationSessionStopConfig` | none |
+| deprovision | `IsolationSessionDeprovisionConfig` | none |
```
bbonaby, line 41: question (non-blocking): so with the addition of the user config that takes in the upn and wamToken in the provision phase, iso session doesn't need to use them for deprovisioning at all?

sdk/src/state-aware-types.ts
```diff
@@ -25,9 +25,31 @@ export type StateAwareSandboxingMethod = Extract<SandboxingMethod, 'isolation_se
...
   string & { readonly __mxcBrand: 'SandboxId'; readonly __mxcBackend: C };
 
+const ISO_USER_INSPECT = Symbol.for('nodejs.util.inspect.custom');
+
+/**
+ * Entra credentials, supplied at provision to opt into an Entra-backed
+ * sandbox and at start to authenticate the session. `wamToken` is treated
+ * as a secret: `util.inspect` and `console.log` redact it. `JSON.stringify`
+ * is unaffected — the wire envelope carries the token verbatim.
+ */
```
bbonaby, line 38: question: so we redact it in the typescript at runtime but pass it as is in the json that get passed to wxc-exec? Am I understanding that right? If so I wonder if there is a way for you to coordinate with wxc-exec to read it from credential manager versus having to pass the token directly in the json. Not sure that's possible without admin though.

docs/isolation-session/state-aware-typescript-initial-plan.md
adpa-ms: This is the username for the "agent user", so the user that is used to log in to the isolation-session. It's coming from a workstream that manages Azure Active Directory users I believe. They have a concept of an "agentic identity" which we are attempting to leverage. I've left this all under experimental so that we can plausibly update this in the future potentially in a "breaking" fashion if we need to. I've also kept it clearly scoped within IsolationSession for now (unclear to me just yet if this could spread to all containers). At least this way, we keep all our options open.

docs/isolation-session/state-aware-typescript-initial-plan.md
adpa-ms: The upn is hidden in the sandbox id which we do pass to all subsequent calls :)

docs/isolation-session/state-aware-typescript-initial-plan.md
adpa-ms: I think everyone is scrambling to get a test setup - it needs a test tenant in Azure and some other external requirements. I will also attempt to manually test this if possible once I understand the requirements. Couldn't put in an automated test for it though (at least not yet).

sdk/src/state-aware-types.ts
adpa-ms: Totally agree, I think this is likely to evolve more. This is just the current shape. Even more reasons I wanted to keep this whole end to end as "experimental" for the time being.

bbonaby (approved):

PR #281, opened 2026-05-12
phase0: add filesystem.allowDaclFallback config option

MGudgin:
## 📖 Description

Introduces a new boolean field `filesystem.allowDaclFallback` (default: `true`) on the MXC dev schema. When the BaseContainer OS API is absent and `bfscfg.exe` is unavailable, MXC may fall back to applying DACL ACEs directly on policy paths (Tier 3 fallback). Because this modifies host filesystem security descriptors, operators need a way to refuse the fallback even though the original DACLs are restored on exit.

The default is `true` so that pre-GE Windows 11 builds — which currently have no other downlevel containment path — continue to work out of the box. Setting the field to `false` causes runs that would require Tier 3 to fail fast with a clear error (wired up in a later phase).

Plumbed through:

- `schemas/stable/mxc-config.schema.0.5.0-alpha.json` (new property)
- `RawFilesystem` in `config_parser.rs` (serde field, optional)
- `ContainerPolicy` in `models.rs` (new bool field; manual `Default` impl so the field defaults to `true` rather than `false`)
- Mapping in `load_request` copies the value through, preserving the `true` default when the field is omitted
- Three new unit tests covering default, explicit-true, explicit-false
- `docs/schema.md` updated with a new Filesystem Policy table that documents all four fields

No runner code is touched in this phase; consumers of `allow_dacl_fallback` are added in subsequent phases (phase1+).

This is **phase 0 of the downlevel containment stack** (phase0 → phase5). Subsequent PRs build on this one.

## 🔗 References

Part of the downlevel containment series.

## 🔍 Validation

- New unit tests in `config_parser.rs` cover the three field states (default, explicit true, explicit false).
- No runner behavior changes in this PR — purely schema/config plumbing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -42,7 +42,8 @@ production configs and the dev schema when working on experimental features:
     "filesystem": {
         "readwritePaths": ["C:\\temp"],     // Read-write access
         "readonlyPaths": ["C:\\data"],      // Read-only access
-        "deniedPaths": ["C:\\Windows"]      // Blocked paths
+        "deniedPaths": ["C:\\Windows"],     // Blocked paths
+        "allowDaclFallback": true           // Allow Tier 3 DACL fallback (default true)
     },
 
     "network": {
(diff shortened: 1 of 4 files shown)


MGudgin:
Closing: the rebased phase0 commit modifies `schemas/stable/mxc-config.schema.0.5.0-alpha.json` instead of the intended `schemas/dev/mxc-config.schema.0.5.0-dev.json`. Will fix the schema path and reopen.

PR #282, opened 2026-05-12
phase0: add fallback.allowDaclMutation config option

MGudgin:
## 📖 Description

Introduces a new top-level `fallback` section on the MXC dev schema (`0.6.0-dev`), with a single occupant: `allowDaclMutation` (default: `true`). When the BaseContainer OS API is absent and `bfscfg.exe` is unavailable, MXC may fall back to applying DACL ACEs directly on policy paths (Tier 3 fallback). Because this modifies host filesystem security descriptors, operators need a way to refuse the fallback even though the original DACLs are restored on exit.

The `fallback` section names **operator consent for host-impacting containment fallbacks**. It is intentionally a section rather than a filesystem-scoped flag so future host-mutating fallbacks can land alongside as siblings without re-shuffling the schema. (Addresses review feedback on the prior shape.)

The default is `true` so that pre-GE Windows 11 builds — which currently have no other downlevel containment path — continue to work out of the box. Setting the field to `false` causes runs that would require Tier 3 to fail fast with a clear error (wired up in a later phase).

Plumbed through:

- `schemas/dev/mxc-config.schema.0.6.0-dev.json` (new top-level `fallback` object with `allowDaclMutation` property)
- `RawFallback` in `config_parser.rs` (new serde struct, optional on both `RawConfig` and the state-aware request envelope)
- `FallbackPolicy` in `models.rs` (new struct with manual `Default` impl so `allow_dacl_mutation` defaults to `true`); `ContainerPolicy` gains a `fallback: FallbackPolicy` field
- Mapping in `load_request` copies the value through, preserving the `true` default when the field is omitted
- Three new unit tests covering default, explicit-true, explicit-false
- `docs/schema.md` updated with a new Fallback Policy section

No runner code is touched in this phase; consumers of `fallback.allow_dacl_mutation` are added in subsequent phases (phase1+).

This is **phase 0 of the downlevel containment stack** (phase0 → phase5). Subsequent PRs build on this one.

## 🔗 References

Part of the downlevel containment series. Schema-doc coverage tracked separately in #285.

## 🔍 Validation

- New unit tests in `config_parser.rs` cover the three field states (default, explicit true, explicit false).
- No runner behavior changes in this PR — purely schema/config plumbing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -45,6 +45,10 @@ production configs and the dev schema when working on experimental features:
         "deniedPaths": ["C:\\Windows"]      // Blocked paths
     },
 
+    "fallback": {
+        "allowDaclMutation": true          // Allow Tier 3 DACL fallback (default true)
+    },
+
     "network": {
         "defaultPolicy": "block",          // "allow" or "block"
         "enforcementMode": "firewall",     // "capabilities", "firewall", or "both"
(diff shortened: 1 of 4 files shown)


docs/schema.md
```diff
@@ -42,7 +42,8 @@ production configs and the dev schema when working on experimental features:
     "filesystem": {
         "readwritePaths": ["C:\\temp"],     // Read-write access
         "readonlyPaths": ["C:\\data"],      // Read-only access
-        "deniedPaths": ["C:\\Windows"]      // Blocked paths
+        "deniedPaths": ["C:\\Windows"],     // Blocked paths
+        "allowDaclFallback": true           // Allow Tier 3 DACL fallback (default true)
```
shschaefer: As we continue to move forward with features UI, outside-in, etc...  we will find more than just the need to fallback DACLs.  Can we promote this to a property of the process container, or a general property on virtualization fallback?

docs/schema.md
```diff
@@ -77,6 +78,17 @@ production configs and the dev schema when working on experimental features:
 }
 ```
 
+### Filesystem Policy
```
shschaefer, line 84: This document does not define all policies.  Adding filesystem is ragged.  We should probably add an issue to add documentation on all of the schema sections.

shschaefer (changes requested):

docs/schema.md
MGudgin: Filed #285 to track full schema documentation coverage. Will keep this PR's scope limited to the new `allowDaclFallback` field.

docs/schema.md
MGudgin: Force-pushed [e1fb78c](https://github.com/microsoft/mxc/commit/e1fb78c) which lifts this out of `filesystem` into a new top-level `fallback` section. The field is now `fallback.allowDaclMutation`. Reasoning: we considered making it a single boolean at the container root vs. a structured section, and went with the section so future host-impacting fallbacks (UI, outside-in, etc.) can land as siblings without re-shuffling — and so an operator who has a different answer per axis isn't forced into a coupled one. Open to going coarser if you'd rather have a single root-level consent gate.

shschaefer (approved):

PR #283, opened 2026-05-12
fix(lxc): lxc tty

caarlos0:
## 📖 Description

Make the LXC backend usable end-to-end for an interactive sandboxed shell driven over a host pty (the model the GitHub Copilot CLI uses on Linux for its `bash` tool). Four stacked fixes — each was necessary, none was sufficient on its own.

### 1. `attach_run`: inherit parent stdio (a3f2a66)

Wires `lxc-attach`'s stdin/stdout/stderr to the parent process so the inner shell shares whatever drives `lxc-exec`. Without this, the child gets a closed stdin and exits immediately on EOF — interactive shells can't survive their first read. Mirrors the AppContainer runner's pty sharing on Windows.

### 2. `destroy()`: skip the redundant `lxc-stop` (562cd5a)

`lxc-destroy -f` already force-stops a running container. We were also calling `lxc-stop` first, which waits up to **60 s** for graceful shutdown. On distros that run systemd as PID 1 in unprivileged userns (anything ubuntu-class), init never cleanly responds to SIGPWR — the wait always times out before the force-stop step. Cleanup drops from ~60 s to ~0.2 s; alpine is unaffected.

### 3. `attach_run`: bridge through an inner pty with a detached ctty (49fbbb9)

The inherit-stdio fix from #1 got bash booting and printing its prompt, but the shell then ignored everything the host wrote. Two compounding root causes:

- **Pre-buffered input was discarded.** Bash's readline init calls `tcsetattr` on its stdin, which can flush bytes the parent buffered into the pty before bash got there. Anything the CLI pre-sent (e.g. its shell-init wrapper) was lost.
- **`lxc-attach` bypassed our slave fds.** When it inherits a controlling terminal from the parent, it forwards the inner container pty's I/O straight to `/dev/tty` instead of using the stdio slots. Bash's prompt reached our screen but the master fd we kept stayed empty, so there was no way to forward host stdin into bash either.

Fix:

- Allocate an inner pty pair via `nix::pty::openpty`. Give the slave to `lxc-attach` (and thus bash) for stdin/stdout/stderr; keep the master in the parent.
- In `pre_exec`: `setsid()` to drop the inherited controlling terminal, then `ioctl(TIOCSCTTY)` on fd 0 so `lxc-attach` adopts our slave as its new ctty. This is what makes the slave fds actually carry the data.
- Output thread copies master → host stdout, signaling "ready" on the first byte from inside the container (bash is past readline init at that point and safe to feed).
- Input thread waits for "ready" (capped at 5 s; the higher-level command timeout takes over from there) before forwarding host stdin → master.

Adds the `term` feature to the workspace `nix` dependency so `pty::openpty` is available.

### 4. `lxc-exec`: destroy the active container on fatal signals (f146e9f)

`LxcScriptRunner::run_internal` only calls `container.destroy()` after `attach_run` returns. When the runner is killed by a fatal signal — its parent exited, or it received SIGHUP/SIGTERM/SIGINT — the in-flight `attach_run` is interrupted and the destroy block is never reached, leaving the container running indefinitely (`lxc-ls` lists it long after the CLI session that spawned it is gone).

Install a `signal_cleanup` watchdog in `lxc-exec`'s `main`:

- Block SIGHUP/SIGTERM/SIGINT in the calling thread before any other thread is spawned, so all later threads inherit the blocked mask.
- A dedicated thread `sigwait`s for any of those signals, destroys the currently active container, and exits with `128 + signo` so the parent sees a normal signal-style exit.

`LxcScriptRunner::run_internal` registers the active container name with the watchdog right after resolving it, so the cleanup window covers create/start/attach as well as the normal post-attach destroy path.

## 🔗 References

Surfaced while wiring up Linux sandbox support in the GitHub Copilot CLI runtime. Tested against the [`createConfigFromPolicy` migration in copilot-agent-runtime#7767](https://github.com/github/copilot-agent-runtime/pull/7767).

## 🔍 Validation

- `cargo test -p lxc_common --lib` — 25/25 unit tests green on host (macOS arm64) and target (linux/aarch64-unknown-linux-gnu).
- `cargo check -p lxc_common --target aarch64-unknown-linux-gnu` clean.
- End-to-end manual test on Ubuntu Resolute (arm64, OrbStack VM) running the Copilot CLI's sandboxed `bash` tool: container creates, bash boots, the CLI's init wrapper executes, commands run, output streams back, container destroys in <1 s. Repeated successfully across 10+ invocations with no `ABORTING` start failures or 60 s teardown waits.
- Signal cleanup verified by SIGTERM/SIGINT'ing `lxc-exec` mid-run: container is destroyed before the process exits, `lxc-ls` shows no orphaned containers across repeated runs.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

refs https://github.com/github/copilot-agent-runtime/issues/164

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/283)


src/lxc_common/src/signal_cleanup.rs
```diff
@@ -0,0 +1,88 @@
...
+/// that synchronously waits (`sigwait`) for any of them. On delivery the
+/// watchdog destroys the active container, then exits with `128 + signo`.
+///
+/// MUST be called once, early in `main`, before any other threads are
+/// spawned: `pthread_sigmask` only changes the mask of the calling thread,
+/// but new threads inherit the mask at creation time. If a thread starts
+/// before `install()`, that thread's mask leaves the signals unblocked and
+/// the kernel may deliver them there instead of to the watchdog.
+#[cfg(target_os = "linux")]
+pub fn install() -> Result<(), String> {
```
bbonaby, line 83: suggestion: you can probably do something like:

```rust
static INSTALLED: OnceLock<()> = OnceLock::new();

pub fn install() -> Result<(), String> {
    if INSTALLED.set(()).is_err() {
        return Ok(());
    }
// 
```
so you return `( )` everytime after the first one

src/lxc_common/src/signal_cleanup.rs
```diff
@@ -0,0 +1,88 @@
...
+    mask.add(Signal::SIGHUP);
+    mask.add(Signal::SIGTERM);
+    mask.add(Signal::SIGINT);
+    mask.thread_block()
+        .map_err(|e| format!("pthread_sigmask: {}", e))?;
+
+    thread::Builder::new()
+        .name("lxc-signal-cleanup".into())
+        .spawn(move || run_watchdog(mask))
+        .expect("spawn lxc-signal-cleanup thread");
```
bbonaby: issue:  `expect` can panic here, we could  do `.map_err(|err| format!("spawn lxc-signal-cleanup thread: {err}"))?;` instead.

caarlos0:
@bbonaby thanks for the review, updated!

caarlos0:
failing tests are on main as well - ignoring.

bbonaby:
@caarlos, I merged a PR that should fix those errors you were getting. If you merge with main and fix the conflict we should see all green after

caarlos0:
@bbonaby done! thank you

src/lxc_common/src/lxc_bindings.rs
```diff
@@ -207,24 +208,168 @@ impl LxcContainer {
...
+                // state, which is what actually matters for lxc-attach.
+                let _ = libc::ioctl(0, libc::TIOCSCTTY as _, 0);
+                Ok(())
+            });
+        }
 
-        let output = cmd
-            .output()
-            .map_err(|e| format!("Failed to run lxc-attach: {}", e))?;
+        let mut child = cmd
```
bbonaby, line 303: Noticed the PR pipeline SDK integration tests timed out, so I dug into this a bit more and I think we have a PTY lifetime bug causing a deadlock.

We create three slave-fd duplicates of the inner PTY and pass them to:

```rust
cmd.stdin(...)
   .stdout(...)
   .stderr(...);
```

What I missed initially is that `std::process::Command` retains ownership of those `Stdio` handles until the `Command` itself is dropped.

So while `cmd` is still in scope:

* the child process has its inherited slave PTY fds
* the parent process also still has slave PTY fds owned by `cmd`

When `lxc-attach` exits, only the child's copies are closed. The parent's copies remain open, so the PTY master never receives EOF/HUP. That leaves `master_reader.read()` blocked forever, and consequently:

```rust
let _ = output_thread.join();
```

never returns.

I think the fix is simply to drop `cmd` immediately after `spawn()` so the parent releases its slave-fd references:

```rust
let mut child = cmd
    .spawn()
    .map_err(|e| format!("Failed to spawn lxc-attach: {}", e))?;

drop(cmd);
```

You should be able to repro locally with:

```bash
sudo node --test dist/linux-process-container.test.js
```

I believe `build.sh` is currently broken (there’s an open PR attempting to fix it), but I verified this under WSL ubuntu after building with `build.bat`. In the meantime here is the npm package that was built from this PR:

bbonaby: [microsoft-mxc-sdk-0.2.0.zip](https://github.com/user-attachments/files/27742560/microsoft-mxc-sdk-0.2.0.zip)

Not sure why I can't upload .tgz files, so i just compressed it to a zip file.

caarlos0: 🤔 that makes sense, will try your approach and e2e test in a VM as well 

thank you!

caarlos0: yes, that was it. e2e tested it in a VM and works like a charm, too.

thank you 🙏🏻

caarlos0:
Addressed the latest review pass:

- **Signals stay blocked in the inner shell** (`lxc_bindings.rs:276`) → 13d644d. `pre_exec` now restores the default mask before exec so Ctrl-C / termination work inside the attached shell.
- **`set_active` runs even when `destroyOnExit=false`** (`lxc_runner.rs:103`) → d752012. Registration is gated on `self.destroy_on_exit` so a SIGTERM no longer deletes a container the caller asked to keep.
- **Failed watchdog spawn leaves signals blocked** (`signal_cleanup.rs:69`) → 1a9d6a4. `install()` calls `thread_unblock()` on spawn failure, defers `INSTALLED` until the watchdog is actually running, and `lxc-exec`'s `main` now exits 1 on install failure instead of warning and continuing.
- **iptables chain/FORWARD hook leaks on fatal signal** (`signal_cleanup.rs:91`) → 7783829. Watchdog now tracks the host-side veth alongside the container name and tears down iptables before destroying the container; new `NetworkIptablesManager::force_cleanup` does the recovery without owning the original manager.
- **Comment overstated timeout protection** (`lxc_bindings.rs:333`) → folded into 13d644d. Comment now says the 5 s gate only delays the stdin forwarder; real timeout enforcement is still tracked by the existing TODO in `lxc_runner`.

`cargo fmt`, `cargo clippy --target aarch64-unknown-linux-gnu -- -D warnings`, and `cargo test -p lxc_common` (25 passed) all green.

bbonaby (approved):
Thanks!

bbonaby:
We have a few more PRs we probably want to get in before we spin up another release.  #284, #288, and #308. potentially #306 and #307 as well. Will keep you posted.

caarlos0:
cool, FWIW seems like PTY is not working on macos seatbelt as well, working on a fix 🙏🏻

bbonaby:
Thanks for the catch! I created this issue: https://github.com/microsoft/mxc/issues/311 feel free to update it however you wish with more details if you want. CC: @richiemsft, not sure if #288 will fix it.

PR #284, opened 2026-05-12
User/saulg/macos ci pipeline

richiemsft:
## 📖 Description

Adding macOS to the ci/cd pipeline. it will now build and run tests for every PR that targets main. Still pending is binary signing, @bbonaby requested the certs, still waiting on to hear back from ESRP folks.

## 🔍 Validation

Here is a test run https://github.com/microsoft/mxc/actions/runs/25749997112

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [x] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/284)


richiemsft:
@bbonaby, should I move this to the azure pipeline then?

bbonaby:
> @bbonaby, should I move this to the azure pipeline then?

yea that's right. You'd want to add a mac  option to the matrix here https://github.com/microsoft/mxc/blob/88b7487ec3dc8ce3877bb651c59643ba293585a7/.azure-pipelines/1ES.Build.yml#L61

then in https://github.com/microsoft/mxc/blob/main/.azure-pipelines/templates/Rust.Build.Job.yml you can follow the pattern for what I did for windows and linux with the compile time if statements. 

So you'd replicate a similar thing for windows like I do here: (you'll see linux is similar)  
https://github.com/microsoft/mxc/blob/88b7487ec3dc8ce3877bb651c59643ba293585a7/.azure-pipelines/templates/Rust.Build.Job.yml#L42

I'll update this: https://github.com/microsoft/mxc/blob/88b7487ec3dc8ce3877bb651c59643ba293585a7/.azure-pipelines/templates/Rust.Build.Job.yml#L113 with the ESRP signing stuff once approved so in this PR you can just focus on making sure it builds.

Once: https://github.com/microsoft/mxc/pull/297 goes in new pushes should kick off a build in the azure pipelines so you can confirm your changes are working.

.azure-pipelines/templates/Mac.Build.Job.yml
```diff
@@ -0,0 +1,80 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
bbonaby, line 1: issue: so with 1ES pipelines, if you don't use a centralized feed we will end up getting an SFI bug. What this means is that you can't use `cargo build|test` directly unless you first specify a centralized feed to pull crates from.

Thankfully, we have this already. If you look at the `Rust.Build.Job.yml` in this task https://github.com/microsoft/mxc/blob/16269fd5fd06f1699f7527247f59efe69ea284da/.azure-pipelines/templates/Rust.Build.Job.yml#L94

 It will append the content in `.azure-pipelines/.cargo/config.toml` : https://github.com/microsoft/mxc/blob/16269fd5fd06f1699f7527247f59efe69ea284da/.azure-pipelines/.cargo/config.toml#L2

At the end of the repo-wide config.toml file (`.cargo/config.toml`): https://github.com/microsoft/mxc/blob/16269fd5fd06f1699f7527247f59efe69ea284da/.cargo/config.toml#L1

We have it separated this way because you actually need to authenticate with the centralize private feed before you can start calling cargo to pull crates down.  Note: we do a similar thing for npm as well [here](https://github.com/microsoft/mxc/blob/16269fd5fd06f1699f7527247f59efe69ea284da/.azure-pipelines/templates/Package.NpmSdk.Job.yml#L45).

The pipeline internally does the authentication  as the feed and pipeline are located in the same ADO project. Outside of the pipeline we let devs use the public crates.io registry  so features can be developed fast.

FYI on how I know this will result in a SFI bug:
In the ADO pipeline log located [here](https://microsoft.visualstudio.com/Dart/_build/results?buildId=146870852&view=logs&jobId=1d24f4d5-38bd-5e08-8057-38328d0a84c0&j=1d24f4d5-38bd-5e08-8057-38328d0a84c0&t=06794591-8118-5f8f-d7a0-fdcbcd29ff7a) it has these lines:

##[warning]sdk/package.json - CFS0001: Missing sibling .npmrc file. (https://aka.ms/cfs/npm)
##[warning]sdk/tests/integration/package.json - CFS0001: Missing sibling .npmrc file. (https://aka.ms/cfs/npm)
##[warning].cargo/config.toml - CFS0042: Missing "replace-with" key in the "crates-io" section. (https://aka.ms/cfs/cargo)

.azure-pipelines/templates/Mac.Build.Job.yml
```diff
@@ -0,0 +1,80 @@
...
+    - script: cargo fmt --all -- --check
+      displayName: Check formatting
+      workingDirectory: $(workingDirectory)
+
+    # Clippy is scoped to the macOS-specific crates because the full
+    # workspace includes Windows-only crates that do not compile on macOS.
+    # wxc_common (a workspace dependency) has Windows-only diagnostic code
+    # that produces warnings on macOS; allow those specific warnings here
+    # to avoid false failures. The Windows CI lints the full workspace.
+    - script: cargo clippy -p mxc_darwin -p seatbelt_common --all-targets -- -D warnings -A unreachable-code -A unused-variables
```
bbonaby: suggestion: Would be preferable to add these overrides to places in the code where they happen with attributes like `#[allow(unused_variables)]` and `#[allow(unreachable_code)]` above said code

.azure-pipelines/templates/Package.NpmSdk.Job.yml
```diff
@@ -38,10 +38,17 @@ jobs:
...
     # Pipeline artifact downloads strip Unix execute permissions.
-    # Restore +x on Linux binaries so the tarball preserves them.
+    # Restore +x on Linux and macOS binaries so the tarball preserves them.
     - script: chmod +x $(sdkDirectory)/bin/*/lxc-exec
       displayName: Restore execute permission on lxc-exec
 
+    - script: |
+        set -euo pipefail
+        # Use `|| true` so the step succeeds when the mac binary is not
+        # present (e.g. the mac build job did not run for some reason).
```
bbonaby: issue: actually we'd want to fail if the mac build doesn't happen. Nicer this way so we don't accidently ship a package without the mac bits.

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
```diff
@@ -88,3 +97,19 @@ jobs:
...
+        # Restore execute bit on the mac binary; pipeline artifact
+        # downloads (via npm pack/unpack) drop Unix execute permissions.
+        - script: |
+            set -euo pipefail
+            BIN=$(find node_modules/@microsoft/mxc-sdk/bin -name mxc-exec-mac -print -quit)
+            if [ -n "$BIN" ]; then chmod +x "$BIN"; else echo "mxc-exec-mac not found in installed package" && exit 1; fi
+          workingDirectory: $(integrationDirectory)
+          displayName: Restore execute permission on mxc-exec-mac
+
+        - script: node --test dist/macos-seatbelt.test.js
```
bbonaby: issue: we'd still want to do `npm test` here. The windows/linux dependent ones shouldn't run and if they do we should make sure in the tests if the platform isn't windows or linux they're skipped. `npm test` mainly because there are common tests that we want to run as well like:
- https://github.com/microsoft/mxc/blob/main/sdk/tests/integration/common.test.ts. Note: the commandline for these have cmd.exe, but that's ok, these are dry-run tests that will confirm the flow of SDK -> executable -> config validation -> return. Without needing to actually spawn the sandbox. It's a good way for us to make sure the sdk can actually invoke the executable.
- and https://github.com/microsoft/mxc/blob/main/sdk/tests/integration/package.test.ts. This one makes sure all binaries are in their expected place and accounted for.

we'd need to make sure macos is covered in those.

.azure-pipelines/1ES.Build.yml
```diff
@@ -119,3 +127,10 @@ extends:
       - template: .azure-pipelines/templates/SDK.Integration.Test.Job.yml@self
         parameters:
           debug: ${{ parameters.debug }}
+          targets:
+            - name: windows
+              os: windows
+            - name: linux
+              os: linux
+            - name: macos
+              os: macos
```
bbonaby: issue: no need to explicitly add the targets here, you can add macos at the end of `.azure-pipelines/templates/SDK.Integration.Test.Job.yml`'s targets paramater list at the top, which will have the same effect.

bbonaby:
whoops! wanted to try some stuff out and accidently pushed to origin. Will revert

.azure-pipelines/templates/Mac.Build.Job.yml
```diff
@@ -0,0 +1,80 @@
...
+    targetTriple: ${{ parameters.triplet }}
+    targetTripleDir: $(CARGO_TARGET_DIR)/$(targetTriple)/release
+    outputDirectory: $(Build.SourcesDirectory)/out
+    artifactBaseName: mxc-binaries-$(targetTriple)
+    workingDirectory: $(Build.SourcesDirectory)/src
+
+  steps:
+    - checkout: self
+
+    - script: |
```
bbonaby: note: for supplychain attacks we'd want to be using the Microsoft version of the toolchain. See: https://github.com/microsoft/mxc/blob/796cbc3e1678da3b0d968d2706af64dbc88ffe29/.azure-pipelines/templates/Rust.Build.Job.yml#L75

See: https://eng.ms/docs/coreai/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/buildworkflows/rust. I believe you should still be able to use it in this Microsoft pool.

.azure-pipelines/templates/Mac.Build.Job.yml
richiemsft: The warnings are a pre-existing issue, is there a task on fixing them for all other pipelines as well?

bbonaby: Sorry let me explain it better. So, the scanning (which leads to the warnings) happens before any of our jobs run. Initially looks at the `config.toml` file before any of our jobs run and never again. What we do as one of the first steps in the rust build job, is update the repositories `config.toml` file with the info about the private feed. I brought it up in this PR because I saw that your Mac.Build.Job didn't update the `config.toml` file, and that crates we're being downloaded from crates.io instead of our private feed. This would have resulted in an eventual SFI bug.

here is a quick example of a before and after now that you've updated your PR to use the config.toml during the mac build: 

**Your previous MAC clippy job**: https://microsoft.visualstudio.com/Dart/_build/results?buildId=146968710&view=results

Image

**Now your current MAC clippy job**: https://microsoft.visualstudio.com/Dart/_build/results?buildId=146968710&view=logs&j=1d24f4d5-38bd-5e08-8057-38328d0a84c0&t=a5f9eab3-2ee5-51d4-6f6e-ace83bbcc340

Image


Let me know if that makes sense.

.azure-pipelines/templates/Mac.Build.Job.yml
bbonaby: I guess you couldn't update this because it was 1ES specific? Can you try using this guide: https://eng.ms/docs/more/languages-at-microsoft/rust/articles/gettingstarted/install/installcicd and see if it works?  You should be able to use the same nuget feed that is used in the Rust.Build.Job.yml file.

Specifically this part:

image

You'll want to make sure you move this task in your current PR: https://github.com/microsoft/mxc/blob/eb1397f5d2f299d6b1214a84503f81d4bc017c58/.azure-pipelines/templates/Mac.Build.Job.yml#L55 to be the first thing you do. (the update of the config.toml file). All else fails they do have a script in the guide we can use.

.azure-pipelines/templates/Mac.Build.Job.yml
richiemsft: got it! thanks I was confused on the first link and warnings you shared above, I kept seeing those after the change.

bbonaby (approved):

PR #288, opened 2026-05-12
add GUI app support via sandbox_init

richiemsft:
## 📖 Description
Add GUI application support for the macOS Seatbelt sandbox backend and simplify the primary execution mechanism to use sandbox_init().

Key changes:

 - Unified execution path (exec): Replace sandbox-exec CLI with sandbox_init() via pre_exec for the primary sandbox execution path. This eliminates tempfile management and drops the tempfile dependency. sandbox_init is a strict superset. Everything that worked with sandbox-exec continues to work.
 - GUI app support: Add guiAccess option to experimental.seatbelt that enables GUI applications to create windows inside the sandbox. When enabled, the profile adds wildcard mach-lookup, mach-register, and iokit-open rules so apps can connect to WindowServer, CoreAnimation, and GPU services.
 - LaunchServices path (launchMethod: "open"): For Apple system apps (e.g., Terminal.app) that enforce Launch Constraints and are killed when exec'd by unauthorized parents. 
Launches via open -n -W -a Terminal, then applies the sandbox to the inner shell via /usr/bin/sandbox-exec in a helper script. sandbox-exec is unavoidable here because Terminal.app owns the fork, we cannot inject pre_exec into another app's process spawning. Currently supports Terminal.app only.

How it works:

 - exec path: sandbox_init() applies the Seatbelt profile inside pre_exec (after fork(), before exec()). The child inherits the parent's Mach bootstrap namespace, allowing GUI services to be reachable. The profile string is passed directly.
 - open path: A secure temp helper script (created with O_EXCL, restrictive permissions) wraps the user command with sandbox-exec -f <profile>. Terminal.app executes this .command file via LaunchServices.

## 🔍 Validation
 - Unit tests: All 20 seatbelt tests pass (17 existing + 3 new GUI tests)
 - Examples: All macOS examples (15–26) verified.
 - Manual GUI testing:
  - Terminal.app (launchMethod: "open"): clean launch, network blocked, sandbox confirmed
  - Alacritty (launchMethod: "exec"): window rendered, network blocked
  - Copilot CLI inside sandboxed Terminal: functional with network allowed
  - Calculator/TextEdit: confirmed killed by Launch Constraints (documented limitation)
  - Custom Swift NSWindow app: window rendered successfully inside sandbox
  - Clipboard test (example 24/25): UI disabled → exit 1, clipboard enabled → exit 0

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/288)


src/seatbelt_common/src/profile_builder.rs
```diff
@@ -232,6 +277,21 @@ fn write_ui_rules(out: &mut String, request: &CodexRequest) {
     }
 }
 
+/// Expand a leading `~` or `~/` to the current user's home directory.
+/// Passes through paths that don't start with `~` unchanged.
+fn expand_tilde(path: &str) -> String {
+    if path == "~" {
+        std::env::var("HOME").unwrap_or_else(|_| "~".to_string())
```
bbonaby: question (non-blocking): so if HOME isn't in the environment variables, we just return the `~`. Is that ok? Just wonder if we should error out at this point since returning the tilde defeats the purpose of the function.

bbonaby (approved):

src/seatbelt_common/src/profile_builder.rs
richiemsft: highly unlikely that HOME is missing from macOS. I can error out if HOME is missing instead of returning ~

bbonaby (approved):

PR #294, opened 2026-05-12
Phase 2: fallback_detector module for tier selection

MGudgin:
## 📖 Description

Adds a pure detection module that produces a `TierDecision` from policy + runtime probes (BaseContainer API, `bfscfg.exe` presence, `WRITE_DAC`), plus security hardening for the Tier 2 (AppContainer + BFS) execution path.

### Module

- **Three-tier model:** BaseContainer / AppContainerBfs / AppContainerDacl
- `allow_dacl_fallback=false` blocks any tier requiring host DACL modification (including T1/T2 deny-augmentation)
- `MXC_FORCE_TIER` injection seam (debug builds only) for testing
- **Pure module:** no `Logger` dependency, no telemetry emission. Phase 4 wires this into the dispatcher.

## 🛡️ Security hardening

1. **Environment-driven Tier 2 → Tier 3 downgrade.** The original probe read the `SystemRoot` env var when locating `bfscfg.exe`, letting an attacker who could influence the process environment (parent process, persisted user env, service definition) force a silent demotion to Tier 3 (DACL) even with the real binary in System32. Release builds now resolve via `GetWindowsDirectoryW` exclusively — the OS populates it from boot configuration, not the process environment.

2. **Executable-search-order hijacking.** `FileSystemBfsManager` previously invoked `CreateProcessW` with `lpApplicationName = NULL` and a bare `bfscfg.exe` command line, exposing the standard image-dir → CWD → System32 → System → Windows → PATH search order. An attacker who could drop `bfscfg.exe` next to `wxc-exec.exe`, in CWD, or earlier on `PATH` would have their binary executed as the BFS broker regardless of what the probe concluded. `find_bfscfg_exe` now returns the absolute path resolved at probe time; `FileSystemBfsManager` passes it as `lpApplicationName` so Windows loads exactly that binary and skips the search order.

Probe and execution share one resolution via `TierDecision.bfscfg_path` — single source of truth, no TOCTOU gap between tier decision and broker invocation.

**Debug-only test seam:** `MXC_BFSCFG_PATH` overrides resolution under `#[cfg(debug_assertions)]`. Release builds never read it, so it cannot be used as an attack vector in production.

A follow-up TODO in `fallback_detector.rs` flags `BaseContainerRunner`s `LoadLibrary` on `processmodel.dll` as the next surface to audit for the same DLL-search-order family of attacks (Tier 1 surface).

## 🔗 References

Stacked on top of #293 (phase1_integrate → phase1d).

## 🔍 Validation

- `cargo test -p wxc_common` — 357 tests pass (16 `fallback_detector` incl. 4 new for the `MXC_BFSCFG_PATH` seam; 8 `filesystem_bfs` incl. 1 new for absolute-path quoting)
- End-to-end coverage lives in `Win25H2Safe-Tests.ps1` in phase5 (held)

### DIFF
--- a/src/wxc/src/main.rs
+++ b/src/wxc/src/main.rs
@@ -151,8 +151,23 @@ fn print_error_envelope(error: &MxcError) {
 }
 
 fn delete_app_container_profile(name: &str, logger: &mut Logger) -> bool {
-    // Clear BFS policy first
-    let mut bfs = FileSystemBfsManager::new(name.to_string());
+    // Clear BFS policy first. We need an absolute path to `bfscfg.exe`
+    // here for the same security reason as the runner — pass an
+    // authoritative path to `CreateProcessW` rather than a bare name.
+    // If resolution fails (rare; only on hosts where
+    // `GetWindowsDirectoryW` itself returns 0), we log and skip the BFS
+    // clearing step; deleting the AppContainer profile below is still
+    // worth attempting.
+    let bfscfg_path = match wxc_common::fallback_detector::find_bfscfg_exe() {
+        Ok(p) => p,
+        Err(e) => {
+            logger.log_line(&format!(
+                "Skipping BFS policy clear: could not resolve bfscfg.exe ({e})"
+            ));
+            None
+        }
+    };
+    let mut bfs = FileSystemBfsManager::new(name.to_string(), bfscfg_path);
     bfs.remove_configuration(logger);
 
     // Delete the AppContainer profile
--- a/src/wxc_common/src/fallback_detector.rs
+++ b/src/wxc_common/src/fallback_detector.rs
@@ -1,0 +1,729 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Fallback tier detector.
+//!
+//! Pure detection module that, given a parsed [`ContainerPolicy`] and a few
+//! runtime probes, produces a [`TierDecision`]. Tiers are described in
+//! `docs/proposals/downlevel_support/basecontainer-fallback-plan-v2.md`:
+//!
+//! 1. **Tier 1 — BaseContainer** (`Experimental_CreateProcessInSandbox`)
+//! 2. **Tier 2 — AppContainer + BFS** (`bfscfg.exe`-driven filesystem policy)
+//! 3. **Tier 3 — AppContainer + DACL** (host-side DACL ACE augmentation)
+//!
+//! This module does not log, emit telemetry, or have any side effects. It is
+//! intentionally Logger-free so it can be unit-tested in isolation. Phase 4
+//! will wire it into the dispatcher in `main.rs`.
+
+use std::path::{Path, PathBuf};
+
+use crate::models::ContainerPolicy;
+
+/// Selected isolation tier. The variant order corresponds to descending
+/// security strength.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum IsolationTier {
+    /// Tier 1 — `Experimental_CreateProcessInSandbox` from `processmodel.dll`.
+    BaseContainer,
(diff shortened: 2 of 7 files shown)


src/wxc_common/src/fallback_detector.rs
```diff
@@ -0,0 +1,499 @@
...
+    crate::base_container_runner::BaseContainerRunner::is_base_container_api_present().is_ok()
+}
+
+/// Returns `true` when `%SystemRoot%\System32\bfscfg.exe` is on disk.
+///
+/// We deliberately do not look in `SysWOW64`: `bfscfg.exe` is shipped only
+/// in the native System32 directory.
+pub(crate) fn is_bfscfg_present() -> bool {
+    let system_root = std::env::var("SystemRoot").unwrap_or_else(|_| "C:\\Windows".to_string());
+    let mut p = PathBuf::from(system_root);
```
shschaefer: Are you at all concerned about a non-standard Windows location?  Removal of the environment variable will cause BFS to downgrade.

MGudgin: I don't understand the question. The code looks for SystemRoot (which on my machine says 'C:\windows'). If for some reason SystemRoot is not defined, it will fall back to c:\windows. Do you think I should do a heuristic walk of the disk(s) if SystemRoot is not defined?

src/wxc/src/main.rs
```diff
@@ -151,8 +151,23 @@ fn print_error_envelope(error: &MxcError) {
 }
 
 fn delete_app_container_profile(name: &str, logger: &mut Logger) -> bool {
```
jsidewhite, line 153: >delete_app_container_profile [](http://example.com/codeflow?start=3&length=28)

Annoying the cleanup function isn't in the same file as the setup function (AC_runner.rs).

src/wxc_common/src/fallback_detector.rs
```diff
@@ -0,0 +1,729 @@
...
+/// is the **absolute** path the caller MUST pass to
+/// `CreateProcessW`'s `lpApplicationName` (or as a quoted absolute
+/// argv[0]) so probe and execution agree on which binary they're
+/// talking about.
+///
+/// Resolution policy:
+///
+/// - **Release builds** consult `GetWindowsDirectoryW` exclusively. The
+///   `SystemRoot` environment variable is deliberately ignored to deny
+///   an attacker who can scrub or rewrite the process environment a
```
jsidewhite, line 289: Is this a real threat?  If an attacker can manipulate the environment of wxc-exec.exe, then it's it's already in.

Is this merely defense in depth?

src/wxc_common/src/filesystem_bfs.rs
```diff
@@ -183,8 +233,23 @@ fn test_for_root_path(path: &str) -> bool {
...
+    let mut cmd_line = if exe_path.contains(' ') {
+        let escaped = if exe_path.ends_with('\\') {
+            format!("{}\\", exe_path)
+        } else {
+            exe_path.to_string()
+        };
+        format!("\"{escaped}\"")
+    } else {
+        exe_path.to_string()
+    };
```
jsidewhite, line 252: can't you use the path crate?

```

use std::path::Path;

let p = Path::new(r"C:\Windows").join("System32").join("kernel32.dll");

```

or

```

use std::path::PathBuf;

let mut path = PathBuf::from(r"C:\");
path.push("Windows");
path.push("System32");
path.set_extension("dll");

println!("{}", path.display());

```

src/wxc_common/src/fallback_detector.rs
jsidewhite: ohhh it's moreso this,

```
preventing
    /// executable-search-order hijacking by an attacker who can plant a
    /// rogue `bfscfg.exe` next to `wxc-exec.exe`
```

jsidewhite (approved):
:shipit:

PR #295, opened 2026-05-12
Phase 3: filesystem_dacl module - crash-safe DACL manager

MGudgin:
## Description

Adds `DaclManager` for granting/denying AppContainer SID access on
host filesystem paths, with full restoration on `Drop`, explicit
`restore()`, and orphaned-state recovery on MXC startup.

Key design points:

- **Inheritable ACEs** (`OI|CI` on directories); `SetNamedSecurityInfoW`
  propagation handles add and remove.
- **Atomic state file** at `%LOCALAPPDATA%\Microsoft\MXC\dacl-restore\<run-id>.json`
  (stage → fsync → rename). Recovery on startup reaps files whose
  owning PID is dead, comparing `GetProcessTimes` against
  `started_at_filetime` to defeat PID reuse.
- **`prior_state` capture**: every explicit ACE for our SID is
  recorded before apply; restore issues `REVOKE_ACCESS` for the
  SID and re-grants the captured entries, surviving
  `SetEntriesInAclW`'s rights-coalescing for the same trustee.
- **Per-path named mutex** serializes concurrent MXC instances and is
  held across scan + persist + apply.
- **Per-entry retain-on-failure** for restore; transient failures
  retry on the next `restore()` or next-startup recovery.
- **Network paths rejected** upfront.

Tests scope `MXC_DACL_STATE_DIR` to a per-test tempdir (serialized
within a process by a static mutex) and run as part of default
`cargo test --workspace`. No elevation required.
`test_scripts\run_dacl_tests.ps1` is a focused convenience runner.

Recovery is invoked unconditionally at MXC startup. Tier-specific
`DaclManager` use lands in Phase 4 dispatcher integration.

## References

Stacked on top of #294 (phase2 → phase1_integrate).

## Validation

- `cargo test -p wxc_common filesystem_dacl::` → 26/26 pass, 0 ignored
- Full `cargo test --workspace` clean
- Crash-recovery end-to-end coverage lives in `Win25H2Safe-Tests.ps1` in phase5 (held)

### DIFF
--- a/src/wxc_common/src/filesystem_dacl.rs
+++ b/src/wxc_common/src/filesystem_dacl.rs
@@ -1,0 +1,1800 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Crash-safe DACL manager for the AppContainer + DACL tier (T3) and for
+//! deny-ACE augmentation under T1/T2.
+//!
+//! See `docs/proposals/downlevel_support/basecontainer-fallback-plan-v2.md`.
+//!
+//! # Design
+//!
+//! - **Inheritable ACEs**: on directories we apply `OBJECT_INHERIT_ACE |
+//!   CONTAINER_INHERIT_ACE`. We rely on `SetNamedSecurityInfoW` automatic
+//!   propagation for both add AND remove — no manual descendant walk.
+//! - **Crash safety**: every applied ACE is appended to a per-process state
+//!   file under `%LOCALAPPDATA%\Microsoft\MXC\dacl-restore\<run-id>.json`
+//!   *before* the Win32 apply call. On startup, [`recover_orphaned_state`]
+//!   inspects every state file and reaps entries owned by dead processes.
+//! - **Concurrency**: each path acquires a named mutex
+//!   (`Local\Microsoft.MXC.Dacl.<hash16>`) for the duration of the
+//!   scan/persist/apply sequence so two concurrent MXC instances on the
+//!   same path serialize cleanly.
+//! - **ACE-merge resilience**: `SetEntriesInAclW(GRANT)` coalesces rights
+//!   for the same trustee into a single ACE. Before each apply we
+//!   capture **every** explicit (non-inherited) ACE for our SID — of
+//!   either type — into [`AppliedAce::prior_state`]. On restore we
+//!   issue `REVOKE_ACCESS` for the SID (atomic strip of all our
+//!   contributions, merged or otherwise) and then re-grant/deny each
+//!   prior entry verbatim, restoring the host to its exact prior state.
+//!
+//! # State directory
+//!
+//! The default state directory is `%LOCALAPPDATA%\Microsoft\MXC\dacl-restore`.
+//! Tests and specialized deployments can override this via the
+//! `MXC_DACL_STATE_DIR` environment variable. The directory contains
+//! up to three file species:
+//!
+//! - `<run-id>.json` — a fully written, owner-active state file.
+//! - `<run-id>.json.tmp` — an in-progress atomic write; cleaned up at
+//!   the start of the next write to the same path, otherwise harmless
+//!   leftover from a crash mid-write.
+//! - `<run-id>.json.corrupt` — quarantine of a state file that failed
+//!   to parse during recovery, preserved for operator inspection.
+//!
+//! # Errors
+//!
+//! Apply paths surface [`DaclError`] on failure. [`DaclManager::restore`]
+//! collects per-path failures into the manager's warning list and only
+//! returns an outer error for catastrophic state-file I/O. Entries
+//! whose Win32 restore failed are **retained** in the in-memory list
+//! and on disk so a future `restore()` call or
+//! [`recover_orphaned_state`] on the next startup can retry.
+//!
+//! # Drop
+//!
+//! [`DaclManager`] best-effort restores on drop; errors are swallowed and
+//! logged to stderr.
+
+// Note: the module-level `#[cfg(target_os = "windows")]` lives on the
+// `mod filesystem_dacl;` declaration in `lib.rs`. Repeating it here as
+// an inner `#![cfg(...)]` trips clippy's `duplicated_attributes` lint
+// on toolchains 1.78+.
+
+use std::ffi::c_void;
+use std::fs;
+use std::io::{self, Write};
+use std::os::windows::ffi::OsStrExt;
+use std::path::{Path, PathBuf};
+use std::ptr;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use serde::{Deserialize, Serialize};
+
+use windows::core::{PCWSTR, PWSTR};
+use windows::Win32::Foundation::{
+    CloseHandle, GetLastError, LocalFree, ERROR_SUCCESS, FILETIME, HANDLE, HLOCAL, WAIT_ABANDONED,
+    WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT, WIN32_ERROR,
+};
+use windows::Win32::Security::Authorization::{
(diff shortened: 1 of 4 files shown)


src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -0,0 +1,1176 @@
...
+            }
+        }
+    }
+
+    fn persist_state(&self) -> Result<(), DaclError> {
+        let state = StateFile {
+            run_id: self.run_id.clone(),
+            pid: std::process::id(),
+            image_name: current_image_basename(),
+            started_at_unix: SystemTime::now()
```
shschaefer: Would using Windows time be better than std?  Are you using this to test against file creation times or just as a marker?

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -0,0 +1,1176 @@
...
+    };
+    let s = canonical.to_string_lossy();
+    // canonicalize emits `\\?\C:\...` for local drives and `\\?\UNC\...` for
+    // network paths. Reject the latter.
+    if s.starts_with(r"\\?\UNC\") || s.starts_with(r"\\?\unc\") {
+        return Err(DaclError::NetworkPathRejected(canonical));
+    }
+    // Also reject raw `\\server\share` that didn't go through `\\?\UNC`
+    // (rare; canonicalize on Windows usually rewrites it).
+    if s.starts_with(r"\\") && !s.starts_with(r"\\?\") {
```
shschaefer: What about DOS Device paths?  \\.\Volume{}\...

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -0,0 +1,1176 @@
...
+// Win32: SID parsing (RAII)
+// -------------------------------------------------------------------------
+
+/// Owned PSID returned by [`ConvertStringSidToSidW`]. Frees via `LocalFree`
+/// on drop.
+struct OwnedSid(PSID);
+
+impl OwnedSid {
+    fn parse(s: &str) -> Result<Self, DaclError> {
+        let wide: Vec<u16> = s.encode_utf16().chain(std::iter::once(0)).collect();
```
shschaefer, line 738: Any guards here?  Length, etc...?

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -0,0 +1,1176 @@
...
+    fn acquire(canonical: &Path) -> Result<Self, DaclError> {
+        let name = mutex_name_for(canonical);
+        let wide: Vec<u16> = name.encode_utf16().chain(std::iter::once(0)).collect();
+        let handle = unsafe { CreateMutexW(None, false, PCWSTR(wide.as_ptr())) }.map_err(|e| {
+            DaclError::Win32 {
+                path: canonical.to_path_buf(),
+                reason: format!("CreateMutexW: {e}"),
+            }
+        })?;
+        let wait = unsafe { WaitForSingleObject(handle, INFINITE) };
```
shschaefer: Do you want INFINITE?

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -0,0 +1,1176 @@
...
+        let parsed: StateFile = serde_json::from_slice(&bytes).unwrap();
+        assert_eq!(parsed.run_id, s.run_id);
+        assert_eq!(parsed.pid, s.pid);
+        assert_eq!(parsed.applied.len(), 1);
+        assert_eq!(parsed.applied[0].access_mask, 0x12_3456);
+        assert_eq!(parsed.applied[0].ace_type, AceType::Allow);
+    }
+
+    #[test]
+    fn unc_path_rejected() {
```
shschaefer: This is not testing your canonicalize function....

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -0,0 +1,1176 @@
...
+        // Synthesize a UNC-style canonical path. We can't use canonicalize
+        // (no real share) so we exercise the suffix-string check directly.
+        let p = PathBuf::from(r"\\?\UNC\server\share\foo");
+        // Mimic canonicalize_local's post-canonicalize check.
+        let s = p.to_string_lossy();
+        assert!(s.starts_with(r"\\?\UNC\"));
+    }
+
+    #[test]
+    fn raw_unc_rejected() {
```
shschaefer: Same here.

shschaefer (changes requested):

src/wxc_common/src/filesystem_dacl.rs
MGudgin: Switched to a Windows FILETIME (100-ns ticks since 1601-01-01 UTC) and renamed the field to `started_at_filetime`. It's still just a debug/observability marker (we never compare it against filesystem timestamps in the recovery path), but the stored value is now in the Windows-native representation so a future `CreateFile` / NTFS timestamp comparison is trivial. I went with a small std-time -> FILETIME conversion rather than pulling in the `Win32_System_SystemInformation` feature to keep the workspace dependency surface unchanged. Fixed in 884289a.

src/wxc_common/src/filesystem_dacl.rs
MGudgin: Good catch — the old check `starts_with(\\\\) && !starts_with(\\\\?\\)` incorrectly classified DOS device paths like `\\.\Volume{GUID}\...` as network. The check is now extracted into `ensure_local_canonical_prefix`: `\\?\X:\...` (Win32 file namespace) and `\\.\...` (DOS device namespace) are both accepted as local; only `\\?\UNC\...` (case-insensitive) and bare `\\server\share\...` are rejected. New unit tests in 884289a cover the volume-GUID and `\\.\C:\` cases.

src/wxc_common/src/filesystem_dacl.rs
MGudgin: Folded into the `ensure_local_canonical_prefix_rejects_raw_unc` test in 884289a, which now exercises the production helper with `\\server\share\foo`.

src/wxc_common/src/filesystem_dacl.rs
MGudgin: No — replaced `INFINITE` with a bounded 30 s timeout (`PATH_MUTEX_WAIT_MS`) plus a new `DaclError::MutexTimeout { path, timeout_ms }` variant. The acquire path now also handles `WAIT_ABANDONED` (previous holder died without releasing) by accepting the mutex with a stderr note — orphan recovery on startup reconciles the on-disk DACL state anyway. `WAIT_FAILED` still surfaces as a `DaclError::Win32` with the `GetLastError` reason. Fixed in 884289a.

src/wxc_common/src/filesystem_dacl.rs
MGudgin: Added an explicit length cap (`MAX_SID_STRING_LEN = 256`) and an empty-string rejection up front, before allocating the UTF-16 buffer or calling into Win32. Well-formed SID strings are well under 200 bytes (15 sub-authorities ≈ 184 chars max), so 256 is a generous bound that still rejects obviously-malformed input early. New `sid_parse_rejects_empty_and_oversized` test covers both cases. Fixed in 884289a.

src/wxc_common/src/filesystem_dacl.rs
MGudgin: Right — the old test was re-implementing the suffix check rather than calling the production code. The post-canonicalize validation is now extracted into `ensure_local_canonical_prefix(&Path) -> Result<(), DaclError>` (called by `canonicalize_local`), and the new tests `ensure_local_canonical_prefix_accepts_local`, `ensure_local_canonical_prefix_rejects_unc_namespace`, and `ensure_local_canonical_prefix_rejects_raw_unc` exercise that helper directly. Fixed in 884289a.

shschaefer (approved):

shschaefer (approved):

PR #300, opened 2026-05-13
phase4.5: remove Tier 2 (AppContainer + bfscfg.exe) support

MGudgin:
## 📖 Description

Tier 2 (AppContainer + `bfscfg.exe`-driven filesystem policy) is removed from the fallback stack. The remaining algorithm is **Tier 1 → Tier 3**: BaseContainer when its API is present, else AppContainer + DACL.

**Why T2 has to go:**

- On Windows 11 **25H2**, `bfscfg.exe` is present but fatal — calling it hard-locks `bfs.sys`. The `Win25H2Safe-Tests.ps1` harness (phase 5) was already written to avoid T2 for exactly this reason.
- On **post-25H2** builds, Tier 1 (`Experimental_CreateProcessInSandbox`) is the preferred path; T2 was only ever a stepping stone.
- On **pre-T1** builds with a working T2, Tier 3 (AppContainer + DACL) is also viable and is what we ship anyway as the universal fallback.

Net: there is no environment where selecting T2 is the right answer.

## 🔍 What changes

| File | Change |
|---|---|
| `telemetry.rs` | Drop `IsolationTier::AppContainerBfs` variant + matching arm + test |
| `fallback_detector.rs` | Drop T2 selection arm, `is_bfscfg_present()` probe, `appcontainer-bfs` `MXC_FORCE_TIER` value, T2-specific tests; rework the invalid-force-tier test to assert the new T3-only fallback |
| `dispatcher.rs` | Drop `AppContainerBfs` dispatch arm; T3 path uses `AppContainerScriptRunner::new()` directly; `DispatchError` message no longer mentions `bfscfg.exe` |
| `appcontainer_runner.rs` | Drop the `FilesystemMode` enum, `with_filesystem_mode` constructor, both BFS call sites in `execute()` (configure + cleanup) |
| `filesystem_bfs.rs` | Deleted in full (-278 lines) |
| `lib.rs` | Drop `pub mod filesystem_bfs;` |
| `wxc/src/main.rs` | Drop the unconditional `FileSystemBfsManager::remove_configuration` call from `delete_app_container_profile`. This call ran on every profile delete and would have hit the same `bfs.sys` hang on 25H2 that the fallback path is designed to avoid — silent latent bug that goes away here |

**Net diff:** ~35 added, ~450 deleted.

## ⚠️ Upgrade-leak note

On systems where an older mxc left BFS rules on an AppContainer profile, those rules are no longer cleaned up by `delete_app_container_profile`. The rules are AppContainer-scoped and become inert once the profile is deleted by the OS lifecycle, so this is judged acceptable. If we ever want to clean them up, a one-shot migration tool is the right place — not the hot path of every profile delete.

## 🔗 References

Stacked on top of #296 (phase4). Phase 5 (held; not yet PR'd) currently re-introduces the `appcontainer-bfs` string in `probe.rs` and the TypeScript SDK `IsolationTier` type; phase 5 will be rebased on top of phase 4.5 and those references removed at that time.

## 🔍 Validation

- `cargo build --workspace` clean
- `cargo test -p wxc_common --lib` — 375/375 pass
- End-to-end coverage (phase 5 harness already exercises only T1/T3, not T2) will continue to pass once phase 5 is rebased.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/300)


MGudgin:
Superseded by PR #410

PR #321, opened 2026-05-15
BaseContainer: Ephemeral sandbox lifecycle cleanup (Part 1)

jsidewhite:
## BaseContainer: Ephemeral sandbox lifecycle tracking

Implements proper AppContainer profile cleanup for the BaseContainer runner when `destroy_on_exit: true` is configured.

### What it does

- **Ephemeral identities**: Every BaseContainer launch gets a unique `sandbox-{random hex}` identity, ensuring no profile collisions between concurrent sandboxes.
- **Registry tracking**: Before launch, writes a tracking entry under `HKCU\Software\Classes\Local Settings\Software\Microsoft\Windows\CurrentVersion\ProcessSandboxes\Mappings\{SID}` with metadata (identity, SID, origin, requested identity, timestamps). A volatile `Active` subkey marks in-flight sandboxes for crash detection.
- **Cleanup on exit**: ==FUTURE==
- **Deferred cleanup**: If a network proxy is configured, cleanup is skipped (proxy state can't be torn down yet). A `CleanupDeferred` marker is written to the tracking entry for future tooling.
- **Ctrl+C handling**: Registers a console control handler so ==FUTURE== cleanup still runs on interrupt.
- **No-op when `destroy_on_exit: false`**: No tracking, no cleanup -- the profile persists across runs.

### Why

Previously, BaseContainer never cleaned up AppContainer profiles. Over time this leaks registry state and BFS policies. This in the ==FUTURE== brings BaseContainer in line with LXC's existing cleanup behavior and follows the Tessera sandbox lifecycle spec.

This PR adds ephemeral sandbox lifecycle cleanup to the BaseContainer runner. When destroy_on_exit: true, each sandbox gets a unique random identity (sandbox-{hex}), a registry tracking entry under HKCU\...\ProcessSandboxes\Mappings\{SID} for crash recoverability.  It also installs a Ctrl+C handler so cleanup runs on interrupt, and defers cleanup when a network proxy is active since proxy state can't yet be torn down.

### Architecture
```
 ┌─────────────────────────────────────────────────────────┐
 │              base_container_runner.rs                    │
 │  (ScriptRunner::run_script)                             │
 │                                                         │
 │  1. Generate ephemeral identity ──┐                     │
 │  2. Derive SID                    │                     │
 │  3. Write tracking entry ─────────┼──► Registry         │
 │  4. Register Ctrl+C handler       │    (HKCU\...\       │
 │  5. CreateProcessInSandbox        │     Mappings\{SID}) │
 │  6. WaitForSingleObject           │                     │
 │  7. run_sandbox_cleanup() // FUTURE ────────┘                     │
 │       ├─► filesystem_bfs::clear_policy()                │
 │       ├─► DeleteAppContainerProfile()                   │
 │       └─► delete_tracking_key()                         │
 │                                                         │
 │  [proxy?] ──► mark_cleanup_deferred()                   │
 │  [Ctrl+C] ──► ctrl_handler() ──► cleanup or defer      │
 └─────────────────────────────────────────────────────────┘
```

### DIFF
--- a/src/wxc_common/src/base_container_runner.rs
+++ b/src/wxc_common/src/base_container_runner.rs
@@ -69,16 +69,16 @@ const ERROR_CALL_NOT_IMPLEMENTED: u32 = 120;
 /// SandboxSpec FlatBuffer schema version embedded in every spec payload.
 const SANDBOX_SPEC_VERSION: &str = "0.1.0";
 
-/// Best-effort sandbox cleanup: deletes the AppContainer profile and tracking
-/// entry unless a network proxy is configured (proxy state can't be torn down
-/// yet, so the tracking entry is retained with a deferral marker).
-fn run_sandbox_cleanup(identity: &str, sid_string: &str, proxy_enabled: bool, logger: &mut Logger) {
-    let _ = writeln!(logger, "{EMOJI_SECTION} SECTION: Lifecycle cleanup");
-    if !proxy_enabled {
-        sandbox_tracking::cleanup_sandbox(identity, sid_string, logger);
-    } else {
-        sandbox_tracking::mark_cleanup_deferred(sid_string, "network proxy configured", logger);
-    }
+/// Sandbox cleanup stub. The actual cleanup (DeleteAppContainerProfile, BFS
+/// policy removal, registry tracking deletion) is currently disabled because
+/// wxc-exec only tracks the main AppContainer process handle -- child processes
+/// may still be running when we reach this point. The tracking entry and
+/// ephemeral identity features remain active for diagnostics and future use.
+fn run_sandbox_cleanup(_identity: &str, _sid_string: &str, _proxy_enabled: bool, logger: &mut Logger) {
+    let _ = writeln!(
+        logger,
+        "{EMOJI_SECTION} SECTION: Lifecycle cleanup (skipping -- child process tracking not yet implemented)"
+    );
 }
 
 impl BaseContainerRunner {
--- a/src/wxc_common/src/base_container_runner.rs
+++ b/src/wxc_common/src/base_container_runner.rs
@@ -471,7 +471,7 @@ impl ScriptRunner for BaseContainerRunner {
         // Identity: when destroy_on_exit is true we generate a random ephemeral
         // identity so each sandbox gets a unique, cleanable AppContainer profile.
         // Otherwise we honour whatever the caller passed in (or the default).
-        let (identity, sid_string, tracking_written) = if request.lifecycle.destroy_on_exit {
+        let (identity, sid_string) = if request.lifecycle.destroy_on_exit {
             let ephemeral = sandbox_tracking::generate_sandbox_identity();
             let _ = writeln!(
                 logger,
@@ -493,25 +493,19 @@ impl ScriptRunner for BaseContainerRunner {
             };
 
             // Write registry tracking entry before launch so it survives crashes.
-            let written = if !sid.is_empty() {
+            if !sid.is_empty() {
                 let entry = TrackingEntry {
                     identity: ephemeral.clone(),
                     sid_string: sid.clone(),
                     destroy_on_exit: true,
                     requested_identity: request.container_id.clone(),
                 };
-                match sandbox_tracking::write_tracking_entry(&entry, logger) {
-                    Ok(()) => true,
-                    Err(e) => {
-                        let _ = writeln!(logger, "warning: tracking entry write failed: {}", e);
-                        false
-                    }
+                if let Err(e) = sandbox_tracking::write_tracking_entry(&entry, logger) {
+                    let _ = writeln!(logger, "warning: tracking entry write failed: {}", e);
                 }
-            } else {
-                false
-            };
+            }
 
-            (ephemeral, sid, written)
+            (ephemeral, sid)
         } else {
             let id = if request.container_id.is_empty() {
                 sandbox_tracking::generate_sandbox_identity()
@@ -523,7 +517,7 @@ impl ScriptRunner for BaseContainerRunner {
                 "destroy_on_exit=false; using identity '{}', no tracking",
                 id
             );
-            (id, String::new(), false)
+            (id, String::new())
         };
         let identity_wide = string_util::to_wide(&identity);
 
@@ -568,7 +562,7 @@ impl ScriptRunner for BaseContainerRunner {
             }
             // The OS may have created the AppContainer profile before failing,
             // so run the same cleanup logic used on normal exit.
-            if tracking_written {
+            if request.lifecycle.destroy_on_exit {
                 run_sandbox_cleanup(
                     &identity,
                     &sid_string,
@@ -594,7 +588,7 @@ impl ScriptRunner for BaseContainerRunner {
 
         // Register Ctrl+C handler so cleanup runs if wxc-exec is interrupted
         // while waiting for the child process.
-        if tracking_written {
+        if request.lifecycle.destroy_on_exit {
             sandbox_tracking::register_ctrl_c_cleanup(
                 &identity,
                 &sid_string,
@@ -630,9 +624,8 @@ impl ScriptRunner for BaseContainerRunner {
         let _ = writeln!(logger, "process exited with code {exit_code}");
 
         // 6. Sandbox cleanup: delete AppContainer profile and tracking entry.
-        //    Only runs when tracking was written (i.e., destroy_on_exit was true).
         //    Deferred if a network proxy is configured (proxy state can't be cleaned up yet).
-        if tracking_written {
+        if request.lifecycle.destroy_on_exit {
             run_sandbox_cleanup(
                 &identity,
                 &sid_string,
--- a/src/wxc_common/src/base_container_runner.rs
+++ b/src/wxc_common/src/base_container_runner.rs
@@ -74,7 +74,12 @@ const SANDBOX_SPEC_VERSION: &str = "0.1.0";
 /// wxc-exec only tracks the main AppContainer process handle -- child processes
 /// may still be running when we reach this point. The tracking entry and
 /// ephemeral identity features remain active for diagnostics and future use.
-fn run_sandbox_cleanup(_identity: &str, _sid_string: &str, _proxy_enabled: bool, logger: &mut Logger) {
+fn run_sandbox_cleanup(
+    _identity: &str,
+    _sid_string: &str,
+    _proxy_enabled: bool,
+    logger: &mut Logger,
+) {
     let _ = writeln!(
         logger,
         "{EMOJI_SECTION} SECTION: Lifecycle cleanup (skipping -- child process tracking not yet implemented)"
--- a/src/wxc_common/src/base_container_runner.rs
+++ b/src/wxc_common/src/base_container_runner.rs
@@ -30,6 +30,7 @@ use crate::models::{
     CodexRequest, NetworkEnforcementMode, NetworkPolicy, ProxyAddress, ScriptResponse,
 };
 use crate::proxy_coordinator::ProxyCoordinator;
+use crate::sandbox_tracking::{self, TrackingEntry};
 use crate::script_runner::{get_timeout_milliseconds, ScriptRunner};
 use crate::string_util;
 use sandbox_spec::base_container_layout::{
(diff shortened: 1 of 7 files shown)


src/wxc_common/src/base_container_runner.rs
```diff
@@ -545,6 +629,20 @@ impl ScriptRunner for BaseContainerRunner {
 
         let _ = writeln!(logger, "process exited with code {exit_code}");
 
+        // 6. Sandbox cleanup: delete AppContainer profile and tracking entry.
+        //    Only runs when tracking was written (i.e., destroy_on_exit was true).
+        //    Deferred if a network proxy is configured (proxy state can't be cleaned up yet).
+        if tracking_written {
+            run_sandbox_cleanup(
```
jsidewhite, line 638: >run_sandbox_cleanup [](http://example.com/codeflow?start=12&length=19)

@bbonaby - do I need an RAII-like container for this cleanup in case someone early returns or an exception happens?

(early return on line 617, for example)

src/wxc_common/src/base_container_runner.rs
```diff
@@ -455,11 +468,62 @@ impl ScriptRunner for BaseContainerRunner {
...
+            };
+
+            (ephemeral, sid, written)
         } else {
-            request.container_id.clone()
+            let id = if request.container_id.is_empty() {
+                sandbox_tracking::generate_sandbox_identity()
+            } else {
+                request.container_id.clone()
+            };
```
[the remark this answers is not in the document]
jsidewhite: This makes no sense.  Many callers sharing a default name will stomp each others' security policies in real time.

src/wxc_common/src/base_container_runner.rs
jsidewhite: @copilot do i need to wrap `run_sandbox_cleanup` in an RAII-like type?

src/wxc_common/src/sandbox_tracking.rs
```diff
@@ -0,0 +1,390 @@
...
+
+/// State captured so the console-ctrl handler can run cleanup on Ctrl+C or
+/// console close without any context pointer (Win32 handler is a bare fn).
+struct ActiveCleanup {
+    identity: String,
+    sid_string: String,
+    proxy_enabled: bool,
+}
+
+static ACTIVE_CLEANUP: OnceLock<Mutex<Option<ActiveCleanup>>> = OnceLock::new();
```
huzaifa-d, line 244: Suggestion - A Mutex<HashMap<String, ActiveCleanup>> with SID keys would clean up all active sandboxes.

src/wxc_common/src/base_container_runner.rs
```diff
@@ -68,6 +69,18 @@ const ERROR_CALL_NOT_IMPLEMENTED: u32 = 120;
...
 
+/// Best-effort sandbox cleanup: deletes the AppContainer profile and tracking
+/// entry unless a network proxy is configured (proxy state can't be torn down
+/// yet, so the tracking entry is retained with a deferral marker).
+fn run_sandbox_cleanup(identity: &str, sid_string: &str, proxy_enabled: bool, logger: &mut Logger) {
+    let _ = writeln!(logger, "{EMOJI_SECTION} SECTION: Lifecycle cleanup");
+    if !proxy_enabled {
+        sandbox_tracking::cleanup_sandbox(identity, sid_string, logger);
+    } else {
+        sandbox_tracking::mark_cleanup_deferred(sid_string, "network proxy configured", logger);
```
huzaifa-d: Where is the deferred cleanup handled?

src/wxc_common/src/base_container_runner.rs
```diff
@@ -455,14 +468,69 @@ impl ScriptRunner for BaseContainerRunner {
...
+            let ephemeral = sandbox_tracking::generate_sandbox_identity();
+            let _ = writeln!(
+                logger,
+                "{EMOJI_WARNING} destroy_on_exit=true: overriding caller identity '{}' -> '{}' for ephemeral cleanup",
+                request.container_id, ephemeral
+            );
+
+            // Derive the AppContainer SID for registry tracking.
+            // This is deterministic and does not require the profile to exist yet.
+            let sid = match sandbox_tracking::derive_sid_string(&ephemeral) {
```
bbonaby, line 489: Future: Today CreateProcessInSandbox takes a boolean parameter to say whether to use AppContainer or not. We don't plumb this through the schema, was your intention for us to eventually plumb it through and give folks the option to not use appcontainer? If so we'd need to remember to not do this in that case (the sid deriving).

src/wxc_common/src/base_container_runner.rs
```diff
@@ -455,14 +468,69 @@ impl ScriptRunner for BaseContainerRunner {
...
+                Ok(s) => {
+                    let _ = writeln!(logger, "derived SID: {}", s);
+                    s
+                }
+                Err(e) => {
+                    let _ = writeln!(logger, "warning: could not derive SID: {}", e);
+                    String::new()
```
huzaifa-d, line 496: Safe to return empty string?

src/wxc_common/src/sandbox_tracking.rs
```diff
@@ -0,0 +1,390 @@
...
+    let key_path = format!("{}\\{}", TRACKING_BASE, sid_string);
+
+    let hkcu = RegKey::predef(HKEY_CURRENT_USER);
+
+    // Delete Active subkey first (registry requires no child keys for deletion).
+    let active_path = format!("{}\\Active", key_path);
+    let _ = hkcu.delete_subkey(&active_path);
+
+    // Delete the tracking key itself.
+    match hkcu.delete_subkey(&key_path) {
```
huzaifa-d: Suggestion: Use delete_subkey_all?

src/wxc_common/src/sandbox_tracking.rs
```diff
@@ -0,0 +1,390 @@
...
+    // SAFETY: SID was allocated by the OS and must be freed with FreeSid.
+    unsafe {
+        FreeSid(psid);
+    }
+
+    sid_string
+}
+
+/// Convert a PSID to its SDDL string representation.
+fn sid_to_sddl(psid: PSID) -> Result<String, String> {
```
bbonaby: suggestion: unless my eyes deceive me other than the error handling difference, this is the same as the `sid_to_string` method in our string_util module [here](https://github.com/microsoft/mxc/blame/1739b40b0d9f71ea16c471ecc4c1b555cdaf544f/src/wxc_common/src/string_util.rs#L26). Think we can use that one?

src/wxc_common/src/sandbox_tracking.rs
```diff
@@ -0,0 +1,390 @@
...
+    let _ = writeln!(logger, "cleanup deferred: {}", reason);
+}
+
+/// Clean up a sandbox: clear BFS filesystem policies, delete the AppContainer
+/// profile, and remove the tracking registry entry. Best-effort and idempotent
+/// -- failures are logged but do not propagate as errors.
+///
+/// Order matters: BFS policies must be cleared before the AppContainer profile
+/// is deleted (BFS needs the SID, which is derived from the profile identity).
+pub fn cleanup_sandbox(identity: &str, sid_string: &str, logger: &mut Logger) {
```
bbonaby, line 169: suggestion: you can use the widestring crate to do this in a more cleaner way. Something like:

```rust
use widestring::U16CString;
use windows::core::PCWSTR;
use windows::Win32::Security::AppContainer::DeleteAppContainerProfile;

pub fn cleanup_sandbox(identity: &str, sid_string: &str, logger: &mut Logger) {
    // Step 1: Clear BFS filesystem policies (must happen before profile deletion).
    crate::filesystem_bfs::FileSystemBfsManager::clear_policy(identity, logger);

    // Step 2: Delete the AppContainer profile.
    let Ok(wide_identity) = U16CString::from_str(identity) else {
        let _ = writeln!(
            logger,
            "warning: failed to convert AppContainer identity to UTF-16: {}",
            identity
        );
        return;
    };

    // SAFETY: `wide_identity` is a valid null-terminated UTF-16 string.
    match unsafe { DeleteAppContainerProfile(PCWSTR(wide_identity.as_ptr())) } {
        Ok(()) => {
            let _ = writeln!(logger, "deleted AppContainer profile: {}", identity);
        }
        Err(e) => {
            let _ = writeln!(
                logger,
                "warning: DeleteAppContainerProfile('{}') failed: {} (may already be deleted)",
                identity,
                e
            );
        }
    }

    // Step 3: Delete the registry tracking entry.
    delete_tracking_key(sid_string, logger);
}
```

src/wxc_common/src/sandbox_tracking.rs
```diff
@@ -0,0 +1,390 @@
...
+                key_path, e
+            );
+        }
+    }
+}
+
+// --- Time helper ---
+
+/// Get the current time as a FILETIME u64 value.
+/// Uses `std::time::SystemTime` to avoid needing `Win32_System_SystemInformation` feature.
```
bbonaby: question (non-blocking): any reason we want to avoid the `Win32_System_SystemInformation` feature? I was thinking we could do something like the following in a common module:
```rust
use windows::Win32::Foundation::FILETIME;
use windows::Win32::System::SystemInformation::GetSystemTimeAsFileTime;

fn get_current_filetime() -> u64 {
    let mut file_time= FILETIME::default();

    unsafe {
        // SAFETY:
        // - GetSystemTimeAsFileTime requires a valid pointer to writable memory.
        GetSystemTimeAsFileTime(&mut file_time);
    }

    ((file_time.dwHighDateTime as u64) << 32) | (file_time.dwLowDateTime as u64)
}
```
or does that not work?

src/wxc_common/src/sandbox_tracking.rs
```diff
@@ -0,0 +1,390 @@
...
+    // SAFETY: Free the OS-allocated string buffer.
+    unsafe {
+        let _ = LocalFree(Some(HLOCAL(string_sid.0 as *mut std::ffi::c_void)));
+    }
+
+    Ok(result)
+}
+
+/// Information about a tracked sandbox entry, used for logging and cleanup.
+pub struct TrackingEntry {
```
bbonaby, line 83: question (non-blocking): so, the secondary purpose, maybe I'd call it "passive cleanup" is to remove AppContainer profiles that were created by us with `destroy_on_exit = true`, but where wxc-exec exited unexpectedly (crash or forced termination)? These would then be cleaned up on the next invocation of wxc-exec, using the persisted registry tracking entries.

src/wxc_common/src/sandbox_tracking.rs
jsidewhite: my copilot says
My take: Currently we only run one sandbox per wxc-exec.exe process, so the single Option<ActiveCleanup> slot is sufficient. A HashMap would
add complexity for a multi-sandbox future that doesn't exist yet. I'd dismiss this as YAGNI unless you want to future-proof.

src/wxc_common/src/base_container_runner.rs
jsidewhite: The answer: deferred cleanup isn't handled automatically yet — the CleanupDeferred marker is
written to the registry so future tooling (or a manual cleanup pass) can find and clean up orphaned entries. It's a breadcrumb, not an
auto-retry.

src/wxc_common/src/base_container_runner.rs
jsidewhite: This is about derive_sid_string failing and returning String::new(). Currently, when the SID is empty we skip tracking but still proceed with
the launch. With the comment 3 fix (cleanup gated on destroy_on_exit not tracking_written), cleanup will still attempt 
DeleteAppContainerProfile by identity name, which works fine without a SID string — just the registry tracking entry removal becomes a no-op.

src/wxc_common/src/sandbox_tracking.rs
jsidewhite: love it

huzaifa-d (approved):

src/wxc_common/src/base_container_runner.rs
jsidewhite: No - plan is not to plumb it through.  CPIS is primitives.  Just because they exist doesn't mean we should give all of them to agentic things.  If copilot requires it, then we revisit.

src/wxc_common/src/sandbox_tracking.rs
jsidewhite: taking it!

src/wxc_common/src/sandbox_tracking.rs
jsidewhite: The answer is yes, that's the intent -- the tracking entries enable future passive/deferred
cleanup tooling.

(I'd say it's same as primary purpose, on a different timeline)

src/wxc_common/src/base_container_runner.rs
bbonaby: Just seeing this but I believe so, for what it's worth you might be able to use the scopeguard crate: https://crates.io/crates/scopeguard like you would with `wil::scope_exit`

bbonaby (approved):

PR #322, opened 2026-05-15
feat: nested ptys, forward sigwinch, keychain access

caarlos0:
## 📖 Description

Three independent fixes that together unblock interactive agent workflows under the macOS seatbelt backend. The baseline profile breaks anything that needs to spawn a real shell (tests, `git`, `gh`, REPLs) or talk to the system Keychain, and the pty bridge corrupts TUIs whenever the host wraps `mxc-exec-mac` in its own pty.

**`feat(seatbelt): add nestedPty + keychainAccess experimental options`**

Two new opt-in/opt-out fields under `experimental.seatbelt`:

- `nestedPty` (default `true`): allow the inner workload to call `posix_openpt`. Emits `(allow pseudo-tty)`, `(allow iokit-open)`, and read/write/ioctl on `/dev/ptmx`. Default-on because almost any agent shell tool needs it.
- `keychainAccess` (default `false`): allow `Security.framework`/`keytar` to reach the system keychain stack. Emits Mach lookups for `SecurityServer`, `securityd`, `trustd`, `cfprefsd.daemon`, `xpcd`, and the `com.apple.lsd.*` family (via `global-name-regex`); read access to `/Library/Keychains` + `/private/var/db/mds`; read+write on `$HOME/Library/Keychains` + `/private/var/folders`; and `(allow iokit-open)`.

Both blocks are emitted before `write_ui_rules` so the HID `iokit-open` deny still wins under Seatbelt's last-match-wins semantics. `nestedPty` is suppressed when `guiAccess` already emits a strict superset.

**`fix(pty): make nested ptys work end-to-end under macOS`**

Four gaps in `mxc_pty::run_with_pty` that broke any nested-pty workload under a host-allocated outer pty:

- Outer pty slave was left in cooked mode → input bytes echoed back, control chars rendered as `^X`. Added a RAII `RawSlaveGuard` that puts the outer slave into raw mode and restores termios on drop. No-op when fd 0 isn't a tty.
- Inner pty was created at the kernel default 0×0. Now inherits the outer pty's `TIOCGWINSZ`.
- SIGWINCH was never propagated. Added a self-pipe SIGWINCH handler + a forwarder thread that re-reads the outer winsize and pushes it to the inner pty master via `TIOCSWINSZ`. Self-pipe rather than `sigwait` so we don't depend on which thread the runtime lets the signal land on.
- Inner child inherited the parent's sigmask (signal_cleanup blocks signals via `sigwait`). `pre_exec` now unconditionally adds `SIGWINCH` to the unblock set — `execve` resets the handler but preserves the inherited mask.

## 🔗 References

https://github.com/github/copilot-agent-runtime/issues/164

## 🔍 Validation

- `cargo fmt --all -- --check` clean.
- `cargo clippy -p mxc_pty -p seatbelt_common -p wxc_common -p mxc_darwin --all-targets -- -D warnings` clean.
- `cargo test -p mxc_pty -p seatbelt_common -p wxc_common` — 4 + 33 + 190 = 227 tests pass. New unit tests cover: `nestedPty` defaults / on / off / interaction with `guiAccess` / HID deny ordering; `keychainAccess` mach + FS + iokit + HID deny ordering; parser pass-through of both fields.
- Manual: ran Copilot CLI inside the seatbelt sandbox on macOS — `gh`, `git`, `node` REPLs, and resize-on-the-fly all work; `keytar` reads/writes Keychain entries with `keychainAccess: true`.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [x] Feature
- [ ] Task


caarlos0:
Addressed Copilot reviewer feedback:

- **SIGWINCH self-pipe blocking write** (`mxc_pty/src/lib.rs:342`): fixed in bf9baa0 — set `O_NONBLOCK` on the write end before installing the handler, so the handler can't deadlock if the pipe fills. The existing "drop on full pipe" comment now matches actual behavior.

- **Per-invocation global SIGWINCH state** (`mxc_pty/src/lib.rs:352`): not addressing in this PR. The function lives in a single-shot binary (`mxc-exec-mac` and `lxc-exec` both call `run_with_pty` exactly once before exiting); there is no concrete caller doing repeated or concurrent runs. Fixing it correctly means either a once-init module-level handler or per-session registration, both of which add real complexity for no current user. Happy to revisit when there's a concrete caller that needs it.

Also fixed the x64 WXC CI failure in 6eb69e7 — the keychain tests need `$HOME` which isn't set on Windows CI; gated them to `cfg(target_os = "macos")` since the feature is macOS-only anyway.

src/mxc_pty/src/lib.rs
```diff
@@ -171,6 +213,14 @@ pub fn run_with_pty(mut command: Command, options: PtyOptions) -> Result<PtyOutc
         .map_err(|e| format!("dup master: {}", e))?;
     let mut master_reader = master;
 
+    // Resize forwarder: when the host's terminal resizes, the kernel
+    // delivers SIGWINCH to us (because our fd 0 is the outer pty
+    // slave). Read the new size off fd 0 and push it to the inner pty
+    // master via TIOCSWINSZ — that delivers SIGWINCH to the inner
+    // child, so TUIs reflow correctly.
+    let master_fd_for_winch = master_writer.as_raw_fd();
+    let _winch_thread = spawn_sigwinch_forwarder(master_fd_for_winch);
```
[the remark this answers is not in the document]
richiemsft: was this fixed?

src/seatbelt_common/src/profile_builder.rs
```diff
@@ -303,6 +315,112 @@ fn write_ui_rules(out: &mut String, request: &CodexRequest) {
...
+    out.push_str("(allow pseudo-tty)\n");
+    // /dev/ptmx is the master multiplexer; opening it is what posix_openpt
+    // does under the hood. The TTY_ALLOW baseline already grants access to
+    // /dev/ttysNNN (the slave side).
+    out.push_str("(allow file-read* file-write* file-ioctl\n");
+    out.push_str("    (literal \"/dev/ptmx\"))\n");
+    // posix_openpt → grantpt → ioctls bottom out in IOKit user-clients
+    // (IOSerialBSDClient and friends). Without iokit-open the kernel
+    // returns EPERM and the master fd is unusable.
+    out.push_str("(allow iokit-open)\n");
```
richiemsft: isn't this too broad? Is it possible to narrow it down to only what's needed?

src/seatbelt_common/src/profile_builder.rs
```diff
@@ -303,6 +315,112 @@ fn write_ui_rules(out: &mut String, request: &CodexRequest) {
...
+            .to_string()
+    })?;
+    let user_keychains = format!("{home}/Library/Keychains");
+    out.push_str(";; --- keychainAccess: user keychain DB + XPC/folder caches (read+write) ---\n");
+    out.push_str("(allow file-read* file-write*\n");
+    let _ = writeln!(out, "    (subpath {})", quote_scheme(&user_keychains));
+    out.push_str("    (subpath \"/private/var/folders\"))\n");
+
+    out.push_str(";; --- keychainAccess: iokit-open for crypto accelerators ---\n");
+    out.push_str("(allow iokit-open)\n");
```
richiemsft: same comment about iokit-open in here.

src/mxc_pty/src/lib.rs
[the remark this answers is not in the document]
caarlos0: Yes — fixed properly in [`07a3db6`](https://github.com/microsoft/mxc/pull/322/commits/07a3db6). The forwarder was holding a borrowed `RawFd` from `master_writer`. The input thread later moves and drops `master_writer` mid-session when stdin closes, so any later `TIOCSWINSZ` could ioctl on a closed (and possibly reused) descriptor. Now `spawn_sigwinch_forwarder` takes a `try_clone` of the master `File` and `mem::forget`s it inside, so the resize fd has the same lifetime as the signal handler that targets it.

src/seatbelt_common/src/profile_builder.rs
caarlos0: Yes — verified empirically and dropped entirely in [`cb21f36`](https://github.com/microsoft/mxc/pull/322/commits/cb21f36). I tested on macOS 15 (Apple Silicon) under `sandbox-exec` with `(allow pseudo-tty)` + `/dev/ptmx` but **no** `iokit-open`: `posix_openpt` → `grantpt` → `unlockpt` → `ptsname` → slave open all succeed. Then end-to-end through `mxc-exec-mac` itself with the new profile builder — `posix_openpt` returns `OK slave=/dev/ttys014, opened slave fd=7`. So the rule is removed from `write_nested_pty_rules` along with the now-stale ordering test and `build_profile` comment.

src/seatbelt_common/src/profile_builder.rs
caarlos0: Same fix in [`cb21f36`](https://github.com/microsoft/mxc/pull/322/commits/cb21f36). Verified `SecItemAdd` + `SecItemCopyMatching` on a generic-password item under `sandbox-exec` with the keychain Mach surface (securityd, trustd, ocspd, cfprefsd, xpcd, lsd.*) and the FS surface (`~/Library/Keychains`, `/private/var/folders`, `/private/var/db/mds`) but **no** `iokit-open` — both return status `0`. Dropped the rule. Also dropped the redundant `/Library/Keychains` and `/System/Library/Keychains` reads since the baseline `(subpath "/Library")` and `(subpath "/System")` already cover them, and updated the docs to match.

richiemsft (approved):

PR #325, opened 2026-05-15
User/saulg/fuzzer

richiemsft:
## 📖 Description

Adds continuous fuzzing infrastructure for MXC's config-parsing surface using cargo-fuzz (libFuzzer) + OneFuzz.

**What's included:**
- 3 fuzz targets: `config_parser`, `base64_decode`, `validator`
- Seed corpus derived from existing test configs and examples
- Azure Pipeline (`Fuzz.Build.yml`) with daily cron schedule on `main`
- OneFuzzConfig.json with ADO bug routing to Tessera area path
- Documentation in `docs/fuzzing.md` and `src/fuzz/README.md`

**Pipeline behavior:**
- `trigger: none` / `pr: none` — only runs on daily cron (00:00 UTC)
- Builds with AddressSanitizer on Windows x64
- Stages drop directory and submits to OneFuzz

## 🔍 Validation

- All 3 fuzz targets run locally on macOS (nightly toolchain) without crashes
- config_parser: ~959K runs in 11s ✅
- base64_decode: ~1.3M runs in 11s ✅
- validator: ~893K runs in 11s ✅
- YAML validated as syntactically correct
- OneFuzzConfig paths aligned with pipeline staging layout

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type
- [x] New feature


.azure-pipelines/Fuzz.Build.yml
```diff
@@ -0,0 +1,49 @@
...
+# Licensed under the MIT License.
+#
+# Daily fuzzing pipeline for MXC.
+#
+# Builds the mxc_fuzz crate with AddressSanitizer and submits the resulting
+# drop directory to OneFuzz. Bugs found are filed against the SDL fuzzing
+# work item 62294501 via the routing fields in src/fuzz/OneFuzzConfig.json.
+#
+# Schedule: daily at 00:00 UTC on `main`. PRs do not trigger this pipeline,
+# but pushes to `user/saulg/Fuzzer` are allowed during initial bring-up so
```
bbonaby: issue: I think we can remove this line

.azure-pipelines/Fuzz.Build.yml
```diff
@@ -0,0 +1,49 @@
...
+# we can validate the build + drop staging before merging.
+
+pr: none
+trigger: none
+
+schedules:
+  - cron: "0 0 * * *"
+    displayName: Daily fuzzing submission
```
bbonaby, line 17: suggestion: you can actually override this in the ADO UI actually. Go to pipeline > edit > 3 vertical dots > triggers. That should navigate to the UI where you can set a CI schedule versus having to update the yaml itself

src/fuzz/fuzz_targets/validator.rs
```diff
@@ -0,0 +1,26 @@
...
+// explore the validator's behaviour on real, well-shaped requests rather
+// than only garbage that the parser rejects.
+
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+use wxc_common::config_parser::load_mxc_request;
+use wxc_common::logger::{Logger, Mode};
+use wxc_common::state_aware_request::MxcRequest;
+use wxc_common::validator::validate_common;
```
bbonaby, line 24: Note: each runner should also have their own validation function via implementing the `validate_runner` trait function. Should we be attempting to call these? Currently if you pass in --dry-run to the executables/in the request each runner should actually go through parsing the requesting + validation but return without actually creating its sandbox. That might be useful here

src/fuzz/OneFuzzConfig.json
```diff
@@ -0,0 +1,117 @@
...
+      "JobNotificationEmail": "anisk@microsoft.com",
+      "FuzzerProperties": {
+        "Fuzzer": {
+          "$type": "libfuzzer"
+        },
+        "FuzzerExe": "validator\\validator.exe",
+        "JobDependencies": [
+          "validator\\clang_rt.asan_dynamic-x86_64.dll"
+        ],
+        "Seeds": [
```
bbonaby, line 89: question: I know above we have a bunch of copied config files to use as the corpus. What I'm wondering is whether we can just point these seed paths to `<repo-root>/test_configs`

.azure-pipelines/templates/Fuzz.Build.Job.yml
```diff
@@ -0,0 +1,105 @@
...
+#
+# Build the mxc_fuzz crate (libFuzzer + AddressSanitizer) and submit the
+# resulting drop directory to OneFuzz via `onefuzz-task@0`.
+#
+# Nightly Rust toolchain: cargo-fuzz requires `-Z sanitizer=address`, which
+# only nightly accepts. The Mxc-Azure-Feed (`ms-prod-*`) toolchains are
+# stable-only today, so we install nightly via `rustup-init` here.
+# TODO(SFI): switch to a Microsoft-published nightly via msrustup once one
+# is available in Mxc-Azure-Feed; until then keep cargo registry pointed at
+# Mxc-Azure-Feed so all *crate* sources still come from the internal feed.
```
bbonaby, line 12: question: Is this something I need to add? I can check and get back to you

richiemsft: unsure, haven't been able to test the azure side since in order to generate the pipeline the code has to be check in. I couldn't reference a non checked in file.

.azure-pipelines/Fuzz.Build.yml
richiemsft: leaving it here, I dont expect we move the auto-run schedule that often.

src/fuzz/README.md
```diff
@@ -0,0 +1,55 @@
+# mxc_fuzz
```
bbonaby, line 1: suggestion: Might be worth it for us to have a table at the bottom with what each base64_decode  relates to since they won't be human readable

.azure-pipelines/templates/Fuzz.Build.Job.yml
bbonaby: all good, lets check it in so you can try it out. There is a way to do it without having to check it in though if I recall but it's ok lets get it in.

bbonaby (approved):

PR #331, opened 2026-05-16
[IsolationSession] Filter system-folder paths from filesystem-policy requests

adpa-ms:
## 📖 Description

`IsoSessionOps::ShareFolderBatchAsync` applies ACEs with subtree inheritance. Granting these on broad system folders is expensive (the OS service must walk and apply the inherited ACE to every file and directory in the subtree, which can be prohibitively slow on deep system trees like `C:\Windows`) and unsafe (the agent gains the granted right across every file in the subtree).

This PR adds a client-side filter inside `IsolationSessionManager::share_folders` that silently drops a fixed set of paths from `readwritePaths` and `readonlyPaths` before forwarding to the OS API: drive roots, `SystemRoot`, parent of `USERPROFILE`, `ProgramFiles`, `ProgramFiles(x86)`, and `ProgramData`. Caller-visible behavior is unchanged — no error, no metadata field, no wire-format change.

The filter region in `isolation_session_runner.rs` is bracketed by `BEGIN:` / `END:` markers, with a matching one-line marker at the call site. All three carry the tracking issue number; removal is a mechanical delete of the bracketed region plus the call-site line.

### Related Issues

- #330

## 🔍 Validation

Local 7-gate × 2-mode (build / test / clippy / fmt with and without `--features isolation_session`); release build of `wxc-exec` with `isolation_session`; ARM64 Windows cross-compile (`--target aarch64-pc-windows-msvc`) clean in both modes.

Test counts: 451 default-feature, 532 with `isolation_session` (+19 new filter unit tests covering normalizer rules, filter-set construction with synthetic env-var fixtures, and bypass-attempt spellings including the `\\?\` long-path verbatim-disk form).

VM integration on Windows `26619.260514-1700`: state-aware 65/65 (+6 new filter assertions verifying protected folders receive no agent ACE and a positive-control non-protected folder does); one-shot 13/13 (+1 new `isolation_session_filtered.json` fixture).

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/331)

### DIFF
--- a/src/wxc_common/src/isolation_session_runner.rs
+++ b/src/wxc_common/src/isolation_session_runner.rs
@@ -499,6 +499,235 @@ pub struct IsolationSessionManager {
     ops: IsoSessionOps,
 }
 
+// =============================================================================
+// BEGIN: filesystem-policy path filter (emergency mitigation, MXC issue #330)
+//
+// Silently drops a fixed set of system-folder paths from filesystem-policy
+// requests before they reach `IsoSessionOps::ShareFolderBatchAsync`. The OS
+// API applies grants with subtree inheritance, so granting these top-level
+// folders propagates an agent SID's ACE through their entire subtrees —
+// catastrophic for drive roots, the Windows directory, the Users root,
+// ProgramFiles, and ProgramData.
+//
+// The proper fix belongs in the OS API. When that lands, delete this entire
+// region and the single call site in `IsolationSessionManager::share_folders`.
+//
+// Known not covered (caller would have to actively pursue these spellings to
+// evade the text match): 8.3 short names (e.g. `PROGRA~1`), symlinks /
+// junctions (no `canonicalize` disk access), UNC paths, `CommonProgramFiles`
+// / `CommonProgramFiles(x86)`. The `\\?\` long-path prefix on a disk path
+// (`\\?\C:\foo`) IS handled — it normalizes to the same canonical form as
+// `C:\foo` so it cannot be used as a textual bypass.
+// =============================================================================
+
+/// Returns the lazily-computed, cached set of canonical paths the filter
+/// rejects. Contents:
+///
+/// - 26 drive roots (`A:\` through `Z:\`) — static, no `GetLogicalDrives`
+///   call so future-mounted drives are also caught.
+/// - The result of normalizing each available env-var-derived path:
+///   `SystemRoot`, parent of `USERPROFILE`, `ProgramFiles`,
+///   `ProgramFiles(x86)`, `ProgramData`, plus the aliases `windir`,
+///   `ProgramW6432`, `AllUsersProfile`, `SYSTEMDRIVE` (each of which
+///   collapses to one of the preceding via normalization-and-dedup).
+///
+/// Missing or empty env vars are silently skipped — pathological host
+/// configuration is not the runner's failure mode.
+fn protected_paths_set() -> &'static std::collections::HashSet<String> {
+    static SET: std::sync::OnceLock<std::collections::HashSet<String>> = std::sync::OnceLock::new();
+    SET.get_or_init(|| build_protected_paths_set(|var| std::env::var(var).ok()))
+}
+
+/// Construction logic for the protected-paths set, parameterised on an
+/// env-var lookup so tests can inject synthetic environments without
+/// touching the process-wide `OnceLock` cache.
+fn build_protected_paths_set(
+    env: impl Fn(&str) -> Option<String>,
+) -> std::collections::HashSet<String> {
+    let mut set = std::collections::HashSet::new();
+    // 26 drive roots, regardless of current mount state.
+    for letter in b'A'..=b'Z' {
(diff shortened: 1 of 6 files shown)


test_scripts/run_isolation_session_state_aware_tests.ps1
```diff
@@ -807,6 +807,123 @@ try {
...
+try {
+    $filterProvisionedOk = Run-StateAwareTest "filter: provision (protected + non-protected paths)" {
+        $r = Invoke-StateAware -ConfigFile 'isolation_session_state_aware_provision_with_filter.json' -Experimental
+        $envObj = Parse-Envelope -Stdout $r.Stdout
+        $arm = Envelope-Arm $envObj
+        if ($arm -ne 'result') {
+            Write-Host "  Envelope arm: $arm" -ForegroundColor Red
+            Write-Host "  Stdout: $($r.Stdout)" -ForegroundColor Gray
+            Write-Host "  Stderr: $($r.Stderr)" -ForegroundColor Gray
+            Assert-True $false "filter provision returned a result envelope (filter is silent -- no error expected)"
```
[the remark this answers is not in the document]
adpa-ms: Thanks for the suggestion. Leaving the phrasing as-is to stay consistent with the existing pattern used across this file: `Assert-True $false "<expected state>"` appears in 12 other places in `run_isolation_session_state_aware_tests.ps1` (lines 352, 389, 595, 632, 697, 715, 778, 791, 946, 967, 994, 1010), where the message describes the asserted state and the `FAIL:` prefix from `Assert-True` indicates it didn't hold. Reworking the convention itself would be a worthwhile cleanup but cross-cutting across every state-aware test site — better suited to a follow-up PR than a one-off change in this site.

src/wxc_common/src/isolation_session_runner.rs
```diff
@@ -499,6 +499,235 @@ pub struct IsolationSessionManager {
...
+/// - Forward slashes → backslashes.
+/// - Trailing slash discipline: drive `C:` → `c:\`; non-root paths drop the
+///   trailing `\` via component re-assembly.
+/// - `.` / `..` collapse, clamped at root (so `C:\..` stays `c:\`).
+/// - Per-component trailing whitespace and dots trim (Win32 silently strips
+///   these at the file-open boundary, making them a real bypass shape).
+/// - Case-insensitive ASCII compare via lowercase-on-output.
+/// - `\\?\` long-path prefix on a disk path (`\\?\C:\foo`): the `\\?\` is
+///   dropped and the rest normalizes the same as `C:\foo`.
+fn normalize_protected_path(input: &str) -> Option<String> {
```
jeffstall, line 601: Does Rust have the equivalent of this in C++?

```
#include <filesystem>
#include <iostream>

int main() {
    std::filesystem::path p1 = L"\\path\\to\\dir1";
    std::filesystem::path p2 = L"\\path\\to\\dir2";
    std::cout << std::filesystem::equivalent(p1, p2) << std::endl;
}
```

[https://learn.microsoft.com/en-us/cpp/standard-library/filesystem-functions?view=msvc-170](https://learn.microsoft.com/en-us/cpp/standard-library/filesystem-functions?view=msvc-170)

Can you ask Copilot?

jeffstall (approved):
:shipit:

src/wxc_common/src/isolation_session_runner.rs
adpa-ms: Rust's closest analog is `std::fs::canonicalize` + string compare. I considered it but stayed textual because the filter isn't really a path-equivalence check — it's a string-set membership test against the specific shapes whose forwarding to `ShareFolderBatchAsync` triggers expensive recursive ACE walks. `canonicalize` answers a different question.

Two concrete obstacles to migrating:

1. **`canonicalize` requires the path to exist.** A caller passing a directory to be created in-session (e.g., `C:\NewWorkDir`) hits `Err`. The remaining choices are all bad: *drop* silently strips caller intent, *pass-through* can re-introduce the shapes we're filtering (Rust-std's strict canonicalize can fail on inputs the OS API would still resolve), and *new error code* is wire churn for a stopgap. The textual approach has none of these failure modes.

2. **Drive roots.** Always-26 catches every letter the caller writes today or after a future mount (USB, BitLocker, `net use`, mapped network drives). Live detection via `GetLogicalDrives`/`sysinfo` saves ~150 bytes at the cost of a race window between init and post-init mounts.

The code does use Rust's standardized path primitives where they fit: `Path::components`, `Prefix::{Disk, VerbatimDisk, …}`, `HashSet`, `OnceLock`. The textual normalization on top is the right primitive for set membership.

The "Known not covered" cases (8.3 short names, symlinks pointing at protected dirs, UNC) are out of scope: the caller is first-party, not constructing evasion shapes. The `BEGIN:`/`END:` region and 19 hermetic unit tests are structured to extend incrementally if needed.

PR #347, opened 2026-05-19
Address PR #294 review feedback

MGudgin:
Move `delete_app_container_profile` from `src/wxc/src/main.rs` into `src/wxc_common/src/appcontainer_runner.rs`, next to `AppContainerScriptRunner` (which owns the create/setup path). Both ends of the AppContainer profile lifecycle now live in the same module.

While moving the function, replace the open-coded
`FileSystemBfsManager::new(...) + remove_configuration(...)` block with `FileSystemBfsManager::clear_policy(name, logger)`. The original code created a fresh manager and called `remove_configuration` without first setting `configured = true`, so the early-return in `remove_configuration` made the BFS-clear step a silent no-op for `wxc-exec --delete`. `clear_policy` is the helper purpose-built for externally-created sandboxes and sets `configured = true` itself, so the cleanup actually runs now.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/347)

### DIFF
--- a/src/wxc/src/main.rs
+++ b/src/wxc/src/main.rs
@@ -7,12 +7,10 @@ use std::process;
 use std::time::Instant;
 
 use clap::Parser;
-use windows::Win32::Security::Isolation::DeleteAppContainerProfile;
-use wxc_common::appcontainer_runner::AppContainerScriptRunner;
+use wxc_common::appcontainer_runner::{delete_app_container_profile, AppContainerScriptRunner};
 use wxc_common::base_container_runner::BaseContainerRunner;
 use wxc_common::config_parser::{is_base_container_version, load_mxc_request, ParseError};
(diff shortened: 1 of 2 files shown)


jsidewhite:
>         let success = delete_app_container_profile(name, &mut logger);

discussion: ideally this delete block would act like a callback probably

`active_backend::delete(...)`

(but given that we can't do call cleanup for AC today anyway, this comment is merely discussion - do not implement.)

---
Refers to: src/wxc/src/main.rs:232 in 9658f45. [](commit_id = 9658f45a24bf3b7b6cba38b3297d95b8add11280, deletion_comment = False)

jsidewhite (approved):
:shipit:

PR #353, opened 2026-05-20
Surface launch diagnostics to SDK callers (e.g. copilot.exe)

jsidewhite:

-----

## Summary

Add post-failure launch diagnostics that detect known environment issues after `CreateProcessW` or `Experimental_CreateProcessInSandbox` fails (or the child exits non-zero), and surface actionable remediation to callers of both wxc-exec.exe and the SDK.

**Detected conditions:**
- **packaged_app** -- MSIX-packaged executables (e.g. Store-installed pwsh.exe) cannot run inside sandboxed containers.
- **missing_filesystem_access** -- pwsh.exe < 7.7 requires read-only access to the root drive (`C:\`) which the sandbox policy may not grant.

**Key changes:**
- New shared Rust module `launch_diagnostics.rs` with detection heuristics, PATH resolution, and unit tests.
- Both AppContainer and BaseContainer runners enrich error responses and log diagnostics with \Error:\ prefix for red highlighting in the diagnostic console.
- wxc-exec emits a structured JSON error envelope on stderr for one-shot flows.
- SDK parses the JSON envelope and rejects with a typed `MxcError`.

**Example messages:**
-	Needs UI
 
image

-	Need C:\ for powershell
 
image

-	Packaged app
 
image

-	Need velocity (Gudge had put in a variant of this in – though it is not appearing in copilot.exe for some reason)

image

..and they will also show up with contextual messages in the mxc-diagnostic-console.
 
image

Closes #348
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/353)


docs/launch-diagnostics.md
```diff
@@ -0,0 +1,120 @@
+# Launch Diagnostics: Surfacing Errors to Callers (#348)
```
bbonaby: suggestion: can we move this under a new folder called `diagnostics` under `docs`? Also can we have an AI agent add a collapable section in the `sdk/README.md` file about error handling based on this doc?  This would appear in the README file in https://www.npmjs.com/package/@microsoft/mxc-sdk and would help future users.

docs/launch-diagnostics.md
```diff
@@ -0,0 +1,120 @@
...
+/// Attempts to diagnose *why* a process launch failed.
+/// Returns `None` if no known condition is detected (the original error passes through unchanged).
+pub fn diagnose_launch_failure(
+    exe_path: &Path,
+    readonly_paths: &[String],
+    exit_code: Option<u32>,
+) -> Option<LaunchDiagnostic> { ... }
+```
+
+**Checks (in order):**
```
bbonaby: note:  We probably need a follow up for Linux + MacOS I assume. Might need to add a GitHub issue for that. That said, not sure this would work for ones like wslc, nanix or hyperlight.

docs/launch-diagnostics.md
```diff
@@ -0,0 +1,120 @@
...
+Both runners already return `ScriptResponse::error(...)` for failures, so this enriches existing error paths without changing control flow.
+
+### Phase 2: Rust -- Structured error envelope for one-shot flows
+
+**File:** `src/wxc/src/main.rs`
+
+When a `ScriptResponse` contains a diagnostic, emit a JSON error envelope on **stderr**:
+
+```json
+{"error": {"code": "backend_error", "message": "...", "details": {"kind": "packaged_app", "remediation": "..."}}}
```
bbonaby: thought: We need to watch out for code paths that just print an error and then do exit(-1), if we're now formalizing how we handle errors.

docs/launch-diagnostics.md
```diff
@@ -0,0 +1,120 @@
...
+---
+
+## Detection Heuristics
+
+### Packaged app detection
+
+```rust
+fn is_packaged_app(exe_path: &Path) -> bool {
+    let normalized = exe_path.to_string_lossy().to_lowercase();
+    normalized.contains("\\windowsapps\\")
```
bbonaby: question: For this snippet and below can't remember but do we allow forward slashes in the path? if so we probably need to detect which one they're using and then do the check right?

docs/launch-diagnostics.md
```diff
@@ -0,0 +1,120 @@
...
+```rust
+fn missing_root_readonly(exe_path: &Path, readonly_paths: &[String]) -> bool {
+    let filename = exe_path.file_name().unwrap_or_default().to_string_lossy().to_lowercase();
+    if filename != "pwsh.exe" { return false; }
+    let root = format!("{}\\", &exe_path.to_string_lossy()[..2]);
+    !readonly_paths.iter().any(|p| p.eq_ignore_ascii_case(&root) || p == "\\")
+}
+```
+
+Version detection is deferred -- the simpler heuristic is path + filename based. Note from @asklar: this does not apply to `powershell.exe` (inbox Windows PowerShell 5.x), only to `pwsh.exe` < 7.7.
```
bbonaby: issue: we probably can remove the Note: part

docs/launch-diagnostics.md
```diff
@@ -0,0 +1,120 @@
...
+
+---
+
+## Open Questions
+
+1. **New error code vs `backend_error` + details?** Adding a first-class code (e.g. `launch_diagnostic`) requires both Rust and SDK changes to the closed union. Using `backend_error` + `details.kind` is additive-only.
+2. **PTY path reliability** -- In PTY mode stdout/stderr are merged. Should the envelope use a unique prefix (e.g., `MXC_ERROR:`) for reliable parsing?
```
bbonaby: thought: I think errors being structured like this `{"error": {"code": "backend_error", "message": "...", "details": {"kind": "packaged_app", "remediation": "..."}}}` probably inherently provides this.

sdk/src/sandbox.ts
```diff
@@ -601,6 +602,15 @@ export function spawnSandboxAsync(
       ptyProcess.onExit((event: { exitCode: number; signal?: number }) => {
         // Note: wxc-exec doesn't separate stdout/stderr when using PTY
         // All output is combined
+        //
+        // Check for structured error envelopes from wxc-exec on failure.
+        if (event.exitCode !== 0) {
```
bbonaby, line 703: issue: don't we also need this in these places?

1. https://github.com/microsoft/mxc/blob/f1b455b31e4bd9126ebfd7768fdf2456c6f21dfc/sdk/src/sandbox.ts#L416
2. https://github.com/microsoft/mxc/blob/f1b455b31e4bd9126ebfd7768fdf2456c6f21dfc/sdk/src/sandbox.ts#L526

sdk/src/sandbox.ts
```diff
@@ -612,3 +622,27 @@ export function spawnSandboxAsync(
     }
   });
 }
+
+/**
+ * Scans a multi-line string for a JSON error envelope emitted by wxc-exec
+ * on stderr. Returns the first matching envelope, or null if none found.
+ * The envelope format is: `{"error": {"code": "...", "message": "...", ...}}`
+ */
+function tryParseErrorEnvelopeFromLines(output: string): MxcError | null {
```
bbonaby, line 727: suggestion: can this be moved out of `sandbox.tx` and into the `helper.ts` function? this seems like it can be reused by the state-aware.ts file as well.

src/wxc_common/src/base_container_runner.rs
```diff
@@ -598,11 +599,31 @@ impl ScriptRunner for BaseContainerRunner {
...
+            }
+            return ScriptResponse::error(&msg);
         }
 
+        // Helper: resolve exe path from the command line for post-exit diagnostics.
+        let resolved_exe = {
+            let bare =
+                std::path::Path::new(request.script_code.split_whitespace().next().unwrap_or(""));
+            crate::launch_diagnostics::resolve_exe_on_path(bare)
+        };
```
bbonaby: suggestion: seems like this can move line `573` since it's used inside the `success == 0` and ` success != 0` lines

src/wxc_common/src/base_container_runner.rs
```diff
@@ -654,11 +675,33 @@ impl ScriptRunner for BaseContainerRunner {
...
+            ) {
+                let _ = writeln!(
+                    logger,
+                    "Error: Launch diagnostic [{}]: {}",
+                    diag.kind, diag.message
+                );
+                let _ = writeln!(logger, "Error: Remediation: {}", diag.remediation);
+                let diagnostic_text = format_diagnostic(&diag);
+                error_message = diagnostic_text.trim().to_string();
+                standard_err = diagnostic_text;
```
bbonaby: suggestion: also seems like these lines are good candidates to put in a common helper function  since they're also repeated for both the success and failure cases

src/wxc_common/src/launch_diagnostics.rs
```diff
@@ -0,0 +1,223 @@
...
+/// Extract the drive root (e.g. `C:\`) from an absolute path. Falls back to
+/// `C:\` when the path is relative or otherwise cannot be parsed.
+fn drive_root(exe_path: &Path) -> String {
+    let s = exe_path.to_string_lossy();
+    if s.len() >= 3 && s.as_bytes()[1] == b':' {
+        format!("{}\\", &s[..2])
+    } else {
+        "C:\\".to_string()
+    }
+}
```
bbonaby, line 397: note: see my suggestion about forward slashes too in the doc

docs/launch-diagnostics.md
jsidewhite: i deleted this file - it makes no sense for each minor feature to have an .md. i forgot to nuke it earlier.

docs/launch-diagnostics.md
jsidewhite: ScriptResponse is pre-existing, but end-to-end should be formalized yes.  it's very confusing.

docs/launch-diagnostics.md
jsidewhite: nuked file. sry!

docs/launch-diagnostics.md
jsidewhite: fxied

src/wxc_common/src/base_container_runner.rs
jsidewhite: refactored this mess to be smaller and better

sdk/src/sandbox.ts
jsidewhite: wontfix. because no time.

sdk/src/sandbox.ts
jsidewhite: i dont think you can because PTY

bbonaby: Hmm but there is a `ptyProcess` return in `spawnWithConfig` though. Users Can use `spawnSandbox` by itself rather than `spawnSandboxAsync` . Same thing with `spawnSandboxFromConfig` there is a `ptyProcess` option in there too.

EDIT: Oh wait, I might have misinterpreted what you said. I thought you meant, that we can't **because** they use PTY

bbonaby (approved):

PR #359, opened 2026-05-20
Fix: deniedPaths now overrides readonlyPaths/readwritePaths in BFS

SohamDas2021:
The AppContainer BFS backend (bfscfg.exe) is an allow-list system with no deny primitive. Previously, paths listed in both deniedPaths and readonlyPaths/readwritePaths would still be granted access via BFS since the denied list was never consulted.

Now the configure() method skips any readwrite or readonly path that is equal to or a child of a denied path, ensuring deny semantics are honored. Comparison is case-insensitive to match Windows filesystem behavior.

Also adds a test config for integration validation.

Fixes #145

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/359)

### DIFF
--- a/src/wxc_common/src/filesystem_bfs.rs
+++ b/src/wxc_common/src/filesystem_bfs.rs
@@ -57,15 +57,14 @@ impl FileSystemBfsManager {
                 ));
                 continue;
             }
-            let mut inherit = test_for_root_path(path);
+            let inherit = test_for_root_path(path);
             if inherit && has_denied_children(path, &policy.denied_paths) {
                 logger.log_line(&format!(
-                    "Warning: readwrite path {:?} has denied sub-paths; \
-                     disabling inheritance to limit scope. \
-                     BFS cannot deny individual sub-paths under a granted parent.",
+                    "Warning: readwrite path {:?} has denied sub-paths that \
+                     cannot be enforced. BFS has no deny primitive — sub-paths \
+                     under a granted parent with inheritance will remain accessible.",
                     path
                 ));
-                inherit = false;
             }
             if let Err(e) = self.add_bfs_path(path, inherit, logger) {
                 self.remove_configuration(logger);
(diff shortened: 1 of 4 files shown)


src/wxc_common/src/filesystem_bfs.rs
```diff
@@ -68,15 +49,24 @@ impl FileSystemBfsManager {
...
+            }
+            let mut inherit = test_for_root_path(path);
+            if inherit && has_denied_children(path, &policy.denied_paths) {
+                logger.log_line(&format!(
+                    "Warning: readwrite path {:?} has denied sub-paths; \
+                     disabling inheritance to limit scope. \
+                     BFS cannot deny individual sub-paths under a granted parent.",
+                    path
+                ));
+                inherit = false;
```
MGudgin: Doesn't this end up breaking access to everything under the r/w path? e.g. if C:\Users\Gudge was r/w but c:\Users\Gudge\Secret was deny, then disabling inherit would also mean that C:\Users\Gudge\NotSecret would also not be visible inside the container (because C:\Users\Gudge no longer has inherit)

src/wxc_common/src/filesystem_bfs.rs
```diff
@@ -85,7 +75,23 @@ impl FileSystemBfsManager {
...
+            }
+            let mut inherit = test_for_root_path(path);
+            if inherit && has_denied_children(path, &policy.denied_paths) {
+                logger.log_line(&format!(
+                    "Warning: readonly path {:?} has denied sub-paths; \
+                     disabling inheritance to limit scope. \
+                     BFS cannot deny individual sub-paths under a granted parent.",
+                    path
+                ));
+                inherit = false;
```
MGudgin: Same issue here, I suspect

test_configs/appcontainer_denied_override.json
```diff
@@ -0,0 +1,19 @@
+{
```
MGudgin, line 1: It seems we don't actually have a test that runs this. Can you add one (or maybe several, I'm guessing we may want to verify more cases than just the one in this file)?

MGudgin (changes requested):

src/wxc_common/src/filesystem_bfs.rs
SohamDas2021: thanks for the catch, fixing.

PR #360, opened 2026-05-20
[mxc_pty] Set FD_CLOEXEC on openpty fds to prevent macOS pty leak

caarlos0:
## 📖 Description

On macOS, `openpty(3)` returns the master and slave fds without `FD_CLOEXEC`. As a result, the master fd was inherited by the sandboxed child across `exec`: when mxc-exec-mac later died, the slave never returned EOF (the child itself was still holding a master ref), so the inner shell blocked forever on `read(0)`, got reparented to launchd, and pinned a pty pair until the host hit `kern.tty.ptmx_max` (default 511) — at which point all new sandbox spawns failed.

Fix: right after `openpty`, fcntl `FD_CLOEXEC` on both originals. Subsequent dups via `File::try_clone` already set CLOEXEC on macOS (`F_DUPFD_CLOEXEC`), so the originals are the only fds that need a fixup.

Tests: added `master_fd_does_not_leak_into_child_across_exec`, which runs a child under `run_with_pty` and uses `lsof -p $$` to assert no inherited fd resolves to `/dev/ptmx`. Verified the test fails without the fix (`LEAK fd=3u target=/dev/ptmx`). Existing pty tests are guarded with a shared mutex so concurrent forks don't race the openpty → CLOEXEC window; production code never invokes `run_with_pty` from multiple threads (each mxc-exec-mac is a one-shot CLI invocation), so the mutex faithfully mirrors real usage rather than masking a race.

## 🔗 References

https://github.com/github/copilot-agent-runtime/issues/164

## 🔍 Validation

Tests, manual test with and without copilot.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/360)

### DIFF
--- a/src/mxc_pty/src/lib.rs
+++ b/src/mxc_pty/src/lib.rs
@@ -34,7 +34,7 @@ pub struct PtyOptions {
     pub timeout: Option<Duration>,
 
     /// How long to wait for the inner process to print its first byte
-    /// before forwarding host stdin to the pty master. The delay matters
+    /// before forwarding host stdin to the pty primary. The delay matters
     /// because an interactive shell calls `tcsetattr` during readline
     /// init, which can flush bytes the parent buffered in the pty before
     /// the shell got there. Set to `Duration::ZERO` to forward stdin
@@ -83,8 +83,8 @@ pub enum PtyOutcome {
 /// Spawn `command` attached to a freshly-allocated pty pair and bridge
 /// it to the host's stdin/stdout/stderr.
 ///
-/// The slave end becomes the child's stdin/stdout/stderr; the master end
-/// stays in this process and is forwarded to/from the host fds on
+/// The secondary end becomes the child's stdin/stdout/stderr; the primary
+/// end stays in this process and is forwarded to/from the host fds on
 /// background threads. All of the child's output has therefore been
 /// streamed to the host stdio by the time this function returns;
 /// callers needing captured output should write it to a file in cwd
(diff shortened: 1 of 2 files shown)


src/mxc_pty/src/lib.rs
```diff
@@ -139,6 +139,25 @@ pub fn run_with_pty(mut command: Command, options: PtyOptions) -> Result<PtyOutc
     let pty_pair =
         openpty(inner_winsize.as_ref(), None).map_err(|e| format!("openpty failed: {}", e))?;
 
+    // Mark both pty fds close-on-exec. macOS' `openpty(3)` leaves them
+    // without `FD_CLOEXEC`, so without this fixup the master fd would
```
richiemsft: I have a policheck PR out that removes the terminology master/slave, can you do primary = master and secondary = slave where applicable? the upstream dependencies are fine since we cannot override them.

richiemsft: https://github.com/microsoft/mxc/pull/358

caarlos0: ✅

caarlos0:
done @richiemsft! thanks for the review. Would love to ship this fast-ish, since folks are already testing this in copilot-cli. Let me know if I can help in any way. Thanks 🙏🏻

richiemsft (approved):

PR #363, opened 2026-05-20
User/dalegg/iso create test package for integration testing workflow with IsolationSession

danlegg:
## 📖 Description
Adds support for building wxc-exec with the --features isolation_session Rust feature flag, which is required for IsolationSession integration tests. The result of the workflow is a test package which can be deployed to a Windows VM to test the latest IsolationSession builds.

The shared build.yml workflow intentionally omits this flag since it's not needed for general MXC builds. This PR updates package-isolation-tests.yml to build x64 wxc-exec locally with the feature enabled, while arm64 continues to use the shared build output.

Also adds CI scripts for packaging and running isolation session tests on Windows test VMs.

## 🔗 References
N/A

## 🔍 Validation
 - Workflow runs successfully on user/dalegg/iso-test-package branch — all build jobs pass
 - Built artifact deployed to test VM isoenv-test and executed end-to-end: 78/79 tests passed, 0 failed, 1 skipped (investigation TBD)
 - Verified build.yml is unchanged from main

## ✅ Checklist

It sounds like these may not be applicable to this PR, please advise if I'm mistaken.
- [ ] Linked to an issue

## 📋 Issue Type
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/363)


.azure-pipelines/templates/IsolationSession.TestBundle.Job.yml
```diff
@@ -0,0 +1,117 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
bbonaby, line 1: issue: The Rust.Build.Job.yml already builds all features in wxc-exec. If you want to use the wxc-exec.exe file later in the `1ES.Build.yml` then this `.yml` file needs to download the artifact that is produced from it. No need to attempt to rebuild it. Here is the artifacts page for this PR: https://microsoft.visualstudio.com/Dart/_build/results?buildId=147480309&view=artifacts&pathAsName=false&type=publishedArtifacts. In there you should eventually see something like this:

Image

You need to download the wxc-binaries-x86_64-pc-windows-msvc artifact in this yml file, which will have wxc-exec.exe. Then you can use it here if you want. That said, do note that the 1ES pools that our pipelines use, use win server 2022.

That said, I believe what we should be doing is distributing MXC into the OS via a vpack, and having a testpass that uses the latest vpack from MXC in uixp/internal branches. That vpack can contain supplemental files of course like powershell scripts and files but I don't believe MXC should be doing anything other than providing the vpack. 

 I can setup the OS vpack pipeline for you and you can create the OS test passes that use it. Does that work?

.github/workflows/package-isolation-tests.yml
```diff
@@ -0,0 +1,149 @@
+# -----------------------------------------------------------------------
```
bbonaby, line 1: issue: this yml file uses build.yml. Note we don't use build.yml anymore. It use to be used for PRs in GitHub and now we've moved to using ADO 1ES pipelines.

bbonaby (changes requested):

PR #385, opened 2026-05-22
BaseContainer: pass explicit env block to CreateProcessInSandbox

jsidewhite:
When process.env is specified in the config, encode it as a UTF-16 environment block and pass it to Experimental_CreateProcessInSandbox with CREATE_UNICODE_ENVIRONMENT. When empty, pass NULL (OS default).

SDK: inject env dict into config.process.env instead of node-pty process inheritance to prevent secret leakage into the sandbox.

#218 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/385)

### DIFF
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -438,6 +438,28 @@ export interface SandboxSpawnOptions {
 }
 
 /**
+ * Inject environment variables into the config's `process.env` field as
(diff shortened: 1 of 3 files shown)


bbonaby (approved):

jsidewhite:
Closes: #218

PR #386, opened 2026-05-22
AppcontainerRunner: pass explicit env block to CreateProcess

jsidewhite:
When process.env is specified in the config, encode it as a UTF-16 environment block and pass it to CreateProcess with CREATE_UNICODE_ENVIRONMENT. When empty, pass NULL (OS default).

#346
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/386)

### DIFF
--- a/src/wxc_common/src/appcontainer_runner.rs
+++ b/src/wxc_common/src/appcontainer_runner.rs
@@ -43,35 +43,15 @@ const HRESULT_ERROR_ALREADY_EXISTS: i32 = 0x8007_00B7u32 as i32;
 /// Proxy-related env var names to strip/override when building the child env block.
 const PROXY_VAR_NAMES: &[&str] = &["HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "ALL_PROXY"];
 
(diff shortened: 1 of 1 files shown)


src/wxc_common/src/appcontainer_runner.rs
```diff
@@ -387,20 +397,35 @@ impl AppContainerScriptRunner {
...
+            // No explicit env but proxy is active -- inject only proxy vars.
+            let proxy_url = addr.to_url();
+            let entries = vec![
+                ("HTTP_PROXY".to_string(), proxy_url.clone()),
+                ("HTTPS_PROXY".to_string(), proxy_url),
+            ];
+            Some(encode_env_block(&entries))
+        } else {
+            None
+        };
```
bbonaby, line 417: thought: we should probably do the same thing for base container. Adding the proxy  env variables I mean, if they give it to us. If you can we can add it to #385 before we check it in

bbonaby (approved):

jsidewhite:
Closes: #346

PR #387, opened 2026-05-22
Phase 3.5: --prepare-system-drive flag for AppContainer host setup

MGudgin:
## ≡ƒôû Description

Adds `wxc-exec --prepare-system-drive` (and its inverse `--unprepare-system-drive`) ΓÇö a one-time, host-wide setup step that grants the well-known AppContainer SIDs `S-1-15-2-1` (`ALL APPLICATION PACKAGES`) and `S-1-15-2-2` (`ALL RESTRICTED APPLICATION PACKAGES`) a tiny metadata-only access mask on the system-drive root (`C:\`).

### Why

AppContainer processes can't, by default, read directory metadata of the system-drive root via APIs such as `GetFileAttributesW`, `_stat`, or `[IO.DirectoryInfo]::GetAccessControl`. Common tools (`cmd.exe`, `powershell.exe`, `pwsh.exe`, `node.exe`) hit these during startup and fail with `ERROR_ACCESS_DENIED` inside an AppContainer. This blocks most non-trivial agent workloads on the T2/T3 paths landed in phases 3 and 4.

### What we grant

| Trustee | SID | Access mask | Inheritance |
| --- | --- | --- | --- |
| `ALL APPLICATION PACKAGES` | `S-1-15-2-1` | `FILE_READ_ATTRIBUTES \| FILE_READ_EA \| READ_CONTROL \| SYNCHRONIZE` (`0x00120088`) | none |
| `ALL RESTRICTED APPLICATION PACKAGES` | `S-1-15-2-2` | same | none |

The mask is the metadata-read set used by `GetFileAttributesW`, `_stat`, and `GetAccessControl`. It explicitly excludes `FILE_LIST_DIRECTORY` (the container still cannot enumerate `C:\`) and `FILE_READ_DATA`. Inheritance is `None`, so descendant files and subdirectories are unaffected ΓÇö the change is scoped to the directory object itself.

### Code shape

- New `wxc_common::system_drive_prep` module (Windows-only).
- `filesystem_dacl::apply_ace` refactored into a thin wrapper over a new `pub(crate)` helper `apply_explicit_ace(path, sid, mask, type, inheritable)`. Behavior-preserving.
- New `pub(crate)` helper `revoke_specific_aces_for_sid` for tuple-precise revoke (only ACEs matching the apply tuple are removed; non-matching ones ΓÇö e.g. third-party `icacls /grant` ACEs ΓÇö are preserved).
- `wxc/src/main.rs` adds three flags (two visible, one hidden) with clap `conflicts_with_all` against the other top-level "do this and exit" flags.

### Hardenings (from the adversarial review on `330fabd5`)

- **Precise tuple-matching revoke** (MF-1): `--unprepare-system-drive` no longer issues a blunt `SetEntriesInAclW(REVOKE_ACCESS)` for the SID. It scans the existing DACL, partitions explicit ACEs into matches/non-matches against the apply tuple, and only revokes matches. PowerShell scripts and Rust path now genuinely mirror each other.
- **clap mutual exclusion** (MF-2): the new flags carry `conflicts_with_all` against `--setup-hyperlight`, `--setup-wslc`, `--delete`, `--dry-run`, and each other. Invalid combos are rejected at parse time instead of silently dropping the other flag.
- **Elevated child diagnostics + WaitForSingleObject check** (MF-3): the elevated child appends its console output to `%TEMP%\wxc-exec-prepare-system-drive.log`; the unelevated parent reads and prints the contents on non-zero child exit so failures surface a real diagnostic instead of an opaque exit code. `WaitForSingleObject` return value is now checked against `WAIT_OBJECT_0`.
- **`--internal-target-path` for path-resolution hardening** (SF-1): the unelevated parent resolves `%SystemDrive%` once and passes the resolved path to the elevated child as a hidden CLI arg. The child does not re-read `%SystemDrive%` from a potentially attacker-controlled environment and validates the received path is a literal drive root (`X:\`) before touching its DACL.

### Elevation model

Modifying the DACL on `C:\` needs `WRITE_DAC`, which normal users don't hold. Windows does not allow a running process to elevate its own token, so the unelevated parent re-launches itself via `ShellExecuteExW(runas)` with a hidden `--internal-elevated-helper` flag, waits for the elevated child, and propagates the exit code. A UAC prompt is expected on the first invocation.

## ≡ƒöù References

Prerequisite for #296 (Phase 4: dispatcher integration for BaseContainer fallback). Phase 4 will need to be rebased onto this once it lands.

Related to #304 ("T3: directory enumeration fails even on granted paths") ΓÇö does not resolve it (that's about ancestor `FILE_TRAVERSE` rights for `FindFirstFile`, a different problem), but is on the same axis and helps unblock the same "cmd/pwsh inside AppContainer can't start up" symptom.

Stacks on top of #295 (Phase 3: `filesystem_dacl` module). Phase 3 is already merged to `main`, so this PR targets `main` directly.

### Automated

- `cargo fmt --all -- --check` Γ£à
- `cargo check --workspace --all-targets` Γ£à
- `cargo clippy --workspace --all-targets -- -D warnings` Γ£à
- `cargo test -p wxc_common --lib -- --test-threads=1` Γ£à (345/345 pass, including 4 new tests in `system_drive_prep::tests`: round-trip, override seam, **MF-1 precise-revoke regression**, **SF-1 drive-root validator**)

### Manual (Windows 11 25H2)

Verified end-to-end with `scripts/host-prep/Prepare-SystemDriveForAppContainer.ps1` on real `C:\`. Before the prep, `icacls C:\` showed no explicit ACEs for the AppContainer SIDs. After `wxc-exec --prepare-system-drive`, icacls confirmed both ACEs landed with mask `(Rc,S,REA,RA)` (= `0x00120088`) and no inheritance flags. Re-running `--prepare-system-drive` was idempotent. `--unprepare-system-drive` removed both. The C:\ DACL restored byte-for-byte to baseline (SDDL comparison).

UAC handshake (from a non-elevated PowerShell): observed UAC consent dialog ΓåÆ elevated child runs ΓåÆ unelevated parent prints `Elevated helper completed successfully.` ΓåÆ exit 0.

## Γ£à Checklist

- [x] Linked to an issue

(Copilot instructions: added `docs/host-prep.md` to the `Key documentation → Core references` list so future agents discover the new doc and the `scripts/host-prep/` hand-test helpers.)

## ≡ƒôï Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -142,6 +142,7 @@ Core references:
 - `docs/authoring-a-new-feature.md` — step-by-step guide for adding experimental features (which files to touch, in what order)
 - `docs/examples.md` — annotated configuration examples (see also `examples/` and `test_configs/`)
 - `docs/diagnostics.md` — diagnostic logging knobs (env vars, log file format)
+- `docs/host-prep.md` — `wxc-exec --prepare-system-drive` / `--unprepare-system-drive` host setup (metadata-only ACEs on the system-drive root for the AppContainer well-known SIDs, so `cmd.exe` / `pwsh.exe` / `node.exe` can stat `C:\` inside an AppContainer). Hand-test helpers live in `scripts/host-prep/`.
 - `docs/sandbox-policy/v1/policy.md` — sandbox policy v1 specification
 
 Per-backend guides:
--- a/docs/host-prep.md
+++ b/docs/host-prep.md
@@ -1,0 +1,109 @@
+# Host preparation: `--prepare-system-drive`
+
+The `wxc-exec --prepare-system-drive` flag adds two persistent allow ACEs to
+the system-drive root (typically `C:\`) so that AppContainer-isolated
+processes can read directory metadata of the drive. Many common tools —
+`cmd.exe`, `powershell.exe`, `pwsh.exe`, `node.exe` — call APIs like
+`GetFileAttributesW("C:\\")`, `_stat("C:\\")`, or
+`[IO.DirectoryInfo]::GetAccessControl` during startup and fail with
+`ERROR_ACCESS_DENIED` inside an AppContainer because the well-known
+AppContainer SIDs are not granted any rights on the system-drive root
+by default.
+
+This is a one-time, host-wide setup step — it does not need to be re-run
+per container or per session.
+
+## What it grants
+
+| Trustee | SID | Access mask | Inheritance |
+| --- | --- | --- | --- |
+| `ALL APPLICATION PACKAGES` | `S-1-15-2-1` | `FILE_READ_ATTRIBUTES \| FILE_READ_EA \| READ_CONTROL \| SYNCHRONIZE` (`0x00120088`) | none |
+| `ALL RESTRICTED APPLICATION PACKAGES` | `S-1-15-2-2` | same | none |
+
+What this **does not** grant:
+
+- `FILE_LIST_DIRECTORY` — the container still cannot enumerate `C:\`.
+- `FILE_READ_DATA` — irrelevant on a directory, but explicit.
+- any write rights.
+
+Because the ACEs are non-inheriting, descendant files and subdirectories
+of `C:\` are not affected — the change is scoped to the directory object
+itself.
+
+## Usage
+
+```
+wxc-exec --prepare-system-drive
+```
+
+The flag self-elevates via UAC. If the current process is not already
+elevated, Windows shows a UAC prompt; the elevated child does the DACL
+write and exits, and the unelevated parent reports success.
+
+```
+wxc-exec --unprepare-system-drive
+```
+
+Removes ACEs the prepare step added. Uses **precise tuple matching**:
+only ACEs whose `(access mask, ACE type, inheritance flags)` exactly
+match what `--prepare-system-drive` would have authored are removed.
+Other explicit ACEs for the same SIDs — e.g. an existing
+`icacls C:\ /grant "ALL APPLICATION PACKAGES":(R)` written by a
+third-party tool — are preserved. This is symmetric with the PowerShell
+`Unprepare-SystemDriveForAppContainer.ps1` script's
+`RemoveAccessRuleSpecific` semantics.
+
+Re-running either flag is safe: `SetEntriesInAclW(GRANT_ACCESS)` merges
+into the existing ACE if one already exists, and the precise-revoke
+path is idempotent — running it again finds no matching ACE and reports
+"no matching ACE; nothing to do".
+
+## Verifying the result
+
+After running `--prepare-system-drive`, the two ACEs should be visible
+in the system-drive root's DACL:
+
+```powershell
+(Get-Acl C:\).Access |
+    Where-Object { [uint32]$_.FileSystemRights -eq 0x00120088 -and -not $_.IsInherited }
+```
+
+You should see two entries — one each for the two AppContainer groups —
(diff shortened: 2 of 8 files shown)


MGudgin:
End-to-end verification on real `C:\` (Windows 11 25H2, elevated PowerShell): **15/15 pass**, including the MF-1 acceptance test against a pre-seeded third-party ACE.

```
PASS :: Step 4: prepare exits 0
PASS :: Step 4: AAP ACE landed (icacls)                          found=1
PASS :: Step 4: RAP ACE landed (icacls)                          found=1
PASS :: Step 4: AAP idempotent after re-prepare                  found=1
PASS :: Step 4: RAP idempotent after re-prepare                  found=1
PASS :: Step 5: third-party inheritable AAP ACE seeded           found=1
PASS :: Step 5: our metadata-only AAP ACE present after prep     found=1
PASS :: Step 5: seeded inheritable AAP ACE STILL present after prep  found=1
PASS :: Step 5: our metadata-only AAP ACE removed by unprep      found=0
PASS :: Step 5: seeded inheritable AAP ACE PRESERVED (MF-1)      found=1
PASS :: Step 6: validator rejects '\''C:\Windows'\''
PASS :: Step 6: validator rejects '\''\\server\share'\''
PASS :: Step 6: validator rejects '\''C:'\''
PASS :: Step 6: validator rejects '\''foobar'\''
PASS :: Cleanup: C:\ SDDL matches baseline
```

The MF-1 procedure was: `icacls C:\ /grant *S-1-15-2-1:(OI)(CI)(R)` (creates an inheritable AAP allow ACE with `FILE_GENERIC_READ`), then `wxc-exec --prepare-system-drive` (adds a no-inheritance AAP allow ACE with the metadata-only `0x00120088` mask — kept distinct by `SetEntriesInAclW` because inheritance flags differ), then `wxc-exec --unprepare-system-drive`. After unprep, icacls confirms only the seeded inheritable `(OI)(CI)(R)` ACE remains for AAP — our no-inheritance one is gone, the third-party one untouched. Precise tuple-matching revoke confirmed.

`C:\` SDDL byte-compared back to baseline after the cleanup step.

MGudgin:
Closing without merging.

Phase 3.5 was motivated by the claim that cmd/pwsh/node fail to start inside an AppContainer because they can't `stat C:\`. End-to-end testing on Windows 11 25H2 + pwsh 7.6.2 + node 24.13.0 falsified that hypothesis:

- cmd.exe runs in an AppContainer without any host-prep.
- pwsh.exe 7.6.2 handles missing `C:\` metadata-read access via [PR #27266's `SafeDoesPathExist`](https://github.com/PowerShell/PowerShell/pull/27266) catch — runs without host-prep.
- node.exe runs without host-prep.

The actual unblocker for pwsh/node was setting `ui.disable=false` (the wxc default is `true`, which applies the Win32k mitigation policy and breaks coreclr DLL init) — a separate concern from C:\ DACLs.

Phase 3.5 still has *one* confirmed empirical effect: cmd.exe's `if exist C:\` construct flips from `HIDDEN` (exit 99) to `SEEN` (exit 42) with prep applied. That's narrower than the motivating problem and not worth ~600 lines of new code (CLI flag, elevation handshake, module, doc, scripts) for the present audience.

The branch `user/gudge/downlevel_phase3_5` is preserved for revival if a real use case emerges (e.g. legacy tools that don't defensively handle `UnauthorizedAccessException` on the drive root).

PR #391, opened 2026-05-22
Fix: SDK validator accepts legacy containment wire values

asklar:
The Rust parser (wxc-exec) accepts deprecated wire values like `appcontainer` and `macos_sandbox` via serde aliases (PR #268), but the TypeScript SDK validator rejected them before the config reached the binary.

Adds a `LegacyContainmentAliases` map that resolves deprecated wire values to their canonical `ContainmentBackend` before the platform check, matching the Rust parser's backward-compat behavior.

**Changes:**
- `sdk/src/types.ts`: Add `LegacyContainmentAliases` map (`appcontainer` -> `processcontainer`, `macos_sandbox` -> `seatbelt`)
- `sdk/src/helper.ts`: Resolve through aliases before platform check
- `sdk/src/index.ts`: Export the new map

All 131 SDK tests pass.

Fixes #390
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/391)

### DIFF
--- a/sdk/src/index.ts
+++ b/sdk/src/index.ts
@@ -31,6 +31,7 @@ export {
   SandboxingMethod,
   ContainmentType,
   ContainmentTypes,
+  LegacyContainmentAliases,
   ContainmentBackend,
   ExperimentalBackends,
   ContainerConfig,
--- a/sdk/src/helper.ts
+++ b/sdk/src/helper.ts
@@ -6,7 +6,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { randomBytes } from 'crypto';
 import { FileLogger } from './logger.js';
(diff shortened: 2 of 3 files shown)


sdk/src/index.ts
```diff
@@ -31,6 +31,7 @@ export {
   SandboxingMethod,
   ContainmentType,
   ContainmentTypes,
+  LegacyContainmentAliases,
```
bbonaby, line 34: note: not sure we need to export this type to the consumer of the SDK though.For now seems like an internal structure.

bbonaby (approved):

PR #407, opened 2026-05-23
sdk: tighten legacy containment alias handling (follow-up to #391)

MGudgin:
This PR fixes the SDKs legacy-containment-alias resolution so every containment check in `resolveExecutableAndArgs` sees the canonical backend, not just the platform-availability check. Without this, `containment: "macos_sandbox"` bypasses the SDKs experimental-mode gate because `ExperimentalBackends` lists `seatbelt` but not its alias. Follow-up to #391.

## Details

- Resolve `config.containment` through `LegacyContainmentAliases` once at the top of `resolveExecutableAndArgs` and drive every containment check from the result. Wire payload still forwarded to wxc-exec unchanged.
- Tighten the alias-map type to `Readonly<Partial<Record<...>>>` so the sparse-map contract is enforced at compile time.
- Drop `LegacyContainmentAliases` from the public `index.ts` re-exports - its a deprecation-window mechanism, not API surface.
- Add a deduplicated `diagLog` deprecation hint when a legacy value is accepted (mirrors the Rust-side `Logger` emission).
- Document the version-agnostic deprecation-alias policy in `docs/versioning.md` so SDK and parser dont drift again.

## Tests

- New `legacy containment aliases` describe block in `sdk/tests/unit/sandbox.test.ts`:
  - `appcontainer` accepted on Windows
  - `appcontainer` rejected on non-Windows with the canonical "not available on this platform" error
  - `macos_sandbox` triggers the experimental-mode gate (the regression this PR fixes)
  - `macos_sandbox` accepted on macOS with `experimental: true`
  - Legacy wire value forwarded to wxc-exec unchanged (envelope re-decoded from `--config-base64`)
- `npm test` (sdk): 141 pass, 0 fail, 3 platform-conditional skips. macOS-only and Linux-only branches will run on CI.

Refs #390.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/407)

### DIFF
--- a/sdk/src/helper.ts
+++ b/sdk/src/helper.ts
@@ -35,6 +35,31 @@ export function diagLogVersion(): void {
     }
 }
 
+const legacyContainmentWarned = new Set<string>();
+
+/**
+ * Emit a one-shot deprecation hint via the diagnostic console when a caller
+ * uses a legacy containment wire value. Dedup'd per legacy value per process
+ * so the message doesn't flood the diag stream on repeated spawns.
+ *
+ * Exposed for tests so the latch can be reset between describes.
+ */
+export function warnLegacyContainmentOnce(legacy: string, canonical: string): void {
+    if (!legacyContainmentWarned.has(legacy)) {
+        legacyContainmentWarned.add(legacy);
+        diagLog(
+            `Containment value '${legacy}' is deprecated; use '${canonical}' instead. ` +
+            `The legacy spelling is accepted via a backward-compatibility alias and may be ` +
+            `removed in a future SDK release.`
+        );
+    }
+}
+
+/** @internal Reset the legacy-containment dedup latch. Intended for unit tests. */
+export function _resetLegacyContainmentWarnedForTesting(): void {
+    legacyContainmentWarned.clear();
+}
+
 /**
  * Result of preparing a sandbox spawn — includes the resolved binary,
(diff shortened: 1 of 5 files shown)


sdk/src/helper.ts
```diff
@@ -35,6 +35,31 @@ export function diagLogVersion(): void {
...
+ * Emit a one-shot deprecation hint via the diagnostic console when a caller
+ * uses a legacy containment wire value. Dedup'd per legacy value per process
+ * so the message doesn't flood the diag stream on repeated spawns.
+ *
+ * Exposed for tests so the latch can be reset between describes.
+ */
+export function warnLegacyContainmentOnce(legacy: string, canonical: string): void {
+    if (!legacyContainmentWarned.has(legacy)) {
+        legacyContainmentWarned.add(legacy);
+        diagLog(
```
bbonaby, line 50: queston (non-blocking): think it would be worth it for us to have structured info, warn, and error logs so the AI / developers can write code that can scan std out for specific  data?

bbonaby (approved):

sdk/src/helper.ts
MGudgin: Yes, I think that would be very useful, although we'd have to be careful not to 'pollute' the output of the process being contained.

PR #409, opened 2026-05-23
sdk: route containment validation through a version-aware registry

MGudgin:
> Stacked on top of #407 — review that first; this PR's diff is best read against `user/gudge/sdk-legacy-aliases-followup`.

## Summary

Replaces the scattered `LegacyContainmentAliases` map / `ExperimentalBackends` list in the SDK validator with a single declarative registry and a pure `judge()` function. The registry is the one place where every containment value the SDK has ever accepted is annotated with its `addedIn` / `deprecatedSince` / `renamedTo` / `removeIn` / `experimental` / `abstract` metadata; the helper calls `judge()` once and switches on the verdict (`ok` / `ok-deprecated` / `removed` / `too-new` / `unknown`).

Behaviour is preserved end-to-end. Legacy aliases (`appcontainer`, `macos_sandbox`) are still accepted on every supported `policy.version`, the on-wire payload still carries the raw user-supplied value verbatim, and the deprecation diagnostic still fires once per process per value — but the warning string is now driven by registry metadata, and flipping `appcontainer` from accept-with-warn to reject becomes a one-line registry edit once the deprecation window closes.

Also folds in a small drift fix: `hyperlight` was in the SDK `ContainmentBackend` union but missing from the `0.6.0-dev` schema enum.

Related Issues

- #391
- #390

## Why this shape

We considered three shapes during design:

- **A — keep `LegacyContainmentAliases`, add a separate `judge()` helper.** Smaller diff, but leaves two related APIs that can drift.
- **B — single registry + `judge()` (this PR).** One source of truth for every value's version metadata; the alias map becomes a derived view if anyone still wants it.
- **C — bundle the JSON schemas into the SDK and run ajv at runtime.** Duplicates the Rust parser's job and ties the SDK to schema-file churn.

We picked B because it consolidates the four lookups (alias rewrite, experimental gate, abstract recognition, schema-version awareness) behind one entry point.

## What's intentionally not done

- `too-new` is plumbed through the verdict union but not enforced in the helper today — the helper treats it identically to `ok`, matching pre-PR behaviour. The plumbing means enforcement is a one-line flip if/when we want it.
- The on-wire `config.containment` is still forwarded verbatim; only the SDK-local `effectiveContainment` is rewritten on `ok-deprecated`. The "forward the legacy wire value to the binary unchanged" test from #407 continues to pass.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/409)

### DIFF
--- a/sdk/src/helper.ts
+++ b/sdk/src/helper.ts
@@ -6,10 +6,11 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { randomBytes } from 'crypto';
 import { FileLogger } from './logger.js';
-import { ContainerConfig, ContainmentBackend, ContainmentTypes, ExperimentalBackends, LegacyContainmentAliases } from './types.js';
+import { ContainerConfig, ContainmentBackend, ContainmentTypes } from './types.js';
 import { findWxcExecutable, findLxcExecutable, findSeatbeltExecutable, getPlatformSupport } from './platform.js';
 import { SandboxSpawnOptions } from './sandbox.js';
 import { diagLog } from './diagnostic.js';
+import { ContainmentRegistry, judge, Verdict } from './version-registry.js';
 
 /** SDK version read from package.json at module load time. */
 export const SDK_VERSION: string = (() => {
@@ -42,15 +43,25 @@ const legacyContainmentWarned = new Set<string>();
  * uses a legacy containment wire value. Dedup'd per legacy value per process
  * so the message doesn't flood the diag stream on repeated spawns.
  *
+ * The verdict carries the rename target and version metadata sourced from
+ * `ContainmentRegistry`, so the message stays in sync with the registry
+ * without further plumbing.
+ *
  * Exposed for tests so the latch can be reset between describes.
  */
-export function warnLegacyContainmentOnce(legacy: string, canonical: string): void {
+export function warnLegacyContainmentOnce(
+    legacy: string,
+    verdict: Extract<Verdict, { kind: 'ok-deprecated' }>,
+): void {
     if (!legacyContainmentWarned.has(legacy)) {
         legacyContainmentWarned.add(legacy);
+        const removalNote = verdict.removeIn
+            ? ` and will be removed in ${verdict.removeIn}`
+            : '';
         diagLog(
-            `Containment value '${legacy}' is deprecated; use '${canonical}' instead. ` +
-            `The legacy spelling is accepted via a backward-compatibility alias and may be ` +
-            `removed in a future SDK release.`
+            `Containment value '${legacy}' is deprecated since schema ${verdict.deprecatedSince}${removalNote}; ` +
+            `use '${verdict.canonical}' instead. The legacy spelling is accepted via a ` +
+            `backward-compatibility alias.`
         );
     }
 }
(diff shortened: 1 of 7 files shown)


sdk/src/helper.ts
```diff
@@ -160,33 +171,56 @@ export function resolveExecutableAndArgs(
...
+  let effectiveContainment: string | undefined = rawContainment;
+  let isExperimental = false;
+  let isIntent = false;
+  if (rawContainment) {
+    const verdict = judge(ContainmentRegistry, rawContainment, config.version);
+    switch (verdict.kind) {
+      case 'ok-deprecated':
+        warnLegacyContainmentOnce(rawContainment, verdict);
+        effectiveContainment = verdict.canonical;
+        isExperimental = !!verdict.entry.experimental;
```
bbonaby, line 191: suggestion: for these `!!` lines, since they are all booleans I think it'd be clearer to just use the `=== true` pattern so it'd clear at a glance.

sdk/src/helper.ts
```diff
@@ -160,33 +171,56 @@ export function resolveExecutableAndArgs(
...
+        isIntent = !!verdict.entry.abstract;
+        break;
+      case 'ok':
+      case 'too-new':
+        effectiveContainment = verdict.canonical;
+        isExperimental = !!verdict.entry.experimental;
+        isIntent = !!verdict.entry.abstract;
+        break;
+      case 'removed':
+        throw new Error(
```
bbonaby, line 201: thought (non-blocking): I do wonder if we should be throwing https://github.com/microsoft/mxc/blob/9ceb814873cf0c29f7b115d96ce06f7ae77b7943/sdk/src/errors.ts#L31  in the SDK now so we're consistent. I know we don't use it consistently, but figured I'd ask. What do you think?

sdk/src/version-registry.ts
```diff
@@ -0,0 +1,152 @@
...
+ * versions the SDK supports. One entry per value the schemas have ever
+ * accepted — canonical names and legacy aliases alike.
+ *
+ * When adding a value: set `addedIn` to the schema version that first
+ * accepted it. When renaming a value: leave the legacy entry in place with
+ * `deprecatedSince`, `renamedTo`, and (optionally) `removeIn` set, and add
+ * a fresh entry for the canonical name with `addedIn` = the rename
+ * version.
+ */
+export interface RegistryEntry {
```
bbonaby, line 24: thought (non-blocking): This is a nice interface but at first I thought we were talking about the Windows Registry. Not sure if others would get confused at first without coming to this file to check. That said, not sure what other names we could use. Would `SchemaValueEntry` be acceptable?

sdk/src/version-registry.ts
```diff
@@ -0,0 +1,152 @@
...
+ * later without changing the registry shape.
+ */
+export function judge(reg: FieldRegistry, value: string, declaredVersion?: SemverString): Verdict {
+    const entry = reg[value];
+    if (!entry) {
+        return { kind: 'unknown', rawValue: value };
+    }
+
+    const canonicalName = entry.renamedTo ?? value;
+    const canonicalEntry = entry.renamedTo ? (reg[entry.renamedTo] ?? entry) : entry;
```
bbonaby, line 117: suggestion: I think we can just do
```typescript
const canonicalName = entry.renamedTo ?? value;
const canonicalEntry = reg[canonicalName] ?? entry;
```

sdk/src/version-registry.ts
```diff
@@ -0,0 +1,152 @@
...
+ * `'unknown'` is returned for values that are not in the registry. Callers
+ * should treat that as "no version-aware information; fall back to the
+ * static TS union as the entry gate" — the registry is additive, not
+ * replacing the TS union.
+ *
+ * `'too-new'` (anachronism) is plumbed through but not enforced by any
+ * caller today. The verdict carries the metadata to flip to rejection
+ * later without changing the registry shape.
+ */
+export function judge(reg: FieldRegistry, value: string, declaredVersion?: SemverString): Verdict {
```
bbonaby, line 110: suggestion: `judge` might be a bit vague. Would `getSchemaValueVerdict` convey this better?

sdk/tests/unit/version-registry.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      assert.strictEqual(verdict.kind, 'ok');
+    });
+  });
+});
+
+describe('ContainmentRegistry coverage', () => {
+  it('has a registry entry for every value in the ContainmentBackend union', () => {
+    // Defensive: if a new backend lands in the TS union without a registry
+    // entry, the SDK falls back to 'unknown' and loses experimental gating
+    // for that value. The hyperlight drift bug (#TBD) is exactly this.
```
bbonaby, line 154: note: you can probably remove "The hyperlight drift bug (#TBD) is exactly this."

bbonaby (approved):

kanismohammed:
Closing the pull request for repo release

PR #412, opened 2026-05-26
Dedup DACL masks into a single source of truth

MGudgin:
This PR collapses three local definitions of the RW / RO access masks
(in dispatcher, fallback_detector, and filesystem_dacl) into one set
of `pub(crate)` constants owned by filesystem_dacl — the module that
actually stamps them. The three sites now import the same constants
instead of re-deriving the same bit pattern.

Details:
- Add `RW_MASK` / `RO_MASK` as `pub(crate) const` in filesystem_dacl,
  next to `grant_appcontainer_access` which applies them.
- Remove `T3_RW_MASK` / `T3_RO_MASK` from dispatcher and `rw_mask()` /
  `ro_mask()` from fallback_detector. Both modules now import from
  filesystem_dacl.
- Add a module-level `const _: () = assert!(...)` that pins the
  bit-pattern of each mask at compile time, so a future drift in
  either constant fails the build rather than silently shifting the
  filter / precheck behavior.

Test:
- wxc_common cargo test: 388/388 (unchanged — pure refactor; values
  byte-identical to before).
- cargo clippy --all-targets -- -D warnings: clean.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/412)

### DIFF
--- a/src/wxc_common/src/dispatcher.rs
+++ b/src/wxc_common/src/dispatcher.rs
@@ -73,7 +73,7 @@ use crate::appcontainer_runner::{derive_sid_string, AppContainerScriptRunner, Fi
 use crate::base_container_runner::BaseContainerRunner;
 use crate::error::WxcError;
 use crate::fallback_detector::{self, FallbackError, IsolationTier};
-use crate::filesystem_dacl::{DaclError, DaclManager};
+use crate::filesystem_dacl::{DaclError, DaclManager, RO_MASK, RW_MASK};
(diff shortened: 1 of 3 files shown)


alexsnitkovskiy (approved):
thank you

PR #421, opened 2026-05-26
[Bubblewrap] Network proxy for Bubblewrap- unprivileged proxy based network filtering

SohamDas2021:
## 📖 Description

Adds a userspace-proxy enforcement path to the Bubblewrap backend so per-host `allowedHosts` / `blockedHosts` can be applied **without root**, removing the iptables-only restriction that made the "unprivileged" backend require privileges for any non-trivial network policy.

When `network.proxy` is configured on `containment: "bubblewrap"`, the runner:

1. Spawns a new Linux-side `linux-test-proxy` (HTTP CONNECT + forward, host filtering) preloaded with the request's allow/block lists and default policy.
2. Injects `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` into the sandbox via `bwrap --setenv` so cooperating clients (curl, git, pip, npm, requests, gh, …) route through the proxy.
3. Skips `--unshare-net` (sandbox needs loopback to reach the proxy) and skips the iptables path entirely.

Honor-system enforcement is documented as a known limitation — a hostile binary that ignores proxy env vars can still talk to the host network. Users needing hard enforcement on Linux keep using LXC. For the agent / CI workloads bwrap targets, this is a strict improvement over the previous "either root-iptables or all-or-nothing `--unshare-net`" choice.

### What's in this PR

**New Rust crate**
- `src/linux_test_proxy/` — Linux-native HTTP CONNECT + forward proxy with pluggable `HostFilter` (allow/block/default-policy decisions). Tokio-based, exits on SIGTERM or parent death (`PR_SET_PDEATHSIG`).

**Bubblewrap runner integration**
- `bwrap_common/src/bwrap_runner.rs` — new decision tree: proxy > iptables > `--unshare-net` > shared netns.
- `bwrap_common/src/bwrap_command.rs` — accepts optional `proxy_port` and emits `--setenv` for the three env vars; suppresses `--unshare-net` when proxy is in use.

**Config / parser**
- `network.proxy` is now accepted on `containment: "bubblewrap"` (previously `processcontainer`-only); schema description updated.
- New `wxc_common/src/linux_proxy_coordinator.rs` mirrors the Windows proxy coordinator: locates `linux-test-proxy` next to `lxc-exec`, spawns it, reads the ready-file, tears down on exit.

**Tests**
- `sdk/tests/integration/linux-bubblewrap.test.ts` — three new cases:
   externally launched proxy, `builtinTestServer`, allowedHosts enforcement.
   Gated on `isLinuxBubblewrap` (no root required for the proxy path).
- `test_configs/bubblewrap_network_proxy_{builtin,allowlist,blocklist}.json`
   + `test_scripts/run_bwrap_network_proxy_test.sh` for native E2E coverage.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/421)

### DIFF
--- a/docs/bwrap-support/bubblewrap-backend.md
+++ b/docs/bwrap-support/bubblewrap-backend.md
@@ -143,6 +143,74 @@ Standard `process` fields work as expected:
 }
 ```
 
+## Network proxy (cooperative, unprivileged)
+
+Bubblewrap supports an **unprivileged, cooperative network proxy** that
+enforces `allowedHosts` / `blockedHosts` at the proxy layer instead of via
+iptables. This is the **recommended** way to do per-host filtering on
+Bubblewrap because it requires **no root and no `CAP_NET_ADMIN`**.
+
+### How it works
+
+1. When `network.proxy` is set, the runner launches an unprivileged HTTP
+   proxy on loopback (`127.0.0.1:N`). For tests, the bundled
+   `linux-test-proxy` binary is used (`builtinTestServer: true`); in
+   production callers supply their own proxy via `localhost: <port>`.
+2. The sandbox is then started **without** `--unshare-net` so the sandbox
+   shares the host network namespace and can reach the loopback proxy.
+3. The command builder sets `HTTP_PROXY`, `HTTPS_PROXY`, `http_proxy`,
+   `https_proxy`, and `NO_PROXY=localhost,127.0.0.1` inside the sandbox
+   via `bwrap --setenv` (any caller-supplied values for these keys are
+   stripped before injection).
+4. Cooperative tools (curl, wget, Python `requests`, Node `https`, etc.)
+   honor the env vars and traffic flows through the proxy, which applies
+   the `allowedHosts` / `blockedHosts` lists.
+
+### Example: builtin test proxy with allowlist
+
+```json
+{
+  "version": "0.6.0-alpha",
+  "platform": "linux",
+  "containment": "bubblewrap",
+  "process": {
+    "commandLine": "curl -fsSL https://api.github.com/zen && echo OK"
+  },
+  "network": {
+    "defaultPolicy": "allow",
+    "proxy": { "builtinTestServer": true },
+    "allowedHosts": ["api.github.com"]
+  }
+}
+```
+
+### Example: external proxy on loopback
+
+```json
+{
+  "version": "0.6.0-alpha",
+  "containment": "bubblewrap",
+  "process": { "commandLine": "curl -fsSL https://example.com" },
+  "network": {
+    "proxy": { "localhost": 8080 }
+  }
+}
+```
+
+### Caveats
+
+- **Cooperative model**: only well-behaved clients that honor `HTTP_PROXY`
+  / `HTTPS_PROXY` are filtered. Tools that bypass these env vars (raw
+  sockets, custom HTTP clients) are **not enforced**. This is a deliberate
+  trade-off for zero-privilege operation.
+- **Mutually exclusive with iptables enforcement**: setting
+  `network.proxy` together with `network.enforcementMode` of `"firewall"`
+  or `"both"` is rejected at config-parse time because iptables-based
+  enforcement requires root and would defeat the proxy's privilege story.
+- **HTTPS via CONNECT**: the proxy uses HTTP `CONNECT` tunnels for TLS, so
+  certificate validation continues to work end-to-end (the proxy does not
+  see plaintext).
+
 ## Comparison with LXC
 
 | Aspect | LXC | Bubblewrap |
@@ -182,8 +250,10 @@ Test configs are in `test_configs/bubblewrap_*.json`.
 - **Linux only** — Bubblewrap requires Linux kernel namespaces
 - **Host filesystem** — the sandbox sees the host's files (read-only by
   default); there is no separate rootfs
-- **Network filtering** — per-host `allowedHosts`/`blockedHosts` requires
-  root for iptables; without host lists, only all-or-nothing isolation is
-  available
+- **Network filtering** — per-host `allowedHosts`/`blockedHosts` is best
+  done via the cooperative env-var **network proxy** (no privilege
+  required, see above). The legacy iptables path
+  (`network.enforcementMode: "firewall"` / `"both"`) still works but
+  requires root and is mutually exclusive with the proxy.
 - **No state-aware lifecycle** — Bubblewrap implements `ScriptRunner` only
   (one-shot), not `StatefulSandboxBackend`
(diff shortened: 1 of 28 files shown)


docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -143,6 +143,74 @@ Standard `process` fields work as expected:
...
+
+- **Cooperative model**: only well-behaved clients that honor `HTTP_PROXY`
+  / `HTTPS_PROXY` are filtered. Tools that bypass these env vars (raw
+  sockets, custom HTTP clients) are **not enforced**. This is a deliberate
+  trade-off for zero-privilege operation.
+- **Mutually exclusive with iptables enforcement**: setting
+  `network.proxy` together with `network.enforcementMode` of `"firewall"`
+  or `"both"` is rejected at config-parse time because iptables-based
+  enforcement requires root and would defeat the proxy's privilege story.
+- **HTTPS via CONNECT**: the proxy uses HTTP `CONNECT` tunnels for TLS, so
```
bbonaby, line 231: question: To confirm this is only for the test proxy right? in the `bring your own proxy` case they can do whatever they want correct?

schemas/dev/mxc-config.schema.0.6.0-dev.json
```diff
@@ -143,7 +143,7 @@
         },
         "proxy": {
           "type": "object",
-          "description": "Proxy configuration. Only supported with processcontainer backend.",
+          "description": "Proxy configuration. Supported with the 'processcontainer' (Windows) and 'bubblewrap' (Linux) backends. On Bubblewrap the proxy URL is exposed inside the sandbox as HTTP_PROXY / HTTPS_PROXY / NO_PROXY environment variables (cooperative env-var proxying), and is mutually exclusive with network.enforcementMode=='firewall'/'both'.",
```
bbonaby: issue: I think this description has too much implementation details. I'd stop after the bubblewrap comment

sdk/src/sandbox.ts
```diff
@@ -87,21 +87,30 @@ function buildWslcContainerConfig(
...
- * so the Linux runner actually applies iptables rules. Without this,
- * `network_enforcement_mode` falls back to the parser default of
- * `Capabilities` and the runner silently skips iptables, dropping
- * `allowedHosts` / `blockedHosts` on the floor.
+ * Apply iptables-mode network enforcement on configs that use
+ * Bubblewrap or LXC and have per-host filtering.
+ *
+ * Without setting `enforcementMode: 'firewall'` explicitly, the Linux
+ * runner falls back to `Capabilities`, silently drops `allowedHosts` /
+ * `blockedHosts`, and never invokes iptables.
```
bbonaby: question: Capabilities doesn't actually exist on linux no? that's just a app container/process container mechanism on windows. So we might need to say it also does nothing on linux, however we should log a warning to the console though.

sdk/src/sandbox.ts
```diff
@@ -87,21 +87,30 @@ function buildWslcContainerConfig(
...
+ * server-side).
+ *
+ * NOTE: when `network.proxy` is configured on Bubblewrap, host filtering
+ * is enforced at the proxy layer (unprivileged, no CAP_NET_ADMIN). The
+ * Rust config parser explicitly rejects `bubblewrap + proxy + firewall`
+ * since the iptables path requires privilege the bwrap backend
+ * deliberately avoids. Callers in that mode must leave enforcementMode
+ * at its default.
  */
 function applyIptablesNetworkEnforcement(config: ContainerConfig): void {
```
bbonaby: issue: I'm thinking this function should be renamed at this point right? We can probably moved it to the helpers.ts file to keep this file lean.

src/bwrap_common/src/bwrap_command.rs
```diff
@@ -79,10 +104,36 @@ pub fn build_args(request: &CodexRequest) -> Vec<String> {
     args.push("--clearenv".into());
     for env_str in &request.env {
         if let Some((key, value)) = env_str.split_once('=') {
+            // When the proxy is active, drop any caller-supplied proxy env
+            // entries so they cannot override the values we set below.
+            if proxy_address.is_some() && PROXY_ENV_KEYS.contains(&key) {
```
bbonaby, line 109: note: Something I've thought about is that the script that is in the commandline section of the schema can actually update these environment variables  in the script itself even after we do all this work. So we can't 100% prevent them using some other proxy. I don't think there is anything we can do other than scan the commandline for theses environment variable name and if we find them throw an error. That said it's tricky since we'd have to be sure they are setting them, and we'd only throw if the config itself tells us the user wants to use a proxy. Just thinking aloud, I don't think we can make that change in this PR.

src/linux_test_proxy/src/proxy.rs
```diff
@@ -0,0 +1,344 @@
...
+
+//! Minimal HTTP proxy for testing — supports CONNECT tunnels and HTTP
+//! forwarding, with optional allow / block host filtering enforced at the
+//! proxy.
+//!
+//! Adapted from `wxc_test_proxy::proxy` and extended with host filtering so
+//! the Bubblewrap backend can enforce `allowedHosts` / `blockedHosts` at the
+//! proxy layer instead of via privileged iptables rules.
+
+use std::sync::Arc;
```
bbonaby, line 12: question: any way we can kind of merge the test proxy for wxc and this? I mean like sure we'll still have the separate executables but they can share a single library right? That would be nice for when we introduce a MacOS test proxy too.

src/wxc_common/src/linux_proxy_coordinator.rs
```diff
@@ -0,0 +1,673 @@
...
+    fs::create_dir(&dir).map_err(|err| {
+        WxcError::NetworkProxy(format!(
+            "Failed to create proxy temp dir {}: {}",
+            dir.display(),
+            err
+        ))
+    })?;
+
+    // Best-effort 0700 chmod so other users cannot snoop the ready file.
+    #[cfg(unix)]
```
bbonaby: suggestion: probably want ` #[cfg(linux)]` here

MGudgin (approved):

src/linux_test_proxy/src/proxy.rs
SohamDas2021: Slightly bigger change, filed https://github.com/microsoft/mxc/issues/440 to address this later.

src/bwrap_common/src/bwrap_command.rs
SohamDas2021: I agree with your opinion here, something for the future.

MGudgin (approved):

PR #425, opened 2026-05-27
Add wxc-host-prep.exe

MGudgin:
This PR adds a dedicated, manifest-elevated Windows host-prep binary
(`wxc-host-prep.exe`) and removes the `wxc-exec --prepare-system-drive` /
`--unprepare-system-drive` self-elevation path. `wxc-exec` no longer
spawns elevated copies of itself.

Details

* New workspace crate `src/wxc_host_prep/` -> `wxc-host-prep.exe`.
  Embedded `requireAdministrator` manifest in release builds (debug
  builds skip it so `cargo test` keeps working) plus a runtime
  `TokenElevation` check. `wxc-` prefix matches `wxc-exec`; `mxc-`
  in this tree is reserved for cross-platform / macOS code.
* Subcommands: `prepare-system-drive` / `unprepare-system-drive`
  (migrated from `wxc_common::system_drive_prep`) and
  `prepare-null-device` / `verify-null-device` / `dump-null-device`
  (new — applies MXC's managed SD to `\Device\Null`, which the
  kernel resets to an AppContainer-hostile default at every boot).
* `wxc/src/main.rs`: removed the 5 self-elevation clap fields and
  their dispatch block; `~1000 LOC` of `system_drive_prep` deleted
  from `wxc_common`.
* `build.bat` copies `wxc-host-prep.exe` into `sdk/bin/<arch>/`;
  `docs/host-prep.md` rewritten end-to-end; `.github/copilot-instructions.md`
  updated.

Tests

* `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`,
  `cargo clippy --workspace --all-targets -- -D warnings` -- all clean.
* `cargo test --workspace --exclude wxc_e2e_tests` -- pass,
  including 17 new `wxc_host_prep` unit tests (SDDL round-trip,
  ACE extraction, drift detection, exit-code mapping, JSON
  serialisation, RFC 3339 formatting, `MapGenericMask`
  normalisation).
* End-to-end on a real Win11 host: `prepare-null-device` ->
  `applied` on first run, `no-change` on second; `verify-null-device`
  -> `match`; `dump-null-device` round-trips.


MGudgin:
Force-pushed `1bbb4bb` addressing the six review comments:

1. **`cli.rs:13` exit-code table** — trimmed the "or applied-after-drift" wording to "or generic non-fatal error"; that row is just the catch-all for the verify path's no-match exit. (`cli.rs`)

2. **`Cargo.toml:18-19` unused `serde` / `serde_json` deps** — kept *both* and actually wired them up rather than dropping them. The hand-rolled `escape_for_json()` had a real reviewer-spotted hot spot (Windows error messages can contain backslashes, quotes, paths, locale text), so I introduced a typed `ApplyLogRecord` struct with `#[derive(Serialize)]` and switched the three apply sites to `serde_json::to_string`. Verify and dump JSON branches now use the `serde_json::json!{}` macro. `escape_for_json()` is gone. New unit tests pin the wire-format field names and the `"`/`\` escaping behaviour.

3. **`docs/host-prep.md:184` dump-null-device --json** — trimmed the description to match the actual `{"op":"dump-null-device","sddl":"…"}` output and pointed triage to `verify-null-device --json` for drift detail.

4. **`docs/host-prep.md:218` log example timestamp** — addressed via #6 below; the example shape (`2026-05-27T03:06:56Z`) is now what the code actually emits.

5. **`docs/host-prep.md:224` system-drive logging** — removed the "logs to the same directory" sentence and added an explicit note that the system-drive subcommands only write to stdout/stderr today.

6. **`null_device/mod.rs:273` `now_rfc3339()` returning `"epoch:<secs>"`** — replaced with a real pure-std RFC 3339 UTC formatter using Howard Hinnant's civil-from-days algorithm. No new dependency added. Locked behaviour with unit tests for `1970-01-01`, `2000-01-01`, `2020-01-01`, and the `2020-02-29` leap day.

Verification ladder run on the rewritten tip:

- `cargo check --workspace --all-targets` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace --no-fail-fast` — all suites pass, including 15 `mxc_host_prep` unit tests (12 original + 3 new for the JSON / RFC 3339 work)

The previous run's `WXC-Exec Hyperlight` failure was a transient runner-side `ERROR_NO_SYSTEM_RESOURCES` during Hyperlight install (other PRs in the same hour saw the same flake and passed on retry). The new CI run on this push should clear it; if it doesn't, I'll trigger `gh run rerun --failed`.

MGudgin:
Follow-up: fixed a real bug that on-device testing surfaced.

**Symptom:** `prepare-null-device` reported `applied` on every run, and `verify-null-device` reported `dacl-differs` even immediately after a successful apply. `dump-null-device` showed all 6 expected DACL ACEs (`WD`, `SY`, `BA`, `RC`, `AC`, `S-1-15-2-2`) plus the integrity label -- so the *write* path was correct.

**Root cause:** the kernel applies `IoFileObjectType`'s generic mapping at `SetKernelObjectSecurity` time. Our target SD (parsed from SDDL) has `GENERIC_READ|WRITE|EXECUTE = 0xE0000000`; the kernel expands those to `FILE_GENERIC_READ|WRITE|EXECUTE = 0x001201BF` before persisting. Our diff was comparing raw mask bytes, so unmapped target vs. mapped current never matched. `FA` ACEs were fine because the SDDL parser already turned that into the specific `0x001F01FF`.

**Fix (`sd.rs`):**

- Added the `FILE_GENERIC_MAPPING` constant matching the kernel's `IoFileObjectType_GenericMapping`.
- Added `normalize_access_mask()` that runs each mask through `MapGenericMask` with that table; idempotent for already-specific masks (`FA`, `SYSTEM_MANDATORY_LABEL_NO_WRITE_UP`, etc).
- Applied normalisation inside `extract_aces` so both sides of the diff are canonicalised before set comparison.
- Two new unit tests: `normalize_expands_file_generics` (pins `GR|GW|GX -> 0x1201BF`, `GR|GX -> 0x1200A9`, `GENERIC_ALL -> 0x1F01FF`) and `normalize_is_idempotent_for_specific_masks` (pins `FA`, integrity-label bit, and zero mask round-tripping).

**End-to-end verification on a real Win11 host (admin shell):**

```
mxc-host-prep.exe prepare-null-device   -> no-change
mxc-host-prep.exe verify-null-device    -> match
mxc-host-prep.exe dump-null-device      -> all 6 DACL ACEs + (ML;;NW;;;LW)
```

Verification ladder also re-run (fmt / check / clippy / test) -- clean; 17 `mxc_host_prep` unit tests pass.

Amended into the same commit and force-pushed.

jsidewhite:
This comment slightly concerning.

```
Not run locally: on-VM end-to-end of the boot-trigger task and
live \Device\Null open/read/write/verify. Those need a
freshly-rebooted Windows host with admin rights and are tracked
as a follow-up before merge.
```

scripts/host-prep/Register-NullDeviceAclTask.ps1
```diff
@@ -0,0 +1,139 @@
+<#
+.SYNOPSIS
+    Register a boot-trigger scheduled task that hardens the security
```
jsidewhite: >Register a boot-trigger scheduled task  [](http://example.com/codeflow?start=4&length=39)

whoa..

scripts/host-prep/Register-NullDeviceAclTask.ps1
```diff
@@ -0,0 +1,139 @@
...
+}
+
+$action = New-ScheduledTaskAction -Execute $exe -Argument 'prepare-null-device --quiet'
+
+$trigger = New-ScheduledTaskTrigger -AtStartup
+
+$principal = New-ScheduledTaskPrincipal `
+    -UserId 'S-1-5-18' `
+    -LogonType ServiceAccount `
+    -RunLevel Highest
```
jsidewhite: Is this script itself attackable?  what if standard-user malware/bot injects a payload into this script, and then user runs it (as admin) - and it puts a SYSTEM task to run at boot.

Should this script,
1. be an exe with a signature/publisher that shows in UAC prompt? or
2. have the powershell signature thingy at the end?

jsidewhite: a tampered binary shows 'Unknown publisher'

jsidewhite (approved):
:shipit:

jsidewhite (approved):

PR #456, opened 2026-05-29
Probe: report Tier 3 (not Tier 2) when BFS is compiled out, with host-prep warnings

MGudgin:
## Summary

Changes how `wxc-exec --probe` (the detection-only path the SDK's `getPlatformSupport()` uses) reports the AppContainer fallback tier, and adds actionable `wxc-host-prep` recommendations for the Tier 3 path.

## Details

* `fallback_detector::detect` now gates the entire Tier 2 (AppContainer + BFS) block behind `cfg!(feature = "tier2_bfs")`. Without the feature, `bfscfg.exe` can never be resolved, so the no-filesystem-policy short-circuit previously mis-reported `appcontainer-bfs` on a binary physically unable to run BFS. It now skips Tier 2 and falls through to Tier 3 (AppContainer + DACL).
* When Tier 3 is selected, `detect` appends `wxc-host-prep` recommendations (`prepare-system-drive`, `prepare-null-device`) to the decision warnings, but only for preparations not already in effect — detected read-only with no elevation or host mutation.
* Added read-only `filesystem_dacl::null_device_appcontainer_grants()`: opens `\\.\NUL`, reads its DACL, and reports whether both AppContainer package SIDs (S-1-15-2-1/-2) hold a non-zero Allow ACE. Everyone (S-1-1-0) is excluded because the kernel default can grant Everyone while still failing AppContainer. The system-drive check reuses `scan_explicit_aces_for_sid`, matching `wxc_host_prep`'s `STAT_ACCESS_MASK` (0x00120088) tuple.
* Documented the probe-driven recommendations in `docs/host-prep.md`.

## Tests

* `cargo test -p wxc_common`: all tests pass (default and `--features tier2_bfs`). Added unit tests for the no-BFS fall-through, the BFS-feature path, warning actionability/bounds, system-drive-root formatting, and a null-device grants smoke test.
* `cargo fmt --all -- --check` clean; `cargo clippy -p wxc_common --all-targets -- -D warnings` clean in both feature configs; `cargo check --workspace --all-targets` clean.
* End-to-end (prep present): `--probe` with no config reports `appcontainer-dacl` with the BFS fall-through warning and no host-prep warnings when both preps are in effect.
* End-to-end (prep removed): after removing the AppContainer ACEs from `C:\`, `--probe` emits the `Run wxc-host-prep prepare-system-drive (elevated)` warning while suppressing the null-device warning (null device still prepared), confirming the read-only state check flips as expected.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/456)

### DIFF
--- a/docs/host-prep.md
+++ b/docs/host-prep.md
@@ -33,6 +33,27 @@ All subcommands require elevation. The binary aborts with exit code
 `65` and a clear message if launched without an elevated token (e.g.
 running a debug build directly without `Run as Administrator`).
 
+### When does MXC need these?
+
+The system-drive and null-device preparations matter for the
+**AppContainer + DACL** isolation tier (Tier 3), which MXC selects on
+hosts without the in-process BaseContainer API (and, for builds that
+ship without the `tier2_bfs` Cargo feature, without AppContainer + BFS
+either). To make the requirement discoverable, `wxc-exec --probe` (the
+detection-only path the SDK's `getPlatformSupport()` uses) emits an
+operator-visible warning recommending the relevant `wxc-host-prep`
+subcommand **whenever Tier 3 is selected and the corresponding prep is
+not already in effect**:
+
+- If the system-drive root is missing the metadata ACEs, the probe
+  recommends `wxc-host-prep prepare-system-drive`.
+- If `\Device\Null` does not grant the AppContainer package SIDs (it
+  resets to an AppContainer-hostile default at every boot), the probe
+  recommends `wxc-host-prep prepare-null-device`.
+
+The probe performs these checks read-only (no elevation, no writes); it
+suppresses a recommendation once the matching prep is detected.
+
 ### `prepare-system-drive`
 
 Adds two persistent, non-inheriting allow ACEs to the system-drive
--- a/src/wxc_common/src/fallback_detector.rs
+++ b/src/wxc_common/src/fallback_detector.rs
@@ -105,9 +105,17 @@ pub enum FallbackError {
 /// 1. If `MXC_FORCE_TIER` is set in a test build, honor it (test seam).
 /// 2. Try Tier 1 (BaseContainer) when `prefer_base_container` is true and the
 ///    API surface is detected.
-/// 3. Otherwise try Tier 2 (AppContainer + BFS) when there's no filesystem
-///    policy at all, or `bfscfg.exe` is on disk.
-/// 4. Otherwise fall back to Tier 3 (AppContainer + DACL).
+/// 3. Otherwise try Tier 2 (AppContainer + BFS), but **only when this binary
+///    was compiled with the `tier2_bfs` Cargo feature**. With the feature on,
+///    Tier 2 is selected when there's no filesystem policy at all, or
+///    `bfscfg.exe` is on disk. With the feature off, `bfscfg.exe` can never be
+///    resolved, so Tier 2 is skipped entirely (rather than mis-reporting
+///    `appcontainer-bfs` for the no-policy case) and we fall through to Tier 3.
+/// 4. Otherwise fall back to Tier 3 (AppContainer + DACL). When Tier 3 is
+///    selected we append `wxc-host-prep` recommendations to the decision's
+///    warnings for any host-side preparation (system-drive metadata ACEs;
+///    `\Device\Null` descriptor) that is read-only-detected as not already in
+///    effect on this machine.
 ///
 /// Any tier that needs to modify host DACLs (T3 always; T1/T2 when
 /// `denied_paths` is non-empty) requires `fallback.allow_dacl_mutation = true`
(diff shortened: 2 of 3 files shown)


src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1405,6 +1405,153 @@ pub(crate) fn compute_appcontainer_effective_access(path: &Path) -> Result<u32,
...
+        // ACCESS_ALLOWED_ACE_TYPE only; deny / audit / object ACEs are
+        // not "grants" and are ignored.
+        if header.AceType != 0x00 {
+            continue;
+        }
+        let allowed = ace_ptr as *const ACCESS_ALLOWED_ACE;
+        let mask = unsafe { (*allowed).Mask };
+        if mask == 0 {
+            continue;
+        }
```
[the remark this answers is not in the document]
MGudgin: This is fine for now, we'll refactor this later.

docs/host-prep.md
```diff
@@ -33,6 +33,27 @@ All subcommands require elevation. The binary aborts with exit code
 `65` and a clear message if launched without an elevated token (e.g.
 running a debug build directly without `Run as Administrator`).
 
+### When does MXC need these?
+
+The system-drive and null-device preparations matter for the
+**AppContainer + DACL** isolation tier (Tier 3), which MXC selects on
+hosts without the in-process BaseContainer API (and, for builds that
+ship without the `tier2_bfs` Cargo feature, without AppContainer + BFS
+either). To make the requirement discoverable, `wxc-exec --probe` (the
```
jsidewhite, line 42: This clause sequence is muddling

jsidewhite (approved):
:shipit:

docs/host-prep.md
```diff
@@ -33,6 +33,27 @@ All subcommands require elevation. The binary aborts with exit code
...
+subcommand **whenever Tier 3 is selected and the corresponding prep is
+not already in effect**:
+
+- If the system-drive root is missing the metadata ACEs, the probe
+  recommends `wxc-host-prep prepare-system-drive`.
+- If `\Device\Null` does not grant the AppContainer package SIDs (it
+  resets to an AppContainer-hostile default at every boot), the probe
+  recommends `wxc-host-prep prepare-null-device`.
+
+The probe performs these checks read-only (no elevation, no writes); it
```
bbonaby, line 54: issue: I just realized but `wxc-host-prep` will not get signed as part of the build and it will not be available and won't be added to the npm package unless you add it to https://github.com/microsoft/mxc/blob/c55dcff29b375d19da7bc880d6a89192394ebae7/.azure-pipelines/templates/Rust.Build.Job.yml#L46

src/wxc_common/src/fallback_detector.rs
```diff
@@ -105,9 +105,17 @@ pub enum FallbackError {
...
-/// 3. Otherwise try Tier 2 (AppContainer + BFS) when there's no filesystem
-///    policy at all, or `bfscfg.exe` is on disk.
-/// 4. Otherwise fall back to Tier 3 (AppContainer + DACL).
+/// 3. Otherwise try Tier 2 (AppContainer + BFS), but **only when this binary
+///    was compiled with the `tier2_bfs` Cargo feature**. With the feature on,
+///    Tier 2 is selected when there's no filesystem policy at all, or
+///    `bfscfg.exe` is on disk. With the feature off, `bfscfg.exe` can never be
+///    resolved, so Tier 2 is skipped entirely (rather than mis-reporting
+///    `appcontainer-bfs` for the no-policy case) and we fall through to Tier 3.
+/// 4. Otherwise fall back to Tier 3 (AppContainer + DACL). When Tier 3 is
```
bbonaby, line 114: question (non-blocking): if I recall we do not compile `tier2_bfs` in release production builds     so effectively it's either base container or appcontainer + dacl for folks in the wild? Am I understanding that right? And for the appcontainer + dacl case consumers will need to use the prep executables at least once  which will throw up a UAC prompt.

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1442,6 +1405,153 @@ pub(crate) fn compute_appcontainer_effective_access(path: &Path) -> Result<u32,
...
+    // the DACL, and asking for more could fail the open spuriously.
+    const READ_CONTROL: u32 = 0x0002_0000;
+
+    // `\\.\NUL` resolves through the I/O manager to `\Device\Null`.
+    let path: Vec<u16> = "\\\\.\\NUL\0".encode_utf16().collect();
+    let share = FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE;
+
+    // SAFETY: standard CreateFileW invocation; every pointer references
+    // local data or is NULL.
+    let handle = unsafe {
```
bbonaby, line 1446: note: you can probably use our RAII wrapper for handles https://github.com/microsoft/mxc/blob/c55dcff29b375d19da7bc880d6a89192394ebae7/src/wxc_common/src/process_util.rs#L55

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1442,6 +1405,153 @@ pub(crate) fn compute_appcontainer_effective_access(path: &Path) -> Result<u32,
...
+        let _ = CloseHandle(handle);
+    }
+    if rc != ERROR_SUCCESS {
+        return None;
+    }
+    if dacl.is_null() {
+        // A NULL DACL grants everyone full access — the device is
+        // reachable by AppContainer, so no prep is needed.
+        unsafe {
+            let _ = LocalFree(Some(HLOCAL(sd.0)));
```
bbonaby, line 1490: suggestion: We should probably see if we can make a RAII wrapper for these too. Sucks there is no WIL in Rust currently :(

src/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1442,6 +1405,153 @@ pub(crate) fn compute_appcontainer_effective_access(path: &Path) -> Result<u32,
...
+            unsafe {
+                let _ = LocalFree(Some(HLOCAL(sd.0)));
+            }
+            return None;
+        }
+    };
+
+    let mut info = ACL_SIZE_INFORMATION::default();
+    let info_sz = std::mem::size_of::<ACL_SIZE_INFORMATION>() as u32;
+    if unsafe {
```
bbonaby, line 1507: note: missing safety statements on functions in this file

PR #457, opened 2026-05-29
Pass STDIO handles directly to child in non-PTY mode

jsidewhite:

-----

## Summary

In pipe mode (SDK non-PTY), forward wxc-exec's own stdin/stdout/stderr to the child process via STARTF_USESTDHANDLES instead of creating intermediate anonymous pipes. Child output now streams to the SDK caller in real time rather than being buffered until exit.

Changes:
- **appcontainer_runner.rs**: Use GetStdHandle + SetHandleInformation + PROC_THREAD_ATTRIBUTE_HANDLE_LIST for direct passthrough in pipe mode
- **base_container_runner.rs**: Same passthrough approach via STARTUPINFOW (inheritHandles stays FALSE per OS API contract - this deviates from regular CreateProcess semantics for explicit STD handle passing)

Closes: #70 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/457)

### DIFF
--- a/src/wxc_common/src/appcontainer_runner.rs
+++ b/src/wxc_common/src/appcontainer_runner.rs
@@ -1,16 +1,21 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use std::io::IsTerminal;
 use std::ptr;
 
 use windows::Win32::Foundation::{
-    GetLastError, LocalFree, ERROR_ALREADY_EXISTS, HLOCAL, WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT,
+    GetLastError, LocalFree, SetHandleInformation, ERROR_ALREADY_EXISTS, HANDLE,
+    HANDLE_FLAG_INHERIT, HLOCAL, WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT,
 };
 use windows::Win32::Security::Authorization::ConvertSidToStringSidW;
 use windows::Win32::Security::Isolation::{
     CreateAppContainerProfile, DeleteAppContainerProfile, DeriveAppContainerSidFromAppContainerName,
 };
 use windows::Win32::Security::{FreeSid, PSID};
+use windows::Win32::System::Console::{
+    GetStdHandle, STD_ERROR_HANDLE, STD_INPUT_HANDLE, STD_OUTPUT_HANDLE,
+};
 use windows::Win32::System::SystemServices::SE_GROUP_ENABLED;
 use windows::Win32::System::Threading::{
     CreateProcessW, DeleteProcThreadAttributeList, GetExitCodeProcess,
@@ -18,8 +23,9 @@ use windows::Win32::System::Threading::{
     WaitForSingleObject, CREATE_SUSPENDED, CREATE_UNICODE_ENVIRONMENT,
     EXTENDED_STARTUPINFO_PRESENT, LPPROC_THREAD_ATTRIBUTE_LIST, PROCESS_CREATION_FLAGS,
     PROCESS_INFORMATION, PROC_THREAD_ATTRIBUTE_ALL_APPLICATION_PACKAGES_POLICY,
-    PROC_THREAD_ATTRIBUTE_MITIGATION_POLICY, PROC_THREAD_ATTRIBUTE_SECURITY_CAPABILITIES,
-    STARTUPINFOEXW, STARTUPINFOW,
+    PROC_THREAD_ATTRIBUTE_HANDLE_LIST, PROC_THREAD_ATTRIBUTE_MITIGATION_POLICY,
+    PROC_THREAD_ATTRIBUTE_SECURITY_CAPABILITIES, STARTF_USESTDHANDLES, STARTUPINFOEXW,
+    STARTUPINFOW,
 };
 use windows_core::{PCWSTR, PWSTR};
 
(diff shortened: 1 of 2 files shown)


src/wxc_common/src/appcontainer_runner.rs
```diff
@@ -522,14 +543,83 @@ impl AppContainerScriptRunner {
...
         let si_ex = STARTUPINFOEXW {
             StartupInfo: STARTUPINFOW {
                 cb: std::mem::size_of::<STARTUPINFOEXW>() as u32,
                 lpDesktop: PWSTR(desktop_wide.as_mut_ptr()),
+                dwFlags: if pipe_mode {
+                    STARTF_USESTDHANDLES
+                } else {
+                    Default::default()
+                },
+                hStdInput: if pipe_mode {
```
MGudgin: Given the else block at lines 590-594, do you need the if/else here, can't you just h_stdin directly (same comment for h_stdout and h_stderr_ #Resolved

src/wxc_common/src/base_container_runner.rs
```diff
@@ -559,9 +567,74 @@ impl ScriptRunner for BaseContainerRunner {
...
+        // STARTUPINFOW -- in pipe mode, pass parent handles via STARTF_USESTDHANDLES
+        // so child output streams directly to the SDK caller.
         let si = STARTUPINFOW {
             cb: std::mem::size_of::<STARTUPINFOW>() as u32,
+            dwFlags: if pipe_mode {
+                STARTF_USESTDHANDLES
+            } else {
+                Default::default()
+            },
+            hStdInput: if pipe_mode {
```
MGudgin: Same question here about whether the if/else is needed.

MGudgin (approved):

src/wxc_common/src/appcontainer_runner.rs
```diff
@@ -128,14 +134,17 @@ struct SecurityCapabilities {
...
-    let mut n = 1;
+fn compute_attr_count(least_privilege_mode: bool, ui_disable: bool, pipe_mode: bool) -> u32 {
+    let mut n = 1; // SECURITY_CAPABILITIES always present
     if least_privilege_mode {
         n += 1;
     }
     if ui_disable {
         n += 1;
     }
+    if pipe_mode {
```
caarlos0, line 145: if you get `2`, how do you know which attrs were set? - and do we need to know?

src/wxc_common/src/appcontainer_runner.rs
jsidewhite: checking twice adds extra certainty.. /s

src/wxc_common/src/appcontainer_runner.rs
jsidewhite: This `+1` is just saying we're going to execute the Update call on line 577..

```c
                UpdateProcThreadAttribute(
                    attr_list,
                    0,
                    PROC_THREAD_ATTRIBUTE_HANDLE_LIST as usize,
                    Some(handle_list.as_ptr() as *const core::ffi::c_void),
                    handle_list.len() * std::mem::size_of::<HANDLE>(),
                    None,
                    None,
                )
```

..to pass in a list of handles.  that list size itself is irrelevant (could be 1, 2, or 3 handles).

But the +1 above is just saying "allocate a spot for the list" - not the list size

MGudgin (approved):

src/wxc_common/src/appcontainer_runner.rs
```diff
@@ -522,14 +543,71 @@ impl AppContainerScriptRunner {
...
+            h_stdout = unsafe { GetStdHandle(STD_OUTPUT_HANDLE) }
+                .map_err(|e| WxcError::Process(format!("GetStdHandle(STDOUT): {e}")))?;
+            h_stderr = unsafe { GetStdHandle(STD_ERROR_HANDLE) }
+                .map_err(|e| WxcError::Process(format!("GetStdHandle(STDERR): {e}")))?;
+
+            // Ensure the handles are inheritable.
+            unsafe {
+                let _ = SetHandleInformation(h_stdin, HANDLE_FLAG_INHERIT.0, HANDLE_FLAG_INHERIT);
+                let _ = SetHandleInformation(h_stdout, HANDLE_FLAG_INHERIT.0, HANDLE_FLAG_INHERIT);
+                let _ = SetHandleInformation(h_stderr, HANDLE_FLAG_INHERIT.0, HANDLE_FLAG_INHERIT);
```
asklar: do we need to emit an error here too if return value is 0?

src/wxc_common/src/base_container_runner.rs
```diff
@@ -634,7 +695,7 @@ impl ScriptRunner for BaseContainerRunner {
                     cmd_wide.as_mut_ptr(),   // commandLine
                     ptr::null(),             // processAttributes (must be NULL)
                     ptr::null(),             // threadAttributes  (must be NULL)
-                    0,                       // inheritHandles    (must be FALSE)
+                    i32::from(false),        // inheritHandles (must be FALSE)
```
asklar: the comment could describe how inheritHandles interacts (or doesn't) with the handles passed in the startupinfo struct (the fix for appcontainer passes inherit = true, why is this one ok to pass false to?)

asklar (approved):

src/wxc_common/src/appcontainer_runner.rs
asklar: i had the same thought, consider adding a comment that says the +1 is "there will be a list", not the list size :)

MGudgin (approved):

caarlos0 (approved):
just validated it through copilot on windows, works great.

Image

Image

src/wxc_common/src/appcontainer_runner.rs
caarlos0: ahh ok, that makes sense. Agreed a comment would help yes 👍🏻

src/wxc_common/src/appcontainer_runner.rs
```diff
@@ -522,14 +543,90 @@ impl AppContainerScriptRunner {
...
+        let h_stdout;
+        let h_stderr;
+
+        if pipe_mode {
+            h_stdin = unsafe { GetStdHandle(STD_INPUT_HANDLE) }
+                .map_err(|e| WxcError::Process(format!("GetStdHandle(STDIN): {e}")))?;
+            h_stdout = unsafe { GetStdHandle(STD_OUTPUT_HANDLE) }
+                .map_err(|e| WxcError::Process(format!("GetStdHandle(STDOUT): {e}")))?;
+            h_stderr = unsafe { GetStdHandle(STD_ERROR_HANDLE) }
+                .map_err(|e| WxcError::Process(format!("GetStdHandle(STDERR): {e}")))?;
```
bbonaby, line 562: thought: I wonder if in the future we can use https://github.com/microsoft/mxc/blob/04236c9e6de941ab4c39acd84e219787347c22fb/src/wxc_common/src/process_util.rs#L184 . Or create a different but common function in that module. Looks like the module is a common module made for us to share this kind of stuff, but just wasn't used in this file directly.

bbonaby (approved):

PR #468, opened 2026-06-01
Gate build-dependent JOB_OBJECT_UILIMIT_* flags via a capability model

MGudgin:
This PR reshapes the UI-restriction handling so the set of enforceable
JOB_OBJECT_UILIMIT_* flags is treated as a host capability that varies by
OS build, rather than a per-exec degradation event. Two flags are
build-gated: JOB_OBJECT_UILIMIT_IME (0x100), accepted on Windows 11 22H2
(22621) and later but rejected on earlier builds (e.g. Server 2022,
20348); and JOB_OBJECT_UILIMIT_INJECTION (0x200), which requires Windows
build 26100 (24H2). On builds below each threshold the corresponding flag
is excluded from the enforceable mask while every other requested UI
restriction still applies. The effective mask is always
requested & supported(build).

Details

* job_object.rs: add ALL_DEFINED_UI_LIMITS (0x03FF); expose pub
  os_build_number(), input_injection_blocking_supported(), and
  supported_ui_limit_mask(); add supported_ui_limit_mask_for_build().
  set_ui_limits() returns Result<(), WxcError> and applies
  requested & supported_ui_limit_mask() instead of stripping-and-warning.
* Build-gate both extended flags: MIN_BUILD_FOR_IME_LIMIT (22621) drops
  IME below 22H2; MIN_BUILD_FOR_INJECTION_LIMIT (26100) drops injection
  below 24H2. UI-limit support is monotonic, so gating at these
  boundaries never hands the kernel a flag it would reject. This fixes
  the all-flags set_ui_limits failure on pre-22H2 CI runners.
* Fail secure on an undetermined build: os_build_number() returns
  u32::MAX when RtlGetVersion can't be resolved, which keeps the more
  restrictive flags rather than silently dropping them.

Tests

* cargo test -p appcontainer_common (release, x86_64-pc-windows-msvc):
  117 passed, including new supported_mask_strips_ime_on_pre_22h2
  (20348 -> 0x00FF), supported_mask_keeps_ime_on_22h2 (22621),
  supported_mask_strips_injection_on_downlevel (22631 -> 0x01FF),
  supported_mask_keeps_injection_on_uplevel (26100 -> 0x03FF), and
  supported_mask_fails_secure_on_unknown_build. set_ui_limits all-flags
  test verified passing on 23H2 (22631) and 25H2 machines.
* cargo fmt --all --check, clippy -p appcontainer_common
  --all-targets -D warnings, and cargo check -p wxc
  --features "hyperlight isolation_session microvm wslc" all clean.

### DIFF
--- a/src/backends/appcontainer/common/src/job_object.rs
+++ b/src/backends/appcontainer/common/src/job_object.rs
@@ -14,6 +14,7 @@
 
 use core::ffi::c_void;
 use std::mem::size_of;
+use std::sync::OnceLock;
 
 use windows::Win32::Foundation::{CloseHandle, HANDLE};
 use windows::Win32::System::JobObjects::{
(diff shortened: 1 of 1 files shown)


src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -29,11 +30,80 @@ use windows_core::PCWSTR;
...
+/// lifetime.
+fn os_build_number() -> u32 {
+    *OS_BUILD_NUMBER.get_or_init(version_detect::get_os_build_number)
+}
+
+/// Returns `true` when the current OS build supports the
+/// `JOB_OBJECT_UILIMIT_INJECTION` flag in `SetInformationJobObject`.
+fn supports_injection_limit() -> bool {
+    os_build_number() >= MIN_BUILD_FOR_INJECTION_LIMIT
+}
```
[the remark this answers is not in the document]
MGudgin: Leaving the flag in *is* the safe option.

src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -91,9 +161,31 @@ impl UiJobObject {
...
+        // JOB_OBJECT_UILIMIT_INJECTION is only supported on build 26100+.
+        // On downlevel hosts, strip the bit and warn the caller so it appears
+        // in diagnostic output.
+        if (mask & JOB_OBJECT_UILIMIT_INJECTION) != 0 && !supports_injection_limit() {
+            mask &= !JOB_OBJECT_UILIMIT_INJECTION;
+            warnings.push(format!(
+                "JOB_OBJECT_UILIMIT_INJECTION (0x200) is not supported on this OS build \
+                 ({}); stripping flag — input-injection blocking is not enforced",
+                os_build_number(),
+            ));
```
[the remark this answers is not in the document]
MGudgin: Incorrect analysis.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -795,17 +795,20 @@ impl AppContainerScriptRunner {
...
             job.assign_process(process_handle.get())?;
-            Ok(job)
+            Ok((job, warnings))
         })() {
-            Ok(job) => {
+            Ok((job, warnings)) => {
+                for w in &warnings {
+                    logger.log_line(&format!("warning: {w}"));
+                }
                 logger.log_line("UI Job Object assigned to child process");
```
jsidewhite, line 809: >UI Job Object assigned to child process [](http://example.com/codeflow?start=33&length=39)

:P this sentence feels backwards

jsidewhite:
>             SetInformationJobObject(

why not just allow SIJO to fail with `ERROR_INVALID_PARAMETER` and then when handling error, call `supports_injection_limit()`.

then,
1. if support ever gets sent downlevel, this will light up automatically
2. we don't pay the cost of loading ntdll.dll for every sandbox on an upstream build

---
Refers to: src/backends/appcontainer/common/src/job_object.rs:194 in af2f8d3. [](commit_id = af2f8d359da7f99939a30470340a4c483f388d35, deletion_comment = False)

alexsnitkovskiy (approved):

src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -29,11 +30,80 @@ use windows_core::PCWSTR;
...
+            let rtl_get_version: RtlGetVersionFn = std::mem::transmute(proc);
+            let mut info = OSVERSIONINFOW {
+                dwOSVersionInfoSize: size_of::<OSVERSIONINFOW>() as u32,
+                ..Default::default()
+            };
+            let status = rtl_get_version(&mut info);
+            if status.is_ok() {
+                info.dwBuildNumber
+            } else {
+                u32::MAX
```
bbonaby, line 78: question: Should we actually just print and return an error  if `rtl_get_version` fails? This way at least instead of the eventual "paramater invalid` message that would happen if the flag isn't supported on the machine the error users will get will be specific. I think this would also help with our diagnostics as well.

src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -91,9 +161,31 @@ impl UiJobObject {
...
+    ) -> Result<Vec<String>, WxcError> {
+        let mut warnings: Vec<String> = Vec::new();
+        let mut mask = to_job_object_uilimit_mask(restrictions);
+
+        // JOB_OBJECT_UILIMIT_INJECTION is only supported on build 26100+.
+        // On downlevel hosts, strip the bit and warn the caller so it appears
+        // in diagnostic output.
+        if (mask & JOB_OBJECT_UILIMIT_INJECTION) != 0 && !supports_injection_limit() {
+            mask &= !JOB_OBJECT_UILIMIT_INJECTION;
+            warnings.push(format!(
```
bbonaby: question: would the right move here be to return an error instead of a warning? To me if I'm GitHub copilot cli I would rather the tech I'm using not continue, if the config I passed into it will not use it fully. As a consumer, wouldn't they rather receive the error and then update their config accordingly now that they know the specific security mechanism is not available to them?

jsidewhite (approved):
:shipit:

MGudgin:
@jsidewhite, @bbonaby I've removed the warning propagation from this PR. We now just have an overall mask we apply for ui.disable=true and that mask differs on Ni vs Ge. I'll add warnings to the --probe path in a follow-up PR tomorrow (and probably also surface a warning/error if someone explicitly tries to block input injection on Ni)

bbonaby (approved):

bbonaby (approved):

PR #469, opened 2026-06-01
fix: route schema 0.4.0 through the tier dispatcher

jsidewhite:

-----
## Issue
0.4.0 schemas always choose Tier 2 - even when it's not compiled in.

## Summary

Schema versions below 0.5.0 previously bypassed the fallback dispatcher and constructed a bare `AppContainerScriptRunner` with the default `FilesystemMode::Bfs`. Without the `tier2_bfs` feature compiled in, `bfscfg.exe` could never be resolved, causing a hard error on any config that declared filesystem policy paths.

Now all `ProcessContainer` requests go through `dispatch_with_fallback` with an `allow_base_container` flag derived from schema version / `--experimental`. Legacy schemas pass `false` so the dispatcher skips Tier 1 (BaseContainer) and correctly falls through to T2 (BFS, if compiled) or T3 (DACL).

Also refactors the dispatcher API: `dispatch_with_fallback` now takes an explicit `allow_base_container: bool` parameter instead of hardcoding `true`, eliminating the need for a separate `dispatch_appcontainer_only` function.

### DIFF
--- a/src/core/wxc/src/main.rs
+++ b/src/core/wxc/src/main.rs
@@ -7,9 +7,7 @@ use std::process;
 use std::sync::{Mutex, OnceLock};
 use std::time::Instant;
 
-use appcontainer_common::appcontainer_runner::{
-    delete_app_container_profile, AppContainerScriptRunner,
-};
+use appcontainer_common::appcontainer_runner::delete_app_container_profile;
 use clap::Parser;
 #[cfg(all(feature = "hyperlight", target_arch = "x86_64"))]
 use hyperlight_common::HyperlightScriptRunner;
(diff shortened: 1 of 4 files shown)


src/core/wxc/src/main.rs
```diff
@@ -806,54 +804,61 @@ fn main() {
...
                         }
-                        eprint!("{}", logger.get_buffer());
-                        process::exit(1);
                     }
+                    eprint!("{}", logger.get_buffer());
+                    process::exit(1);
                 }
-            } else {
-                Box::new(AppContainerScriptRunner::new())
             }
```
bbonaby, line 857: issue: I think if I'm reading this right `AppContainerScriptRunner` would never be called due to this removal. From what I see in the before PR case `ProcessContainer` == (`Appcontainer` **OR** `BaseContainer`) based on if BaseContainer is available on the machine or not. Let me know if I'm just reading this wrong, (using GitHub UI after all)

jsidewhite: i think yo'ure reading wrongly - Gudge's dispatcher a few lines up will select AC runner (internally) in the Tier logic

PR #482, opened 2026-06-03
fix(bwrap): default to deny-by-default filesystem (mirror seatbelt)

caarlos0:
## 📖 Description

Change the Bubblewrap backend's default filesystem posture from "host root mounted read-only" to **deny-by-default**, matching the macOS Seatbelt backend's `(deny default)` baseline.

### Why

`bwrap_command::build_args` used to emit:
```rust
args.extend(["--ro-bind".into(), "/".into(), "/".into()]);
```
That bind-mounted the entire host root read-only into every sandbox, so the caller's `$HOME/.aws/credentials`, `$HOME/.ssh/id_*`, browser cookies, etc. were readable inside the sandbox by default. The Seatbelt backend on macOS starts from `(deny default)` and only allows narrow system paths (`SYSTEM_READ_ALLOW` in `src/backends/seatbelt/common/src/profile_builder.rs`), so the two backends had a meaningful asymmetry in the confidentiality guarantees they offered. This PR closes that gap.

### What changes

**New baseline (`BASELINE_RO_BIND_PATHS`)** — mirrors seatbelt's `SYSTEM_READ_ALLOW`:

- Top-level dirs: `/bin`, `/sbin`, `/lib`, `/lib32`, `/lib64`, `/libx32` (symlinks under `/usr` on merged-usr distros; bwrap follows source-side symlinks so both real-dir and symlinked distros work).
- `/usr` subpaths: `/usr/bin`, `/usr/sbin`, `/usr/lib`, `/usr/lib32`, `/usr/lib64`, `/usr/libexec`, `/usr/share` — deliberately **not** `/usr` wholesale, so `/usr/local` is not implicitly exposed.
- `/etc` — whole, like seatbelt's `/private/etc`. Files with restrictive perms (`/etc/shadow`, `/etc/sudoers`, `/etc/ssh/ssh_host_*_key`) stay unreadable to a non-root caller because user-namespace UID mapping does not bypass kernel DAC.
- DNS stub-resolver dirs: `/run/systemd/resolve`, `/run/NetworkManager`, `/run/resolvconf` — needed when `/etc/resolv.conf` is a symlink. Narrow subpaths so `/run/user/<uid>` (D-Bus session, keyring, ssh-agent sockets) stays hidden.
- `/etc/resolv.conf` symlinks **outside** `/run`: also synthesise a `/var/run -> /run` compat symlink (for `/var/run/...`-routed targets — older RHEL/CentOS-era and some container images) and `--ro-bind-try` `/mnt/wsl/resolv.conf` (for WSL), so DNS keeps working under deny-by-default without exposing host `/var` or `/mnt` contents.

All emitted via `--ro-bind-try` so missing paths are silently skipped (e.g. `/lib32` on x86_64-only systems, `/run/systemd/resolve` on hosts without systemd-resolved).

### What disappears from sandbox by default

`$HOME`, `/root`, `/home/*`, `/opt`, `/srv`, `/mnt`, `/media`, `/var`, `/sys`, `/usr/local`, `/run/user/<uid>`, `/run/dbus`. Callers who legitimately need any of these must list them under `readonlyPaths` or `readwritePaths`.

### What's preserved

- `readwritePaths` / `readonlyPaths` / `deniedPaths` semantics — unchanged.
- `--unshare-*` flags, network policy handling, proxy env-var injection, working-dir, env clearing — unchanged.
- Standard `--dev /dev` / `--proc /proc` / `--tmpfs /tmp` overlay — unchanged.

### Drive-by build fix

The second commit (`fix(nanvix): compile as build-dep from non-Linux/Windows hosts`) adds empty/zero fallbacks for `REQUIRED_BINARIES` and `NANVIXD_BINARY` so `nanvix_common` compiles on macOS hosts when pulled in as a `[build-dependency]` of `lxc` / `wxc` during cross-compile. Zero runtime impact on supported platforms — the consuming build scripts already gate the surrounding logic behind `cfg(target_os = "linux"/"windows")` and `feature = "microvm"`. Separated out so it can be reviewed (or split into its own PR) independently.

### Breaking change for users

This is a behavior change. Configs that implicitly relied on `$HOME` (or `/opt`, `/var`, `/usr/local`, …) being readable will start failing. The migration is to list the directory in `readonlyPaths`:

```json
{
  "filesystem": {
    "readonlyPaths": ["/home/alice/project", "/usr/local"]
  }
}
```

Documented in the updated "How It Works → Deny-by-default filesystem" and "Limitations" sections of `docs/bwrap-support/bubblewrap-backend.md`.

## 🔗 References

No tracking issue — this came out of a direct comparison between the seatbelt and bwrap baselines while reviewing the two unprivileged backends.

Related follow-up (out of scope for this PR):

- microsoft/mxc#507 — readable preflight errors when a path isn't in the deny-by-default baseline (raised in review).

## 🔍 Validation

**Unit tests** (`cargo test -p bwrap_common` from `src/`) — 25/25 pass, including new regression tests covering the new contract:

- `baseline_does_not_bind_mount_host_root` — regression test for the old `--ro-bind / /` default.
- `baseline_emits_required_ro_bind_try_paths` — `/bin`, `/sbin`, `/lib`, `/lib64`, `/usr/bin`, `/usr/lib`, `/usr/share`, `/etc` all emitted.
- `baseline_does_not_expose_usr_local` — no `--ro-bind /usr /usr` and no explicit `/usr/local` entry.
- `baseline_excludes_confidential_paths` — no `/home`, `/root`, `/opt`, `/srv`, `/var`, `/sys`, `/run/user`, `/run/dbus` bind-mounts.
- `baseline_includes_dns_stub_resolver_dirs` — all three DNS dirs emitted via `--ro-bind-try`.
- `baseline_mounts_precede_policy_mounts` — policy mounts can still shadow baseline.
- `baseline_recreates_var_run_compat_symlink` — emits `--symlink /run /var/run` (and never binds host `/var`) so `/var/run/...`-routed `resolv.conf` symlinks resolve.
- `baseline_includes_wsl_resolv_conf` — emits `--ro-bind-try /mnt/wsl/resolv.conf` (and never exposes `/mnt` wholesale) so WSL DNS works.

Plus updated `filesystem_policy_produces_correct_mounts` to match the new contract (a bare `--ro-bind /data /data` is now unambiguously the policy mount).

**Lint / format** — `cargo clippy -p bwrap_common --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean.

**Bubblewrap behavior** — verified empirically against bwrap 0.8.0 that the `/var/run -> /run` symlink makes `/var/run/NetworkManager/resolv.conf` resolve into the bound `/run/NetworkManager`, that a WSL-style `/etc/resolv.conf -> /mnt/wsl/resolv.conf` is readable, and that the control case (no symlink) fails — reproducing the original gap.

**Linux VM verification** — cross-compiled `lxc-exec` for `aarch64-unknown-linux-gnu` and ran a 6-config smoke suite on a Linux VM (see `src/target/vm-test-bundle/` locally — gitignored). The suite plants `TOP_SECRET=hunter2` in `/home/SENTINEL_DO_NOT_LEAK.txt` on the host and verifies the secret does **not** appear in sandbox output without an explicit `readonlyPaths: ["/home"]`, then verifies the opt-in *does* expose it. Also covers `/opt`/`/var`/`/sys`/`/root`/`/usr/local` being hidden, DNS resolution working with network allowed, and `/etc/shadow` staying unreadable via DAC. *(Will paste the run output as a PR comment once the VM run is complete.)*

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/482)


docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -63,6 +68,41 @@ Bubblewrap creates a namespace-isolated process by:
...
+  not readable from the sandbox.
+- `/opt` and `/usr/local` tooling is not on PATH; list either path under
+  `readonlyPaths` if the script depends on it.
+- `working_directory` must live under the baseline or a policy path — a
+  `cwd` of `~/project` without a matching `readonlyPaths` entry will fail.
+- DNS works on systemd-resolved, NetworkManager, and resolvconf hosts
+  because the corresponding `/run/...` directories are bound. Hosts where
+  `/etc/resolv.conf` symlinks somewhere else need that target listed in
+  `readonlyPaths`.
+
```
bbonaby, line 107: question: Should we return readable errors in these cases if we know beforehand that these things may not work? Also it'd probably be worth it to create a GitHub issue if you think these can be resolved at some point in the future

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -63,6 +68,41 @@ Bubblewrap creates a namespace-isolated process by:
...
+Files in `/etc` that contain secrets (`/etc/shadow`, `/etc/sudoers`,
```
bbonaby, line 108: note: same question as above

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -22,6 +22,69 @@ const PROXY_ENV_KEYS: &[&str] = &[
...
+///   ssh-agent socket. We only bind the well-known DNS stub-resolver
+///   directories so name resolution still works when `/etc/resolv.conf`
+///   is a symlink (the default on systemd-resolved hosts).
+/// - `/etc` is bound whole because cherry-picking files (`passwd`,
+///   `nsswitch.conf`, `ssl/`, `ld.so.conf*`, …) is fragile and breaks
+///   tools that read other config files. Files with sensitive contents
+///   (`/etc/shadow`, `/etc/sudoers`, `/etc/ssh/ssh_host_*_key`) are mode
+///   `0400` / `0640` root and remain unreadable to a non-root caller —
+///   user-namespace UID mapping does not bypass kernel DAC.
+const BASELINE_RO_BIND_PATHS: &[&str] = &[
```
bbonaby, line 64: question: Is the addition of this a breaking change for what we have now? E.g These were suppose to be default deny  but they were not prior to this change. We'd want to know so we can call this out via SDK versioning.

bbonaby: Oh wait, I may have read this wrong initially. It looks like you want to allow these read only paths and disallow everything else unless specified right? If so, then I think we're good here.

docs/bwrap-support/bubblewrap-backend.md
caarlos0: Good call — I've filed microsoft/mxc#507 to add readable preflight errors for these cases (validating `working_directory` against the baseline up front, and detecting an out-of-baseline `/etc/resolv.conf` symlink target). Keeping it out of this PR so the behavior change stays focused.

docs/bwrap-support/bubblewrap-backend.md
caarlos0: Same — tracked in microsoft/mxc#507 (the `/etc` secret files stay unreadable via DAC by design, but where we can predict a failure we should surface a clear error). 👍

src/backends/bubblewrap/common/src/bwrap_command.rs
caarlos0: Right, exactly — allow this narrow read-only baseline and deny everything else unless the caller opts in via `readonlyPaths`/`readwritePaths`. The previous default bind-mounted all of host `/` read-only, so this does tighten the default; the PR description calls it out as a behavior change for SDK versioning.

src/backends/nanvix/common/src/lib.rs
```diff
@@ -28,6 +38,10 @@ pub const NANVIXD_BINARY: &str = "nanvixd.exe";
 #[cfg(target_os = "linux")]
 pub const NANVIXD_BINARY: &str = "nanvixd.elf";
 
+/// Fallback for non-Windows/Linux hosts. See `REQUIRED_BINARIES` above.
+#[cfg(not(any(target_os = "windows", target_os = "linux")))]
+pub const NANVIXD_BINARY: &str = "";
```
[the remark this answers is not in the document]
caarlos0: Fixed in 69cd0f7 — `NANVIXD_BINARY` on unsupported hosts is now the descriptive sentinel `"nanvixd-unsupported-host"` instead of `""`, so any accidental `Command` construction fails with a named program. (This const is only consumed by Linux/Windows-gated build scripts, so it's compile-time-only in practice.)

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -62,13 +125,15 @@ pub fn build_args(request: &ExecutionRequest, proxy_address: Option<&ProxyAddres
...
-    // always win, including when policy paths overlap a standard mount such
-    // as `/tmp` (e.g. `readwritePaths: ["/tmp/workspace"]`).
-    args.extend(["--ro-bind".into(), "/".into(), "/".into()]);
+    // earlier ones. We therefore lay the baseline + standard virtual
+    // filesystems down first, then apply user-supplied policy mounts last
+    // so they always win when paths overlap (e.g. `readwritePaths:
+    // ["/tmp/workspace"]` must beat the standard `--tmpfs /tmp`).
+    for path in BASELINE_RO_BIND_PATHS {
+        args.extend(["--ro-bind-try".into(), (*path).into(), (*path).into()]);
+    }
```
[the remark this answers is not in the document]
caarlos0: Documented in 69cd0f7 — added a note to the bwrap docs Prerequisites that the deny-by-default baseline needs bwrap 0.3.0+ for `--ro-bind-try` (released 2017; every supported distro ships newer). Kept `--ro-bind-try` rather than `--ro-bind` because the baseline intentionally tolerates absent paths (e.g. `/lib32` on x86_64-only, `/run/systemd/resolve` on non-systemd hosts).

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -206,17 +271,13 @@ mod tests {
...
-        assert!(ro_positions.len() >= 2);
-        let data_pos = *ro_positions.last().unwrap();
-        assert_eq!(args[data_pos + 1], "/data");
+        // ro — baseline paths are emitted via --ro-bind-try, so a bare
+        // --ro-bind must correspond to the user's readonlyPaths entry.
+        let ro_pos = args
+            .windows(3)
+            .position(|w| w[0] == "--ro-bind" && w[1] == "/data" && w[2] == "/data")
+            .expect("readonly policy path /data should produce a --ro-bind mount");
+        assert!(ro_pos > 0);
```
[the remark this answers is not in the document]
caarlos0: Fixed in 69cd0f7 — dropped `assert!(ro_pos > 0)`; the preceding `.expect(...)` already guarantees the mount exists, so the assertion added nothing and was order-fragile.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -401,4 +462,156 @@ mod tests {
...
+        assert!(
+            !usr_whole,
+            "baseline must bind /usr subpaths individually so /usr/local is \
+             not implicitly exposed; got: {:?}",
+            args
+        );
+        // And no explicit /usr/local entry either.
+        let usr_local = args.iter().any(|a| a == "/usr/local");
+        assert!(!usr_local, "baseline must not expose /usr/local by default");
+    }
```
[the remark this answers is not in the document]
caarlos0: Fixed in 69cd0f7 — the `/usr/local` check now scans only mount-argument windows (`--bind`/`--ro-bind`/`--ro-bind-try` with source `/usr/local`), so a script body mentioning the path can't trigger a false positive.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -22,6 +22,69 @@ const PROXY_ENV_KEYS: &[&str] = &[
...
+    // Top-level executable / library dirs (symlinks under /usr on
+    // merged-usr distros, real directories on Alpine and older Debian).
+    "/bin",
+    "/sbin",
+    "/lib",
+    "/lib32",
+    "/lib64",
+    "/libx32",
+    // /usr subpaths — mirrors seatbelt's baseline exactly, intentionally
+    // excluding /usr/local.
```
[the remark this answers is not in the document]
caarlos0: Fixed in 69cd0f7 — softened the comment from "mirrors seatbelt's baseline exactly" to "aligned with" to avoid implying exact, lasting parity (the Linux `/run/...` DNS dirs already differ).

bbonaby (approved):

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -22,6 +22,69 @@ const PROXY_ENV_KEYS: &[&str] = &[
...
+    // System configuration (ld.so config, certs, resolv.conf, hosts,
+    // passwd, group, machine-id, …). See module-level note on DAC.
+    "/etc",
+    // DNS stub-resolver directories. /etc/resolv.conf is usually a
+    // symlink into one of these on modern Linux distros (systemd-resolved
+    // / NetworkManager / resolvconf). We bind the narrow subdirectories
+    // rather than all of /run to avoid exposing /run/user/<uid>.
+    "/run/systemd/resolve",
+    "/run/NetworkManager",
+    "/run/resolvconf",
```
MGudgin, line 91: ### ⚠️ Deny-by-default can silently break DNS when `/etc/resolv.conf` is a symlink to a path outside the baseline

**Severity:** Medium (the failure is *silent* and cryptic, which is what keeps it from Low)

**Mechanism**

`/etc` is bound whole, so a `/etc/resolv.conf` *symlink* is preserved as a symlink — bwrap doesn't dereference it. When a sandboxed process opens it, the kernel resolves the target **inside the sandbox mount namespace**, where only the baseline paths exist. Because `/var` and `/mnt` are never mounted, any resolv.conf target routed through them is a dangling link → glibc falls back to `nameserver 127.0.0.1` (usually nothing listening) → name resolution fails with no obvious cause. The sandbox still starts successfully, so this surfaces as a confusing "DNS doesn't work" rather than a clear error.

A subtle trap: binding the canonical `/run/NetworkManager` here does **not** rescue a link written as `/var/run/NetworkManager/resolv.conf`, because the intermediate `/var/run` compat symlink lives in the unmounted `/var`.

**What actually works vs. breaks**

The common modern configs are fine (the baseline was clearly designed around them):

| Config | `/etc/resolv.conf` | Result |
|---|---|---|
| systemd-resolved (Ubuntu 18.04+, most cloud images) | → `/run/systemd/resolve/stub-resolv.conf` | ✅ works (`/run/systemd/resolve` bound) |
| resolvconf | → `/run/resolvconf/resolv.conf` | ✅ works (`/run/resolvconf` bound) |
| NetworkManager default mode | real file written into `/etc` | ✅ works (inside bound `/etc`) |
| NM symlink mode, canonical | → `/run/NetworkManager/resolv.conf` | ✅ works (`/run/NetworkManager` bound) |
| static `/etc/resolv.conf` | real file | ✅ works |
| **WSL** | → `/mnt/wsl/resolv.conf` | ❌ breaks (`/mnt` unmounted) |
| **`/var/run`-routed** (older RHEL/CentOS-era) | → `/var/run/NetworkManager/resolv.conf` | ❌ breaks (`/var` unmounted) |
| **admin-custom target** outside the 17 baseline paths | → e.g. `/etc/dns/resolv.conf` on a non-bound mount | ❌ breaks |

So this is a real but **minority** break, not the broad regression it might first appear — the dominant systemd-resolved / resolvconf / NM-default paths are all covered. Worth noting the PR's VM smoke test ("DNS working with network allowed") almost certainly ran on a systemd-resolved host, i.e. the case that *does* work, which is why the gap wasn't surfaced.

**Suggested fix** (cheap, and worth doing regardless):

- Preferred: at runtime, `readlink` `/etc/resolv.conf` and `--ro-bind-try` its canonical target before emitting the baseline.
- Or statically: add `--symlink run /var/run` (rescues the entire `/var/run/...` family without exposing `/var`'s contents) plus `/mnt/wsl/resolv.conf` to the baseline.

Either way, a regression test that points `/etc/resolv.conf` at a `/var/run/...` target and asserts the resolved target ends up reachable would lock this in.

MGudgin (changes requested):

src/backends/bubblewrap/common/src/bwrap_command.rs
caarlos0: Fixed in 9ae2486 — went with the static option since `build_args` is deliberately pure/platform-agnostic and unit-testable on every host:

- **`/var/run/...`-routed targets:** synthesise a `--symlink /run /var/run` compat symlink. `/var/run/NetworkManager/resolv.conf` (and the rest of the `/var/run/...` family) now resolve into the already-bound `/run/*` DNS dirs. bwrap synthesises an empty `/var` for the symlink, so no host `/var` contents are exposed.
- **WSL:** `--ro-bind-try /mnt/wsl/resolv.conf` (single file, skipped on non-WSL hosts — no `/mnt` exposure).

Added two regression tests (`baseline_recreates_var_run_compat_symlink`, `baseline_includes_wsl_resolv_conf`) and updated the backend docs. Verified the symlink + bind resolution empirically against bwrap 0.8.0 (and confirmed the control case fails without the symlink).

Truly custom out-of-baseline targets (e.g. an admin-set `/etc/dns/resolv.conf` on an unbound mount) still need a `readonlyPaths` entry; that residual caveat is now documented.

MGudgin (approved):

PR #485, opened 2026-06-03
Fix stale crate paths in .fbs code gen

MGudgin:
The regenerate.ps1 script and README referenced src\generated\... but the crate lives at src\core\generated\base_container_specification. Update both to the correct path.

Microsoft reviewers: PR builds don't auto-run (ADO policy). Comment `/azp run`
to start `MXC-PR-Build`. See [docs/pull-requests.md](../docs/pull-requests.md).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/485)

PR #487, opened 2026-06-03
Make .fbs regen script path-robust and simplify README

MGudgin:
## Summary

Follow-up to #485 addressing review feedback on the FlatBuffers regeneration tooling.

This PR changes the `.fbs` regeneration script to be path-robust and simplifies the README so the manual steps are no longer duplicated.

Details

* `regenerate.ps1`: derive the crate dir from `$PSScriptRoot` instead of a hardcoded repo-relative string, so the path can never go stale if the crate moves (the class of bug #485 fixed).
* `README.md`: replace the duplicated manual flatc/reorg/patch/fmt steps with a single call to `regenerate.ps1`, which already performs all of them.

Tests

* Ran `regenerate.ps1` (flatc 25.12.19) from both the repo root and `src/`; the regenerated output is byte-identical to the committed bindings (LF/CRLF only) and `cargo check -p sandbox_spec` passes.

## Related Issues

- #485

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/487)

### DIFF
--- a/src/core/generated/base_container_specification/README.md
+++ b/src/core/generated/base_container_specification/README.md
@@ -9,30 +9,12 @@ The `base_container_specification` crate contains Rust bindings auto-generated f
 
 ## Steps
 
-From the repo root, run in PowerShell:
+From the repo root, run the regeneration script in PowerShell:
 
 ```powershell
-# Clean old generated output
-Remove-Item src\core\generated\base_container_specification\src\ -Recurse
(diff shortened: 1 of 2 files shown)


jsidewhite (approved):
:shipit:

PR #490, opened 2026-06-03
Surface portable UI capability facts via --probe

MGudgin:
This PR changes the platform probe and SDK support API to expose portable UI
capability facts instead of Windows-specific UI-limit details.

Details

* Emit a uiCapabilities object from the Rust probe that maps native Windows
  UI-limit support into platform-neutral restriction capabilities.
* Expose PlatformSupport.uiCapabilities and UiCapabilitySupport in the SDK, with
  TypeScript validation for probe output.
* Cover missing input-injection and input-method support cases in Rust and SDK
  tests.

Tests

* npm test in sdk/ passed.
* npm run build in sdk/ passed.
* cargo fmt --all -- --check in src/ passed.
* cargo test -p appcontainer_common probe in src/ passed.
* cargo test -p appcontainer_common supported in src/ passed.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/490)

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -79,6 +79,8 @@ The default `processcontainer`, `bubblewrap`, and `lxc` backends work out of the
 
 > **Hyperlight** is an opt-in build flavor (Linux x64 and Windows x64) gated by the `--with-hyperlight` cargo feature. Default shipped binaries do not include it; build from source with `build.bat --with-hyperlight` (Windows) or the equivalent cargo invocation on Linux.
 
+`getPlatformSupport()` reports backend availability and, when the native probe can determine it, `uiCapabilities`: a platform-neutral view of which UI restrictions the host can enforce. This is currently populated only by the Windows native probe, where it is derived from `JOB_OBJECT_UILIMIT_*` support; Linux and macOS omit the field until their probes expose equivalent data.
+
 **Node.js:** ≥ 18.
 
 ---
@@ -363,6 +365,9 @@ getPlatformSupport() → PlatformSupport
 getAvailableToolsPolicy(env?, options?) → FilesystemPolicyResult
 getUserProfilePolicy()                  → FilesystemPolicyResult
 getTemporaryFilesPolicy(env?)           → FilesystemPolicyResult
+
+// Capability types
+UiCapabilitySupport
 
 // Errors (typed wire-format errors from wxc-exec)
 ErrorCode, MxcError, mxcErrorFromCode(code)
--- a/sdk/src/types.ts
+++ b/sdk/src/types.ts
@@ -408,6 +408,39 @@ export type IsolationTier =
   | 'appcontainer-dacl';
 
 /**
+ * Host support for enforcing sandbox UI restrictions.
+ *
+ * The fields describe platform-agnostic restriction intents, not the
+ * OS-specific primitive used to enforce them. For example, Windows derives
+ * these values from `JOB_OBJECT_UILIMIT_*` support. The SDK currently
+ * receives this object only from the Windows native probe; other platforms
+ * omit `PlatformSupport.uiCapabilities` until they expose equivalent probe
+ * data.
+ */
+export interface UiCapabilitySupport {
+  /** Whether the host can block reads from the clipboard. */
+  canBlockClipboardRead: boolean;
+  /** Whether the host can block writes to the clipboard. */
+  canBlockClipboardWrite: boolean;
+  /** Whether the host can block synthetic keyboard/mouse input. */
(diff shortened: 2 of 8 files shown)


src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -141,14 +141,46 @@ fn supported_ui_limit_mask_for_build(build: u32) -> u32 {
     supported
 }
 
+fn has_ui_limit(mask: u32, flag: u32) -> bool {
```
bbonaby, line 145: suggestion: I think you can probably make this an inline function if you want. You'd just add `#[inline(always)]` on top of the function. Seems like a good candidate for it from what I see. #Resolved

sdk/README.md
```diff
@@ -79,6 +79,8 @@ The default `processcontainer`, `bubblewrap`, and `lxc` backends work out of the
 
 > **Hyperlight** is an opt-in build flavor (Linux x64 and Windows x64) gated by the `--with-hyperlight` cargo feature. Default shipped binaries do not include it; build from source with `build.bat --with-hyperlight` (Windows) or the equivalent cargo invocation on Linux.
 
+`getPlatformSupport()` reports backend availability and, when the native probe can determine it, `uiCapabilities`: a platform-neutral view of which UI restrictions the host can enforce. On Windows this is derived from `JOB_OBJECT_UILIMIT_*` support; on other platforms the same shape can be populated from their native sandbox mechanisms.
```
bbonaby: question: should we say this only works on Windows currently? I don't think the probe functionality works for linux or MacOs yet right? #Resolved

sdk/src/types.ts
```diff
@@ -407,6 +407,37 @@ export type IsolationTier =
...
 
+/**
+ * Host support for enforcing sandbox UI restrictions.
+ *
+ * The fields describe platform-agnostic restriction intents, not the
+ * OS-specific primitive used to enforce them. For example, Windows derives
+ * these values from `JOB_OBJECT_UILIMIT_*` support; other platforms can
+ * populate the same shape from their own sandbox mechanisms.
+ */
+export interface UiCapabilitySupport {
```
bbonaby, line 420: note: we're likely ok here since this is all generic but I wonder what our plans would be for linux and macOS. I'm wondering if it would be better to have a per backend UI capability support and a generic one. E.g `canBlockDesktopSwitching` not sure what the bubblewrap or lxc equivalent would be. #Resolved

MGudgin: We'll populate what makes sense on linux/macOS when we implement --probe there. We might need to change the shape of uiCapabilities and/or the type of the fields at that time.

MGudgin:
/azp run
 #Resolved

src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -141,14 +141,47 @@ fn supported_ui_limit_mask_for_build(build: u32) -> u32 {
     supported
 }
 
+#[inline(always)]
+fn has_ui_limit(mask: u32, flag: u32) -> bool {
```
jsidewhite, line 145: >has_ui_limit [](http://example.com/codeflow?start=3&length=12)

(aside: it's confusing that this is in the `backends/appcontainer` folder - I just realized that holds the basecontainer_runner too)

sdk/src/types.ts
```diff
@@ -407,6 +407,39 @@ export type IsolationTier =
...
+ * Host support for enforcing sandbox UI restrictions.
+ *
+ * The fields describe platform-agnostic restriction intents, not the
+ * OS-specific primitive used to enforce them. For example, Windows derives
+ * these values from `JOB_OBJECT_UILIMIT_*` support. The SDK currently
+ * receives this object only from the Windows native probe; other platforms
+ * omit `PlatformSupport.uiCapabilities` until they expose equivalent probe
+ * data.
+ */
+export interface UiCapabilitySupport {
```
jsidewhite, line 420: >UiCapabilitySupport [](http://example.com/codeflow?start=17&length=19)

@copilot Is there a recommended way to document the API surface (doxygen or some such) that the caller is expected to interact with?  i.e. they look up getPlatformSupport() and see this type

jsidewhite (approved):
:shipit:

src/backends/appcontainer/common/src/job_object.rs
MGudgin: Yes, I think we should rename the directory to processcontainer or some such

PR #503, opened 2026-06-08
Stop npm junctioning the integration-test SDK dependency

MGudgin:
## 📖 Description

npm was creating a junction (Windows) / symlink at
`sdk/tests/integration/node_modules/@microsoft/mxc-sdk` because the integration
tests declare `"@microsoft/mxc-sdk": "file:../../"`. The junction points back at
the repo's `sdk` root, which breaks `git clean` on Windows (it follows the
reparse point into the live tree) and self-references the repo.

This PR makes the integration tests install the SDK as a **packed copy** instead
of a symlink, on every install path, and ensures existing junctions are replaced
in place so dev machines self-heal and fresh clones never create one.

* Add `sdk/tests/integration/.npmrc` with `install-links=true` so the `file:`
  dependency installs as a real directory on every install in that directory —
  including a bare `npm install` — and replaces any pre-existing junction. This
  `.npmrc` is the single source of truth for the behavior.
* Track that `.npmrc` by dropping it from the integration `.gitignore`.
* Regenerate `sdk/tests/integration/package-lock.json`: the SDK is now a resolved
  copy (no `link: true`) with its runtime closure (`node-pty`, `node-addon-api`)
  materialized; `semver` is no longer dev-only.
* Set `overWrite: true` on the CopyFiles task in `SDK.Integration.Test.Job.yml`
  so the agent's Azure-feed auth `.npmrc` replaces the committed one. The
  integration job installs the SDK from a packed tarball and does not need
  install-links.

The `file:` protocol and `../..` path are unchanged and cross-platform; this only
changes the materialization mode (copy vs symlink), so macOS/Linux builds are
unaffected.

## 🔗 References

None.

## 🔍 Validation

On Windows, verified that each of these produces a real directory with zero
reparse points under `sdk/`:

* fresh clone (no `node_modules`) + bare `npm install`
* dev machine with a pre-existing junction + bare `npm install` (self-heals)
* explicit `npm install --install-links`

Also confirmed `git clean -fdx sdk/tests/integration` removes `node_modules`
cleanly (exit 0) without following into the repo tree, that the committed
lockfile is in sync (a re-install produces no drift), and that the integration
tests still build against the copied SDK (`npm run build` / tsc, exit 0).

### DIFF
--- a/.azure-pipelines/templates/SDK.Integration.Test.Job.yml
+++ b/.azure-pipelines/templates/SDK.Integration.Test.Job.yml
@@ -75,6 +75,10 @@ jobs:
           SourceFolder: '$(Build.SourcesDirectory)/.azure-pipelines/.npm'
           Contents: '.npmrc'
           TargetFolder: '$(integrationDirectory)'
+          # Overwrite the repo's committed .npmrc (install-links=true) with the
+          # Azure feed/auth .npmrc. The integration job installs the SDK from a
+          # packed tarball, so it does not need install-links.
+          overWrite: true
 
       - task: NpmAuthenticate@0
         displayName: NPM Authenticate
--- a/sdk/tests/integration/.gitignore
+++ b/sdk/tests/integration/.gitignore
@@ -1,3 +1,2 @@
 node_modules/
 dist/
-.npmrc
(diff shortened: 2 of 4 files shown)


jsidewhite:
To obviate the need to git clean, would be nice if test artifacts were spewed outside the repo.

jsidewhite (approved):

jsidewhite (approved):

PR #508, opened 2026-06-09
versioning phase 0: schema-version sync + config validation CI gates

MGudgin:
## 📖 Description

**Phase 0 of the versioning remediation** (first in a stacked series). This change is a pure-additive **safety net** for the config versioning scheme — **no runtime behavior changes**. It establishes ground truth and CI enforcement before the later phases tighten the parser.

What it does:

- Adds **`schemas/schema-version.json`** as the single canonical source of truth for the **schema**-version constants (`min` / `maxSupported` / `stateAware` / `stableLatest` / `devSchemaFile`), kept explicitly separate from the **product** version (Cargo / npm).
- Adds **`scripts/versioning/check-schema-versions.js`**, which asserts the Rust parser (`CURRENT_SCHEMA_VERSION`, `SUPPORTED_VERSION` bounds), the SDK (`SUPPORTED_VERSION`, `MIN_VERSION`, `STATE_AWARE_VERSION`), and the stable/dev schema filenames all agree with the canonical source. The existing `scripts/check-version-sync.js` (product Cargo == npm) is unchanged.
- Adds **`scripts/versioning/validate-configs.js`** (ajv), validating the `tests/examples` + `tests/configs` corpus against the dev JSON schema, with an exemption list for intentionally-invalid negative tests.
- Wires all three checks into a new **`Versioning Checks`** CI job in `build.yml` (none of them ran in CI before).
- **Triage fix**: scopes the single-backend-section `allOf` clauses in the dev schema to one-shot requests (`not: { required: ["phase"] }`) so state-aware per-phase configs — which carry `containment` only at `provision` — validate correctly. Schema-only change, no runtime impact.
- Fixes stale schema-version numbers in `copilot-instructions.md`.

## 🔗 References

Part of a stacked versioning-remediation series (Phase 0 of 6). No issue to auto-close.

## 🔍 Validation

- All three gates pass locally: product `0.6.1`; schema `maxSupported 0.7.0-alpha (min 0.4.0-alpha, state-aware 0.6.0-alpha, stable 0.6.0-alpha)`; 150/150 configs validate.
- **Negative-tested** both checkers: drifting a canonical constant makes `check-schema-versions.js` fail with precise messages; an invalid-enum config makes `validate-configs.js` fail. Restored after.
- Triage outcome: only 3 of 150 configs initially failed — all the same legitimate state-aware schema gap, fixed in the schema; no configs were wrong.
- `scripts/check-version-sync.js` is byte-identical to `main` (product check untouched).

[...]

PR #509, opened 2026-06-09
versioning phase 1: reject unknown top-level config fields

MGudgin:
## 📖 Description

**Phase 1 of the versioning remediation** (stacked on #508, now merged). This adds **structural enforcement at the trust boundary**: the parser previously dropped unrecognised top-level config fields silently, so typos (e.g. `fileSystem` vs `filesystem`) and stray sections meant the associated policy was *never applied*.

What it does:

- **Parser** (`config_parser.rs`): captures unrecognised top-level keys via a `#[serde(flatten)]` map on both `RawConfig` (one-shot) and `RawStateAwareRequest` (state-aware) and rejects them with a precise error. `$schema` and `_comment` are allowed as ignored annotations. Schema-version validation is moved to the front of conversion so an unsupported version fails fast.
- **Dev schema**: sets root `additionalProperties: false` and adds the previously-undocumented `$schema`, `_comment`, `platform`, and `appContainer` (alias) top-level properties so editor/CI validation matches the parser exactly.
- **Latent bugs fixed** (surfaced by the new enforcement):
  - `05_custom_python.json`: the bogus `fileSystem.executablePath` block (no such field; silently ignored) is converted to a valid `filesystem.readonlyPaths` grant for the anaconda install dir, preserving the author's intent (mirrors the `readonlyPaths: ["/opt/homebrew"]` pattern in `21_mac_python_info.json`).
  - The three `isolation_session_concurrent_*` configs: top-level `policy` renamed to `filesystem` (the readwrite path was never applied).
- **+6 parser unit tests**.

## 🔗 References

Part of the stacked versioning-remediation series (Phase 1 of 6). Follows #508.

## 🔍 Validation

- `cargo test -p wxc_common`: **334 pass**; new parser tests: **158 pass** (+6). `cargo test --workspace --exclude wxc_e2e_tests` passes (the single `wxc_host_prep` failure is a pre-existing `requireAdministrator`-manifest elevation requirement, confirmed identical on the base branch).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean. `cargo fmt --all -- --check`: clean.
- Both Phase 0 CI gates stay green; **162/162** configs validate against the tightened dev schema (root `additionalProperties:false`).

## Deferred (intentional)

Version-introduced (back-dating) enforcement — rejecting a config that declares an older schema version while using newer fields — is **not** in this PR. The config corpus itself back-dates (e.g. `processContainer` declared at `0.4`), and normalizing config versions to enforce the gate would silently change backends while `is_base_container_version` couples backend selection to the version string. So that gate depends on Phase 3 (backend decoupling) and is tracked separately.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type
- [ ] Bug fix
- [ ] Feature
- [x] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/509)

### DIFF
--- a/schemas/dev/mxc-config.schema.0.7.0-dev.json
+++ b/schemas/dev/mxc-config.schema.0.7.0-dev.json
@@ -20,10 +20,6 @@
     "_comment": {
       "description": "Optional human-readable annotation. Accepted but ignored by the parser."
     },
-    "platform": {
-      "type": "string",
-      "description": "Target platform hint (e.g. 'windows', 'linux'). Defaults to 'windows' when omitted."
-    },
     "appContainer": {
       "$ref": "#/properties/processContainer"
     },
--- a/schemas/dev/mxc-config.schema.0.7.0-dev.json
+++ b/schemas/dev/mxc-config.schema.0.7.0-dev.json
@@ -13,6 +13,20 @@
         "0.7.0-alpha"
       ]
     },
+    "$schema": {
+      "type": "string",
+      "description": "Optional JSON Schema reference for editor validation. Accepted but ignored by the parser."
+    },
+    "_comment": {
+      "description": "Optional human-readable annotation. Accepted but ignored by the parser."
+    },
+    "platform": {
+      "type": "string",
+      "description": "Target platform hint (e.g. 'windows', 'linux'). Defaults to 'windows' when omitted."
+    },
+    "appContainer": {
+      "$ref": "#/properties/processContainer"
+    },
     "containerId": {
       "type": "string",
       "description": "Externally assigned container identifier."
@@ -572,6 +586,9 @@
       }
     }
   },
+
+  "additionalProperties": false,
+
   "allOf": [
     {
       "$comment": "Enforce single backend section: each per-backend section requires the matching `containment` value (concrete backend or matching abstract intent). Requiring `containment` to be present when a backend section is provided also rules out the ambiguous case where two backend sections appear with no `containment` set.",
--- a/src/core/wxc_common/src/config_parser.rs
+++ b/src/core/wxc_common/src/config_parser.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use std::collections::HashMap;
 use std::fmt::Write;
 use std::fs;
 
@@ -231,6 +232,12 @@ struct RawConfig {
     /// Top-level seatbelt config.
     #[serde(alias = "macos_sandbox")]
     seatbelt: Option<RawSeatbelt>,
+    // Captures any top-level key not matched above so unrecognised fields can
+    // be rejected at the trust boundary instead of being silently dropped.
+    // Allowed annotation keys ($schema, _comment) are filtered out in
+    // `reject_unknown_top_level_fields`.
+    #[serde(flatten)]
+    unknown: HashMap<String, serde_json::Value>,
 }
 
 // State-aware request shape. `phase` is required (no `#[serde(default)]` on
@@ -260,6 +267,10 @@ struct RawStateAwareRequest {
     ui: Option<RawUi>,
     #[serde(default)]
     experimental: Option<serde_json::Value>,
+    // See `RawConfig::unknown`. Mirrored here so unrecognised top-level keys are
+    // rejected on the state-aware path too.
+    #[serde(flatten)]
+    unknown: HashMap<String, serde_json::Value>,
 }
 
 // Untagged enum: serde tries `StateAware` first (requires `phase`), falls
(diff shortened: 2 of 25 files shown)


jsidewhite:
> use serde::Deserialize;

I'm still not clear on why we need this custom parser in lieu of "Full JSON Schema validator supporting modern drafts"

```
[dependencies]
jsonschema = "0.18"
serde_json = "1"
```

```rust
use jsonschema::JSONSchema;
use serde_json::json;

fn main() {
    let schema = json!({
        "type": "object",
        "properties": {
            "name": { "type": "string" },
            "age": { "type": "integer" }
        },
        "required": ["name"]
    });

    let instance = json!({
        "name": "Jeff",
        "age": 42
    });

    let compiled = JSONSchema::compile(&schema).unwrap();

    match compiled.validate(&instance) {
        Ok(_) => println!("valid"),
        Err(errors) => {
            for error in errors {
                println!("validation error: {}", error);
            }
        }
    }
}
```

---
Refers to: src/core/wxc_common/src/config_parser.rs:8 in 5a52c4f. [](commit_id = 5a52c4ff20ada3bf27fc5f12e8063c47a5acb466, deletion_comment = False)

schemas/dev/mxc-config.schema.0.7.0-dev.json
```diff
@@ -494,6 +508,8 @@
     }
   },
 
+  "additionalProperties": false,
```
jsidewhite, line 586: what is this one?

schemas/dev/mxc-config.schema.0.7.0-dev.json
```diff
@@ -11,6 +11,20 @@
...
       "examples": ["0.6.0-alpha", "0.7.0-alpha"]
     },
+    "$schema": {
+      "type": "string",
+      "description": "Optional JSON Schema reference for editor validation. Accepted but ignored by the parser."
+    },
+    "_comment": {
+      "description": "Optional human-readable annotation. Accepted but ignored by the parser."
+    },
+    "platform": {
```
jsidewhite: >platform [](http://example.com/codeflow?start=5&length=8)

is it used for anything anywhere?

schemas/dev/mxc-config.schema.0.7.0-dev.json
```diff
@@ -11,6 +11,20 @@
...
+      "type": "string",
+      "description": "Target platform hint (e.g. 'windows', 'linux'). Defaults to 'windows' when omitted."
```
jsidewhite: >Defaults to 'windows' when omitted." [](http://example.com/codeflow?start=70&length=36)

why can't the default be specified here like it is below for some properties,

```json
"default": "windows",
```

jsidewhite (dismissed):
:shipit:

schemas/dev/mxc-config.schema.0.7.0-dev.json
MGudgin: This is the schema property that says 'the schema is closed'

MGudgin:
Addressed the review feedback in bac86d2 (kept as a separate commit, not an amend):

1. **`IgnoredAny` for the flattened `unknown` map** (both `RawConfig` and `RawStateAwareRequest`) — keys are still captured for rejection, but values are discarded during deserialization, so a large/complex unknown section costs nothing at the trust boundary.
2. **`reject_unknown_top_level_fields` is now generic** over the map value type — it only inspects keys, so it no longer couples to `serde_json::Value`.
3. **`containerId` is forwarded into the state-aware surrogate** so the inner `ExecutionRequest` preserves it (was silently dropped); added a `state_aware_forwards_container_id` test.
4. **`platform` removed entirely** rather than modeled into state-aware. It was vestigial: parsed, stored, and only emitted to a diagnostic log line — backend resolution uses the binary's compile-time `target_os` (`cfg!`), never this field, and the SDK never put it on the wire. It was also never part of any 0.4–0.7 schema. Removed from `RawConfig`, `ExecutionRequest` (now derives `Default`), the `wxc` log line, the dev schema, parser tests, and the 17 internal test configs that set it (those would otherwise trip the new unknown-field enforcement). This answers the ""is platform used anywhere?"" and ""why not a schema default?"" threads — the cleanest answer was to delete it.

Verified: `cargo test -p wxc_common` 338 pass, `cargo clippy --workspace --all-targets -D warnings` clean, `cargo fmt --check` clean, and both versioning CI gates green (163 configs validate).

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1281,6 +1336,7 @@ fn convert_raw_state_aware(
         // one-shot RawExperimental; it is preserved separately on
         // ParsedStateAwareRequest as raw JSON.
         experimental: None,
+        unknown: HashMap::new(),
     };
```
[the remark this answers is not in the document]
MGudgin: Done in bac86d2 for `containerId`: it's now modeled on `RawStateAwareRequest` and forwarded into the state-aware surrogate so the inner `ExecutionRequest` preserves it, with a `state_aware_forwards_container_id` test. `platform` is *not* forwarded — it was vestigial (only logged, never used in resolution) and is removed entirely in the same commit, so there's nothing to forward. See the `platform` thread for details.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -228,6 +229,12 @@ struct RawConfig {
     network: Option<RawNetwork>,
     ui: Option<RawUi>,
     experimental: Option<RawExperimental>,
+    // Captures any top-level key not matched above so unrecognised fields can
+    // be rejected at the trust boundary instead of being silently dropped.
+    // Allowed annotation keys ($schema, _comment) are filtered out in
+    // `reject_unknown_top_level_fields`.
+    #[serde(flatten)]
+    unknown: HashMap<String, serde_json::Value>,
```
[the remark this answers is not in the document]
MGudgin: Done in bac86d2 — both flatten maps are now `HashMap<String, IgnoredAny>`. Keys are still captured so we can name them in the rejection error, but the values are discarded during deserialization, so a large/complex unknown section no longer allocates at the trust boundary.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -257,6 +264,10 @@ struct RawStateAwareRequest {
     ui: Option<RawUi>,
     #[serde(default)]
     experimental: Option<serde_json::Value>,
+    // See `RawConfig::unknown`. Mirrored here so unrecognised top-level keys are
+    // rejected on the state-aware path too.
+    #[serde(flatten)]
+    unknown: HashMap<String, serde_json::Value>,
```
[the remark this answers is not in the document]
MGudgin: Resolved in bac86d2 on all points: (a) `containerId` is now modeled on `RawStateAwareRequest` and forwarded into the surrogate (no longer dropped); (b) `platform` is removed everywhere rather than modeled, so there's no remaining mismatch; (c) the flatten map is now `IgnoredAny`. Net: schema and parser agree for state-aware again, and the previously-silent drop of `containerId` is fixed with a test.

schemas/dev/mxc-config.schema.0.7.0-dev.json
MGudgin: Good catch — effectively no. It was parsed and stored on `ExecutionRequest`, but its only consumer was a single diagnostic log line (`wxc/main.rs`). Backend resolution for the abstract `process`/default intent uses the binary's compile-time `cfg!(target_os)`, **not** this field — and the SDK never emitted it on the wire. It was also never part of any 0.4–0.7 schema. So I removed it entirely in bac86d2 (parser, model, log line, schema, tests, and the 17 internal configs that set it). (Not to be confused with the WSLC `target_os` field, which is a different thing — the OS *inside* the WSL container.)

schemas/dev/mxc-config.schema.0.7.0-dev.json
MGudgin: You're right that a prose-only default is worse than a machine-readable `"default"`. It's moot now, though: since `platform` does nothing (see the adjacent thread), I removed it rather than document a default for a field with no effect. If we ever reintroduce a real platform/target selector, it'll get a proper `"default"`.

jsidewhite (approved):
:shipit:

PR #511, opened 2026-06-10
feat: lazy load node-pty only when usePty: true

caarlos0:
## 📖 Description
Lazy load `node-pty` only when `usePty: true`.

`node-pty` moves from a regular `dependency` to an **optional `peerDependency`** and is `require`d lazily at the PTY spawn call sites, so pipe-only (`usePty: false`) consumers no longer pull in or load the native addon. Because this is a breaking install change for PTY consumers, the SDK is bumped to **0.7.0** (Rust workspace version kept in sync). `loadPty()` surfaces an actionable error when the peer dependency is missing.

PS: should we make `usePty` default to `false`?

## 🔍 Validation
tests

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

Microsoft reviewers: PR builds don't auto-run (ADO policy). Comment `/azp run`
to start `MXC-PR-Build`. See [docs/pull-requests.md](../docs/pull-requests.md).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/511)

### DIFF
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import type pty from 'node-pty';
+import type * as pty from 'node-pty';
 import * as os from 'os';
 import { spawn, ChildProcess } from 'child_process';
 import { randomBytes } from "crypto";
--- a/sdk/src/sandbox.ts
+++ b/sdk/src/sandbox.ts
@@ -1,12 +1,13 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-import pty from 'node-pty';
+import type pty from 'node-pty';
 import * as os from 'os';
 import { spawn, ChildProcess } from 'child_process';
 import { randomBytes } from "crypto";
 import { parse as semverParse } from 'semver';
 import { SandboxPolicy, ContainerConfig, ContainmentType, ContainmentBackend } from './types.js';
+import { loadPty } from './lazyPty.js';
 import { prepareSpawn, diagLogVersion, applyLinuxNetworkPolicy } from './helper.js';
 import { diagLog } from './diagnostic.js';
 import { MxcError, mxcErrorFromCode } from './errors.js';
@@ -493,7 +494,7 @@ function injectEnvIntoConfig(
 /**
  * Internal helper: resolves the executor binary path and spawns a PTY process.
  */
-function spawnWithConfig(
+function spawnPtyWithConfig(
   config: ContainerConfig,
   options: SandboxSpawnOptions,
   workingDirectory?: string,
@@ -516,9 +517,9 @@ function spawnWithConfig(
       cwd: workingDirectory || process.cwd(),
     };
 
-    diagLog(`spawnWithConfig: spawning PTY process, cwd=${ptyOpts.cwd}`);
+    diagLog(`spawnPtyWithConfig: spawning PTY process, cwd=${ptyOpts.cwd}`);
 
-    const ptyProcess = pty.spawn(executablePath, args, ptyOpts);
+    const ptyProcess = loadPty().spawn(executablePath, args, ptyOpts);
 
     ptyProcess.onExit((event) => {
       logger?.log('info', 'mxc.spawn.exit', {
@@ -567,7 +568,7 @@ export function spawnSandbox(
   env?: { [key: string]: string | undefined },
 ): pty.IPty {
   const config = buildSandboxPayload(script, policy, workingDirectory, containerName);
-  return spawnWithConfig(config, options, workingDirectory, env);
+  return spawnPtyWithConfig(config, options, workingDirectory, env);
 }
 
 /**
@@ -645,7 +646,7 @@ export function spawnSandboxFromConfig(
   }
 
   diagLogVersion();
-  return spawnWithConfig(config, options, workingDirectory, env);
+  return spawnPtyWithConfig(config, options, workingDirectory, env);
 }
 
 /**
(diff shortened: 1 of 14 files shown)


MGudgin:
This doesn't seem to work for default TypeScript consumers. A consumer importing a type from sandbox.ts or state-aware.ts fails with error TS2307: Cannot find module 'node-pty' or its corresponding type declarations.

I'm also unsure that NOT including node-pty in our package is a good idea, as any consumers that do use PTY will now fail at runtime.

caarlos0:
@MGudgin thanks for the feedback!

I did a couple of changes:

- replicated the node-pty types in a new file
- make `usePty` default to `false`

tested it and seems to work properly now. My previous test "worked" on something that shouldn't because I had node-pty coming as a dep of something else 🤦

sdk/src/sandbox.ts
```diff
@@ -672,46 +699,119 @@ export function spawnSandboxFromConfig(
...
+      child = spawnChildWithConfig(config, options, workingDirectory);
     } catch (error) {
       reject(error);
+      return;
     }
+
+    let stdout = '';
+    let stderr = '';
+    child.stdout?.on('data', (d: Buffer | string) => {
+      stdout += typeof d === 'string' ? d : d.toString('utf-8');
```
MGudgin, line 778: This will do bad things if d happens to start/end *within* a UTF-8 multi-byte encoding sequence.

sdk/tests/integration/test-helpers.ts
```diff
@@ -242,30 +242,60 @@ export function createTempDir(prefix: string = 'mxc-test'): string {
...
       reject(err);
+      return;
     }
+    let stdout = '';
+    let stderr = '';
+    child.stdout?.on('data', (d: Buffer) => {
+      stdout += d.toString();
+    });
+    child.stderr?.on('data', (d: Buffer) => {
+      stderr += d.toString();
```
MGudgin, line 293: Same UTF-8 multi-byte issue here

sdk/src/sandbox.ts
```diff
@@ -580,72 +632,47 @@ export function spawnSandbox(
...
-      });
-      child.on('error', () => {
-        logger?.close();
-      });
-      return child;
-    } catch (err) {
-      logger?.close();
-      throw err;
-    }
+): IPty | ChildProcess {
```
MGudgin, line 669: I think the covariant return here is going to cause issues. There are cases where the compiler would infer ChildProcess at the call site but we would return IPty, resulting in a run-time crash.

sdk/tests/integration/test-helpers.ts
```diff
@@ -242,30 +242,60 @@ export function createTempDir(prefix: string = 'mxc-test'): string {
...
   config: sdkNamespace.ContainerConfig,
   options: sdkNamespace.SandboxSpawnOptions = {},
   workingDirectory?: string,
 ): Promise<{ stdout: string; stderr: string; exitCode: number }> {
+  if (options.usePty === true) {
+    return new Promise((resolve, reject) => {
+      try {
+        const ptyProcess = sdkNamespace.spawnSandboxFromConfig(
+          config,
+          { ...options, usePty: true as const },
```
MGudgin, line 263: I suspect that the 'as const' here is in order to work around the covariance issue noted above.

MGudgin:
> // PTY mode (default) — IPty, merged stdout+stderr

No longer the default?

---
Refers to: sdk/README.md:131 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin:
> The SDK provides three entry points. **Prefer the config-based path** (`createConfigFromPolicy` + `spawnSandboxFromConfig`) — it gives you backend selection, backend-specific tuning, and (with `usePty: false`) separated stdout/stderr.

This is now the default?

---
Refers to: sdk/README.md:102 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin:
> Quick path for **process-isolation only** (`processcontainer` on Windows, `lxc` on Linux, `seatbelt` on macOS). Returns a `node-pty` `IPty` with merged stdout/stderr.

Now returns a ChildProcss?

---
Refers to: sdk/README.md:145 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin:
> const pty = spawnSandboxFromConfig(config, { experimental: true });

Now childProcess?

---
Refers to: sdk/README.md:222 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin:
> ### PTY APIs merge stdout and stderr

This section needs fixup

---
Refers to: sdk/README.md:313 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin:
> ## API Surface

This section needs fixup

---
Refers to: sdk/README.md:353 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin:
> const pty = spawnSandboxFromConfig(config, { experimental: true });

Similar fixups are needed in wsl-container-getting-started.md, versioning,md, copilot-instructions.md

---
Refers to: sdk/README.md:222 in 5b2b8b1. [](commit_id = 5b2b8b1b50c808a85ade674637d0af78e73aa78b, deletion_comment = False)

MGudgin (changes requested):

caarlos0:
with #524 this will not be needed anymore

PR #515, opened 2026-06-11
Select BaseContainer tier by capability, not symbol presence

### LINKED ISSUE #495: MXC does not fall back to Tier 3 AppContainer+DACL on Windows 11 25H2 when BaseContainer is velocity-gated and bfscfg.exe is unavailable
### Relevant area(s)

Windows

MXC does not fall back to Tier 3 AppContainer+DACL on a Windows host where BaseContainer appears to be present but is not usable because required velocity keys are disabled, and where `bfscfg.exe` is unavailable.

### Steps to reproduce

1. On Windows 11 Pro 25H2 non-preview, run a simple processcontainer sandbox with schema `0.6.0-alpha`
2. Observe the BaseContainer failure.
3. Change the same config to schema `0.4.0-alpha`.
4. Run the same command again.

### Expected behavior

MXC should select a working fallback tier, ideally AppContainer+DACL Tier 3, when:

- BaseContainer launch fails with `E_NOTIMPL` due to disabled velocity keys, and
- AppContainer+BFS is unavailable because `bfscfg.exe` is missing.

Alternatively, the probe should detect this host state up front and report that Tier 3 is selected or explain why Tier 3 cannot be used.

### Actual behavior

Observed on Windows 11 Pro 25H2, latest non-preview build.

```text
PowerShell: 7.6.2
OsName: Microsoft Windows 11 Pro
OsVersion: 10.0.26200
OsBuildNumber: 26200
DisplayVersion: 25H2
UBR: 8524
```

With schema `0.6.0-alpha`, MXC selects BaseContainer and fails at launch:

>`Experimental_CreateProcessInSandbox returned E_NOTIMPL. The following velocity keys are not enabled: 61389575, 61155944. Enable them and retry, or use schema version '0.4.0-alpha' to fall back to the AppContainer backend.`
>`{"error":{"code":"backend_error","extended_error":"Experimental_CreateProcessInSandbox failed: WIN32_ERROR(120)","message":"Experimental_CreateProcessInSandbox returned E_NOTIMPL. The following velocity keys are not enabled: 61389575, 61155944. Enable them and retry, or use schema version '0.4.0-alpha' to fall back to the AppContainer backend."}}`

With schema `0.4.0-alpha`, MXC fails because BFS is unavailable:

> `Filesystem policy error: bfscfg.exe is not available on this Windows build. Your config uses schema version '0.4.0-alpha', which requires BFS support. Either update your Windows build to one that includes bfscfg.exe, or update your config to schema version '0.6.0-alpha' or later (which uses the BaseContainer backend and does not require bfscfg.exe).`

This leaves no working fallback path on this Windows 11 25H2 non-preview host, even though documentation suggests Tier 3 AppContainer+DACL should be available when neither BaseContainer nor BFS can be used.

MGudgin:
## Summary

This PR fixes the BaseContainer (Tier 1) fallback so a Windows host where
the `Experimental_CreateProcessInSandbox` symbol resolves but the OS
feature is disabled falls back to Tier 3 (AppContainer + DACL) instead of
failing the run with `ERROR_CALL_NOT_IMPLEMENTED`.

Closes #495

### Details

* Gate Tier 1 on a capability check (`is_base_container_usable`) instead
  of symbol presence: query `Experimental_QuerySandboxSupport` when
  present (authoritative), else probe the create API with invalid
  arguments (no launch), else treat a missing symbol as down-level. The
  result is cached in a process-wide `OnceLock`.
* `fallback_detector::detect` selects Tier 1 only when usable, so it
  never picks a Tier 1 that cannot launch. Removed the duplicate
  `is_base_container_api_present` detector export; `probe.rs` calls the
  runner directly, keeping the raw-symbol fact distinct from the tier
  decision.
* Added `FailurePhase::BackendUnavailable` to classify a
  present-but-disabled backend distinctly from a hard launch failure;
  extracted a shared `is_api_not_implemented` helper and capture
  `GetLastError` immediately after the FFI call.
* Refreshed docs (`sdk/README.md`, `host-prep.md`, `schema.md`,
  `playground-limitations.md`) that described Tier 3 as an
  API-absent-only path.
* Reworked the Win25H2Safe harness to key off the selected tier, so it
  runs unchanged on both fallback hosts (BaseContainer symbol absent, or
  present but feature-disabled).

### Tests

* `cargo test -p wxc_common -p appcontainer_common -p wxc`: 332 + 125 +
  13 passed, including new tests: detect skips/selects Tier 1 by forced
  capability (`MXC_FORCE_BC_USABLE` seam), dispatch falls back to Tier 3
  when unusable, the not-implemented classifier, the query-bitmask
  decoder, and a `BackendUnavailable` serde round-trip.
* `cargo fmt --all -- --check` and
  `cargo clippy -p wxc_common -p appcontainer_common -p wxc --all-targets -- -D warnings`
  clean.
* Manual E2E on a 25H2 host with Tier 1 present-but-disabled: `--probe`
  reports `appcontainer-dacl` and a schema-0.5 run lands directly on
  Tier 3 with no `E_NOTIMPL`; replayed the harness Phase 1 / Phase 4c
  logic (assertions pass, Phase 4c skips).


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -135,6 +150,108 @@ impl BaseContainerRunner {
...
+    /// `None` means the export is absent (ambiguous), so the caller must probe
+    /// another way rather than assume "unusable".
+    fn query_sandbox_create_capability() -> Option<bool> {
+        let query = Self::load_query_sandbox_support()?;
+        let mut capabilities: u64 = 0;
+        // SAFETY: `query` is the resolved export; `capabilities` is a valid
+        // out-param. A FALSE return means nothing is enforced.
+        let ok = unsafe { query(&mut capabilities) };
+        Some(Self::decode_create_capability(ok, capabilities))
+    }
```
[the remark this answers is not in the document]
MGudgin: Good catch — fixed in 4872aac. `query_sandbox_create_capability` now returns `None` when the `Experimental_QuerySandboxSupport` call fails (`ok == 0`), so we fall through to the invalid-arg create-API probe instead of treating a failed query as "disabled". Only a successful (`ok != 0`) result decodes the capability bit.

src/core/wxc_common/src/models.rs
```diff
@@ -616,6 +616,10 @@ pub enum FailurePhase {
     LaunchFailed,
     /// The process was created but exited with a non-zero code.
     ProcessExited,
+    /// The selected containment backend is present but not usable on this host
+    /// (e.g. symbol-present but feature-disabled). Distinct from [`LaunchFailed`]
+    /// so callers can fall back to a lower tier rather than hard-fail.
```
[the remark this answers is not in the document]
MGudgin: Fixed in 4872aac. The doc now reads "the API is missing, or present but not usable (e.g. feature-disabled)", covering the symbol-absent path that `validate_runner` also classifies as `BackendUnavailable`.

docs/schema.md
```diff
@@ -101,7 +101,7 @@ The `fallback` section gates the runner's host-impacting fallbacks. Each flag is
 
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
-| `allowDaclMutation` | boolean | `true` | When neither the in-process BaseContainer API nor the OS-side filesystem broker helper is available, allow MXC to apply DACL ACEs on policy paths (Tier 3 fallback). **⚠️ This modifies host filesystem security descriptors**; original DACLs are restored on exit. Set to `false` to refuse this fallback — the run will then fail on machines that require Tier 3 (e.g., Windows 11 builds below 26300 without the BaseContainer API). |
+| `allowDaclMutation` | boolean | `true` | When a usable in-process BaseContainer API and the OS-side filesystem broker helper are both unavailable, allow MXC to apply DACL ACEs on policy paths (Tier 3 fallback). "Unusable" covers both a missing API and one that is present but feature-disabled. **⚠️ This modifies host filesystem security descriptors**; original DACLs are restored on exit. Set to `false` to refuse this fallback — the run will then fail on machines that require Tier 3 (e.g., Windows 11 builds without the BaseContainer API, or newer builds where the BaseContainer feature is disabled). |
```
jsidewhite: >"Unusable" covers both a missing API and one that is present but feature-disabled.  [](http://example.com/codeflow?start=213&length=83)

You're beating the horse dead on 'API existence' _and_ 'Feature on'

I'd just say "If BaseContainer feature is available" - I'm not sure the 3P developer consumers need to really know or care if the API exists.

jsidewhite: (ofc, comments in the Rust code should definitely make this distinction.)

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -135,6 +150,115 @@ impl BaseContainerRunner {
...
+                    let _ = CloseHandle(pi.hProcess);
+                }
+                if !pi.hThread.is_invalid() {
+                    let _ = CloseHandle(pi.hThread);
+                }
+            }
+            return true;
+        }
+
+        !is_api_not_implemented(err.0)
```
jsidewhite, line 259: >is_api_not_implemented [](http://example.com/codeflow?start=9&length=22)

nit: Awkward "not" in the function name. why not invert so you can just write it as

```Rust
is_api_implemented(err.0)
```

jsidewhite: maybe i recant - the function is just checking the err code, not doing anything fancy.

docs/schema.md
MGudgin: Good call — simplified in 32e2966. The `allowDaclMutation` row now reads "When the BaseContainer feature and the OS-side filesystem broker helper are both unavailable..." and drops the API-existence-vs-feature-disabled detail. The missing-vs-disabled distinction stays in the Rust code comments (`base_container_runner.rs` / `models.rs`) where it matters.

src/backends/appcontainer/common/src/probe.rs
```diff
@@ -109,7 +109,9 @@ impl From<EffectiveUiRestrictions> for UiCapabilitySupport {
 /// summary. The detector is always asked to prefer BaseContainer (Tier 1).
 pub fn run_probe(policy: &ContainerPolicy) -> ProbeOutput {
     let probes = ProbeFacts {
-        base_container_api_present: fallback_detector::is_base_container_api_present(),
+        base_container_api_present:
+            crate::base_container_runner::BaseContainerRunner::is_base_container_api_present()
```
jsidewhite, line 113: >is_base_container_api_present [](http://example.com/codeflow?start=63&length=29)

how come this one isn't calling `is_base_container_usable`?

jsidewhite (approved):
:shipit:

PR #516, opened 2026-06-11
Centralize the tempfile dev-dependency in workspace deps

MGudgin:
This PR lifts the duplicated tempfile dev-dependency into the workspace so crates inherit a single pinned version instead of each pinning `tempfile = "3"` independently.

## Details

* Add `tempfile` to `[workspace.dependencies]` in `src/Cargo.toml`.
* Switch `appcontainer/common`, `wslc/common`, and `wxc_host_prep` from `tempfile = "3"` to `tempfile.workspace = true`.

## Tests

* `cargo metadata` resolves the workspace cleanly.

### DIFF
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -109,5 +109,5 @@ sandbox_spec = { path = "core/generated/base_container_specification" }
 widestring = "1"
 url = "2"
 winreg = "0.55"
-mxc_build_common = { path = "core/mxc_build_common" }
 tempfile = "3"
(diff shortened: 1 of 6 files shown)


shschaefer (approved):

jsidewhite (approved):

bbonaby:
closing and re-opening to see if that kicks off the GitHub builds since that's blocking Gudge from checking this in. Approvals should stays as is I believe.

PR #517, opened 2026-06-11
Prune orphaned-state DACL ACEs whose target no longer exists

MGudgin:
This PR fixes a perpetual DACL-recovery error: `recover_orphaned_state()` retained any ACE whose restore failed, including ACEs whose target path was already gone, so every startup re-attempted the restore and failed with `PATH_NOT_FOUND` forever.

## Details

* Prune an ACE when its target is confirmed missing (`try_exists() == Ok(false)`); an ambiguous `Err` (e.g. access denied) is still retried, with a re-check after a failed restore to cover the delete-during-recovery race.
* Add `RecoveryReport.aces_pruned_missing` and surface the count in the recovery diagnostic line.

## Tests

* New unit test: a dead-pid orphan whose ACE targets a missing path is pruned rather than errored, and the fully-pruned state file is removed.
* `cargo check -p wxc` passed.

### DIFF
--- a/src/core/wxc/src/main.rs
+++ b/src/core/wxc/src/main.rs
@@ -447,9 +447,10 @@ fn main() {
         Ok(report) => {
             if report.files_processed > 0 || !report.errors.is_empty() {
                 eprintln!(
-                    "DACL recovery: {} file(s), {} ACE(s) restored, {} error(s)",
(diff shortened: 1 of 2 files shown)


shschaefer (approved):

jsidewhite (approved):
:shipit:

bbonaby:
closing and re-opening to see if that kicks off the GitHub builds since that's blocking Gudge from checking this in. Approvals should stays as is I believe.

PR #519, opened 2026-06-11
Add NANVIX_BIN env var for offline nanvix binary prefetching

### LINKED ISSUE #488: Enable nanvix binary prefetching

The current build system calls `curl` at compilation time to download the nanvix files.
These files are downloaded into a subdirectory of `$OUT_DIR`, which is not a predictable location.
[source](https://github.com/microsoft/mxc/blob/5a330ce014e8636fc59568267b6cd2483fda40cc/src/backends/nanvix/binaries/build.rs#L48).

The `build.rs` script should check for the existence of an environment variable such as `NANVIX_BIN`, and use that directory instead if the environment variable exists.
This enables offline builds, where all dynamic build inputs are pre-fetched.

huzaifa-d:
## 📖 Description

The `nanvix_binaries` build script previously always downloaded NanVix release assets via `curl`/`tar` into an unpredictable `OUT_DIR` subdirectory, so the build required network access. This PR adds a `NANVIX_BIN` environment variable: when set, the build uses that directory of pre-fetched binaries directly and performs **no** network downloads, enabling offline / air-gapped / hermetic builds. Checksums are still verified against `checksums.json`.

Behavior when `NANVIX_BIN` is unset is unchanged.

**Key changes**
- `backends/nanvix/binaries/build.rs`: detect `NANVIX_BIN`; use that (canonicalized) directory as the binary dir; gate both the Windows and Linux download paths behind `!use_prefetched_binaries`; still verify checksums; skip host-local WHP snapshot generation in prefetched mode so a read-only/shared prefetch cache is never mutated; emit `rerun-if-changed` per pre-fetched file (binaries, `bin/`, snapshots) plus `rerun-if-env-changed=NANVIX_BIN`.
- `backends/nanvix/common`: `copy_artifacts_to_target` now *mirrors* snapshots — copies newer files and **purges target snapshots absent from source** — so the runtime never warm-boots a stale snapshot against mismatched binaries (the runner trusts complete exe-side snapshots without a binary-match check). Adds `emit_rerun_for_copied_artifacts`. Online builds always ship every snapshot, so nothing is purged there.
- `core/wxc/build.rs` and `core/lxc/build.rs`: call `emit_rerun_for_copied_artifacts` so an in-place `NANVIX_BIN` content update refreshes/purges the artifacts copied next to the executable (previously they only reran on the source *path* changing).
- `docs/nanvix-microvm/nanvix.md`: documents the offline-build workflow.

**Limitations**
- WHP snapshots are Windows-only; on Linux this mirror logic is a no-op. In prefetched mode without snapshots, the runtime cold-boots on first use (existing fallback).

## 🔗 References

Resolves #488

## 🔍 Validation

- `cargo fmt --check` and `cargo clippy --features microvm --all-targets -- -D warnings` — clean.
- `cargo test -p nanvix_common` — 5 passed.
- Full `cargo build -p wxc --features microvm` — succeeds.
- Offline path (Windows): built normally, copied the staged `nanvix-binaries` dir to a fresh location, set `NANVIX_BIN` to it, rebuilt with `--features microvm` → logged "using pre-fetched binaries (offline)" and performed no download; nonexistent dir → clear panic; relative path → emitted absolute `cargo:BIN_DIR`.
- Snapshot purge: seeded stale target snapshots, ran an offline build whose prefetch dir had no snapshots → stale target snapshots removed; a subsequent online build restored the real 256 MB snapshot.
- Linux: verified via WSL/Docker (see PR thread).

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/519)


src/backends/nanvix/binaries/build.rs
```diff
@@ -69,8 +76,57 @@ fn main() {
...
 
     let out_dir = PathBuf::from(std::env::var("OUT_DIR").unwrap());
-    let bin_dir = out_dir.join("nanvix-binaries");
-    fs::create_dir_all(&bin_dir).expect("failed to create nanvix-binaries dir");
+
+    // When `NANVIX_BIN` is set, use the caller-provided directory of
+    // pre-fetched binaries and skip all network downloads. This enables fully
+    // offline builds where every dynamic build input has been pre-fetched.
+    // Otherwise, download into a subdirectory of OUT_DIR as before.
+    let offline_bin_dir = std::env::var_os("NANVIX_BIN")
```
ppenna: Do we want to move this to `src/backends/nanvix/common/src/lib.rs`?

src/backends/nanvix/binaries/build.rs
```diff
@@ -69,8 +76,57 @@ fn main() {
...
+        .filter(|v| !v.is_empty())
+        .map(PathBuf::from);
+    let use_prefetched_binaries = offline_bin_dir.is_some();
```
ppenna: Nitpick: we already have testing logic for `is_some()` and return dir base on that, we could make it return a tuple that includes this boolean.

src/backends/nanvix/common/src/lib.rs
```diff
@@ -239,70 +239,127 @@ pub fn generate_snapshot(
...
+            dst.display()
+        );
+        if let Err(e) = std::fs::remove_file(dst) {
+            panic!(
+                "nanvix: failed to remove stale snapshot {}: {} — refusing to \
+                 leave an unverified warm-start image that would be booted \
+                 against mismatched binaries",
+                dst.display(),
+                e
+            );
```
ppenna: Instead of panicking, we should return an error and let the caller decide if we should panic (ok for build path, not ok for mainline code).

src/backends/nanvix/common/src/lib.rs
```diff
@@ -239,70 +239,127 @@ pub fn generate_snapshot(
...
+/// [`copy_artifacts_to_target`] reads from `src_dir`. Call this from a
+/// consuming crate's build script so the copy reruns when the source contents
+/// change in place — for example when an offline `NANVIX_BIN` prefetch
+/// directory is updated at the same path. Without it, the consumer only reruns
+/// when the source *path* changes, leaving stale artifacts next to the exe.
+pub fn emit_rerun_for_copied_artifacts(src_dir: &Path) {
+    for (_, rel) in artifact_rel_paths() {
+        println!("cargo:rerun-if-changed={}", src_dir.join(rel).display());
     }
 }
```
ppenna: If this is build only, should we move to respective `build.rs`? If we need common build utils, we should move to

src/core/wxc/build.rs
```diff
@@ -94,8 +94,24 @@ fn copy_nanvix_binaries() {
         }
     };
 
-    nanvix_common::copy_artifacts_to_target(Path::new(&nanvix_bin_dir), target_dir);
+    // WHP snapshots from a prefetched (externally supplied) directory are not
+    // covered by checksums.json, so they must not be trusted/copied; the
+    // producer reports this via `cargo:PREFETCHED`. Default to trusting
+    // (online build) when the flag is absent.
+    let trust_snapshots = std::env::var("DEP_NANVIX_BINARIES_PREFETCHED")
```
ppenna: Should we also move to `src/backends/nanvix/common/src/lib.rs`?

ppenna (approved):
LGTM

jsidewhite (approved):
:shipit:

PR #524, opened 2026-06-12
feat: mxc-sdk lib crate

caarlos0:
## 📖 Description

Adds **`mxc-sdk`** (library name `mxc_sdk`), an importable Rust library crate for starting sandboxes in-process, **without a pty**. Callers can stream sandboxed processes without shelling out to the executor binaries or depending on the `@microsoft/mxc-sdk` TypeScript module.

Supported backends: **Bubblewrap** (Linux), **Seatbelt** (macOS), and **Windows ProcessContainer** (AppContainer + BaseContainer fallback).

The crate's public surface is **decoupled from the internal execution model**: `build_request` / `spawn_sandbox` take and return the crate's own types (`SandboxRequest`, `Sandbox`, and `Error` / `ErrorCode`) rather than the internal `wxc_common` types they map to, so the internals can be refactored without breaking callers.

To share one execution path between the new library (no-pty, caller-owns-stdio) and the existing executor binaries (run-to-completion), the three library backends were unified on a single `SandboxBackend` trait. That part is a refactor of the shared backend layer — not purely additive; see "Unified backend execution model" below.

### What's included

- **Handle-based streaming API** — `spawn_sandbox` takes a `SandboxRequest` (built by `build_request`, see below) and returns a `Sandbox` handle for persistent bidirectional stdio (`take_stdin`/`take_stdout`/`take_stderr`), `try_wait()`, `id()`, `kill()`, and `wait()`. No pty is allocated; `wait()` drains and discards any untaken stdout/stderr (so the child can't block on a full pipe) and returns the exit code as `io::Result<i32>` (`ErrorKind::TimedOut` on timeout). Streaming is implemented for Seatbelt, Bubblewrap, and Windows ProcessContainer. The pipe-deadlock contract (drain stdout/stderr concurrently) is documented on the handle.
- **Unified backend execution model** — the three library backends (Seatbelt, Bubblewrap, Windows ProcessContainer) now implement one `SandboxBackend` trait (`validate` + `spawn(request, logger, StdioMode) -> Box<dyn SandboxProcess>` + a `diagnose_exit` hook that preserves AppContainer launch diagnostics). `StdioMode::Pipes` is what the library uses (the caller takes stdin/stdout/stderr); `StdioMode::Inherit` is what the executor binaries use (the child inherits the binary's own stdio, preserving the TTY when launched under a pty). A single generic `Runner<B>` adapter in `wxc_common::sandbox_process` provides the run-to-completion `ScriptRunner` the binaries dispatch on, by calling `spawn(StdioMode::Inherit)` then `wait()` — so there is no longer any per-backend run-to-completion logic and the old `StreamingRunner` trait is removed. The `wxc-exec` / `lxc-exec` / `mxc-exec-mac` binaries wrap their backends in `Runner` and otherwise behave identically (LXC keeps its own native pty path).
- **Process-tree termination** — `kill()` and timeouts take down the whole tree (Unix process-group signal with an unconditional `SIGKILL` sweep so a racily-forked descendant can't survive; Windows job-object terminate), so descendants can't keep inherited pipe write-ends open and hang `wait()`. A timeout surfaces uniformly as `wait()` returning `io::ErrorKind::TimedOut` across all backends.
- **Clean environment** — sandboxed children never inherit the host process's environment (the Seatbelt backend now always `env_clear()`s, matching Bubblewrap's `--clearenv` and the AppContainer clean block), so an embedding host app's secrets aren't exposed to untrusted code.
- **Ported SDK config-building helpers** — `mxc_sdk::policy` (`SandboxPolicy`, `build_request`, plus `available_tools_policy` / `user_profile_policy` / `temporary_files_policy` discovery) and `mxc_sdk::platform_support`, so callers no longer need the TypeScript SDK for policy construction or platform discovery. `build_request` maps a `SandboxPolicy` to a `SandboxRequest`; set the command with `set_script(...)`, and optionally `set_working_directory(...)` / `set_env(...)`, before spawning. On macOS the request also exposes the Seatbelt knobs a host may need — `seatbelt_extra_mach_lookups()` / `set_seatbelt_extra_mach_lookups(...)` / `set_seatbelt_keychain_access(...)`.
- **Unsupported-host guard** — a clear error on platforms with no MXC backend.

### Limitations

- **LXC is not supported** by the library: the LXC backend has no non-pty capture path (it always streams via a pty and returns empty stdout/stderr), so it can't honour the crate's no-pty contract; it returns `unsupported_containment`. The `lxc-exec` binary keeps its native LXC support.
- **`dry_run`** is rejected by the streaming spawn (there is no process to stream).
- The Windows ProcessContainer integration tests run a real sandbox, so they require an elevated, host-prepped Windows host (see `docs/host-prep.md`) and are `#[ignore]`d.

## 🔭 Follow-ups (separate PRs)

- **An actual `mxc` engine crate** — the crate is named `mxc-sdk` because a follow-up will introduce a lower-level `mxc` crate that both `mxc-sdk` and the existing executor binaries call into. Until then, the in-crate backend dispatch (`dispatch.rs` / `platform.rs`) in `mxc-sdk` is provisional and moves into that engine crate.

The unified `SandboxBackend` / `Runner` model can be extended to the remaining backends incrementally; the cost is set by each backend's I/O model, not by trait plumbing:

- **LXC** and **IsolationSession** are the natural next steps — both drive a real process (IsolationSession already streams live via relay threads + ConPTY), so each is a focused `SandboxProcess` impl that would also bring the backend into the `mxc-sdk` library (Linux-via-LXC; capable Windows hosts).
- **NanVix (MicroVM)** is doable but heavier — it has a host child, but the filesystem copy-back/staging teardown has to move onto the handle's lifecycle.
- **Hyperlight**, **Windows Sandbox**, and **WSLC** run to completion across an in-process FFI / in-VM daemon / COM-SDK boundary with no local child, so real streaming is gated by the underlying runtime/protocol. These can stay on `ScriptRunner` until a concrete need arises.
- **`platform_support` UI / isolation reporting** — the SDK `getPlatformSupport` parity fields (isolation tier, tier-degradation warnings, and the Windows UI-capability probe results) were trimmed from `PlatformSupport` since no consumer reads them yet. Restore them (re-wiring the AppContainer probe) when a caller needs richer host-capability discovery.

## 🔗 References

Follow-up to the TypeScript SDK's one-shot spawn surface; this is its in-process Rust counterpart.

## 🔍 Validation

- `cargo test -p mxc-sdk` on macOS — 41 tests pass (12 lib-unit, `sandbox` 7, `sdk_helpers` 10, `streaming` 11, plus 1 doc-test); the Windows-host ProcessContainer integration tests are `#[ignore]`d (they spawn a real sandbox and need an elevated, host-prepped Windows host).
- `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` clean (macOS + Linux + Windows cross-check).
- Cross-compilation: `cargo check` against `x86_64-pc-windows-msvc` and `aarch64-unknown-linux-gnu`, plus the per-backend crates on each target.
- **Runtime-validated on all three host platforms** after the backend unification:
  - **macOS / Seatbelt** — exec with piped stdio and with inherited stdio, exit-code propagation, the `guiAccess` GUI path, `launchMethod: open` (Terminal), and timeout tree-kill.
  - **Linux / Bubblewrap** — output streaming, non-zero exit propagation, inherited TTY, and timeout tree-kill.
  - **Windows / ProcessContainer** — BaseContainer (schema `0.7.0-alpha`) streams stdout and propagates a non-zero exit code (`exit /b 7` → `EXIT=7`); a 2000 ms `scriptTimeout` fires and tree-kills a process tree blocked on a modal dialog (`before` streamed, `after` never ran). The direct-AppContainer/BFS path needs `bfscfg.exe`, absent on the test host, so it was covered by compile checks only.
- Test coverage spans output streaming via taken stdout/stderr, host-environment scrubbing, env reaching the sandboxed child, streaming bidirectional stdio, process-tree termination (incl. the racily-forked-descendant case), finite-timeout enforcement (uniform `io::ErrorKind::TimedOut` from `wait()`), the `guiAccess` streaming rejection, `dry_run` rejection, version-out-of-range / empty-version rejection, the `build_request` filesystem/network/clipboard/proxy wire mapping, and the macOS Seatbelt `extra_mach_lookups` / `keychain_access` round-trip.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task


caarlos0:
Manually tested end-to-end through Copilot on macOS, Linux, and Windows — exercising the Seatbelt, Bubblewrap, and Windows ProcessContainer backends respectively via the in-process library path.

caarlos0:
## Merged `microsoft/mxc#540` (executor characterization tests) into this branch

The git-level merge was clean (no textual conflicts — #540 only touches `src/testing/wxc_e2e_tests/` + two workflows). But the new tests surfaced a **semantic** conflict: three Seatbelt tests lock in `main`'s run-to-completion behavior, which this branch's `ScriptRunner` → `SandboxBackend`/`Runner` refactor intentionally changed. I grounded each in the code and reconciled them:

| Characterization test | Verdict | Action |
|---|---|---|
| `seatbelt_timeout_kills_before_completion` | **Real regression** in this branch | Fixed the code (`c633932`) |
| `seatbelt_inherits_host_env_when_process_env_empty` | Intentional behavior change | Updated test (`0e94f26`) |
| `seatbelt_runs_in_launcher_cwd_when_process_cwd_empty` | Intentional behavior change | Updated test (`0e94f26`) |

### 1. Timeout regression — fixed (`c633932`)

`main` enforced the run-to-completion timeout by SIGKILLing the child (via the pty bridge), so a timed-out workload stopped immediately. The refactor instead used a graceful group kill (SIGTERM → 2s grace → SIGKILL).

On macOS, `/bin/sh` (bash) parked in a foreground `wait` **defers SIGTERM**: it reaps the killed child, prints its `Terminated` notice, and then runs the *rest* of the script before the SIGKILL escalation fires — so a timed-out job could still execute post-timeout commands. (The streaming/`Pipes` path only avoided this incidentally, via SIGPIPE on the closed stdout pipe; the executor/`Inherit` path inherits the real terminal, so nothing is closed.)

The kill paths are now collapsed into a single primitive: `group_kill` does **one immediate SIGKILL** (no graceful SIGTERM, no separate `hard_kill`), used by every Unix kill path — streaming `kill()`, `Drop` teardown, and the timeout branch. A graceful window was unreliable anyway (the shell defers SIGTERM) and untrusted sandboxed code isn't owed one.

Crucially, it signals the **group leader before the group** (`kill(pid)` then `kill(-pid)`). A `-pgid`-only sweep races: the kernel can terminate the foreground child first, wake the shell, and let it run one more command before the shell's own signal lands. Signalling the leader first makes its SIGKILL already pending when the descendant's death would wake it, so the kernel terminates it on the `wait` return instead of resuming it. This matches what `main` effectively did (SIGKILLing the blocked pty child directly) and makes the test deterministic (10/10 locally — the naïve group-only kill was flaky).

### 2 & 3. Env + cwd — tests updated to the branch's intended behavior (`0e94f26`)

Both were behavior changes the #540 author explicitly anticipated (the cwd test predicted it would "turn RED"; the Windows suite notes the env/cwd changes are "Seatbelt-specific"):

- **env** — the exec path now always starts from a cleared environment (`env_clear()` + a default `PATH`), so the launcher's environment (cloud creds / API tokens) never leaks into untrusted sandboxed code. This makes Seatbelt consistent with bubblewrap's `--clearenv` model (`bubblewrap_clears_host_env_by_default`). Renamed to `seatbelt_clears_host_env_when_process_env_empty` and flipped to assert a cleared env.
- **cwd** — an empty `process.cwd` now resolves to the first readwrite policy path (a profile-allowed directory) instead of inheriting the launcher cwd, which the deny-by-default profile may forbid (making the child's `getcwd()` fail and leak a `getcwd: Operation not permitted` line). Renamed to `seatbelt_runs_in_first_readwrite_path_when_process_cwd_empty` and updated to assert the probe lands in that path.

### Validation (pinned 1.93 toolchain, macOS)

- All **7** Seatbelt characterization tests pass, including `seatbelt_timeout_kills_before_completion` run **10×** with no flakes; `mxc` (40) + `wxc_common` (259) tests pass; the `wxc_e2e_tests` crate compiles.
- `cargo fmt --check` clean; **1.93 clippy** clean on the changed crates and on the Windows/Linux cross-targets.

### Cross-platform caveat

I could only runtime-test Seatbelt locally. The Windows `processcontainer` timeout test asserts the strong behavior (no `CHAR_AFTER`) — Windows job-object termination is atomic, so it should hold, but only Windows CI can confirm. The Linux `bubblewrap` suite is lenient and should pass. CI will validate both.

docs/lxc-support/lxc-backend.md
```diff
@@ -87,7 +87,7 @@ The `process.cwd` and `process.env` fields from the standard schema are honored
...
 
-**Replace semantics.** When `process.env` is non-empty, `lxc-exec` also passes `--clear-env` to `lxc-attach` so the host environment does **not** leak into the sandbox, regardless of how many entries survive the malformed-skip. This matches the Seatbelt backend's `env_clear()`-on-non-empty contract and is the posture `lxc-attach(1)` recommends for sandbox-spawn callers. If a variable is set in both the host and `process.env`, the `process.env` value wins.
+**Replace semantics.** When `process.env` is non-empty, `lxc-exec` also passes `--clear-env` to `lxc-attach` so the host environment does **not** leak into the sandbox, regardless of how many entries survive the malformed-skip. This is the posture `lxc-attach(1)` recommends for sandbox-spawn callers. If a variable is set in both the host and `process.env`, the `process.env` value wins. (Note: the Seatbelt backend now clears the environment **unconditionally** — not only when `process.env` is non-empty — so it no longer shares this empty-`process.env` keep-env behavior.)
```
MGudgin: >(Note: the Seatbelt backend now clears the environment **unconditionally** — not only when `process.env` is non-empty — so it no longer shares this empty-`process.env` keep-env behavior.) [](http://example.com/codeflow?start=389&length=187)

nit: Remove mention of Seatbelt here (this document is about lxc)

sdk/src/policy.ts
```diff
@@ -212,31 +220,60 @@ function getPowerShellPolicy(
...
-    const readwritePaths: string[] = [];
+    const readonlyCandidates: string[] = [];
+
+    // 1. The PowerShell install tree: pwsh.exe, its self-contained .NET runtime,
+    //    and the shipped modules under `$PSHOME\Modules`.
+    readonlyCandidates.push(psHome);
+    readonlyCandidates.push(path.join(psHome, 'Modules'));
+
+    // 2. System DLLs the loader and the .NET runtime resolve from System32.
+    readonlyCandidates.push(path.join(getWindowsDirectory(), 'System32'));
```
MGudgin: Is this actually needed? I've been running pwsh without this in the policy and its been working.

sdk/src/policy.ts
```diff
@@ -212,31 +220,60 @@ function getPowerShellPolicy(
...
+        }
+    }
 
+    const readwritePaths: string[] = [];
     const userProfile = env['USERPROFILE'];
     if (userProfile) {
-        const psReadLineDir = path.join(
-            userProfile, 'AppData', 'Roaming', 'Microsoft', 'Windows', 'PowerShell', 'PSReadLine',
+        // Per-user module directory (read-only).
+        readonlyCandidates.push(path.join(userProfile, 'Documents', 'PowerShell', 'Modules'));
```
MGudgin: >Documents [](http://example.com/codeflow?start=56&length=9)

Why are we adding Documents? This seems wrong.

MGudgin: I suspect this may not work if Documents has been re-directed/moved.

src/backends/appcontainer/common/src/network_manager.rs
```diff
@@ -28,20 +28,80 @@ pub enum DefaultPolicy {
...
+
+impl ComApartment {
+    /// Join (or initialize) an apartment-threaded COM apartment for the current
+    /// thread. `S_OK`/`S_FALSE` both count as an initialization this guard must
+    /// balance; `RPC_E_CHANGED_MODE` reuses an existing apartment without
+    /// taking ownership of its teardown.
+    fn new() -> Result<Self, WxcError> {
+        // SAFETY: `CoInitializeEx` is always safe to call; the matching
+        // `CoUninitialize` runs in `Drop` on this same thread when we own it.
+        let hr = unsafe { CoInitializeEx(None, COINIT_APARTMENTTHREADED) };
```
MGudgin, line 63: >COINIT_APARTMENTTHREADED [](http://example.com/codeflow?start=47&length=24)

Do we have to use STA here?

caarlos0: Good question — STA here predates this PR. `COINIT_APARTMENTTHREADED` is already on `main`; this PR only refactored the per-call init/teardown into the `ComApartment` RAII guard, it didn't pick the apartment model.

You're right that MTA is arguably the better fit: `INetFwPolicy2`/`INetFwRule` are registered `ThreadingModel=Both`, and since the owning sandbox handle is `Send` and teardown can run on a different thread than setup, a process-wide MTA would actually *simplify* the soundness story (no per-thread apartment affinity to design around — which is most of what the `ComApartment` doc comment is justifying).

That said, flipping STA→MTA is a behavioral change to the Windows firewall path that I'd want to validate on a real Windows host, and it's orthogonal to this PR (the `mxc-sdk` crate). I'd rather not change pre-existing, already-validated behavior untested here — I'll open a follow-up to evaluate the switch on its own. Keeping STA for now.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -960,107 +1004,399 @@ impl ScriptRunner for BaseContainerRunner {
...
+        //
+        // The child is created running: the BaseContainer create API
+        // (`Experimental_CreateProcessInSandbox`) does not, on the builds
+        // validated to date, honor a suspended/paused start the way the
+        // AppContainer path does (create suspended → assign → resume), so a
+        // descendant spawned in the brief window between `CreateProcessInSandbox`
+        // returning and `assign_process` completing could still escape the job.
+        // In practice the child is a shell that has not yet run the user
+        // command, so that window is empty; fully closing it would require a
+        // create-suspended path verified on a host-prepped build.
```
MGudgin: >host-prepped [](http://example.com/codeflow?start=47&length=12)

Host prep shouldn't apply to base container

MGudgin: This (creating SUSPENDED) should work (at least from looking at the OS code). Can we at least amend the code above to create suspended and then resume after the job add (where the resume will be a no-op if CPIS is not doing the right thing).

caarlos0:
Superseded by a 3-PR stack that splits this change into reviewable, compile-ordered layers (per the proportionality feedback in review):

1. **#554** — `feat(wxc_common)`: the sandbox execution interfaces (`SandboxBackend` / `SandboxProcess` / `Runner`). Purely additive.
2. **#555** — `refactor(backends)`: migrate the backends onto those interfaces + route the executor binaries via `Runner` (the behavior-affecting refactor, with characterization tests).
3. **#556** — `feat(mxc-sdk)`: the importable in-process library crate, built on top.

Review in order (#554 → #555 → #556); #555/#556 are stacked, so their diffs shrink to just their own layer once the base PR merges. Closing this in favor of the stack.

caarlos0:
(will keep it open for now)

PR #531, opened 2026-06-15
NanVix backend: host networking support

huzaifa-d:

-----

## Summary

Adds host networking support to the NanVix (microvm) backend, wiring the MXC runner to the per-host egress filtering that nanvixd now implements.

The NanVix runner emits `-allow-host-networking` to nanvixd when `network.defaultPolicy` is `"allow"`, and forwards per-host allow/block lists as `-allow-host` / `-block-host`. Egress filtering is allow-XOR-block (an allowlist is deny-by-default; a blocklist is allow-by-default), and host resolution mirrors the `lxc` backend so the two agree on hostname-to-IPv4 mapping at preflight.

The pinned `nanvix-python` build is bumped to `3.12.3-nanvix-0.16.65-fce3620`, which bundles the nanvixd that implements per-host egress filtering, and `checksums.json` is updated to match. The schema gains `microvm` in the `allowedHosts` / `blockedHosts` backend lists, with the microvm-specific mutual-exclusivity and DNS-exemption notes.

Windows and Linux microvm networking e2e tests (`test_microvm_network`) plus their configs are added. The Windows test was run locally against the bumped nanvixd and passes (guest completes a loopback TCP round-trip and prints `NET_OK`).

### Related Issues

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/531)

### DIFF
--- a/src/backends/nanvix/runner/src/lib.rs
+++ b/src/backends/nanvix/runner/src/lib.rs
@@ -519,8 +519,8 @@ impl NanVixScriptRunner {
     ///   dropped because the guest filter is IPv4-only.
     /// - Entries that fail to parse or resolve contribute nothing.
     ///
-    /// Mirrors `lxc::network_iptables::resolve_host` so the two backends agree
-    /// on resolution semantics.
+    /// Mirrors `lxc::network_iptables::resolve_host` for hostname-to-IPv4 mapping,
+    /// and additionally preserves IPv4/CIDR literals for nanvixd.
     fn resolve_hosts(hosts: &[String]) -> Vec<String> {
         let mut out: Vec<String> = Vec::new();
         for host in hosts {
--- a/src/backends/nanvix/runner/src/lib.rs
+++ b/src/backends/nanvix/runner/src/lib.rs
@@ -107,8 +107,28 @@ const ERR_HOSTS_UNRESOLVED: &str = concat!(
     "none of the specified allowedHosts/blockedHosts resolved to an IPv4 address -- ",
     "the NanVix guest filter is IPv4-only; use IPv4 literals/CIDR or hosts with A records",
 );
+const ERR_BLOCKED_HOST_UNRESOLVED: &str = concat!(
+    "a blockedHosts entry did not resolve to any IPv4 address -- ",
+    "the NanVix guest egress filter is static (resolved once at preflight) and IPv4-only, ",
+    "so an unresolvable blocked host cannot be enforced. Silently dropping it would let ",
+    "traffic the policy explicitly blocks flow freely, so the run is rejected (fail-closed). ",
+    "Use an IPv4 literal/CIDR or a host with A records",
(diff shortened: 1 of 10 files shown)


src/backends/nanvix/runner/src/lib.rs
```diff
@@ -483,16 +498,100 @@ impl NanVixScriptRunner {
...
+    /// Host networking is enabled when `network.defaultPolicy = "allow"` OR when
+    /// a per-host allow/block list is present (a list always implies networking,
+    /// regardless of `defaultPolicy`). When enabled, the runner passes
+    /// `-allow-host-networking` to nanvixd; per-host lists are additionally
+    /// forwarded as `-allow-host`/`-block-host` (see [`Self::spawn_nanvixd`]).
+    fn host_networking_enabled(request: &ExecutionRequest) -> bool {
+        request.policy.default_network_policy == NetworkPolicy::Allow
+            || !request.policy.allowed_hosts.is_empty()
+            || !request.policy.blocked_hosts.is_empty()
+    }
```
ppenna, line 532: I assume this logic is sandbox independent and should live in a common place for every backend in MXC?

src/backends/nanvix/runner/src/lib.rs
```diff
@@ -483,16 +498,100 @@ impl NanVixScriptRunner {
...
+            if let Some((addr, prefix)) = entry.split_once('/') {
+                let addr_ok = addr.trim().parse::<std::net::Ipv4Addr>().is_ok();
+                let prefix_ok = prefix
+                    .trim()
+                    .parse::<u8>()
+                    .map(|p| p <= 32)
+                    .unwrap_or(false);
+                if addr_ok && prefix_ok {
+                    out.push(entry.to_string());
+                }
```
ppenna, line 570: If we skip, do you want to log anything or inform the user about it? How do the other sandboxes handle these? Also, is this common logic that can be shared?

tests/configs/microvm_network.json
ppenna, line 1: Should we also include a test cases that test that an endpoint that is not enabled is blocked by default, as well an endpoint that is blocked, is really blocked?

ppenna (approved):
LGTM. My comments concerning common logic can be handled as follow-up work for MXC :)

src/backends/nanvix/runner/src/lib.rs
jsidewhite: agree - can you emit a warning or something?

jsidewhite: should this be a hard error if it's a `blocked` host that doesn't resolve?

jsidewhite (approved):
:shipit:

jsidewhite (approved):

PR #539, opened 2026-06-19
Remove stale TS CLI References

MGudgin:
Remove references to the TypeScript CLI from copilot-instructions.md

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/539)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -27,7 +27,7 @@ build.bat --with-microvm   # Include NanVix micro-VM binaries
 ```
 ./build.sh                 # Release build
 ./build.sh --debug         # Debug build
-./build.sh --rust-only     # Only Rust binaries, skip SDK/CLI
+./build.sh --rust-only     # Only Rust binaries, skip SDK
 ```
 
 ### Full build (macOS)
@@ -52,9 +52,6 @@ cargo build --release -p mxc_darwin --target aarch64-apple-darwin  # macOS only 
 
 # SDK (from sdk/)
 npm install && npm run build
-
-# CLI (from cli/)
-npm install && npm run build
 ```
 
 ### Lint and format
@@ -63,9 +60,6 @@ npm install && npm run build
 # Rust (from src/)
 cargo fmt --all -- --check
 cargo clippy --workspace --all-targets -- -D warnings
-
-# CLI (from cli/)
-npx eslint src --ext .ts
 ```
 
 ### Tests
@@ -79,9 +73,6 @@ cargo test -p wxc_common -- config_parser   # Filter by test name
 # SDK (from sdk/)
 npm test
 npm run test:integration
-
-# CLI (from cli/) — requires build first
-node --test dist/cli.test.js
 
 # Local PowerShell helpers — run from repo root, require built binaries
 tests\scripts\run_test_configs.ps1            # All test configs via wxc_test_driver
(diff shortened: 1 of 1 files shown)


jsidewhite (approved):

bbonaby:
I think for this one it's because we have .md files set to ignore in here: https://github.com/microsoft/mxc/blob/b19e8126f7c5c2562066d602ec78096f136c014d/.github/workflows/Build.yml#L19

But then in the Github actions  in the settings for the repository, I have all the GitHub actions set to `required`. In the azure pipelines we have the same thing as well here:  https://github.com/microsoft/mxc/blob/b19e8126f7c5c2562066d602ec78096f136c014d/.azure-pipelines/1ES.Build.yml#L16
That's why when you commented `azp run` the bot said `Azure Pipelines could not run because the pipeline triggers exclude this branch/path`.

 We don't require the azure pipelines to succeed before merging so that one is fine.  Unfortunately the GitHub pipelines will stay stuck forever waiting on the build to run even though we have these paths set to ignore. E.g Github waiting for required pipelines to succeed but they never run because they're skipped. You can see the required status checks here: https://github.com/microsoft/mxc/rules/16396467 under `Require status checks to pass`.

We could remove the `required` tags from these  Github pipelines in the settings to prevent this, as I believe the pipelines would show up as skipped in this case for .md only PRs. That said, we'd have to be very careful approving pull requests in the future since that would allow PR builds to fail but allow someone to still check it in regardless.

OR

We can add a bypass to the PR rule. Note: the PR rule is really just to make sure the pipelines are required to pass at PR time. There is a Microsoft rule that PRs require at least 1 approver which we also have and isn't bypassable [here](https://github.com/microsoft/mxc/rules/5351760), so effectively bypassing would be just for the PR rule and we can scope it to only a few people with it documented that it should only be used for document changes. 


I'm good with either two. Thoughts? 

Note: This one also has the same issue: https://github.com/microsoft/mxc/pull/545

MGudgin:
> I think for this one it's because we have .md files set to ignore in here:
> 
> https://github.com/microsoft/mxc/blob/b19e8126f7c5c2562066d602ec78096f136c014d/.github/workflows/Build.yml#L19
> 
> But then in the Github actions in the settings for the repository, I have all the GitHub actions set to `required`. In the azure pipelines we have the same thing as well here:
> 
> https://github.com/microsoft/mxc/blob/b19e8126f7c5c2562066d602ec78096f136c014d/.azure-pipelines/1ES.Build.yml#L16
> 
> 
> That's why when you commented `azp run` the bot said `Azure Pipelines could not run because the pipeline triggers exclude this branch/path`.
> We don't require the azure pipelines to succeed before merging so that one is fine. Unfortunately the GitHub pipelines will stay stuck forever waiting on the build to run even though we have these paths set to ignore. E.g Github waiting for required pipelines to succeed but they never run because they're skipped. You can see the required status checks here: https://github.com/microsoft/mxc/rules/16396467 under `Require status checks to pass`.
> 
> We could remove the `required` tags from these Github pipelines in the settings to prevent this, as I believe the pipelines would show up as skipped in this case for .md only PRs. That said, we'd have to be very careful approving pull requests in the future since that would allow PR builds to fail but allow someone to still check it in regardless.
> 
> OR
> 
> We can add a bypass to the PR rule. Note: the PR rule is really just to make sure the pipelines are required to pass at PR time. There is a Microsoft rule that PRs require at least 1 approver which we also have and isn't bypassable [here](https://github.com/microsoft/mxc/rules/5351760), so effectively bypassing would be just for the PR rule and we can scope it to only a few people with it documented that it should only be used for document changes.
> 
> I'm good with either two. Thoughts?
> 
> Note: This one also has the same issue: #545

Is a third option to just run the same gates even for 'none code' PRs? If so, I think that is our least worst option. Yes, it means some unnecessary checks run, but it means that we don't have to worry about PRs being merged that have not cleared gates.

bbonaby:
yea that's cool with me: Can you approve https://github.com/microsoft/mxc/pull/549, merge it and then merge with main? That should get the pipelines running again.

PR #542, opened 2026-06-19
Add dedicated --allow-testing-features gate for builtinTestServer

MGudgin:
## Summary

This PR fixes a fail-open parity gap where the builtin test proxy
(`network.proxy.builtinTestServer`) could be activated on the Windows
process-container backends (AppContainer and BaseContainer) without any gate,
while bubblewrap gated it behind the overloaded `--experimental` flag. It
introduces a dedicated testing-only axis, `--allow-testing-features`, enforced
uniformly across all backends.

The root cause was that `--experimental` is a single overloaded axis (backend
admission + experimental features + testing-only helpers) — on the Windows
process-container family it is even self-contradictory as a gate, since it also
flips AppContainer to BaseContainer. The fix decouples "not-for-production
testing scaffolding" into its own flag.

## Details

* Add `ExecutionRequest.testing_features_enabled`, wired from a new
  `--allow-testing-features` flag on `wxc-exec`, `lxc-exec`, and `mxc-exec-mac`.
* Enforce the gate centrally in `validate_common` (called by `ScriptRunner::run`
  for every backend): reject `network.proxy.builtinTestServer` unless the flag
  is set. This is a distinct axis from `--experimental` ("unstable/new") versus
  "not-for-production testing scaffolding".
* Remove bubblewrap's now-redundant local `--experimental` gate; the central
  check covers it.
* The SDK forwards `--allow-testing-features` automatically when a one-shot
  policy sets `builtinTestServer`, preserving SDK ergonomics while the direct
  CLI/`wxc-exec` surface stays fail-closed.
* `wxc-test-driver` auto-passes the flag for configs that use
  `builtinTestServer`; the bwrap proxy test script is updated.
* Docs updated: bubblewrap backend, examples, schema, policy v1, and SDK type
  JSDoc.

## Tests

* `cargo fmt --check`, `cargo check --workspace --all-targets`, and
  `cargo clippy --workspace --all-targets -- -D warnings` all pass.
* Rust unit tests pass, including two new `validate_common` tests (reject
  without the flag / accept with it) and an updated bwrap test (wxc_common 340
  passed; plus wxc, lxc, appcontainer_common, wxc_test_driver).
* SDK: `tsc` build clean and `npm test` 178 passed, including two new
  flag-forwarding tests.
* macOS `mxc-exec-mac` compiles on Windows but its runtime path was not
  exercised on this host.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/542)

### DIFF
--- a/sdk/src/helper.ts
+++ b/sdk/src/helper.ts
@@ -267,7 +267,28 @@ export function resolveExecutableAndArgs(
     }
   }
 
-  return resolveBinaryAndCommonArgs(JSON.stringify(config), options);
+  // `network.proxy.builtinTestServer` is testing-only, deliberately-permissive
+  // scaffolding that the native binary gates behind `--allow-testing-features`.
+  // Mirror that fail-closed posture at the SDK boundary: the caller must opt in
+  // explicitly via `allowTestingFeatures` (a distinct axis from `experimental`).
+  // Forwarding the flag automatically whenever the policy used the feature would
+  // make the gate meaningless — requesting the dangerous feature would silently
+  // enable the gate that is supposed to guard it.
+  const proxy = config.network?.proxy as { builtinTestServer?: boolean } | undefined;
+  const usesBuiltinTestServer = proxy?.builtinTestServer === true;
+  if (usesBuiltinTestServer && !options.allowTestingFeatures) {
+    throw new Error(
+      "network.proxy.builtinTestServer is a testing-only feature. Set " +
+      "'allowTestingFeatures: true' in SandboxSpawnOptions to enable it. For " +
+      "production, point network.proxy at a real HTTP proxy via 'localhost' or 'url'.",
+    );
+  }
+
+  const resolved = resolveBinaryAndCommonArgs(JSON.stringify(config), options);
+  if (usesBuiltinTestServer) {
+    resolved.args.push('--allow-testing-features');
+  }
+  return resolved;
 }
 
 /**
--- a/src/testing/wxc_test_driver/src/main.rs
+++ b/src/testing/wxc_test_driver/src/main.rs
@@ -52,12 +52,31 @@ fn run_configs(config_path: &std::path::Path, debug: bool) -> anyhow::Result<()>
             cmd.arg("--debug");
         }
 
-        // Windows Sandbox is experimental — pass --experimental when the config uses it.
-        let content = fs::read_to_string(path).unwrap_or_default();
-        if content.contains("\"containment\": \"windows_sandbox\"")
-            || content.contains("\"containment\":\"windows_sandbox\"")
-        {
-            cmd.arg("--experimental");
+        // Derive the flags wxc-exec needs from the config's actual JSON fields
+        // rather than sniffing for substrings. Parsing leniently: a config that
+        // fails to parse just gets no extra flags (wxc-exec will report the real
+        // error), matching the prior best-effort behavior.
+        let config_json: Option<serde_json::Value> = fs::read_to_string(path)
+            .ok()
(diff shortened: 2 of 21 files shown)


MGudgin:
Addressed both review comments in f726343:

- **wxc-test-driver**: narrowed the match from the bare `"builtinTestServer"` substring to the concrete `"builtinTestServer": true` JSON pattern (both spacings), so it can't false-positive on the string appearing in `process.commandLine` and mask fail-closed coverage.
- **policy v1 docs**: added `{ localhost: <port> }` as a proxy option in both the type definition and the network table to match the schema/SDK wire format.

src/testing/wxc_test_driver/src/main.rs
```diff
@@ -60,6 +60,16 @@ fn run_configs(config_path: &std::path::Path, debug: bool) -> anyhow::Result<()>
             cmd.arg("--experimental");
         }
 
+        // builtinTestServer is testing-only scaffolding gated behind
+        // --allow-testing-features — pass it when the config opts in. Match the
+        // concrete JSON pattern (both spacings) rather than the bare key name so
+        // the substring can't false-positive from e.g. a process.commandLine.
+        if content.contains("\"builtinTestServer\": true")
+            || content.contains("\"builtinTestServer\":true")
```
jsidewhite: kind of a messy json parsing :P

sdk/src/helper.ts
```diff
@@ -267,7 +267,19 @@ export function resolveExecutableAndArgs(
...
+  // native binary gates behind `--allow-testing-features`. A caller that sets
+  // it in the policy has already explicitly opted in, so forward the flag
+  // automatically to keep the SDK surface ergonomic while the CLI/wxc-exec
+  // surface stays fail-closed.
+  const proxy = config.network?.proxy as { builtinTestServer?: boolean } | undefined;
+  const usesBuiltinTestServer = proxy?.builtinTestServer === true;
+
+  const resolved = resolveBinaryAndCommonArgs(JSON.stringify(config), options);
+  if (usesBuiltinTestServer) {
+    resolved.args.push('--allow-testing-features');
```
jsidewhite, line 289: If we are adding `--allow-testing-features` implicitly for the caller when they have ask to use the built in test proxy, then aren't we undermining the `--allow-testing-features` gate itself?

jsidewhite: e.g.

caller: `use dangerous feature x`
impl: `enabling all-dangerous-features gate`

jsidewhite (dismissed):
:shipit:

sdk/src/helper.ts
MGudgin: Good catch — this was undermining the gate. Fixed in 1957dfe: the SDK no longer auto-forwards `--allow-testing-features`. Callers now opt in explicitly via a new `allowTestingFeatures` `SandboxSpawnOptions` flag (mirroring the existing `experimental` gate), and `resolveExecutableAndArgs` throws a clear error if `builtinTestServer` is used without it. So requesting the feature no longer silently flips the gate guarding it.

src/testing/wxc_test_driver/src/main.rs
MGudgin: Agreed — cleaned up in 1957dfe. The test-driver now parses each config as JSON (`serde_json`) and reads the real `containment` and `network.proxy.builtinTestServer` fields instead of sniffing for substrings.

jsidewhite (approved):
:shipit:

jsidewhite (approved):

PR #547, opened 2026-06-22
Make the Windows process-container test harness capability-driven

MGudgin:
## 📖 Description

Reworks the Windows process-container test harness to be **capability-driven**:
it runs the correct checks on any Windows build (Nickel 23H2 / Germanium
24H2+25H2) by deriving the expected isolation tier and per-limit expectations
from `wxc-exec --probe` at runtime, instead of branching on the OS version.
Renames the harness accordingly, adds a working
`JOB_OBJECT_UILIMIT_INJECTION` probe, and makes the result output unambiguous.

Highlights:

* **Capability-driven harness + rename.** Tier and per-limit expectations come
  from `--probe`; renamed `Win25H2Safe-Tests.ps1` → `WinProcessContainer-Tests.ps1`
  and updated stale references (`T3-Workloads.ps1`, `wxc` main.rs,
  `filesystem_dacl.rs`).
* **BaseContainer (Tier 1) support.** Tier-aware telemetry/assertions and a new
  `probes.baseContainerSupportsDenyPaths` fact (`SANDBOX_CAP_DENY_PATHS`) so
  `deniedPaths` tests auto-skip where the BaseContainer tier can't yet enforce
  them and auto-enable when the capability ships. Fixed a T1 `New-Config`
  `$null`-array-collapse crash under StrictMode.
* **HANDLES probe** switched from `GetWindowTextW` to `GetWindowThreadProcessId`
  so it reads window-manager state directly and isn't confounded by UIPI or the
  target pumping messages.
* **INJECTION probe** (`JOB_OBJECT_UILIMIT_INJECTION`, build 26100+) now creates
  and foregrounds its **own** top-level window before `SendInput`. The kernel's
  `DoInputCheck` evaluates the foreground-accessible check *before* the injection
  job-limit check and silently skips the input (returning success) when the
  foreground belongs to another inaccessible process, so a contained process on
  an interactive desktop would otherwise misread enforcement as "not enforced".
  Owning the foreground (the kernel permits injecting into one's own window)
  makes the limit actually evaluate. New `INJECTION=INCONCLUSIVE` outcome →
  harness SKIP when the foreground can't be owned, rather than a false verdict.
* **Result reporting.** `Record-Result` gains `skip`/`warn` statuses
  (`[SKIP]`/`[WARN]`, non-failing, yellow) so a not-applicable or not-enforced
  check is never shown as a green `[PASS]`; the summary and JSON carry
  skipped/warnings counts. Probe `PASS`/`FAIL` tokens are translated to semantic
  verbs in output (blocked/allowed for UI/atom probes, allowed/denied for the
  filesystem matrix) so a green `[PASS]` line never contains the word `FAIL`.
* Dropped stale "T3 forced" labels from the Phase 4b/4c headers (those phases run
  on the host's baseline tier).

> **Note:** This PR targets `user/gudge/job_limit_test_fixes` (the head of
> #544). It should be rebased onto `main` and retargeted once #544 merges.

## 🔗 References

- Builds on #544.

## 🔍 Validation

* `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, and
  `cargo clippy --workspace --all-targets -- -D warnings`: all clean.
* `cargo test`: `appcontainer_common` 126 passed, `wxc_common` 338 passed, 0
  failures.
* Direct verification on build 26200: a process self-assigned to a job with
  `JOB_OBJECT_UILIMIT_INJECTION` is blocked by `SendInput`
  (`gle=ERROR_ACCESS_DENIED`), confirming the OS enforces the limit; the
  reworked probe run contained via `wxc-exec` reports `INJECTION=PASS`
  ("owns foreground", injected 0/1 gle=5) where the old probe falsely reported
  not-enforced.
* Harness parse-checked via `[Parser]::ParseFile`; unit-tested the INJECTION
  verdict gate (enforced / not-enforced / inconclusive / build-gated) and the
  verdict-translation helpers (UI / filesystem / atom token sets, plus DIAG and
  `<missing>` pass-through).
* Validated end-to-end on interactive 23H2 (22631), 25H2 (26200), and 25H2+
  (26634) machines — INJECTION correctly skips when the foreground is stolen
  mid-run and passes when left alone. Full Phase 4b/4c not run on the shared dev
  host (EXITWINDOWS carries a logoff risk).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/547)

### DIFF
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -99,6 +99,13 @@ type PfnQuerySandboxSupport = unsafe extern "system" fn(capabilities: *mut u64) 
 /// `SANDBOX_CAP_CREATE_PROCESS_IN_SANDBOX`: when clear, Tier 1 is unusable.
 const SANDBOX_CAP_CREATE_PROCESS_IN_SANDBOX: u64 = 0x0000_0000_0000_0001;
 
+/// `SANDBOX_CAP_DENY_PATHS`: when set, the BaseContainer (Tier 1) backend can
+/// enforce `filesystem.deniedPaths`. Bit 0 reports whether the create API is
+/// callable; deny support is expected to light up on a separate bit (currently
+/// assumed bit 1). When clear, `deniedPaths` is rejected at launch and callers
+/// must rely on default-deny plus explicit `readwrite`/`readonly` grants.
+const SANDBOX_CAP_DENY_PATHS: u64 = 0x0000_0000_0000_0002;
+
 /// True when a Win32 error code signals the BaseContainer feature is not
 /// enabled on this build (symbol present, capability gated off).
 pub(crate) fn is_api_not_implemented(err: u32) -> bool {
@@ -162,6 +169,34 @@ impl BaseContainerRunner {
             Some(enabled) => enabled,
             None => Self::probe_create_process_feature_enabled(),
         }
+    }
+
+    /// Whether the BaseContainer (Tier 1) backend can enforce
+    /// `filesystem.deniedPaths` on this host.
+    ///
+    /// Reads the `SANDBOX_CAP_DENY_PATHS` bit from
+    /// `Experimental_QuerySandboxSupport`. Returns `false` when the query
+    /// export is absent or the bit is clear — the behavior on builds where
+    /// BaseContainer deny support has not yet shipped, where `deniedPaths` is
+    /// rejected at launch. Tier 3 (AppContainer + DACL) enforces `deniedPaths`
+    /// via DENY ACEs independently of this bit.
+    pub fn base_container_supports_deny_paths() -> bool {
+        let Some(query) = Self::load_query_sandbox_support() else {
+            return false;
+        };
+        let mut capabilities: u64 = 0;
+        // SAFETY: `query` is the resolved export; `capabilities` is a valid
+        // out-param.
+        let succeeded = unsafe { query(&mut capabilities) };
+        Self::decode_deny_capability(succeeded, capabilities)
+    }
+
+    /// Decode a `QuerySandboxSupport` result for the deny-paths capability.
+    /// `succeeded` is the export's Win32 `BOOL` return (TRUE = nonzero = the
+    /// query call succeeded), NOT an HRESULT. The capability is present only
+    /// when the call succeeded and the bit is set.
+    fn decode_deny_capability(succeeded: i32, capabilities: u64) -> bool {
+        succeeded != 0 && (capabilities & SANDBOX_CAP_DENY_PATHS) != 0
     }
 
     /// Query `Experimental_QuerySandboxSupport` for the create-process bit.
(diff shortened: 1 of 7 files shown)


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -164,6 +171,32 @@ impl BaseContainerRunner {
...
 
+    /// Whether the BaseContainer (Tier 1) backend can enforce
+    /// `filesystem.deniedPaths` on this host.
+    ///
+    /// Reads the `SANDBOX_CAP_DENY_PATHS` bit from
+    /// `Experimental_QuerySandboxSupport`. Returns `false` when the query
+    /// export is absent or the bit is clear — the behavior on builds where
+    /// BaseContainer deny support has not yet shipped, where `deniedPaths` is
+    /// rejected at launch. Tier 3 (AppContainer + DACL) enforces `deniedPaths`
+    /// via DENY ACEs independently of this bit.
```
jsidewhite, line 182: Can Tier 1 use the Tier 3 `deniedPaths` functionality if the 0x2 bit is not present?

(yep... asking about mixing tiers)

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -164,6 +171,32 @@ impl BaseContainerRunner {
...
+        };
+        let mut capabilities: u64 = 0;
+        // SAFETY: `query` is the resolved export; `capabilities` is a valid
+        // out-param.
+        let ok = unsafe { query(&mut capabilities) };
+        Self::decode_deny_capability(ok, capabilities)
+    }
+
+    /// Decode a `QuerySandboxSupport` result for the deny-paths capability:
+    /// present only when the call succeeded (`ok != 0`) and the bit is set.
```
jsidewhite: >`ok != 0` [](http://example.com/codeflow?start=46&length=9)

I'm confused by `ok`

`S_OK` == 0 -> pass

`ok` > 0  -> pass

?

src/testing/wxc_ui_probe/src/main.rs
```diff
@@ -486,37 +570,267 @@ fn probe_handles(user32: Hmodule, args: &ProbeArgs) {
...
+        },
+    };
+    let inserted = unsafe {
+        send_input(
+            1,
+            &input as *const Input,
+            std::mem::size_of::<Input>() as i32,
+        )
+    };
+    let gle = if inserted < 1 {
```
jsidewhite, line 827: >gle [](http://example.com/codeflow?start=8&length=3)

comment above says "gle=ERROR_ACCESS_DENIED" but you're not actually checking for access denied.

src/testing/wxc_ui_probe/src/main.rs
```diff
@@ -486,37 +570,267 @@ fn probe_handles(user32: Hmodule, args: &ProbeArgs) {
...
 
+/// Probe the INJECTION UI restriction (`JOB_OBJECT_UILIMIT_INJECTION`, build
+/// 26100+). The limit blocks a job process from injecting synthetic input via
+/// `SendInput`, enforced in the kernel against the caller.
+///
+/// Subtlety that this probe must work around: in the kernel's input check
+/// (`DoInputCheck`), the **foreground-accessible** test runs *before* the
+/// injection job-limit test, and `xxxSendInput` silently `continue`s (skips the
+/// input, returns success, leaves `GetLastError` untouched) when the foreground
+/// window belongs to another process the caller cannot access. A contained
```
jsidewhite, line 626: AlexSn should review :P

jsidewhite (dismissed):
:shipit:

src/backends/appcontainer/common/src/base_container_runner.rs
MGudgin: No, we don't mix T1 and T3. On a machine where 0x2 is set, we'll use BFS deny paths for T1. We could support deny ACEs if BFS deny paths are not available, but I expect the window where that will be true to be relatively short.

src/backends/appcontainer/common/src/base_container_runner.rs
MGudgin: Yeah, it's a Win32 BOOL not an HRESULT. Renamed to succeeded.

src/testing/wxc_ui_probe/src/main.rs
MGudgin: Fixed.

src/testing/wxc_ui_probe/src/main.rs
MGudgin: Agreed.

src/backends/appcontainer/common/src/base_container_runner.rs
jsidewhite: yeah, thinking about it for 4 more seconds, it would be madness

jsidewhite (approved):
:shipit:

PR #551, opened 2026-06-23
Add filesystem policy path validation (same-path conflict, existence check)

SohamDas2021:
## 📖 Description
Validate filesystem paths across `readwritePaths`, `readonlyPaths`, and `deniedPaths` at config parse time:

1. **Same-path conflict**: most restrictive path wins — `deny` > `r/o` > `r/w`. If a path appears in multiple lists, it is kept only in the most restrictive one and a diagnostic log line is emitted.
2. **Existence warning**: logs a `WARNING` if a listed path does not exist on the host (advisory; not a hard error, since some backends create mount targets dynamically).

Object-based conflict detection (inode/canonicalization deduplication) has been deferred to a follow-up and will be implemented closer to the enforcement point in each backend runner, per reviewer feedback.

All checks live in `wxc_common` and apply to every backend.

## ✅ Checklist

- <a href="https://cla.opensource.microsoft.com"> ] Signed the [Contributor License Agreement</a>
- [ ] Linked to an issue
- <a> ] Updated [Copilot instructions</a> (if build, architecture, or conventions changed)

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

on a PR with `/azp run`. See <a>docs/pull-requests.md</a>.
 ###### Microsoft Reviewers: <a href="https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/551">Open in CodeFlow</a>


jsidewhite:
Can you add Daniel Dudugjian to this PR?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -612,6 +612,175 @@ fn validate_paths(paths: &[String], logger: &mut Logger) -> Result<(), WxcError>
...
+fn file_identity(path: &str) -> Option<(u64, u64)> {
+    use std::os::unix::fs::MetadataExt;
+    fs::metadata(path).ok().map(|m| (m.dev(), m.ino()))
+}
+
+#[cfg(windows)]
+fn file_identity(path: &str) -> Option<(u64, u64)> {
+    use std::os::windows::io::AsRawHandle;
+
+    let file = fs::File::open(path).ok()?;
```
MGudgin: >let file = fs::File::open(path).ok()?; [](http://example.com/codeflow?start=4&length=38)

This fails if path refers to a directory, the underlying API call returns ACCESS_DENIED unless FILE_FLAG_BACKUP_SEMANTICS is specified (yes, I was surprised too...)

src/core/wxc_common/src/config_parser.rs
```diff
@@ -612,6 +612,175 @@ fn validate_paths(paths: &[String], logger: &mut Logger) -> Result<(), WxcError>
...
+                let entry = max_rank.entry(id).or_insert(0);
+                if rank > *entry {
+                    *entry = rank;
+                }
+            }
+        }
+    }
+
+    policy.readwrite_paths.retain(|path| {
+        if let Some(id) = file_identity(path) {
```
MGudgin: >file_identity [](http://example.com/codeflow?start=26&length=13)

If this fails (see above comment on directories), we will *keep* the path, not drop it.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -612,6 +612,175 @@ fn validate_paths(paths: &[String], logger: &mut Logger) -> Result<(), WxcError>
...
+                     more restrictive entry; applying most-restrictive intent",
+                    path
+                ));
+                return false;
+            }
+        }
+        true
+    });
+    policy.readonly_paths.retain(|path| {
+        if let Some(id) = file_identity(path) {
```
MGudgin: Same issue here. We should only call file_identity once per path in the initial loop at lines 746-752 and store the result, doing a lookup here, rather than a new file_identity call.

MGudgin: Moving to only calling file_identity once per path will also help perf.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -612,6 +612,175 @@ fn validate_paths(paths: &[String], logger: &mut Logger) -> Result<(), WxcError>
     Ok(())
 }
 
+/// Returns a platform-specific identity for the filesystem object a path points to.
```
MGudgin: Despite being posted here, this comment was written *after* all the comments below. 

I actually don't think the config_parser should be doing any file I/O at all. I think this layer should only deal with string paths. So it shouldn't be dealing with de-duping different paths that refer to the same object at all. It can still do the 'you specified the same path in two different groups, so this one wins' work.

The file IO should probably be *much* closer to where the policy is actually applied, perhaps as low as BFS itself for Tier 1. Part of the reason for this is that the results could change between us doing checks in MXC and, in the BaseContainer case, BFS actually doing the enforcement. (e.g. two paths refer to distinct objects when we do the check in MXC, but by the time the container exists, something has tuned one of the paths into a junction/symlink to the other). While that window will, I suspect, always exist, we can shrink it by doing the IO checks closer to the enforcement point.

MGudgin: Note, for some container types we won't be able to push it all the way down into the OS, we'll have to do the work in the relevant runner.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -612,6 +612,175 @@ fn validate_paths(paths: &[String], logger: &mut Logger) -> Result<(), WxcError>
...
+#[cfg(windows)]
+fn file_identity(path: &str) -> Option<(u64, u64)> {
+    use std::os::windows::io::AsRawHandle;
+
+    let file = fs::File::open(path).ok()?;
+    let mut info: windows::Win32::Storage::FileSystem::BY_HANDLE_FILE_INFORMATION =
+        unsafe { std::mem::zeroed() };
+    let handle = windows::Win32::Foundation::HANDLE(file.as_raw_handle());
+    let ok = unsafe {
+        windows::Win32::Storage::FileSystem::GetFileInformationByHandle(handle, &mut info)
```
MGudgin: This needs to be GetFileInfomationByHandleEx and we need to move to 128-bit ids for the file index, ReFS/Dev Drive use 128-bit identifiers.

MGudgin (changes requested):

src/core/wxc_common/src/config_parser.rs
SohamDas2021: Thanks! That is a good point and I have noted it.

I am removing the object-based validation from this PR. We may have to duplicate the logic across runners. I will do that as a follow up.

SohamDas2021:

MGudgin (approved):
:shipit:

PR #559, opened 2026-06-24
Phase 2C: generate the SDK wire types from a Rust emitter

MGudgin:
## Summary

Phase 2C of the versioning remediation. Closes the third drift surface - the Rust wire model, the generated JSON schema, and the SDK TypeScript types - by emitting the SDK wire types from a hand-written Rust emitter (`wxc_common::ts_emit`, exposed via `mxc_schema_gen --ts`) with **no new npm or cargo dependency**, asserting the hand-written public types conform to them, and retiring the hand-maintained `0.5.0-alpha-strict` stable schema view.

A wire-model change now ripples to all three surfaces or fails CI; it cannot drift silently.

### What changed
- **Rust emitter** `src/core/wxc_common/src/ts_emit.rs` walks the generated schema value and emits `sdk/src/generated/wire.ts` (an internal drift oracle, not public API). `wire.rs` is refactored to share a `schema_value()` helper; the JSON schema output stays byte-for-byte identical.
- **`mxc_schema_gen --ts`** mode emits the committed TS; the existing schema mode is unchanged.
- **Conformance test** `sdk/tests/unit/wire-conformance.test.ts` asserts (at `tsc` time) that `sdk/src/types.ts` conforms to the generated wire shape - **bidirectional and exact**: catches public/wire field rename/removal, wire-only field additions the SDK forgot to expose, and enum narrowing *and* widening, at the root and every leaf. Every intentional divergence is pinned to a documented allow-list.
- **CI gate** `scripts/versioning/check-sdk-types-codegen.js` runs the emitter and diffs the committed `wire.ts` (mirrors `check-schema-codegen.js`); wired into the Versioning Checks job.
- Retires `schemas/stable/mxc-config.schema.0.5.0-alpha-strict.json` and updates the three doc references.

### Review
Converged through two independent, model-diverse adversarial review loops (GPT-5.3-Codex and GPT-5.5) before this PR; the bidirectional/exact conformance design is the result of findings from both.

### Verification
- `cd sdk && npm test` -> 181 pass (incl. the conformance oracle)
- `cargo test -p wxc_common --features schema-gen ts_emit` -> 3 pass; clippy + fmt clean
- `node scripts/versioning/check-sdk-types-codegen.js` -> OK
- `node scripts/versioning/check-schema-codegen.js` -> OK (schema byte-identical)
- `node scripts/versioning/validate-configs.js` -> 169/169

### Follow-up (non-blocking)
A companion conformance block for the state-aware public types (`sdk/src/state-aware-types.ts` vs the `IsolationSession*` wire defs), flagged by both reviewers.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/559)


MGudgin:
Addressed both review comments in 74a22c0 (doc/label accuracy, no behavior change):
- `ts_emit.rs` module doc reworded so the `serde_json::Value` is described as built from the `MxcConfig` wire model (the same value `generate_config_schema_json` renders to JSON text), not `produced by` that String-returning function.
- Versioning Checks step renamed to `...in sync with the Rust wire model...` to match the sibling schema-codegen step and reflect that the gate regenerates `wire.ts` from the wire model via `mxc_schema_gen --ts`.

bbonaby:
To generate typescript types from the rust types should we be using https://crates.io/crates/ts-rs ?

schemas/stable/mxc-config.schema.0.5.0-alpha-strict.json
```diff
@@ -1,250 +0,0 @@
-{
-  "$schema": "http://json-schema.org/draft-07/schema#",
```
bbonaby, line 2: note: we'll need to update the sdk readme https://github.com/microsoft/mxc/blob/e1d1e4cf5a4a6392d461c555bd11f55414d30eb2/sdk/README.md?plain=1#L60 to remove the link to this file.

scripts/versioning/check-sdk-types-codegen.js
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env node
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// SDK wire-types codegen gate (Phase 2C, option C): the committed
```
bbonaby: note: we should probably remove mentions of the phases and options from the comments.

sdk/src/generated/wire.ts
```diff
@@ -0,0 +1,530 @@
...
+  test?: TestFeature | null;
+  /**
+   * Windows Sandbox backend config.
+   */
+  windows_sandbox?: WindowsSandbox | null;
+  /**
+   * WSL container backend config.
+   */
+  wslc?: Wslc | null;
+  [k: string]: unknown;
```
bbonaby, line 74: question: what's this for?

docs/schema-codegen.md
```diff
@@ -102,12 +102,30 @@ rules — gaps consciously owned by the parser. The equivalence is not a
 one-time review: the codegen gate regenerates the schema from the types on every
 CI run, and the corpus gate pins the accept-side behavior.
 
+## Generated SDK types (drift oracle, Rust emitter)
+
+The SDK's wire TypeScript types are generated too — by a **Rust emitter**, with
+no third-party generator. `mxc_schema_gen --ts` walks the same generated schema
```
bbonaby, line 108: question: I'm still wondering why we can't have a third party generator? Wouldn't that ease our maintenance burden here? To be honest thinking about it, it makes a lot of sense if the types we expose in type script are generated from the wire rust code. I'm trying to think of a reason why we wouldn't want to do that but can't. Would be great if us devs only need to run a script to update the typescript types and then check in the changes.

side note: I wonder if we should have just used flatbuffers for our schema. It allows json conversion between the flatbuffer schema and json. Just wondering since that would have done the work for us in generating both rust types and typescript types without us making anything new.

scripts/versioning/check-sdk-types-codegen.js
MGudgin: Agreed

sdk/src/generated/wire.ts
MGudgin: It makes the experimental section of the JSON 'open' (allows additional fields that are not defined here).

schemas/stable/mxc-config.schema.0.5.0-alpha-strict.json
MGudgin: Agreed

docs/schema-codegen.md
MGudgin: See (newly added) rational in the schema-codegen.md, but in short, a 3p generator added a lot of crate dependencies, and we already generate the wire.ts from the Rust. The thing we don't auto-gen is the public API surface of the TS SDK, and I think that's the right call.

bbonaby: I hear, I'm just a bit worried about maintenance of our own solution is all. Was hoping there was a library we could leverage just for this purpose, since it seemed like a general problem.

bbonaby (approved):

PR #566, opened 2026-06-24
Phase 2B: deserialize the parser into the wire model

MGudgin:
## Summary

This PR changes the config parser to deserialize directly into the typed wire model (`wxc_common::wire::MxcConfig`) instead of an intermediate set of permissive structs, so the JSON schema source of truth and the parser's trust boundary share a single definition of the wire shape and cannot drift.

### What changed
- The parser deserializes both one-shot and state-aware requests straight into `wire::MxcConfig`, discriminating state-aware requests by the `phase` key and keeping the `experimental` block as a raw value for per-backend dispatch typing.
- The wire structs are the deserialization target; the intermediate structs and their conversion helpers are removed (a large net reduction in `config_parser.rs`).
- The boundary is strict by construction: invalid `containment` / `network` / `clipboard` / `configurationId` values are rejected, and unknown nested fields are rejected (`deny_unknown_fields`). Deprecated spellings (`appcontainer`, `macos_sandbox`, `appContainer`) are accepted via serde aliases.
- The `wire` module is always compiled as the parser target; its `JsonSchema` derive stays behind the `schema-gen` feature.
- The TypeScript SDK and the in-process `mxc-sdk` crate emit only fields the wire objects define: the container id is carried at top-level `containerId` and the teardown / preserve flags at `lifecycle.destroyOnExit` / `lifecycle.preservePolicy`, so `processContainer.name`, `lxc.containerName`, `lxc.destroyOnExit`, and `filesystem.clearPolicyOnExit` are no longer sent.
- Documentation (`versioning.md`, `authoring-a-new-feature.md`, `schema-codegen.md`, `copilot-instructions.md`, and the state-aware lifecycle docs) describes the wire-model parser.

### Verification
- `cargo test --workspace`; `cargo fmt --all -- --check`; `cargo clippy --workspace --all-targets -- -D warnings`
- schema codegen + config corpus (169/169) + schema-version gates
- `cd sdk && npm test`

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/566)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -163,7 +163,7 @@ State-aware lifecycle:
 New features go under the `experimental` JSON section and are only active when `--experimental` is passed. See `docs/authoring-a-new-feature.md` for the full checklist. The pattern:
 
 1. Add the field to the Rust wire model (`src/core/wxc_common/src/wire.rs`) under the `Experimental` section, then regenerate the dev schema (`cargo run --manifest-path src/Cargo.toml -p mxc_schema_gen -- schemas/dev/mxc-config.schema.<dev>.json`) — do not hand-edit the generated schema
-2. Add the matching field to the wire model's `Experimental` struct (`src/core/wxc_common/src/wire.rs`) and the domain `ExperimentalConfig` in `models.rs`, then map wire→domain in `config_parser.rs`
+2. Add the matching field to the wire model's `Experimental` struct (`src/core/wxc_common/src/wire.rs`) and the domain `ExperimentalConfig` in `models.rs`, then map wire→domain in `config_parser.rs` (use `From` impls beside the domain type for trivial enum/struct conversions)
 3. Guard execution behind `if request.experimental_enabled` in the runner
 4. Never modify files in `schemas/stable/` — those are immutable release artifacts
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -109,7 +109,7 @@ The Rust workspace (`src/`) implements multiple sandboxing backends behind the `
 
 ### Config flow
 
-1. User provides JSON config (file or base64) → `config_parser.rs` deserializes into intermediate `Raw*` structs → validates and maps to `ExecutionRequest` (the internal execution model in `models.rs`)
+1. User provides JSON config (file or base64) → `config_parser.rs` deserializes into the typed wire model (`wxc_common::wire`) → validates and maps to `ExecutionRequest` (the internal execution model in `models.rs`)
 2. `ExecutionRequest` includes the containment backend selection, process config, filesystem/network policies, and optional experimental features
 3. The appropriate `ScriptRunner` implementation executes the process and returns `ScriptResponse`
 
@@ -163,7 +163,7 @@ State-aware lifecycle:
 New features go under the `experimental` JSON section and are only active when `--experimental` is passed. See `docs/authoring-a-new-feature.md` for the full checklist. The pattern:
 
 1. Add the field to the Rust wire model (`src/core/wxc_common/src/wire.rs`) under the `Experimental` section, then regenerate the dev schema (`cargo run --manifest-path src/Cargo.toml -p mxc_schema_gen -- schemas/dev/mxc-config.schema.<dev>.json`) — do not hand-edit the generated schema
-2. Add Rust structs to `models.rs` (`ExperimentalConfig`) and `config_parser.rs` (`RawExperimental`)
+2. Add the matching field to the wire model's `Experimental` struct (`src/core/wxc_common/src/wire.rs`) and the domain `ExperimentalConfig` in `models.rs`, then map wire→domain in `config_parser.rs`
 3. Guard execution behind `if request.experimental_enabled` in the runner
 4. Never modify files in `schemas/stable/` — those are immutable release artifacts
 
@@ -191,7 +191,7 @@ The workspace is organized into five top-level directories under `src/`:
 
 ### Config parser pattern
 
-The parser uses two layers of structs: `Raw*` structs (matching JSON with `#[serde(rename)]`) that deserialize permissively, then map to validated domain structs in `models.rs`. This keeps serde attributes separate from the internal model.
+The parser deserializes JSON directly into the typed wire model (`wxc_common::wire`), the single source of truth for the config shape (it also generates the JSON schema). `config_parser.rs` then maps the wire types to the validated domain structs in `models.rs`. The stable surface uses `deny_unknown_fields` (closed); the `experimental` block is permissive.
 
 ### TypeScript conventions
 
--- a/docs/authoring-a-new-feature.md
+++ b/docs/authoring-a-new-feature.md
@@ -112,67 +112,52 @@ Adding a feature may touch these files:
 
 | File | What to change |
 |------|----------------|
-| `schemas/dev/mxc-config.schema.0.8.0-dev.json` | Add `gpuIsolation` as a feature under `experimental` |
+| `src/core/wxc_common/src/wire.rs` | Add a `gpuIsolation` field + `GpuIsolation` struct to the wire `Experimental` model (the schema is generated from this) |
+| `schemas/dev/mxc-config.schema.0.8.0-dev.json` | **Generated** — regenerate with `mxc_schema_gen` after editing the wire model; do not hand-edit |
 | `src/core/wxc_common/src/models.rs` | Add `GpuIsolationConfig` struct, add field to `ExperimentalConfig` |
-| `src/core/wxc_common/src/config_parser.rs` | Add `gpuIsolation` field to `RawExperimental` |
+| `src/core/wxc_common/src/config_parser.rs` | Map the new wire field to the domain struct in `convert_wire_config` |
 | Runner (`appcontainer.rs` or `lxc_runner.rs`) | Feature logic, guarded behind `experimental_enabled` |
 | `tests/configs/` | Test config exercising your feature |
 
-## Step 1: Update the schema
+## Step 1: Add the field to the wire model (the schema source of truth)
 
-In `schemas/dev/mxc-config.schema.0.8.0-dev.json`, the `experimental` section already
-exists with `compartments` as a feature. Add `gpuIsolation` as a new
-feature with its own typed schema:
+The JSON schema is **generated** from the Rust wire model
+(`src/core/wxc_common/src/wire.rs`); you never hand-edit
+`schemas/dev/mxc-config.schema.0.8.0-dev.json`. Add your feature as a typed field
+on the wire `Experimental` struct (which is intentionally permissive — no
+`deny_unknown_fields` — so in-flux experimental shapes stay forward-compatible):
 
-```json
-"experimental": {
-  "type": "object",
-  "description": "Experimental features. Only active when --experimental is passed.",
-  "properties": {
-    "compartments": {
-      "type": "object",
-      "description": "Network compartment isolation (experimental).",
-      "properties": {
-        "namespace": {
-          "type": "string",
-          "description": "Compartment namespace identifier."
-        },
-        "isolationLevel": {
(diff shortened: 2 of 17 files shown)


src/core/wxc_common/src/wire.rs
```diff
@@ -470,6 +504,7 @@ pub struct IsolationUser {
 
 /// Canonical `$id` for the generated dev schema. Bump alongside the dev schema
 /// version/filename (see `schemas/schema-version.json`).
+#[cfg(feature = "schema-gen")]
```
bbonaby, line 508: question: I see a lot of these feature files through this file. Should we just gate the entire file with :
```rust
#![cfg(feature = "schema-gen")]
```
At the top of the file? #Resolved

src/core/wxc_common/src/config_parser.rs
```diff
@@ -476,22 +138,31 @@ pub fn load_mxc_request_with_options(
     let json_str =
         decode_request_input(input, logger, opts.is_base64).map_err(ParseError::Decode)?;
 
-    let raw: RawMxcRequest = serde_json::from_str(&json_str).map_err(|e| {
+    // Parse once into a generic value so we can (a) discriminate one-shot vs
+    // state-aware by presence of the `phase` key and (b) capture the raw
+    // `experimental` block for the state-aware path, where it is typed
+    // per-backend at dispatch time rather than at parse time.
+    let value: serde_json::Value = serde_json::from_str(&json_str).map_err(|e| {
```
bbonaby: nite: we probably want a more descriptive name than value here. #Resolved

src/core/wxc_common/src/config_parser.rs
```diff
@@ -629,59 +300,95 @@ fn validate_paths(paths: &[String], logger: &mut Logger) -> Result<(), WxcError>
     Ok(())
 }
 
-// ---------- Conversion from raw JSON to domain model ----------
+// ---------- Conversion from wire model to domain model ----------
+
+/// Convert a typed `wire::Proxy` block into the validated domain `ProxyConfig`.
+/// Exactly one of `builtinTestServer` / `localhost` / `url` may be set.
+fn convert_wire_proxy(proxy: wire::Proxy) -> Result<ProxyConfig, WxcError> {
```
bbonaby, line 307: thought: I guess all this and below will change when I update how networking works in the schema.  #Resolved

src/core/wxc_common/src/config_parser.rs
```diff
@@ -778,54 +485,165 @@ fn validate_experimental_backend_keys(
...
+        wire::IsolationConfigurationId::Composable => IsolationSessionConfigurationId::Composable,
+    }
+}
+
+/// Map the wire Entra user bundle to the domain struct.
+fn map_wire_isolation_user(u: wire::IsolationUser) -> IsolationSessionUser {
+    // Destructure (no `..`) so a new wire field fails to compile until mapped.
+    let wire::IsolationUser { upn, wam_token } = u;
+    IsolationSessionUser { upn, wam_token }
+}
```
bbonaby: issue: instead of creating free functions here you can actually implement from<T>  for enums.
We could do something like:
```rust
impl From<wire::Phase> for Phase {
    fn from(p: wire::Phase) -> Self {
        match p {
            wire::Phase::Provision => Self::Provision,
            wire::Phase::Start => Self::Start,
            wire::Phase::Exec => Self::Exec,
            wire::Phase::Stop => Self::Stop,
            wire::Phase::Deprovision => Self::Deprovision,
        }
    }
}
```
Then we can do things like:
```rust
let phase: Phase = wire_phase.into();
// OR
let phase = Phase::from(wire_phase);
// instead of 
let phase = map_wire_phase(p);
```

Doc: https://doc.rust-lang.org/std/convert/trait.From.html #Resolved

sdk/src/sandbox.ts
```diff
@@ -105,13 +105,13 @@ function buildBubblewrapConfig(
  */
 function buildLinuxProcessConfig(
     config: ContainerConfig,
-    containerId: string,
 ): ContainerConfig {
+    // The container id is carried at top-level `containerId` and the teardown
```
bbonaby: note: I think we can delete these comments since the parsing error will showcase the behavior #Resolved

docs/authoring-a-new-feature.md
```diff
@@ -198,74 +183,39 @@ pub struct ExperimentalConfig {
...
-    #[serde(rename = "allowCuda")]
-    allow_cuda: Option<bool>,
-}
-```
+## Step 3: Map the wire field to the domain model
 
-In `convert_raw_config()`, map it directly — no name matching needed. Each
-feature should own its parsing via a constructor:
+The parser deserializes JSON directly into the wire model (`wire::MxcConfig`),
+so there is no separate `Raw*` struct to add — your `wire::GpuIsolation` from
```
bbonaby: question: with this change I think we can delete this sentence since it's talking about something that doesn't exist anymore right? Would be weird for future readers to see the `Raw*` part and not know what/where it came from . #Resolved

.github/copilot-instructions.md
```diff
@@ -163,7 +163,7 @@ State-aware lifecycle:
 New features go under the `experimental` JSON section and are only active when `--experimental` is passed. See `docs/authoring-a-new-feature.md` for the full checklist. The pattern:
 
 1. Add the field to the Rust wire model (`src/core/wxc_common/src/wire.rs`) under the `Experimental` section, then regenerate the dev schema (`cargo run --manifest-path src/Cargo.toml -p mxc_schema_gen -- schemas/dev/mxc-config.schema.<dev>.json`) — do not hand-edit the generated schema
-2. Add Rust structs to `models.rs` (`ExperimentalConfig`) and `config_parser.rs` (`RawExperimental`)
+2. Add the matching field to the wire model's `Experimental` struct (`src/core/wxc_common/src/wire.rs`) and the domain `ExperimentalConfig` in `models.rs`, then map wire→domain in `config_parser.rs`
```
bbonaby: issue: See my comment in config parser.rs about how we can convert between wire -> domain right next to where either the wire struct or domain struct are defined. Either place would do. Comment is here: https://github.com/microsoft/mxc/pull/566/changes#r3472388416 

That said, I do have one other question for you though. Why not just use the wire types directly? Why do we need to go from wire type -> domain type when the wire types are already rust types and not json objects? #Resolved

docs/bwrap-support/bubblewrap-backend-plan.md
```diff
@@ -75,11 +75,19 @@ A backend-specific config can be added later under `ExperimentalConfig` if neede
...
-**File:** `src/core/wxc_common/src/config_parser.rs`
-
-- Add `RawBubblewrap` struct (placeholder — empty for now, reserved for future backend-specific fields)
-- Add `bubblewrap` field to `RawExperimental` (optional, for future use)
-- Add `"bubblewrap"` arm to containment match in `convert_raw_config_inner()`
+> **Note (superseded):** This plan predates the wire-model parser rewire. The
+> parser no longer uses `Raw*` structs or `convert_raw_config_inner`; it
+> deserializes into `wire::MxcConfig` (`src/core/wxc_common/src/wire.rs`) and maps
+> to the domain model in `convert_wire_config`. Translate the steps below to the
+> wire model — see [authoring-a-new-feature.md](../authoring-a-new-feature.md).
```
bbonaby: note: i think we can remove this note, since the steps below have been updated and there is no more `Raw*` #Resolved

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -982,70 +982,54 @@ implements one trait, the other, or both, depending on its declared participatio
...
 
-MXC's existing parser at `src/core/wxc_common/src/config_parser.rs` uses private `Raw*`
-intermediate structs that mirror the wire-format JSON shape (with serde renames to handle
-camelCase keys), then converts them into typed domain models via `convert_*` helpers
-(e.g., `RawConfig` → `ExecutionRequest`) before dispatch. The state-aware path extends this
-same pattern.
+MXC's parser at `src/core/wxc_common/src/config_parser.rs` deserializes the
+wire-format JSON directly into the typed wire model in
+`src/core/wxc_common/src/wire.rs` (`wire::MxcConfig`), then maps it into the
+typed domain models via `convert_*` helpers (`convert_wire_config` →
```
bbonaby: note: you'll probably need to update this `via "convert_*" helpers`  when you switch to using the `From<T>`  trait for conversions #Resolved

src/core/wxc_common/src/wire.rs
MGudgin: We can't gate the whole file, because the structs/enums etc are used outside of the feature. They just get the schemars when the feature is enabled (in order to generate the JSON). I will move the functions that are gated on the cfg to a submodule.

src/core/wxc_common/src/config_parser.rs
MGudgin: It's just the parsed JSON value for json_str. Maybe request would work.

src/core/wxc_common/src/config_parser.rs
MGudgin: Yes, the config_parser (which at some point should probably be called policy_parser) will need to be updated whenever we make changes to the Rust structs

src/core/wxc_common/src/config_parser.rs
MGudgin: Neat. Will do that.

sdk/src/sandbox.ts
MGudgin: Agreed.

docs/authoring-a-new-feature.md
MGudgin: Agreed.

docs/bwrap-support/bubblewrap-backend-plan.md
MGudgin: Agreed.

.github/copilot-instructions.md
MGudgin: Wire types represent the JSON, where things might be optional, or structured differently to the model types. For now, I think there is still value in having the two layers. I'm looking at ways of reducing the authoring burden.

MGudgin: versioning.md is updated with a more descriptive version of the above comment.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
MGudgin: Done

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1344,14 +961,14 @@ fn convert_raw_config_inner(
...
-    let seatbelt = raw.seatbelt.map(make_seatbelt_config);
+    let seatbelt = cfg.seatbelt.map(make_seatbelt_config);
 
     // UI section
-    if let Some(raw_ui) = raw.ui {
-        let clipboard = match raw_ui.clipboard.as_deref() {
-            Some("read") => ClipboardPolicy::Read,
-            Some("write") => ClipboardPolicy::Write,
-            Some("all") => ClipboardPolicy::All,
+    if let Some(raw_ui) = cfg.ui {
```
bbonaby, line 967: some of these ones can also have the ``from<T>``  added for them as well.

- `wire::ClipboardPolicy -> ClipboardPolicy`,
- `wire::NetworkEnforcement-> NetworkEnforcementMode`
- `wire::NetworkPolicy-> NetworkPolicy`
- `wire::Containment -> ContainmentBackend`

Then for `wire_ui_isolation_str` we'd probably want a `to_string()` function in the enum itself kind of like what we do here: https://github.com/microsoft/mxc/blob/22f8a463270fa5aa8dc38c129ab5ed7c511dfd4a/src/core/wxc_common/src/models.rs#L50

bbonaby (approved):

PR #567, opened 2026-06-24
[SDK] Ship per-platform native binaries via optional-dependency packages (#512)

### LINKED ISSUE #512: Unnecessary platform binaries are shipped inside @microsoft/mxc-sdk
Can we get these cleaned up as this also leads to increase in the download size for vscode.

Image

MGudgin:
## Summary

Makes `@microsoft/mxc-sdk` install only the host platform's native binaries by
splitting the executor payload into five per-platform packages
(`@microsoft/mxc-sdk-<os>-<arch>`) referenced as exact-pinned, os/cpu-filtered
`optionalDependencies` of a binary-free meta package. An install pulls just the
matching host's payload instead of every platform's binaries.

Presented as **six reviewable commits** (scaffold → discovery → build scripts →
meta/manifest → pipeline → tests/docs), intended to land as a single
**squash-merge**: the phases are a review-time decomposition, not independently
shippable units — dropping the bundled `bin/` for `optionalDependencies` is only
safe alongside the publishing pipeline, so the series must merge atomically to
avoid a binary-less `main`.

Closes #512

## Details

* Five platform packages: `win32-x64`, `win32-arm64`, `linux-x64`,
  `linux-arm64`, `darwin-arm64`. Intel macOS (`darwin-x64`) is intentionally not
  shipped. The `win32-arm64` manifest ships only the arch-independent payload —
  the micro-VM payload (`nanvixd` + kernel assets) is x64-only because `nanvixd`
  is an x64 binary.
* Fixes a pre-existing `wslc_common/build.rs` path bug (the committed WSLC SDK
  nupkg at repo-root `external/wslc-sdk` was resolved two levels too shallow
  after the crate moved under `src/backends/wslc/common`, so `wslcsdk.dll` was
  silently never built). With the path corrected the `wslc` feature build now
  stages `wslcsdk.dll` into the release dir so it ships in the win32 packages.
* Fail-closed runtime resolver (`platform.ts`): in an installed tree (root under
  `node_modules`, case-insensitive) the executor is sourced ONLY from
  `MXC_BIN_DIR` or the identity/version-validated installed platform package —
  planted binaries in legacy `bin/<arch>` or `src/target` paths are never
  executed, and a planted `../src/Cargo.toml` cannot re-enable dev mode. The
  bundler/CJS sibling fallback fails closed when the meta version is unreadable
  (a name-only match is not trusted). `SUPPORTED_TUPLES` is the single source of
  truth. An explicit `executablePath` is honored before the unsupported-platform
  gate (both the low-level resolver and the public `resolveExecutableAndArgs`).
* Build scripts stage each host's binaries into its platform package and
  fail-fast on a missing mandatory component. The Windows CI build derives the
  shipped payload from the manifest `files` allowlist
  (`scripts/platform-package-payload.js`), stages it preserving the
  `bin/`+`snapshots/` substructure from an absolute workspace-rooted target path,
  and hard-fails on a missing artifact. The official ADO Windows build stages the
  same manifest payload (manifest-driven, node-free), closing the gap where it
  previously copied only `signPattern` + `wslcsdk.dll`. Dev/test-only helpers
  (proxies, diagnostic console) are excluded from packages.
* The meta package drops bundled `bin/`, gains the five `optionalDependencies`,
  and ships a non-failing `postinstall` that warns when the host's platform
  package is absent OR installed without its native binary.
  `sync-platform-package-versions.js` keeps manifests + pins in lockstep
  (CI `--check`).
* Release pipeline builds, packs, and publishes the five platform packages plus
  the meta package. The filesystem-canonical release-completeness gate treats
  `sdk/platform-packages` as the source of truth: rejects an empty set, requires
  the meta to pin exactly those packages at their versions, requires every
  expected tarball, **verifies every manifest `files` payload entry is actually
  staged on disk** (so `npm pack` cannot silently omit a binary), **hard-fails on
  a malformed platform manifest**, and rejects ANY stray `.tgz` (recursive scan).
* The builtin test-proxy resolver
  (`linux_proxy_coordinator::resolve_sibling_binary`) now honors
  `MXC_TEST_PROXY_DIR` before falling back to a sibling of the executor, so the
  `builtinTestServer` E2E proxy can be sourced out-of-band on Linux — the
  dev/test-only `linux-test-proxy` is intentionally NOT shipped in the
  per-platform package, and co-locating it would trip the package-hygiene tests.
* Docs (`sdk/README.md`, `copilot-instructions.md`, playground) describe the
  per-platform delivery model, cross-arch install via npm `--os`/`--cpu`, and the
  production-vs-dev resolver order.

## Tests

* SDK unit suite: **276 tests, 269 pass / 7 skip / 0 fail** — resolver
  fail-closed paths (planted-competing-binary, unreadable-version, dev preference
  order, identity/version validation, sibling fallback), packaging matrix bound
  bidirectionally to `SUPPORTED_TUPLES`, micro-VM x64-only payload assertions,
  postinstall warn/skip/unsupported + installed-but-binary-less warning +
  `runPostinstall` CLI seam, explicit-`executablePath` escape hatch on an
  unsupported host (both resolver entry points), proxy-availability fail-closed.
* Script tests: release-completeness incl. nested-stray-`.tgz`, on-disk payload
  verification, malformed-manifest hard-fail, payload helper, version-sync.
  `sync --check` OK. `tsc --noEmit` clean.
* **Not verified locally (CI-only):** the ADO `Rust.Build.Job.yml`
  manifest-driven payload-staging step (ESRP/agent environment is pipeline-only).
  The on-disk release-completeness gate is the safety net — a deficient package
  fails the release loudly rather than publishing silently.


sdk/node/src/helper.ts
```diff
@@ -148,45 +195,60 @@ export function resolveBinaryAndCommonArgs(
...
+  // documented escape hatch on an unsupported host — honor it BEFORE the
+  // platform-support gate so a caller can run a self-built binary on, e.g.,
+  // Intel macOS without also having to set skipPlatformCheck.
   if (options.executablePath) {
     if (!fs.existsSync(options.executablePath)) {
       throw new Error(`File not found: ${options.executablePath}`);
     }
     executablePath = options.executablePath;
-  } else if (platform === 'linux') {
+    return { executablePath, args: buildCommonArgs(envelopeJson, options) };
```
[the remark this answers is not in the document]
MGudgin: Addressed: 
esolveBinaryAndCommonArgs now validates the explicit xecutablePath via the shared erifyExecutable() helper (rejects a directory, and on non-Windows a non-executable non-.exe file) and throws an actionable `Not an executable file: …` before returning, rather than only xistsSync. Covered by a new unit test (
ejects an executablePath that exists but is not a runnable file).

.azure-pipelines/templates/Package.NpmSdk.Job.yml
```diff
@@ -96,20 +103,64 @@ jobs:
       workingDirectory: $(sdkDirectory)
       displayName: npm run build
 
-    - script: npm pack
+    # Verify platform package versions and pins are in sync before packing
+    # (CI must not silently mutate tracked files — fail on drift instead).
+    - script: node scripts/sync-platform-package-versions.js --check
```
bbonaby: issue: I think this is might be too late in terms of erroring out. Usually takes ~14 minutes for the build to finish in ADO so I'm thinking these checks should happen in parallel to the build or earlier so we get the errors quicker.

OR do we need a full build completed before we do this check? If so you can ignore this comment.

sdk/src/platform.ts
```diff
@@ -338,96 +410,236 @@ function isSeatbeltAvailable(): boolean {
...
+  // as production, regardless of any sibling marker. A planted/typosquatted
+  // `../src/Cargo.toml` in the install tree must NOT be able to flip dev mode and
+  // re-open the unvalidated dev fallbacks (the executor is the sandbox boundary).
+  // Case-insensitive: Windows/macOS filesystems are case-insensitive, so a
+  // `Node_Modules` segment must classify as production too.
+  if (root.split(/[\\/]+/).some((seg) => seg.toLowerCase() === 'node_modules')) {
+    return false;
+  }
+  try {
+    return fs.existsSync(path.join(root, '..', 'src', 'Cargo.toml'));
```
dhoehna: [Node Documentation](https://nodejs.org/api/fs.html#fsexistssyncpath) does not mention that `existsSync` throws.

Return false on an error will hide any legitimate error that might occur.

sdk/src/platform.ts
```diff
@@ -338,96 +410,236 @@ function isSeatbeltAvailable(): boolean {
...
-    if (verifyWxcExecutable(wxcPath)) {
-      return wxcPath;
+/**
+ * Validate that the package.json resolved for the platform package is genuinely
+ * our package at the expected version, guarding against an unrelated same-named
+ * package resolved from an ancestor `node_modules` being executed as the sandbox
+ * binary. Returns the package directory when valid, otherwise `null`. Exported
+ * (test-only) so the identity/version guard can be exercised directly.
+ */
+export function _validatePlatformPackageDir(pkgJsonPath: string): string | null {
```
dhoehna: Is the starting underscore here on purpose?

sdk/src/platform.ts
```diff
@@ -338,96 +410,236 @@ function isSeatbeltAvailable(): boolean {
...
+  if (platformPackageDirOverride !== undefined) {
+    return platformPackageDirOverride;
+  }
+  const name = getPlatformPackageName();
+  const req = getSdkRequire();
+  if (req) {
+    try {
+      return _validatePlatformPackageDir(req.resolve(`${name}/package.json`));
+    } catch {
+      // Not resolvable via node_modules require — fall through.
```
dhoehna, line 755: I'm new to node and typescript.  Why have a catch with nothing as opposed to no try/catch?  Does an empty catch offer something that is not present when no try/catch is used?

sdk/tests/integration/linux-bubblewrap.test.ts
```diff
@@ -90,6 +91,8 @@ const PROXY_SCHEMA = '0.6.0-alpha';
 describe('Linux Bubblewrap network proxy (schema 0.6.0-alpha)', {
   skip: !isLinuxBubblewrap
     ? 'Linux Bubblewrap proxy tests require Linux with bwrap installed'
+    : !isTestProxyAvailable('linux-test-proxy')
+    ? 'linux-test-proxy unavailable (excluded from shipped package per #512; set MXC_TEST_PROXY_DIR or build the Rust binaries locally)'
```
dhoehna: Please add a link to #512 in a comment.

sdk/tests/integration/package.test.ts
```diff
@@ -6,70 +6,89 @@ import assert from 'node:assert';
...
+const require = createRequire(import.meta.url);
+
 const expectedBinaries: Record<string, string[]> = {
   win32: EXPECTED_WINDOWS_BINARIES,
   linux: EXPECTED_LINUX_BINARIES,
   darwin: EXPECTED_MACOS_BINARIES,
 };
 
-describe('SDK package binaries', () => {
+// Files that legitimately sit beside the binaries in a platform package.
```
dhoehna: I am curious why you used `legitimately`.  Can files not-legitimately sit next to the binary?

sdk/tests/integration/test-helpers.ts
```diff
@@ -30,33 +32,138 @@ export const supportedVersions = [
...
+  try {
+    return path.dirname(require.resolve(`${name}/package.json`));
+  } catch {
+    // not installed; try the dev layout
+  }
+  // 2. Monorepo dev layout: walk up to the SDK root and use its
+  //    platform-packages/<tuple> dir (robust to this file's compiled location).
+  const tuple = `${process.platform}-${sdkArch()}`;
+  let dir = path.dirname(fileURLToPath(import.meta.url));
+  for (let i = 0; i < 8; i++) {
```
dhoehna: Magic number 8.  What does 8 represent here?

sdk/tests/unit/platform.test.ts
```diff
@@ -380,3 +393,350 @@ describe('isolation_session availability gate', () => {
...
+    // Wrong name → rejected.
+    fs.writeFileSync(pkgPath, JSON.stringify({ name: '@evil/squatter', version: '0.7.0' }));
+    assert.strictEqual(_validatePlatformPackageDir(pkgPath), null);
+
+    // Wrong version → rejected.
+    fs.writeFileSync(pkgPath, JSON.stringify({ name, version: '9.9.9' }));
+    assert.strictEqual(_validatePlatformPackageDir(pkgPath), null);
+  });
+
+  // --- Hermetic resolver tests (round-2 P1-2): drive the real resolution paths
```
dhoehna: Why the `---` at the beginning and end on this comment block in particular?

sdk/tests/unit/platform.test.ts
```diff
@@ -380,3 +393,350 @@ describe('isolation_session availability gate', () => {
...
+  // against fixtures via the _setSdkRequire / _setSdkPackageRoot seams. ---
+
+  const META_VERSION = '0.7.0';
```
dhoehna: Does this number need to stay in synch with any other meta versions in this repo?

sdk/tests/unit/proxy-availability.test.ts
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+//
+//   node --test dist-tests/tests/unit/proxy-availability.test.js
```
dhoehna: Does this line describe how to run the tests?  If so, why is this not in a README file?  I don't see this line declaration in any other tests.

scripts/sync-platform-package-versions.js
```diff
@@ -0,0 +1,255 @@
...
+// wrapper at the bottom.
+
+const { readFileSync, writeFileSync, readdirSync, existsSync } = require("fs");
+const { join } = require("path");
+
+/**
+ * Permissive semver check (x.y.z with optional -prerelease / +build). Avoids a
+ * dependency on the `semver` package, which this scripts/ tree does not carry.
+ */
+const VERSION_RE = /^\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
```
dhoehna: What does RE stand for?  I have not seen a variable name with Re at the end.

dhoehna (commented):
Looking through this PR I feel that some of the code can be de-duplicated.  However, I don't know enough of the build system to understand where or how to make that happen.

Otherwise the code looks sound to me.  Just a few comments.

bbonaby:
@mgudgin I'm thinking we should split this up a bit. Perhaps it would be good for us to have a design  of how we want this to look like first. When I initially thought about the optional-dependency packages, my thought was that we wouldn't need this much change. 

That said we currently use ESRP release publishing to publish a single npm package. I'd have to check if we can publish multiple packages through that service or if there is anything extra they need to configure for us on the backend to do that.

MGudgin:
> @MGudgin I'm thinking we should split this up a bit. Perhaps it would be good for us to have a design of how we want this to look like first. When I initially thought about the optional-dependency packages, my thought was that we wouldn't need this much change.
> 
> That said we currently use ESRP release publishing to publish a single npm package. I'd have to check if we can publish multiple packages through that service or if there is anything extra they need to configure for us on the backend to do that.

This PR is a squash of the following branches;

https://github.com/microsoft/mxc/tree/user/gudge/mxc-512-phase1-scaffold
https://github.com/microsoft/mxc/tree/user/gudge/mxc-512-phase2-discovery
https://github.com/microsoft/mxc/tree/user/gudge/mxc-512-phase3-build-scripts
https://github.com/microsoft/mxc/tree/user/gudge/mxc-512-phase4-meta-manifest
https://github.com/microsoft/mxc/tree/user/gudge/mxc-512-phase5-pipeline
https://github.com/microsoft/mxc/tree/user/gudge/mxc-512-phase6-tests-docs

if you want to see smaller diffs.

sdk/src/platform.ts
MGudgin: Fixed.

sdk/tests/integration/linux-bubblewrap.test.ts
MGudgin: Done

sdk/tests/integration/test-helpers.ts
MGudgin: Fixed

sdk/tests/integration/package.test.ts
MGudgin: Fixed.

sdk/tests/unit/platform.test.ts
MGudgin: Fixed.

sdk/tests/unit/platform.test.ts
MGudgin: No. Renamed to FIXTURE_VERSION.

scripts/sync-platform-package-versions.js
MGudgin: RegEx. Renamed to VERSION_REGEX

sdk/tests/unit/proxy-availability.test.ts
MGudgin: Removed.

sdk/src/platform.ts
MGudgin: empty catch will swallow exceptions from the content of the try (so any exceptions thrown by req.resolve and/or _validatePlatformPackageDir). Removing the try/catch would propagate any exception up the stack.

sdk/src/platform.ts
MGudgin: Yeah, we use _ for test/internal 'exports'

sdk/src/platform.ts
dhoehna: > Removing the try/catch would propagate any exception up the stack.
Yeah.  It would, huh.

This will lead to hard to find bugs in the future because the exception is not logged and no one is modified.  Any code that runs after `_validatePlatformPackageDir` will be in a garbage state.  At least the catch block should log the exception.  That way developers can be aware if it throws.

dhoehna:
Thank you for addressing all my comments.  I am concerned with the empty catch.  We can discuss options around it.

sdk/src/platform.ts
MGudgin: The code won't be in a garbage state, the catch block will see to that. Also see the comment at lines 576-584.

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
```diff
@@ -86,9 +96,26 @@ jobs:
...
 
       - pwsh: |
-          $sdkTgz = (Get-Item "$(sdkPkgDirectory)/*.tgz").FullName
+          $sdkTgz = Get-ChildItem "$(sdkPkgDirectory)" -Filter *.tgz | Select-Object -First 1
+          if (-not $sdkTgz) {
+            Write-Error "Meta SDK tarball (*.tgz) not found in $(sdkPkgDirectory)"
+            exit 1
+          }
+          $arch = if ([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture -eq 'Arm64') { 'arm64' } else { 'x64' }
+          $os = if ($IsWindows) { 'win32' } elseif ($IsMacOS) { 'darwin' } else { 'linux' }
```
bbonaby, line 105: question: do we pass these en variables in somewhere? `$IsWindows` and `$IsMacOS`

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
```diff
@@ -102,9 +129,36 @@ jobs:
...
+        - task: DownloadPipelineArtifact@2
+          displayName: Download linux-test-proxy
+          inputs:
+            artifact: lxc-binaries-x86_64-unknown-linux-gnu
+            path: $(Build.SourcesDirectory)/test-proxy-bin
+
+        - script: |
+            set -euo pipefail
+            find "$(Build.SourcesDirectory)/test-proxy-bin" -name linux-test-proxy -exec chmod +x {} +
+          displayName: Restore execute permission on linux-test-proxy
```
bbonaby: note: I think we can probably  combine the download of both, the proxy and the lxc-exec into one and then use a single script to readd the executable  permissions to them.

sdk/node/platform-packages/win32-x64/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "wxc-windows-sandbox-daemon.exe",
+    "winhttp-proxy-shim.exe",
+    "wxc-host-prep.exe",
+    "wslcsdk.dll",
+    "nanvixd.exe",
+    "nanvix_rootfs.img",
+    "python3.initrd",
+    "bin/kernel.elf",
+    "snapshots/kernel.vmem",
+    "snapshots/kernel.whp.cbor",
```
bbonaby: issue: I believe if I recall one of these nanvix files are 100+ MBs in size. We likely don't want to ship these either  and I don't know if nanvixd.exe is signed as well. I believe for nanvix and hyperlight support we'd want them to ship their own npm package that we can depend on rather than shipping their binaries with our stuff.

sdk/node/src/helper.ts
```diff
@@ -138,6 +150,41 @@ export function applyLinuxNetworkPolicy(config: ContainerConfig): void {
...
+ * for this CPU (rather than pointing at a package npm would refuse to install).
+ */
+function missingBinaryMessage(binary: string, extraHint?: string): string {
+  const parts: string[] = [`${binary} not found.`];
+  if (!isSupportedPlatformTuple()) {
+    parts.push(
+      `This host (${process.platform}-${process.arch}) is not a supported MXC`,
+      `target — binaries ship for win32/linux (x64, arm64) and macOS arm64 only`,
+      `(Intel macOS is not supported). Build from source, or pass an explicit`,
+      `options.executablePath, or set both skipPlatformCheck and MXC_BIN_DIR (to`,
```
bbonaby: note: below we say `set the MXC_BIN_DIR "environment variable"`.  but here and others we don't add the `environment variable part. I think adding it to the printed messages is a good idea so folks know upfront what we mean.

sdk/node/tests/integration/package.test.ts
```diff
@@ -6,70 +6,91 @@ import assert from 'node:assert';
...
 };
 
-describe('SDK package binaries', () => {
+// Files allowed to sit beside the binaries in a platform package (its tracked
+// metadata — the manifest, readme, and license — everything else must be a
+// shipped binary from the manifest's `files` allowlist).
+const ALLOWED_NON_BINARY_FILES = ['package.json', 'README.md', 'LICENSE.md'];
+
+describe('per-platform package binaries (#512)', () => {
+  // After #512 the native executor + sandbox helpers ship in the host's
```
bbonaby: note: we can probably remove the reference to the PR here

.azure-pipelines/1ES.Release.yml
```diff
@@ -58,15 +58,43 @@ extends:
...
             pipeline: MXC
             targetPath: '$(Pipeline.Workspace)/packages'
             artifactName: mxc-npm-sdk-package
 
         displayName: Publish to NPM
         steps:
+        # Publish the per-platform binary packages FIRST. The meta package pins
+        # them as exact-version optionalDependencies, so they must exist on the
+        # registry before the meta package is published.
+        - task: EsrpRelease@10
```
bbonaby, line 75: note: Will need to update our ESRP release package names in aka.ms/esrpportal to add:

```
"@microsoft/mxc-sdk-win32-x64"
    "@microsoft/mxc-sdk-win32-arm64"
    "@microsoft/mxc-sdk-linux-x64"
    "@microsoft/mxc-sdk-linux-arm64"
    "@microsoft/mxc-sdk-darwin-arm64"
``` 
Probably will take 3 or so days to propagate in their system. That said, I'm not 100 percent positive that the ESRP task can publish all tarballs in a single task though I feel like it could. We might have to test this out when we do release to see what happens.

.azure-pipelines/templates/Package.NpmSdk.Job.yml
```diff
@@ -48,32 +56,31 @@ jobs:
...
     - script: |
         set -euo pipefail
-        find $(sdkDirectory)/bin -name linux-test-proxy -exec chmod +x {} +
-      displayName: Restore execute permission on linux-test-proxy
-
-    # Drop symbol files from the npm package — symbols ship separately.
-    - script: rm -rf $(sdkDirectory)/bin/*/symbols
-      displayName: Strip symbols/ from npm package
+        find $(platformPackagesDir) -name lxc-exec -exec chmod +x {} +
+        find $(platformPackagesDir) -name mxc-exec-mac -exec chmod +x {} +
```
bbonaby, line 80: note: missing `linux-test-proxy` ?

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
```diff
@@ -102,9 +129,36 @@ jobs:
...
+          displayName: Restore execute permission on lxc-exec
+
+        # The platform package ships the executor but NOT the dev/test-only
+        # proxies (#512). Source linux-test-proxy from the raw build artifact and
+        # REQUIRE the proxy E2E tests to run (no silent green-skip).
+        - task: DownloadPipelineArtifact@2
+          displayName: Download linux-test-proxy
+          inputs:
+            artifact: lxc-binaries-x86_64-unknown-linux-gnu
+            path: $(Build.SourcesDirectory)/test-proxy-bin
```
bbonaby, line 148: question: My understanding was that we'd ship the test proxies for folks to use. If we don't want to do that anymore then that's cool too. I do notice in the package.json in win32-x64 file below we have it added.

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
```diff
@@ -102,9 +129,36 @@ jobs:
...
+
+        - script: |
+            set -euo pipefail
+            find "$(Build.SourcesDirectory)/test-proxy-bin" -name linux-test-proxy -exec chmod +x {} +
```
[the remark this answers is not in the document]
MGudgin: Fixed in 827b4a85: the pipeline now finds and chmods unix-test-proxy, and fails explicitly if the artifact is absent.

tests/playground/src/main/main.ts
```diff
@@ -526,14 +526,15 @@ ipcMain.handle('run-sandbox-raw', (_event, configJson: string, debug: boolean, e
...
+      // per-platform package binaries live under
+      // sdk/node/platform-packages/<os>-<arch>; local dev builds under
+      // src/target/<triple>/{release,debug}.
       const arch = process.arch === 'arm64' ? 'arm64' : 'x64';
       const triple = process.arch === 'arm64' ? 'aarch64-pc-windows-msvc' : 'x86_64-pc-windows-msvc';
       const repoRoot = path.join(__dirname, '..', '..', '..');
       const candidates = [
         execPath,
-        path.join(repoRoot, 'sdk', 'bin', arch),
+        path.join(repoRoot, 'sdk', 'platform-packages', `win32-${arch}`),
```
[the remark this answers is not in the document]
MGudgin: Fixed in 9c2b559a: the compiled playground path now walks four levels to the repository root and includes sdk/node/platform-packages.

build.bat
```diff
@@ -85,35 +85,30 @@ echo   Check linting
...
+        echo   Copied win32-!SDK_ARCH!\wxc-exec.exe
+        :: These components are always built for both Windows arches, so a missing
+        :: one is a real build failure — copy unguarded and let it fail-fast
+        :: rather than silently shipping an incomplete platform package.
+        for %%B in (wxc-windows-sandbox-guest.exe wxc-windows-sandbox-daemon.exe winhttp-proxy-shim.exe wxc-host-prep.exe) do (
+            copy /Y "!BIN_DIR!\%%B" "!DEST!\" >nul || goto :error
+            echo   Copied win32-!SDK_ARCH!\%%B
         )
         if exist "!BIN_DIR!\plm.exe" (
             copy /Y "!BIN_DIR!\plm.exe" "sdk\node\bin\!SDK_ARCH!\" >nul
```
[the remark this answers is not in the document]
MGudgin: Fixed in 58251fd8: plm.exe is now staged beside wxc-exec.exe in the per-platform directory, with copy failure propagated.

scripts/platform-package-payload.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @param {string} manifestPath path to a platform-package package.json
+ * @param {{ readFileSync?: Function }} [io]
+ * @returns {string[]} build-artifact payload files (relative paths)
+ */
+function payloadFiles(manifestPath, { readFileSync = fs.readFileSync } = {}) {
+  const pkg = JSON.parse(readFileSync(manifestPath, "utf8"));
+  const files = Array.isArray(pkg.files) ? pkg.files : [];
+  return files.filter((f) => typeof f === "string" && !NON_ARTIFACT_FILES.has(f));
```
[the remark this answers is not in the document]
MGudgin: Fixed in 827b4a85: payload allowlists must now be non-empty arrays of non-empty strings with at least one build artifact; the related completeness fixtures and tests were updated.

build.bat
```diff
@@ -85,64 +85,59 @@ echo   Check linting
...
+            copy /Y "!BIN_DIR!\plm.exe" "!DEST!\" >nul || goto :error
+            echo   Copied win32-!SDK_ARCH!\plm.exe
         )
         if "%WITH_NANVIX%"=="1" (
             for %%B in (nanvixd.exe nanvix_rootfs.img python3.initrd) do (
                 if exist "!BIN_DIR!\%%B" (
-                    copy /Y "!BIN_DIR!\%%B" "sdk\node\bin\!SDK_ARCH!\" >nul
-                    echo   Copied !SDK_ARCH!\%%B
+                    copy /Y "!BIN_DIR!\%%B" "!DEST!\" >nul || goto :error
+                    echo   Copied win32-!SDK_ARCH!\%%B
```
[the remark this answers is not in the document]
MGudgin: Fixed in c8351f13: --with-microvm now requires every x64 NanVix payload file and propagates any copy failure; ARM64 staging is excluded because that package has no MicroVM payload.

build.bat
```diff
@@ -85,64 +85,59 @@ echo   Check linting
...
+                    echo   Copied win32-!SDK_ARCH!\snapshots\%%S
                 )
             )
         )
         if "%WITH_WSLC%"=="1" (
             if exist "!BIN_DIR!\wslcsdk.dll" (
-                copy /Y "!BIN_DIR!\wslcsdk.dll" "sdk\node\bin\!SDK_ARCH!\" >nul
-                echo   Copied !SDK_ARCH!\wslcsdk.dll
+                copy /Y "!BIN_DIR!\wslcsdk.dll" "!DEST!\" >nul || goto :error
+                echo   Copied win32-!SDK_ARCH!\wslcsdk.dll
```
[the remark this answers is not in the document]
MGudgin: Fixed in c8351f13: --with-wslc now copies wslcsdk.dll unconditionally for each requested architecture and fails the build if it is absent.

sdk/node/src/platform.ts
```diff
@@ -668,90 +878,77 @@ function verifyExecutable(execPath: string): boolean {
...
+  // Dev: prefer local builds, then fall back to the installed package.
+  const sdkRoot = getSdkPackageRoot();
+  const targetDir = path.join(sdkRoot, '..', '..', 'src', 'target');
+  const candidates = [
+    path.join(sdkRoot, 'platform-packages', `${hostId().platform}-${getSdkArch()}`, binary),
+    path.join(targetDir, targetTriple, 'release', binary),
+    path.join(targetDir, targetTriple, 'debug', binary),
+    path.join(targetDir, 'release', binary),
+    path.join(targetDir, 'debug', binary),
+    ...(installed ? [path.join(installed, binary)] : []),
```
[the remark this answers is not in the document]
MGudgin: Fixed in b44c9b85: the IsolationSession test-bundle job now stages wxc-exec.exe into sdk/node/platform-packages/win32-x64, matching the dev resolver even with its external CARGO_TARGET_DIR.

build-mac.sh
```diff
@@ -121,17 +121,28 @@ copy_binary_for_target() {
...
+    # shipped package (no manifest); its binary stays in src/target for local
+    # dev resolution via the monorepo fallback.
+    if [ ! -f "$bin_dir/package.json" ]; then
+        echo "Skipping darwin-$sdk_arch: no platform package; binary remains in src/target."
+        return
+    fi
+
+    # Clean previously-staged binaries so stale artifacts never persist into the
+    # package; keep only the tracked metadata files.
+    find "$bin_dir" -mindepth 1 ! -name package.json ! -name README.md -delete
```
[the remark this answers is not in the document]
MGudgin: Fixed in 4eca72a1: build-mac.sh now leaves unix-test-proxy in src/target for integration-test discovery and stages only mxc-exec-mac into the publishable darwin package.

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
MGudgin: The executor and proxy intentionally remain separate downloads. `lxc-exec` is installed from the packed platform-package artifact under test, while `unix-test-proxy` is a dev/test-only binary sourced from the raw Rust build artifact and must not ship in the npm package. Keeping the downloads separate also lets the package-hygiene tests enforce that boundary. Both downloaded binaries have execute permission restored explicitly before the tests run.

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
MGudgin: Yes. These are built-in automatic variables provided by PowerShell 7 (`pwsh`) on every agent: `$IsWindows`, `$IsMacOS`, and `$IsLinux`. The step uses them to select the npm platform-package tuple, so no pipeline parameter or explicit environment variable is required.

.azure-pipelines/templates/SDK.Integration.Test.Job.yml
MGudgin: We intentionally do not ship test proxies to SDK consumers. The current package manifests exclude both `wxc-test-proxy.exe` and `unix-test-proxy`; integration tests source them from a separate build artifact through `MXC_TEST_PROXY_DIR`. The win32-x64 manifest now contains only runtime payloads, and package tests explicitly reject test-only binaries.

.azure-pipelines/templates/Package.NpmSdk.Job.yml
MGudgin: Addressed using the binary's current name, `unix-test-proxy`. The Linux Rust artifact includes it, and the SDK integration job downloads that artifact separately, locates `unix-test-proxy`, restores execute permission, and fails if it is absent. It is deliberately stripped from the publishable npm platform package.

sdk/node/src/helper.ts
MGudgin: Addressed in 45e880fe. Both recovery branches now explicitly call `MXC_BIN_DIR` an environment variable.

sdk/node/tests/integration/package.test.ts
MGudgin: Addressed in ac91f723. The integration test suite and comments now describe the permanent platform-package behavior without referencing PR #512.

.azure-pipelines/templates/Package.NpmSdk.Job.yml
MGudgin: Addressed in ac91f723. The platform-package version/pin check is now an independent job in the `Build_Binaries` stage, so it runs in parallel with the Windows, Linux, and macOS builds and can fail early. The packaging stage still waits for the build artifacts, but no longer repeats this source-only check after the build.

sdk/node/platform-packages/win32-x64/package.json
MGudgin: Addressed in the rewritten stack ending at d6779e29. NanVix is now an explicit companion package, `@microsoft/mxc-sdk-nanvix-win32-x64`, rather than part of the automatically installed Windows x64 platform package. Consumers opt in with `npm install @microsoft/mxc-sdk @microsoft/mxc-sdk-nanvix-win32-x64`; the companion carries the complete NanVix payload plus a compatible co-located executor and uses an exact SDK peer dependency. It is intentionally not an `optionalDependency`, so ordinary Windows x64 installs do not download the large payload.

.azure-pipelines/1ES.Release.yml
MGudgin: Acknowledged. The release allowlist will need the five platform package names listed here plus the new explicit NanVix companion, `@microsoft/mxc-sdk-nanvix-win32-x64`, before the first release of this layout. That ESRP portal configuration and its propagation are external rollout prerequisites rather than repository changes; the pipeline publishes the binary packages before the meta package so the release order is correct once the names are approved.

bbonaby:
Thinking about this some more I think it would be best if we came up with a design to move the npm SDK to an in proc one rather than using the executables. Likely would be to use the mxc_ffi  native library. We cal probably also do something using https://napi.rs/ as a small wrapper

PR #568, opened 2026-06-25
Phase 2.5: add state-aware SDK wire-type conformance

MGudgin:
## Summary

This PR adds a compile-time conformance oracle for the state-aware SDK types, extending the one-shot wire-conformance coverage to the lifecycle surface so a change to the state-aware wire model that the SDK forgets to mirror fails CI instead of drifting silently.

### What changed
- Add `sdk/tests/unit/wire-conformance-state-aware.test.ts`: type-level assertions that the public state-aware types in `sdk/src/state-aware-types.ts` conform to the generated wire defs — the `Phase` and `IsolationConfigurationId` enums, the Entra user bundle (`IsolationUser`), and the per-phase `IsolationSessionPhase` field set — with drift checked in **both directions** (public→wire and wire→public) so a new optional wire field the SDK omits is also caught. The per-phase configs are asserted to reuse the one-shot public leaf types (`ProcessConfig` / `FilesystemConfig`) rather than re-deriving them.
- Extract the shared assertion helpers into `sdk/tests/unit/conformance-helpers.ts` and reuse them from both the one-shot and state-aware conformance tests.
- Wire the new test into the SDK unit-test run list and document both conformance surfaces in `docs/schema-codegen.md`.

### Verification
- `cd sdk && npm test` → 185 (incl. both conformance oracles)
- Negative probes confirm teeth: a widened sizing enum, a `Phase` change, a new `IsolationSessionPhase` field, and a new optional user-bundle field each make the corresponding assertion fail to compile (drift detected).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/568)

### DIFF
--- a/sdk/tests/unit/wire-conformance-state-aware.test.ts
+++ b/sdk/tests/unit/wire-conformance-state-aware.test.ts
@@ -1,0 +1,160 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// State-aware wire-type conformance oracle (Phase 2.5).
+//
+// The one-shot oracle (`wire-conformance.test.ts`) asserts that
+// `sdk/src/types.ts` conforms to the generated wire types. This companion does
+// the same for the STATE-AWARE lifecycle public types in
+// `sdk/src/state-aware-types.ts`, against the generated wire state-aware defs
+// (`Phase`, `IsolationConfigurationId`, `IsolationUser`, `IsolationSessionPhase`).
+// Without it, a wire-model change to the state-aware surface — a new sizing
+// profile, a field added to the Entra user bundle, a `Phase` change — would
+// regenerate `wire.ts`, pass the codegen gate, and still leave the SDK silently
+// lagging with no CI signal.
+//
+// Mapping note (why this is a separate file, not part of the one-shot oracle):
+// the public per-phase call configs do NOT map 1:1 to a single wire type. Each
+// mixes SDK-level / top-level wire fields with `IsolationSessionPhase` fields:
+//
+//   public field                          wire location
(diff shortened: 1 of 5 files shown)


sdk/tests/unit/wire-conformance-state-aware.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  LiftedPhaseKey
+>;
+type WirePhaseKeys = keyof StripIndex<WireIsolationSessionPhase>;
+
+// A public phase field with no wire `IsolationSessionPhase` counterpart fails
+// (the SDK exposes a field the wire model does not define).
+type _PhasePublicKeys = AssertTrue<Equivalent<Exclude<PublicPhaseKeys, WirePhaseKeys>, never>>;
+// A wire `IsolationSessionPhase` field no phase config exposes fails (the wire
+// model gained a per-phase field the SDK forgot to surface).
+type _PhaseWireKeys = AssertTrue<Equivalent<Exclude<WirePhaseKeys, PublicPhaseKeys>, never>>;
```
huzaifa-d, line 125: Consider a short comment here telling future authors to add a value-type assertion (like _ConfigurationId) whenever they add a new phase field, since the key-set checks won't catch a type mismatch.

huzaifa-d (approved):

sdk/tests/unit/wire-conformance-state-aware.test.ts
MGudgin: See follow-up PR; https://github.com/microsoft/mxc/pull/595

PR #573, opened 2026-06-26
refactor: idiomatic cleanups in wxc_common and the AppContainer/WSLC/LXC backends

MGudgin:
## Summary

Idiomatic-Rust cleanups. No intended behavior change — only ownership,
error-signalling, and stdlib-usage improvements. This is the follow-up to the
behavior-bug PR (#570); the larger related refactors (`WxcError` typed source
errors, SDK/CLI `PathBuf` migration, `mxc_pty` typed error, the WinHTTP-shim
RAII rework) are intentionally **deferred to their own focused PRs**.

**Details**

* **Capability SID ownership.** `get_capability_sid_from_name` now returns an
  owning `OwnedCapabilitySid` (frees on drop) instead of a raw `*mut c_void`
  with a "caller must `LocalFree`" contract. The helper *and* the owner type
  were relocated out of the shared `wxc_common` foundation into the
  `appcontainer` backend (now private) — capability SIDs are an AppContainer
  concept with a single caller, so this both removes the hand-rolled
  `CapabilitySidGuard` and keeps backend-specific logic out of the foundation
  crate.
* **`string_util::sid_to_string`** returns `Option<String>` instead of taking a
  sentinel `default_value`. The `LocalAlloc`'d buffer is still read then freed
  exactly once on the success path; the four call sites map `None` to their
  prior fallback.
* **`ErrorEnvelope.code`** is now the closed `MxcErrorCode` enum rather than a
  `String`. The snake_case wire format is unchanged (serde).
* **`deserialize_config`** deserializes from the borrowed `&Value` instead of
  cloning the backend config subtree.
* **WSLC image enumeration** is allocation-free (`slice::from_raw_parts` + a
  first-NUL scan with full-buffer fallback) instead of manual `.offset(i)`
  pointer walking and a per-entry `Vec<u8>`.
* **LXC `validate_path`** drops a provably-unreachable `\n`/`\r` branch already
  covered by `char::is_whitespace`.

**Tests**

* `cargo test -p wxc_common -p appcontainer_common -p wslc_common` — 126 + 45 +
  346 pass (includes the updated `sid_to_string_null_returns_none` and the
  `ErrorEnvelope` round-trip/serialisation tests).
* `cargo clippy -p wxc_common -p appcontainer_common -p wslc_common
  --all-targets -- -D warnings` and `cargo fmt ... -- --check` clean.
* `cargo check -p wxc --all-targets` clean (full Windows backend dispatch).
* **Not built on the Windows host:** the LXC `validate_path` change
  (`lxc_common` is Linux-only). It is the removal of a provably-dead branch
  with no logic change.

## Related Issues

- (none)

### DIFF
--- a/src/backends/appcontainer/common/src/appcontainer_runner.rs
+++ b/src/backends/appcontainer/common/src/appcontainer_runner.rs
@@ -173,82 +173,93 @@ fn inject_proxy_vars(entries: &mut Vec<(String, String)>, addr: &wxc_common::mod
     entries.push(("HTTPS_PROXY".to_string(), proxy_url));
 }
 
-/// Owned capability SID derived from a capability name via
-/// [`get_capability_sid_from_name`]. Frees the underlying `LocalAlloc`'d SID
-/// with `LocalFree` on drop, so callers don't have to track and free the raw
-/// pointer themselves.
-struct OwnedCapabilitySid(PSID);
+/// Owned capability SID derived from a capability name. Frees the underlying
+/// `LocalAlloc`'d SID with `LocalFree` on drop, so callers don't have to track
+/// and free the raw pointer themselves.
+struct OwnedCapabilitySid {
+    sid: PSID,
+}
 
 impl OwnedCapabilitySid {
+    fn new(sid: PSID) -> Self {
+        Self { sid }
+    }
+
+    /// Derive the capability SID for a given capability name. Capability SIDs
+    /// are an AppContainer concept, so this helper lives in the AppContainer
+    /// backend rather than the shared foundation crate.
+    fn from_capability_name(name: &str) -> Result<Self, WxcError> {
+        let wide_name = string_util::to_wide(name);
+        let pcwstr = PCWSTR(wide_name.as_ptr());
+
+        let mut capability_sids: *mut PSID = std::ptr::null_mut();
+        let mut capability_sid_count: u32 = 0;
+        let mut group_sids: *mut PSID = std::ptr::null_mut();
+        let mut group_sid_count: u32 = 0;
+
+        // SAFETY: the Windows API writes LocalAlloc-compatible SID arrays and
+        // counts. We read only within those counts, free each returned pointer
+        // or array once, and transfer one capability SID into OwnedCapabilitySid.
+        unsafe {
+            DeriveCapabilitySidsFromName(
+                pcwstr,
+                &mut group_sids,
+                &mut group_sid_count,
+                &mut capability_sids,
+                &mut capability_sid_count,
+            )
+            .map_err(|e| {
+                WxcError::Process(format!("DeriveCapabilitySidsFromName({name}) failed: {e}"))
+            })?;
+
+            // Free group SIDs
+            for i in 0..group_sid_count {
+                let sid = *group_sids.add(i as usize);
+                let _ = LocalFree(Some(HLOCAL(sid.0)));
+            }
+            let _ = LocalFree(Some(HLOCAL(group_sids as *mut _)));
+
+            if capability_sid_count == 0 {
+                let _ = LocalFree(Some(HLOCAL(capability_sids as *mut _)));
+                return Err(WxcError::Process(format!(
+                    "No capability SID returned for {}",
+                    name
+                )));
+            }
+
+            // Take the first capability SID
+            let result_sid = *capability_sids;
+
+            // Free remaining capability SIDs (index 1+)
+            for i in 1..capability_sid_count {
+                let sid = *capability_sids.add(i as usize);
+                let _ = LocalFree(Some(HLOCAL(sid.0)));
+            }
+            let _ = LocalFree(Some(HLOCAL(capability_sids as *mut _)));
+
+            Ok(Self::new(result_sid))
+        }
+    }
+
     /// Borrow the raw `PSID` for building a `SID_AND_ATTRIBUTES` array. The
     /// pointer is valid only while `self` is alive — keep the owner around
     /// until the process-creation call has returned.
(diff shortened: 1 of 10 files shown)


src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -479,16 +479,15 @@ impl WSLContainerRunner {
...
+            let images_slice = std::slice::from_raw_parts(images, image_count as usize);
+            for info in images_slice {
+                // `info.name` is a fixed-size, possibly-unterminated C buffer;
+                // reinterpret it as bytes and read up to the first NUL without
+                // allocating. `from_bytes_until_nul` stays in-bounds.
+                let name_bytes =
+                    std::slice::from_raw_parts(info.name.as_ptr().cast::<u8>(), info.name.len());
+                if let Ok(name) = std::ffi::CStr::from_bytes_until_nul(name_bytes) {
+                    if name.to_bytes() == image_name.as_bytes() {
                         image_found = true;
```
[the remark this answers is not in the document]
MGudgin: Good catch — you're right, and it contradicted this PR's no-behavior-change goal. `info.name` is `[c_char; 256]`, and the original `take_while(|&&b| b != 0)` consumed the whole buffer when no NUL was present, so `from_bytes_until_nul` returning `Err` would skip an image whose name exactly fills the field. Fixed in 0631c9c: scan for the first NUL with `position()` and fall back to the full buffer when none is found, still allocation-free.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -173,30 +173,85 @@ fn inject_proxy_vars(entries: &mut Vec<(String, String)>, addr: &wxc_common::mod
...
+fn get_capability_sid_from_name(name: &str) -> Result<OwnedCapabilitySid, WxcError> {
+    let wide_name = string_util::to_wide(name);
+    let pcwstr = PCWSTR(wide_name.as_ptr());
+
+    let mut capability_sids: *mut PSID = std::ptr::null_mut();
+    let mut capability_sid_count: u32 = 0;
+    let mut group_sids: *mut PSID = std::ptr::null_mut();
+    let mut group_sid_count: u32 = 0;
+
+    unsafe {
```
bbonaby: issue: missing safety comment here and throughout

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -173,30 +173,85 @@ fn inject_proxy_vars(entries: &mut Vec<(String, String)>, addr: &wxc_common::mod
...
             }
         }
     }
 }
 
+/// Derive the capability SID for a given capability name, returning an owned
+/// handle that frees the SID on drop. Capability SIDs are an AppContainer
+/// concept, so this helper lives in the AppContainer backend rather than the
+/// shared foundation crate.
+fn get_capability_sid_from_name(name: &str) -> Result<OwnedCapabilitySid, WxcError> {
```
bbonaby: question: should this just be a static member function inside `OwnedCapabilitySid` rather than a free function? Seems like they go hand in hand.

src/core/wxc_common/src/mxc_error.rs
```diff
@@ -133,11 +133,12 @@ impl MxcError {
...
 
-/// Wire shape of the `error` arm. `code` is a snake_case string from
-/// `MxcErrorCode::as_str`; `details` is omitted from JSON when absent.
+/// Wire shape of the `error` arm. `code` is a closed `MxcErrorCode` that
+/// serialises to its snake_case wire string; `details` is omitted from JSON
+/// when absent.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ErrorEnvelope {
-    pub code: String,
+    pub code: MxcErrorCode,
```
bbonaby, line 141: thought: I'm trying to remember why this was a String instead of the enum error code to begin with but I can't remember. It does seem more natural to use the enum though.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -173,30 +173,85 @@ fn inject_proxy_vars(entries: &mut Vec<(String, String)>, addr: &wxc_common::mod
...
 }
 
-impl Drop for CapabilitySidGuard {
+impl Drop for OwnedCapabilitySid {
     fn drop(&mut self) {
-        for &sid in &self.0 {
+        if !self.0 .0.is_null() {
             unsafe {
-                let _ = LocalFree(Some(HLOCAL(sid)));
+                let _ = LocalFree(Some(HLOCAL(self.0 .0)));
```
bbonaby: note: for readability we should be a bit careful with nested tuple structs, since to get the inner value of them we need to use the `.#` syntax. In this case, we have to do `.#.#` to get the windows-rs  PSID value so it can be a hard read if we were doing more than just that.

I'd probably just use a regular struct, and do something like:
```rust
struct OwnedCapabilitySid {
    sid: PSID,
}

impl OwnedCapabilitySid {
    fn new(sid: PSID) -> Self {
        Self { sid }
    }

    fn as_psid(&self) -> PSID {
        self.sid
    }
}

impl Drop for OwnedCapabilitySid {
    fn drop(&mut self) {
        let ptr = self.sid.0;

        if !ptr.is_null() {
            // SAFETY:
            // `OwnedCapabilitySid` owns this SID pointer. It must have been
            // allocated by the Windows API using LocalAlloc-compatible
            // allocation.
            unsafe {
                let _ = LocalFree(Some(HLOCAL(ptr)));
            }
        }
    }
}
```

bbonaby (approved):

PR #576, opened 2026-06-26
Add shared plumbing for the Windows Sandbox rewrite (Phase 0)

MGudgin:
This commit adds the small, additive `wxc_common` pieces that the upcoming
Windows Sandbox rewrite stack builds on, landed first so the larger
lifecycle/one-shot PR stays focused.

Details

* filesystem_dacl: add `set_owner_only_dacl` (+ `owner_is_self` and the
  `current_user_sid` / `current_token_owner_sid` / `self_owner_sids`
  helpers) to stamp an owner-only, optionally-inheritable DACL on a path.
  This is the primitive the rewrite's nonce/daemon-record files use to
  keep other users off them. Fails closed before stamping owner-only
  DACLs on foreign-owned paths, since a Windows object owner can still
  rewrite the DACL and MXC cannot secure a path it does not own.
* filesystem_dacl: factor token SID lookup into a shared helper with
  explicit sizing-probe handling and SAFETY notes. Comments use ASCII
  punctuation so they do not render as mojibake in Windows tooling.
* models: add two `FailurePhase` variants -- `Rejected` (request can't be
  honored without changing input/host) and `PostLaunchFailed` (guest/
  sandbox infra failed after a successful launch) -- and sharpen the
  doc comments on the existing variants. Purely additive: the only
  `match` on `FailurePhase` (mxc-sdk `map_spawn_error`) has a wildcard
  arm, so adding variants is non-breaking.

Tests

* `cargo test -p wxc_common`: passing, including the new
  `set_owner_only_dacl_strips_inheritance_and_grants_owner`,
  `owner_is_self_*`, and `failure_phase_serde_round_trips_all_variants`
  (covers the Timeout failure phase).
* `cargo fmt -p wxc_common -- --check` and
  `cargo clippy -p wxc_common --all-targets -- -D warnings`: clean.
* `cargo check -p mxc-sdk -p appcontainer_common`: clean (verified the
  new enum variants don't break downstream consumers).

### DIFF
--- a/src/core/wxc_common/src/filesystem_dacl.rs
+++ b/src/core/wxc_common/src/filesystem_dacl.rs
@@ -77,20 +77,23 @@ use windows::Win32::Foundation::{
     WAIT_FAILED, WAIT_OBJECT_0, WAIT_TIMEOUT, WIN32_ERROR,
 };
 use windows::Win32::Security::Authorization::{
-    ConvertStringSidToSidW, GetNamedSecurityInfoW, SetEntriesInAclW, SetNamedSecurityInfoW,
-    EXPLICIT_ACCESS_W, GRANT_ACCESS, SE_FILE_OBJECT, TRUSTEE_IS_SID, TRUSTEE_IS_UNKNOWN, TRUSTEE_W,
+    ConvertSidToStringSidW, ConvertStringSidToSidW, GetNamedSecurityInfoW, SetEntriesInAclW,
+    SetNamedSecurityInfoW, EXPLICIT_ACCESS_W, GRANT_ACCESS, SE_FILE_OBJECT, TRUSTEE_IS_SID,
+    TRUSTEE_IS_UNKNOWN, TRUSTEE_W,
 };
 // DENY_ACCESS lives on ACCESS_MODE in windows 0.62
 use windows::Win32::Security::Authorization::DENY_ACCESS;
 use windows::Win32::Security::{
     AclSizeInformation, AddAccessAllowedAceEx, AddAccessDeniedAceEx, AddAce, EqualSid, GetAce,
-    GetAclInformation, GetLengthSid, InitializeAcl, IsValidSid, ACCESS_ALLOWED_ACE,
-    ACCESS_DENIED_ACE, ACE_FLAGS, ACE_HEADER, ACL, ACL_REVISION, ACL_SIZE_INFORMATION,
-    CONTAINER_INHERIT_ACE, DACL_SECURITY_INFORMATION, INHERITED_ACE, OBJECT_INHERIT_ACE,
-    PSECURITY_DESCRIPTOR, PSID,
+    GetAclInformation, GetLengthSid, GetTokenInformation, InitializeAcl, IsValidSid, TokenOwner,
+    TokenUser, ACCESS_ALLOWED_ACE, ACCESS_DENIED_ACE, ACE_FLAGS, ACE_HEADER, ACL, ACL_REVISION,
+    ACL_SIZE_INFORMATION, CONTAINER_INHERIT_ACE, DACL_SECURITY_INFORMATION, INHERITED_ACE,
+    OBJECT_INHERIT_ACE, OBJECT_SECURITY_INFORMATION, OWNER_SECURITY_INFORMATION,
+    PROTECTED_DACL_SECURITY_INFORMATION, PSECURITY_DESCRIPTOR, PSID, TOKEN_INFORMATION_CLASS,
+    TOKEN_OWNER, TOKEN_QUERY, TOKEN_USER,
 };
 use windows::Win32::Storage::FileSystem::{
-    DELETE, FILE_GENERIC_EXECUTE, FILE_GENERIC_READ, FILE_GENERIC_WRITE,
+    DELETE, FILE_ALL_ACCESS, FILE_GENERIC_EXECUTE, FILE_GENERIC_READ, FILE_GENERIC_WRITE,
 };
 use windows::Win32::System::Threading::{
     CreateMutexW, GetCurrentProcess, GetProcessTimes, OpenProcess, QueryFullProcessImageNameW,
@@ -187,6 +190,11 @@ pub enum DaclError {
         /// Win32 error description.
         reason: String,
     },
+    /// The path is owned by another identity. Windows object owners can still
+    /// rewrite the DACL, so MXC cannot secure this path by stamping an
+    /// owner-only DACL.
+    #[error("path is not owned by this process identity: {0}")]
+    ForeignOwner(PathBuf),
     /// Generic Win32 error.
     #[error("Win32 error on {path}: {reason}")]
     Win32 {
@@ -907,9 +915,10 @@ unsafe impl Send for OwnedSid {}
 unsafe impl Sync for OwnedSid {}
 
 /// Process-wide cache of the three well-known AppContainer-membership
-/// SIDs. `compute_appcontainer_effective_access` used to re-parse these
-/// from string form on every invocation — three `ConvertStringSidToSidW`
-/// + `LocalAlloc` + matching `LocalFree` per call.
+/// SIDs, parsed once instead of re-parsing from string form on every
+/// `compute_appcontainer_effective_access` invocation (which would
+/// otherwise cost three `ConvertStringSidToSidW` + `LocalAlloc` +
+/// matching `LocalFree` per call).
 ///
 /// The cache is read-only after first init; OnceLock guarantees the
 /// init closure runs exactly once. The cached `OwnedSid`s deliberately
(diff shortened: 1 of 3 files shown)


src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+    use windows::Win32::Security::{GetTokenInformation, TokenOwner, TOKEN_OWNER, TOKEN_QUERY};
+    use windows::Win32::System::Threading::OpenProcessToken;
+
+    let token_path = Path::new("<process token owner>");
+    unsafe {
+        let mut token = HANDLE::default();
+        OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &mut token)
+            .map_err(|e| win32_err_str(token_path, &format!("OpenProcessToken: {e}")))?;
+
+        let mut len = 0u32;
```
dhoehna: Missed adding the comment ` // First call sizes the buffer (returns ERROR_INSUFFICIENT_BUFFER).`

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+        let _ = GetTokenInformation(token, TokenOwner, None, 0, &mut len);
```
dhoehna: Why is throwing away the GetTokenInformation result okay?  I am suspicious when return codes are ignored.

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+        }
+        OwnedSid::parse(&sid_string?)
+    }
+}
+
+/// The SID set that a filesystem object created by THIS process can legitimately
+/// be owned by: the token user SID and (when elevated) the token owner SID.
+fn self_owner_sids() -> Result<Vec<OwnedSid>, DaclError> {
+    let mut sids = vec![current_user_sid()?];
+    // Best-effort: only meaningful (and only differs) under elevation.
```
dhoehna, line 1286: What is the difference under elevation and why is it meaningful?

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+    if !owner_psid.0.is_null() {
+        for sid in &mine {
+            if unsafe { EqualSid(owner_psid, sid.as_psid()).is_ok() } {
+                is_self = true;
+                break;
+            }
+        }
+    }
+    unsafe {
+        let _ = LocalFree(Some(HLOCAL(sd.0)));
```
dhoehna, line 1343: I'm new to rust and I might not understand the code flow correctly.  Is LocalFree called if the code returns on line 1311?

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+    }
+    Ok(is_self)
+}
+
+///
```
dhoehna, line 1349: NIT: newline with no comment.

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+        0
+    };
+    let mask = FILE_ALL_ACCESS.0;
+    let grant = |sid: &OwnedSid| EXPLICIT_ACCESS_W {
+        grfAccessPermissions: mask,
+        grfAccessMode: GRANT_ACCESS,
+        grfInheritance: ACE_FLAGS(inheritance),
+        Trustee: trustee_for(sid),
+    };
+    let ea = [grant(&user), grant(&system)];
```
dhoehna: What does `ea` represent?  `Each Access`, `Every Access`, `Early Access`?  I can not figure it out from code.

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -2008,7 +2220,43 @@ fn process_creation_filetime() -> Result<u64, DaclError> {
 mod tests {
     use super::*;
 
-    // Note: the analogous compile-time `const _: () = assert!(...)`
+    /// A directory THIS process creates must be reported as owned by us.
+    #[test]
+    fn owner_is_self_true_for_dir_we_create() {
```
dhoehna, line 2255: +1 for including the expected result in the test name. :)

dhoehna (commented):
I do not know the architecture of the code base and I do not feel comfortable approving the PR.

A comment for the change in general.  Consider expanding variable names to describe what they are for instead of the type. Aside from well known monikers like `rc`.

Otherwise the code looks sound.

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
-/// Revoke explicit ACEs on `path` for `sid_str` whose `(access_mask,
-/// ace_type, inherit_flags)` tuple exactly matches the requested one.
+/// The current process token's user SID, as an [`OwnedSid`].
+fn current_user_sid() -> Result<OwnedSid, DaclError> {
+    use windows::Win32::Security::Authorization::ConvertSidToStringSidW;
+    use windows::Win32::Security::{GetTokenInformation, TokenUser, TOKEN_QUERY, TOKEN_USER};
+    use windows::Win32::System::Threading::OpenProcessToken;
+
+    let token_path = Path::new("<process token>");
+    unsafe {
```
bbonaby, line 1196: issue: missing safety comment here and throughout

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
     Ok(())
 }
 
-/// Revoke explicit ACEs on `path` for `sid_str` whose `(access_mask,
-/// ace_type, inherit_flags)` tuple exactly matches the requested one.
+/// The current process token's user SID, as an [`OwnedSid`].
+fn current_user_sid() -> Result<OwnedSid, DaclError> {
```
bbonaby: suggestion: function feels like a general function we could reuse outside of this filesystem_dacl path. Perhaps we should think about moving it to a helpers file.

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+        let tu = &*(buf.as_ptr() as *const TOKEN_USER);
+        let mut pstr = PWSTR::null();
+        ConvertSidToStringSidW(tu.User.Sid, &mut pstr)
+            .map_err(|e| win32_err_str(token_path, &format!("ConvertSidToStringSidW: {e}")))?;
+        let sid_string = pstr
+            .to_string()
+            .map_err(|e| DaclError::InvalidSid(format!("token user SID string not UTF-16: {e}")));
+        if !pstr.is_null() {
+            let _ = LocalFree(Some(HLOCAL(pstr.0 as *mut c_void)));
+        }
```
bbonaby: issue: we can reuse the sid_to_string function here https://github.com/microsoft/mxc/blob/ccbe3980b2c816264406c8bb5e9d310d98667706/src/core/wxc_common/src/string_util.rs#L26 for this purpose

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+        OwnedSid::parse(&sid_string?)
+    }
+}
+
+/// The token's *owner* SID — the SID that newly-created objects get as their
+/// owner. Differs from [`current_user_sid`] (the token USER) only when running
+/// elevated, where the default owner is the Administrators group. Best-effort:
+/// callers fall back to the user SID if this fails.
+fn current_token_owner_sid() -> Result<OwnedSid, DaclError> {
```
bbonaby, line 1270: issue: the body of this function is mostly the same as the one above, we should combine the shared functionality into a helper function. Also this function also benefits from resusing the `sid_to_string` helper function I mentioned in the function above.

src/core/wxc_common/src/filesystem_dacl.rs
```diff
@@ -1176,8 +1178,218 @@ pub fn apply_explicit_ace(
...
+    }
+    Ok(is_self)
+}
+
+///
+/// Unlike [`DaclManager`], this is deliberately NOT journaled: it stamps a
+/// permanent owner-only DACL on a path MXC owns (its own scratch dirs / record
+/// files), so there is nothing to restore on the host afterward.
+///
+/// # Threat model
```
bbonaby, line 1354: thought: do we need a formal threat model here that goes through the right MORSE channels for this and downlevel filesystem support? I see the word `nonce` below and that made me think about that.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: In this case, it was because this call is just to ask how large a buffer we should allocate for the call on line 1240, and the buffer size was returned in len. That said, this code has been re-written.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: Comment updated. Elevated, default owner is Administrators, non-elevated default owner is user.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: No, but if we return at that line, nothing was allocated that needs freeing.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: Renamed to access_entries

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: Will consider for future PR.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: sid_to_string returns a default on failure, which is not what I want here. I'll look for an oppurtunity to combine/refactor in future.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: I think if we do need formal review, it can happen after the Windows Sandbox PR stack is merged.

src/core/wxc_common/src/filesystem_dacl.rs
MGudgin: Done

MGudgin:
@microsoft-github-policy-service rerun

MGudgin:
Closing and re-opening to try to jog license/cla

shschaefer (approved):

PR #577, opened 2026-06-26
Rewrite the Windows Sandbox one-shot path onto an authenticated protocol (Phase 1)

### LINKED ISSUE #560: Windows Sandbox backend misreports "not enabled" when  wxc-exec runs non-elevated
### Relevant area(s)

Windows

Running a `windows_sandbox` config with a non-elevated `wxc-exec.exe` fails with  "Windows Sandbox is not enabled..." even when the feature is enabled. If you run `wxc-exec.exe` in an elevated prompt it runs fine. There is no elevation requirement in the documentation.

### Steps to reproduce

1. On a host where Windows Sandbox is enabled and `dism ... /get-featureinfo` shows `State : Enabled` when run from an **elevated** prompt.
2. From a **non-elevated** terminal, run a `windows_sandbox` config: `wxc-exec.exe --experimental windows_sandbox_echo.json` (config: `{ "containment": "windows_sandbox", "process": { "commandLine": "cmd /c echo hi" } }`)
3. Observe the error.

This error doesn't occur if you run in an elevated prompt

### Expected behavior

`wxc-exec.exe` running non-elevated should correctly detect that Windows Sandbox is enabled and proceed (or, if disabled, say so). Feature-state detection should not require elevation. If elevation is required, it should be documented and the error message should show that.

### Actual behavior

 Non-elevated `wxc-exec.exe` reports:
 `Windows Sandbox is not enabled. Run 'dism /online /enable-feature /featurename:Containers-DisposableClientVM /all'  and reboot.`
 even though the feature is enabled.

MGudgin:
Phase 1 of the Windows Sandbox rewrite stack: rewrite the Windows Sandbox
**one-shot** backend onto a new `windows_sandbox_lifecycle` crate and an
authenticated host<->guest protocol, and fix the non-elevated feature-detection
false negative. The state-aware lifecycle + host daemon (Phase 2) and the SDK
(Phase 3) follow in stacked PRs.

**Details**

* New `windows_sandbox_lifecycle` crate: the transient one-shot
  `WindowsSandboxRunner` (`ScriptRunner`) plus reusable VM lifecycle primitives
  (`vm`, `rendezvous`, `bridge`, `ipc_exec`, `policy`, `teardown`,
  `control_plane` host-VM-lock/ownership + Win32 `os` FFI).
* One-shot now launches a **fresh disposable VM per invocation** with guaranteed
  teardown on every exit path (normal/panic, Ctrl-C/shutdown handler,
  next-run reconcile) -- no warm-VM reuse. The legacy
  `experimental.windows_sandbox.idleTimeoutMs`/`daemonPipeName` fields are still
  parsed but ignored, with a stderr warning
  (`WXC_WSB_ACK_ONESHOT_FRESH_VM=1` silences the per-call notice).
* Authenticated protocol: `windows_sandbox_common::auth` adds a per-launch
  32-byte `Nonce` + 1-byte `ChannelRole` handshake on every TCP channel (boot +
  reconnect); the guest pairs accepted sockets by declared role, not accept
  order, defeating cross-user accept-race hijack. Guest agent rewritten
  (`listener`/`executor`/`job`/`firewall`) onto the new `sandbox_protocol` + auth.
* **Fixes #560**: feature detection is now by `WindowsSandbox.exe` presence
  instead of an elevated `dism /get-featureinfo` query, so a non-elevated
  `wxc-exec` no longer misreports "not enabled".
* Removed the old `windows_sandbox_common::windows_sandbox_runner`; wxc one-shot
  dispatch constructs `WindowsSandboxRunner`. The host daemon is dropped from the
  workspace members for this phase (re-added with the state-aware lifecycle in
  Phase 2); the dormant `control_plane` record/discovery functions are `pub` and
  compile clean for Phase 2 to build on.
* Tests/docs: e2e `test_windows_sandbox` and the renamed
  `run_windows_sandbox_one_shot_tests.ps1` drop the daemon pre-spawn and exercise
  the fresh-VM path. The full dual-mode backend-doc rewrites
  (windows-sandbox.md/reference.md, schema.md, state-aware API) are deferred to
  Phase 2, where they describe the end state; Phase 1 makes only minimal
  P1-accurate touches to `copilot-instructions.md`.

**Tests**

* `cargo check --workspace --all-targets`, `cargo clippy --workspace
  --all-targets -- -D warnings`, and `cargo fmt --all -- --check`: clean.
* `cargo test`: windows_sandbox_lifecycle 110, windows_sandbox_common 26, wxc 13,
  wxc_windows_sandbox_guest 8 -- all passed.
* Verified the dropped daemon doesn't break the full build (build.bat's daemon
  copy is `if exist`-guarded) and that no source references the deleted runner.

Fixes #560

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/577)


src/backends/windows_sandbox/common/src/auth.rs
```diff
@@ -0,0 +1,527 @@
...
+
+use std::path::Path;
+use std::time::Duration;
+
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpStream;
+
+/// Length in bytes of the per-launch nonce. 32 bytes / 256 bits matches the
+/// strength of common KEK / IKM sizes; brute-force is intractable.
+pub const NONCE_LEN: usize = 32;
```
dhoehna: NIT: add `_IN_BYTES` to the name.  Less chance for future developers to assume something other than bytes. #Resolved

src/backends/windows_sandbox/common/src/auth.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// exactly [`NONCE_LEN`] bytes to a socket and not log the result.
+    pub fn as_bytes(&self) -> &[u8; NONCE_LEN] {
+        &self.0
+    }
+
+    /// Compare two nonces in constant time. A naïve `==` would short-circuit
+    /// on the first differing byte, exposing the matched prefix length via
+    /// a timing side channel; over localhost the resolution is sub-µs but
+    /// across many guess attempts the side channel is real.
+    pub fn constant_time_eq(&self, other: &Nonce) -> bool {
```
dhoehna, line 78: I'm confused.  The functions says `Constant_time` and the function iterates over every element.  That is a linear time algorithm. #WontFix

src/backends/windows_sandbox/common/src/auth.rs
```diff
@@ -0,0 +1,527 @@
...
+        let mut diff: u8 = 0;
+        for i in 0..NONCE_LEN {
+            // Wrapping XOR -> OR into accumulator. The whole-array iteration
+            // means the work done does not depend on where the bytes differ.
```
dhoehna: An equality method that does not care about different bytes in the same index.  What does "equal" mean for this method? #Resolved

src/backends/windows_sandbox/common/src/sandbox_protocol.rs
```diff
@@ -121,6 +121,79 @@ pub fn decode_message(buf: &[u8]) -> Result<DecodeResult, serde_json::Error> {
...
+/// Lets the host fail fast (rather than hanging or misframing) if it connects
+/// to something that is not a matching guest agent.
+pub const PROTOCOL_MAGIC: [u8; 4] = *b"WSBP";
+
+/// Wire-protocol version negotiated via the preamble.
+///
+/// The host injects the guest binary, so in normal operation both sides are
+/// always the same build; the version guards against a stale guest left over
+/// from an incompatible image and makes mismatches diagnosable.
+pub const PROTOCOL_VERSION: u32 = 1;
```
dhoehna, line 141: Seems like the number is for a simple "Make sure two binary version match" and not meant as a in-depth versioning system with major/minor/security numbers?

When does this number need to be bumped?  Every time a file is edited, every release, or a different cadence?

What is the expected result if these numbers do not match? #Resolved

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1256 @@
...
+use windows_sandbox_common::auth::{self, ChannelRole, Nonce};
+use windows_sandbox_common::sandbox_protocol::{
+    decode_message, encode_message, validate_preamble, ControlMessage, DecodeResult, ExecRequest,
+    PREAMBLE_LEN,
+};
+
+use crate::ipc_exec::{self, ExecStart, FRAME_STDERR, FRAME_STDOUT};
+
+/// Maximum time (seconds) to wait for the guest's StreamsReady message.
+const STREAMS_READY_TIMEOUT_SECS: u64 = 60;
```
dhoehna: NIT: The comment can be removed.  The constant name is detailed enough to understand this is the wait time is seconds of the stream ready.  

Same with the Reconnect constant below it.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1256 @@
...
+
+/// Maximum time (seconds) for each data stream reconnection attempt.
+const RECONNECT_TIMEOUT_SECS: u64 = 30;
```
dhoehna: Why are these uint64's while the other constants the describe a duration using `Duration::`?

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1256 @@
...
+/// the agent writes immediately after accepting the control connection. Bounded
+/// independently of the caller's (minutes-long) VM-ready timeout so a peer that
+/// accepts the socket but never speaks cannot pin the connect path: it fails
+/// fast instead. Generous enough to tolerate a loaded host where the kernel
+/// accepts the TCP backlog slightly before the userspace agent services it.
+const PREAMBLE_HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Four TCP connections to the guest agent.
+///
+/// TODO: These TCP connections are unencrypted. Verify if this is a concern.
```
dhoehna: If not done already please make a bug for this item and put the link in the comment.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1256 @@
...
+
+    let control = connect(ChannelRole::Control).await?;
+    let stdin_stream = connect(ChannelRole::Stdin).await?;
+    let stdout_stream = connect(ChannelRole::Stdout).await?;
+    let stderr_stream = connect(ChannelRole::Stderr).await?;
+
+    eprintln!("[daemon] 4 TCP connections established to {}", addr);
+
+    // Wait for the READY message from the agent.
+    let mut conn = GuestConnection {
```
dhoehna, line 81: Does the ctor of `GuesConnection` have a wait inside it?

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1256 @@
...
+        control_residual: ctrl_buf,
+        ipc_alive,
+        exit_code,
+        error_message,
+    })
+}
+
+/// Write the terminal exit frame to the IPC client.
+///
+/// Call this only **after** the guest slot has been restored and the
```
dhoehna: Any safe guards in place to make sure the assumption it honored?  

What is the result if `write_exit_frame` is called before the guest slot has been restored and the single-flight lock is released?

src/backends/windows_sandbox/lifecycle/src/control_plane.rs
```diff
@@ -0,0 +1,1413 @@
...
+    /// reconcile: an orphaned VM is only reclaimed (torn down) when the running
+    /// sandbox processes intersect this recorded set. A present record alone is
+    /// never sufficient — without an intersection the VM is treated as foreign
+    /// and left untouched. `#[serde(default)]` keeps older records readable.
+    #[serde(default)]
+    pub vm_processes: Vec<VmProcId>,
+}
+
+/// What a freshly-starting daemon should do about any Windows Sandbox VM it
+/// finds already running. Produced by the pure [`classify_startup`] so the
```
dhoehna: is the first sentence a question?

dhoehna: NVM.  I understand.  It is description why the enum exists. :)

dhoehna: Github continues to fail to resolve this comment.   I'd resolve it if I could.

src/backends/windows_sandbox/lifecycle/src/control_plane.rs
```diff
@@ -0,0 +1,1413 @@
...
+            .stdin(std::process::Stdio::null())
+            .stdout(std::process::Stdio::null())
+            .stderr(std::process::Stdio::null())
+            .spawn()
+            .unwrap();
+        let pid = child.id();
+        let creation_time = process_creation_time(pid).expect("creation time");
+        let killed = terminate_processes(&[VmProcId { pid, creation_time }]);
+        assert_eq!(killed, 1);
+        // The child must actually be reaped.
```
dhoehna: Throughout this PR you use `Kill` and `Reap` in reference to processes.  Are these two words interchangeable or do they have different meanings?

src/backends/windows_sandbox/lifecycle/src/error.rs
```diff
@@ -0,0 +1,172 @@
...
+    RuntimeSetup(String),
+    /// Preparing or launching the VM (`.wsb` generation /
+    /// `WindowsSandbox.exe`), capturing ownership proof, waiting for the
+    /// guest to publish its rendezvous address, or establishing the TCP
+    /// channels to the guest agent. A single variant because the shared
+    /// [`crate::vm::launch_managed_vm`] helper owns the whole boot sequence;
+    /// per-phase variants (`Rendezvous`, `Connect`) would never be observed
+    /// distinctly by callers (`OneShotError` is `pub(crate)`).
+    Launch(String),
+    /// The request's policy cannot be expressed by Windows Sandbox (e.g. host
```
dhoehna: This comment uses `Expressed` while the error string uses `enforced`  Which one is correct?

src/backends/windows_sandbox/lifecycle/src/error.rs
```diff
@@ -0,0 +1,172 @@
...
+//! opaque string) lets the orchestration code attribute a failure to a
+//! specific lifecycle phase and lays the groundwork for richer typed wire
+//! errors in the state-aware backend.
+
+use wxc_common::models::{FailurePhase, ScriptResponse};
+
+/// Failure modes of a single transient one-shot execution.
+#[derive(Debug)]
+pub(crate) enum OneShotError {
+    /// `WindowsSandbox.exe` is not present (optional feature not installed).
```
dhoehna: Which description is correct, the enum comments, or the error strings?  How will future developers make sure to change both?  RECOMMENDATION: remove the enum comments and let the error strings express the enum values.

src/backends/windows_sandbox/lifecycle/src/ipc_exec.rs
```diff
@@ -0,0 +1,282 @@
...
+//! Binary framing (rather than JSON/base64) keeps arbitrary binary stdout/stderr
+//! byte-exact and avoids ~33% base64 bloat on the hot path. The small structured
+//! payloads ([`ExecStart`], [`ExecExit`]) stay JSON for forward-compatibility.
+
+use std::io::{self, Read, Write};
+
+use serde::{Deserialize, Serialize};
+use tokio::io::{AsyncRead, AsyncReadExt};
+
+/// Data-frame kind: raw stdout bytes from the guest child process.
```
dhoehna: Please use an enum to keep compile time safety.

src/backends/windows_sandbox/lifecycle/src/one_shot.rs
```diff
@@ -0,0 +1,446 @@
...
+            .spawn(capture_host_stdin)
+            .map_err(|e| {
+                OneShotError::RuntimeSetup(format!("spawn host stdin spool thread: {e}"))
+            })?;
+
+        // The lifecycle primitives are async; bridge them with a dedicated
+        // current-thread runtime. Declared after the guard so it is dropped
+        // first, leaving the guard to run teardown once the runtime is gone.
+        //
+        // This must not be called from within an existing tokio runtime
```
dhoehna: Any way to assert this requirement?

src/backends/windows_sandbox/lifecycle/src/one_shot.rs
```diff
@@ -0,0 +1,446 @@
...
+/// * `set_ownership` -> pushes the state into the process-global
+///   teardown slot ([`teardown::set_vm_ownership`]) so the
+///   teardown guard / console handler tear down the right VM.
+/// * `persist_proof` -> rewrites the per-run marker file with the
+///   captured proof so a later one-shot run can reclaim this VM if
+///   the current run hard-dies.
+/// * `note_empty_proof` -> the one-shot-specific warning calling
+///   out the loss of marker-based reclaim and the fallback to
+///   enumerate-kill at exit.
+struct OneShotLaunchObserver<'a> {
```
dhoehna, line 195: why tilde-a?  Is that different from a lone a?

src/backends/windows_sandbox/lifecycle/src/teardown.rs
```diff
@@ -0,0 +1,1072 @@
...
+    vm_processes: Vec<VmProcId>,
+}
+
+/// Per-marker liveness/proof state, distilled from an [`OneShotMarker`] for the
+/// pure [`classify_reconcile`].
+#[derive(Debug, Clone)]
+struct MarkerState {
+    /// Whether the launching process is *strongly* alive (PID present AND
+    /// creation time matches). A missing creation time is treated as not-alive.
+    launcher_alive: bool,
```
dhoehna: Consider splitting this up into their constituent parts.  Unless you are confident that this program will never need to test PID present and creation time matches individually.

src/backends/windows_sandbox/lifecycle/src/vm.rs
```diff
@@ -0,0 +1,663 @@
...
+                    if is_real_python(dir) {
+                        return Ok(dir.to_path_buf());
+                    }
+                }
+            }
+        }
+    }
+
+    // Common install locations.
+    let candidates = [
```
dhoehna: What happens when Python3.13 comes out?  Is this the only location with a version specifier?

src/backends/windows_sandbox/lifecycle/src/vm.rs
```diff
@@ -0,0 +1,663 @@
...
+                let dir = entry.path();
+                if is_real_python(&dir) {
+                    return Ok(dir);
+                }
+            }
+        }
+    }
+
+    anyhow::bail!(
+        "Python installation not found on host. Install Python and ensure python.exe is on PATH."
```
dhoehna: Error message does not take into the case where a python installation can be found and the version is incorrect.

src/backends/windows_sandbox/lifecycle/src/vm.rs
```diff
@@ -0,0 +1,663 @@
...
+    )
+}
+
+/// Returns true if `dir` contains a real Python installation (not a
+/// Windows Store stub).  The WindowsApps stub passes `exists()` but
+/// fails when invoked, so we run `python --version` to verify.
+fn is_real_python(dir: &Path) -> bool {
```
dhoehna: Method name suggests the method checks if a python installation is a "real" python. and `Path` is the path to python.exe.

The comments say it checks if the directory has python.exe.

NIT: change the name to notify developers the function finds python.exe in the passed in directory.

src/backends/windows_sandbox/lifecycle/src/vm.rs
```diff
@@ -0,0 +1,663 @@
...
+        }
+    }
+
+    anyhow::bail!(
+        "Python installation not found on host. Install Python and ensure python.exe is on PATH."
+    )
+}
+
+/// Returns true if `dir` contains a real Python installation (not a
+/// Windows Store stub).  The WindowsApps stub passes `exists()` but
```
dhoehna: > The WindowsApps stub passes `exists()` but
/// fails when invoked, so we run `python --version` to verify.

the use of a comma suggests that the second half is dependent on the first and `python --version` is ran regardless if python is from the store or not.

However, the function returns false once it finds python in the WindowsApp location and no `--version` is ran in that case.

src/backends/windows_sandbox/lifecycle/src/vm.rs
```diff
@@ -0,0 +1,663 @@
...
+        return false;
+    }
+    // WindowsApps stubs live under Microsoft\WindowsApps — skip them
+    // outright since they always fail when invoked from another context.
+    if dir.to_string_lossy().contains("Microsoft\\WindowsApps") {
+        eprintln!("[daemon] skipping Windows Store stub at {:?}", dir);
+        return false;
+    }
+    // Belt-and-suspenders: verify it actually runs.
+    std::process::Command::new(&python)
```
dhoehna: What happens if Python does not run?

src/backends/windows_sandbox/lifecycle/src/vm.rs
dhoehna: I don't see any checks in the code above to verify if python is on the path on not.  Did I miss it?

src/backends/windows_sandbox/lifecycle/src/vm.rs
```diff
@@ -0,0 +1,663 @@
...
+    let bootstrap_content = format!(
+        r#"@echo off
+set "LOG={rendezvous}\bootstrap.log"
+:: Truncate any prior bootstrap.log content on this boot so its
+:: contents cannot accumulate across runs on the host.
+type nul > "%LOG%"
+
+echo [bootstrap] Starting at %date% %time% >> "%LOG%" 2>&1
+
+echo [bootstrap] Adding mapped Python to PATH... >> "%LOG%" 2>&1
```
dhoehna: Is this the PATH for the sandbox?

src/core/wxc/src/main.rs
```diff
@@ -946,13 +947,56 @@ fn main() {
...
+            // concern.
+            //
+            // Warn unconditionally: one-shot launches a fresh VM per call
+            // (boot tax ~12-15s vs ~ms for a reused-VM exec), so a caller
+            // relying on warm reuse pays a large wall-clock cost. The warning
+            // is not gated on the non-default `experimental.windows_sandbox`
+            // block, because a typical warm-reuse caller never set those
+            // fields. Callers who acknowledge the model can set
+            // `WXC_WSB_ACK_ONESHOT_FRESH_VM=1` to suppress.
+            if std::env::var_os("WXC_WSB_ACK_ONESHOT_FRESH_VM").is_none() {
```
dhoehna: MXC is in alpha.  Users know that any release could break their workflows.   I suggest to remove this.

src/core/wxc/src/main.rs
```diff
@@ -946,13 +947,56 @@ fn main() {
...
+                eprintln!(
+                    "[wxc-exec] WARNING: one-shot `containment: \"windows_sandbox\"` now \
```
dhoehna: When will this warning be removed?  Is this something that needs to be kept forever?

src/core/wxc/src/main.rs
```diff
@@ -946,13 +947,56 @@ fn main() {
...
+                     launches a FRESH disposable VM per call (boot ~12-15s) instead of the \
+                     legacy warm-reuse daemon. If you were relying on warm reuse for \
+                     repeated calls, switch to the state-aware lifecycle (provisionSandbox / \
+                     startSandbox / execInSandbox / stopSandbox / deprovisionSandbox) -- see \
+                     docs/windows-sandbox/windows-sandbox.md and docs/state-aware-lifecycle/. \
```
dhoehna: `https://github.com/microsoft/mxc/blob/main/docs/windows-sandbox/windows-sandbox.md` has nothing about the state-aware lifecycle.  Is that in a different PR?

src/core/wxc/src/main.rs
```diff
@@ -946,13 +947,56 @@ fn main() {
...
+                     Set WXC_WSB_ACK_ONESHOT_FRESH_VM=1 to silence this warning once you \
+                     have audited your usage."
+                );
+            }
+            //
```
dhoehna: Empty comment line.

src/core/wxc/src/main.rs
```diff
@@ -946,13 +947,56 @@ fn main() {
...
+            // If the caller supplied a non-default
+            // `experimental.windows_sandbox` block, warn separately and more
```
dhoehna: Comment suggests the code is looking for a non-default experimental.windows_sandbox block.  The code searched for two specific attributes.

Does having `idle_timeout_ms` or `daemon_pipe_name` make the block "non-default"?

src/testing/wxc_e2e_tests/tests/e2e_windows.rs
```diff
@@ -535,50 +535,11 @@ fn run_sandbox_case(case: &SandboxCase) {
...
-    }
-}
-
-impl Drop for SandboxDaemon {
-    fn drop(&mut self) {
-        let _ = self.child.kill();
-        let _ = self.child.wait();
-    }
-}
+// The one-shot suite above needs no pre-spawned daemon -- each
```
dhoehna: Because MXC is in alpha and this is removing the use of pre-spawned daemon does the comment need to be here?  A comment that states something has changes?  Isn't that what git is for, to keep the change history?

src/backends/windows_sandbox/guest/src/executor.rs
```diff
@@ -176,36 +248,87 @@ async fn execute_script(
     cmd.stderr(std::process::Stdio::piped());
 
     let mut child = cmd.spawn().context("spawn child process")?;
+    let child_pid = child.id();
+
+    // Assign the child to a Job Object so we can reliably reap its entire
+    // descendant tree later (see `job` module). Best-effort: if job creation or
```
dhoehna: link to a file so we can easily go to the job module?

dhoehna (dismissed):
Looked through the comments and code.

In addition to my comments I am concerned with the amount of in-code comments.  Many comment blocks explain the code increasing the risk future developers will forget to update the comment and the code.

Because we need to get this out in a certain time frame I will not block the PR on comment length of quality.  In the future please keep the length in mind.

src/backends/windows_sandbox/common/src/auth.rs
MGudgin: Given the number of elements iterated over is fixed, it takes the same amount of time every time its called. Prevents side-channel attacks that might be possible if where the nonces differ is discoverable.

src/backends/windows_sandbox/common/src/sandbox_protocol.rs
MGudgin: Comment added.

src/backends/windows_sandbox/common/src/auth.rs
MGudgin: It cares about different bytes at the same index, but it always checks *all* the bytes. See other comment about preventing side-channel attacks.

src/backends/windows_sandbox/lifecycle/src/one_shot.rs
MGudgin: apostrophe a ('a) declares a lifetime parameter and basically says here that a OneShotLaunchObserver cannot have a lifetime that extends past the lifetime of the run_dir Path reference. Or put another way, the lifetime of the run_dir Path reference must be at least as long as the lifetime of the OneShotLaunchObserver struct instance.

MGudgin: A lone a would be a generic type parameter (and are typically capitalized in Rust)

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: All Python related code removed (such paths should be specified by policy external to this code)

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: See above comment.

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: See above comment.

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: See above comment.

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: See above comment.

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: See above comment.

MGudgin:
> Looked through the comments and code.
> 
> In addition to my comments I am concerned with the amount of in-code comments. Many comment blocks explain the code increasing the risk future developers will forget to update the comment and the code.
> 
> Because we need to get this out in a certain time frame I will not block the PR on comment length of quality. In the future please keep the length in mind.

Comments heavily trimmed.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
MGudgin: No, waits are after construction. Comment removed.

src/backends/windows_sandbox/lifecycle/src/vm.rs
MGudgin: See above comment.

src/backends/windows_sandbox/lifecycle/src/error.rs
MGudgin: Enforced. Fixed.

src/backends/windows_sandbox/lifecycle/src/error.rs
MGudgin: Comments removed.

src/backends/windows_sandbox/lifecycle/src/ipc_exec.rs
MGudgin: Fixed.

src/backends/windows_sandbox/lifecycle/src/one_shot.rs
MGudgin: Fixed.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
MGudgin: Will be used/enforced in #578

src/backends/windows_sandbox/guest/src/executor.rs
MGudgin: Reference removed, job module is in an external crate.

src/testing/wxc_e2e_tests/tests/e2e_windows.rs
MGudgin: Fixed

src/core/wxc/src/main.rs
MGudgin: Clarified.

src/core/wxc/src/main.rs
MGudgin: Clarified.

src/core/wxc/src/main.rs
MGudgin: Fixed.

src/core/wxc/src/main.rs
MGudgin: Fixed.

src/backends/windows_sandbox/lifecycle/src/teardown.rs
MGudgin: Renamed.

src/backends/windows_sandbox/lifecycle/src/control_plane.rs
MGudgin: They are different. Kill/terminate requests process termination via TerminateProcess. reap waits for completion and collects the child’s exit status. I updated the test to make that distinction explicit: terminate_processes  issues the termination request, then  child.wait() reaps the test-owned child. Production teardown does not own Child  handles, so it terminates the recorded processes and polls until they are confirmed gone.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1070 @@
...
+            }
+            DecodeResult::Incomplete => {
+                continue;
+            }
+        }
+    }
+}
+
+/// Send an EXEC request to the guest and relay stdin/stdout/stderr.
+pub async fn execute_on_guest(
```
huzaifa-d, line 153: The host has no time limit of its own here. If the guest freezes, or the timeout is set to 0, the host will sit on this call indefinitely.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
```diff
@@ -0,0 +1,1070 @@
...
+                .shutdown()
+                .await
+                .err()
+                .map(|e| anyhow::Error::new(e).context("shutdown stdin"))
+        }
+    };
+
+    // Preserve partial output on failure. A reset before any bytes is an empty
+    // stream; a reset after data is a transport error.
+    let stdout_task = {
```
huzaifa-d, line 191: The output we collect from the guest isn't capped. Can we cap the captured output?

src/backends/windows_sandbox/lifecycle/src/control_plane.rs
```diff
@@ -0,0 +1,1195 @@
...
+    match ownership {
+        VmOwnership::NotLaunched => CleanupAction::Noop,
+        VmOwnership::LaunchInFlight => CleanupAction::LeakUnowned,
+        VmOwnership::LaunchSucceededNoProof => CleanupAction::Teardown(Vec::new()),
+        VmOwnership::Owned(pids) => CleanupAction::Teardown(pids.clone()),
+    }
+}
+
+/// Compute the process identities safe to terminate for a proven VM owner.
+pub fn plan_kill_set(ownership: &VmOwnership, snapshot: &[VmProcId]) -> Option<Vec<VmProcId>> {
```
huzaifa-d, line 165: When building the shutdown list, the Owned branch unions the entire snapshot of WindowsSandbox processes after a single intersection. That means we could kill an unrelated sandbox another user is running.

src/backends/windows_sandbox/lifecycle/src/teardown.rs
```diff
@@ -0,0 +1,889 @@
...
+                match &self.ownership {
+                    VmOwnership::NotLaunched | VmOwnership::LaunchSucceededNoProof => {
+                        clear_marker_dir(&self.run_dir);
+                    }
+                    VmOwnership::LaunchInFlight => {
+                        // Preserve ambiguous ownership for the next reconcile.
+                    }
+                    VmOwnership::Owned(_) => {
+                        // plan_kill_set(Owned, _) is always Some.
+                        unreachable!("Owned with any snapshot always returns Some kill plan");
```
huzaifa-d, line 173: There is an unreachable!() on the teardown path. If that assumption is ever wrong, it turns a minor edge case into a hard crash during cleanup. Prefer logging and continuing here rather than panicking.

src/backends/windows_sandbox/common/src/auth.rs
```diff
@@ -0,0 +1,344 @@
...
+            diff |= self.0[i] ^ other.0[i];
+        }
+        diff == 0
+    }
+}
+
+/// Generate a fresh per-launch nonce using the OS RNG.
+pub fn generate_nonce() -> Nonce {
+    let mut buf = [0u8; NONCE_LEN_IN_BYTES];
+    getrandom::getrandom(&mut buf).expect("OS getrandom failed (kernel CSPRNG broken)");
```
huzaifa-d: getrandom().expect(...) will panic if the RNG ever fails. Low risk, but since this is on the security path, consider returning an error instead of unwrapping.

src/backends/windows_sandbox/lifecycle/src/bridge.rs
MGudgin: Addressed in #578

src/backends/windows_sandbox/common/src/auth.rs
MGudgin: Fixed.

src/backends/windows_sandbox/lifecycle/src/control_plane.rs
MGudgin: Addressed in #578

src/backends/windows_sandbox/lifecycle/src/teardown.rs
MGudgin: Addressed in #578

src/backends/windows_sandbox/lifecycle/src/bridge.rs
MGudgin: Addressed in #578 gugut

huzaifa-d (approved):

PR #580, opened 2026-06-26
Phase 3a: decouple ProcessContainer backend selection from schema version

MGudgin:
This change makes the Windows ProcessContainer AppContainer-vs-BaseContainer
choice entirely host-capability driven: BaseContainer is preferred when usable,
and AppContainer is the downlevel fallback. Schema version and
--experimental no longer affect this selection. It also incorporates updates
that clarify diagnostics and make the streaming fallback explicit.

Details:
- Route the executor through the capability-driven dispatcher unconditionally
  and remove the schema-version/experimental gate and obsolete parser helper.
- Select the SDK streaming backend using the same BaseContainer usability probe;
  on downlevel hosts it explicitly creates an AppContainer runner.
- Reword BaseContainer E_NOTIMPL diagnostics to describe an OS-disabled feature
  without internal terminology, and remove stale schema-version guidance.
- Update backend-agnostic regression coverage and tidy related comments.

Tests:
- cargo test -p wxc_common -p appcontainer_common -p mxc-sdk
- cargo clippy --workspace --all-targets -- -D warnings
- cargo fmt --all -- --check
- check-schema-codegen.js and check-sdk-types-codegen.js

Copilot-Session: 6c58a97a-f189-43eb-b159-78965a028dc9
Generated-with: gpt-5.6-terra

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/580)

### DIFF
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -1234,22 +1234,13 @@ impl SandboxBackend for BaseContainerRunner {
             ));
         }
         Self::is_base_container_api_present().map_err(|e| {
-            let hint = if !request.experimental_enabled {
-                format!(
-                    "BaseContainer API unavailable: {e}\n\
-                     Hint: Config schema version '{}' requires the BaseContainer backend, \
-                     but this OS build does not support it. \
-                     Use schema version '0.4.0-alpha' to fall back to AppContainer.",
-                    request.schema_version
-                )
-            } else {
-                format!(
-                    "BaseContainer API unavailable: {e}\n\
-                     Hint: --experimental requested BaseContainer, but this OS build \
-                     does not support it. Remove --experimental to use the AppContainer \
-                     backend, or use an OS build with BaseContainer support."
-                )
-            };
+            let hint = format!(
+                "BaseContainer API unavailable: {e}\n\
+                 Hint: this OS build does not support the BaseContainer backend. \
+                 MXC selects BaseContainer automatically when the host supports it \
+                 and otherwise uses AppContainer; use an OS build with BaseContainer \
+                 support if you require it."
+            );
             ScriptResponse {
                 // Symbol absent: report BackendUnavailable, not a hard error.
                 failure_phase: FailurePhase::BackendUnavailable,
--- a/src/backends/appcontainer/common/src/launch_diagnostics.rs
+++ b/src/backends/appcontainer/common/src/launch_diagnostics.rs
@@ -167,25 +167,29 @@ fn diagnose_api_not_implemented() -> LaunchDiagnostic {
     let message = if key_status.is_empty() {
         "Experimental_CreateProcessInSandbox returned E_NOTIMPL. \
          The BaseContainer feature is not enabled on this OS build. \
-         Ensure the required velocity keys are enabled, or use schema \
-         version '0.4.0-alpha' to fall back to the AppContainer backend."
+         It may be possible to enable it through the Windows experimental \
+         features settings, or run on a host that supports the BaseContainer \
+         backend (MXC falls back to AppContainer automatically on builds \
+         without it)."
             .to_string()
     } else {
         let disabled: Vec<_> = key_status.iter().filter(|(_, enabled)| !enabled).collect();
         if disabled.is_empty() {
             "Experimental_CreateProcessInSandbox returned E_NOTIMPL. \
-             All known velocity keys appear enabled, but the feature is \
-             still gated. The OS build may require additional enablement. \
-             Use schema version '0.4.0-alpha' to fall back to the AppContainer backend."
+             The BaseContainer feature is not enabled on this OS build; it may \
+             require additional enablement. MXC falls back to AppContainer \
+             automatically on builds without BaseContainer support."
                 .to_string()
         } else {
             let disabled_list: Vec<String> =
                 disabled.iter().map(|(id, _)| id.to_string()).collect();
             format!(
                 "Experimental_CreateProcessInSandbox returned E_NOTIMPL. \
-                 The following velocity keys are not enabled: {}. \
-                 Enable them and retry, or use schema version '0.4.0-alpha' \
-                 to fall back to the AppContainer backend.",
+                 The BaseContainer feature is not enabled on this OS build \
+                 (disabled feature flags: {}). It may be possible to enable it \
+                 through the Windows experimental features settings, or run on a \
+                 host that supports the BaseContainer backend (MXC falls back to \
+                 AppContainer automatically on builds without it).",
                 disabled_list.join(", ")
             )
         }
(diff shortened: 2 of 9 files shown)


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1215,22 +1215,13 @@ impl SandboxBackend for BaseContainerRunner {
...
-                    "BaseContainer API unavailable: {e}\n\
-                     Hint: --experimental requested BaseContainer, but this OS build \
-                     does not support it. Remove --experimental to use the AppContainer \
-                     backend, or use an OS build with BaseContainer support."
-                )
-            };
+            let hint = format!(
+                "BaseContainer API unavailable: {e}\n\
+                 Hint: this OS build does not support the BaseContainer backend. \
+                 MXC selects BaseContainer automatically when the host supports it \
```
bbonaby, line 1240: `BaseContainer`, I guess we should update the name in the future to ProcessContainer

src/backends/appcontainer/common/src/launch_diagnostics.rs
```diff
@@ -167,25 +167,28 @@ fn diagnose_api_not_implemented() -> LaunchDiagnostic {
     let message = if key_status.is_empty() {
         "Experimental_CreateProcessInSandbox returned E_NOTIMPL. \
          The BaseContainer feature is not enabled on this OS build. \
-         Ensure the required velocity keys are enabled, or use schema \
-         version '0.4.0-alpha' to fall back to the AppContainer backend."
+         Ensure the required velocity keys are enabled, or run on a host \
+         that supports the BaseContainer backend (MXC falls back to \
+         AppContainer automatically on builds without it)."
```
bbonaby: suggestion: Lets not talk about velocity. We can say " feature isn't enabled on this build". I do believe there is a way in the Windows Settings app that will allow folks in the future to turn velocity features on in the Windows insider experimental channel. 

Image

src/core/mxc-sdk/tests/sandbox.rs
```diff
@@ -39,7 +39,8 @@ fn seatbelt_request(command: &str, timeout_ms: u32) -> SandboxRequest {
 }
 
 /// A Windows ProcessContainer request exposing `C:\Windows\Temp` read-write.
-/// The policy `version` selects the tier (>= 0.5 implies BaseContainer).
+/// `version` is the schema version stamped on the policy; it no longer selects
```
bbonaby: suggestion: comment can be removed

src/core/mxc-sdk/tests/sandbox.rs
```diff
@@ -241,8 +242,10 @@ fn seatbelt_defaults_cwd_to_allowed_path_without_getcwd_leak() {
 // Windows ProcessContainer (AppContainer + BaseContainer) — integration tests.
 //
 // These exercise the capture and timeout paths that regressed as review items
-// #1 (BaseContainer ran with an already-closed process handle) and #2
-// (AppContainer timeout killed only the direct child, so it never fired).
+// #1 (the process handle was closed before the wait completed) and #2 (a finite
+// timeout killed only the direct child, so it never fired). ProcessContainer
+// resolves to BaseContainer or AppContainer by host capability; these guards
+// hold for whichever backend the host selects.
```
bbonaby: question: are these comments needed?

src/core/wxc/src/main.rs
```diff
@@ -809,59 +806,45 @@ fn main() {
...
+                    } = &e
+                    {
+                        for w in warnings {
+                            eprintln!("  dacl warning: {w}");
                         }
-                        eprint!("{}", logger.get_buffer());
-                        process::exit(1);
                     }
+                    eprint!("{}", logger.get_buffer());
+                    process::exit(1);
```
bbonaby, line 992: suggestion: We should probably start making sure we use the ScriptResponse structure to print errors and warnings so all stderr and stdout messages are structured json and can be parsed and filtered easily. What do you think about that?

sdk/tests/unit/conformance-helpers.ts
bbonaby, line 4: note: Tbh I'm still a bit iffy on these. I kind of wish we didn't have to do this and could rely on third party libraries to get us 90% of the way there between the schema, typescript types and rust types. For example I could imagine us using a flatbuffer schema (internally), and having flatc.exe generate the json schema for us. From there We can generate the typescript and rust types on both sides using flatc.exe, and can also use the flatbuffer SDK to verify the json.

Mostly because on the surface it's hard for me to understand some of these straight away. I won't block on these but we should really think about whether it's worth it for us to do all of this and maintain it versus looking at how other tech handle this stuff.

src/backends/appcontainer/common/src/base_container_runner.rs
MGudgin: No, processcontainer is the platform neutral container identifier, basecontainer is the windows specific process container

src/core/wxc/src/main.rs
MGudgin: Agreed in principle -- structured `ScriptResponse` JSON for all stdout/stderr errors and warnings would make the executor output machine-parseable and give a consistent error surface across backends. It's cross-cutting across every backend's dispatch arm though, unrelated to this PR's backend-selection decoupling, so I've tracked it as a follow-up rather than folding it in here: #612.

sdk/tests/unit/conformance-helpers.ts
MGudgin: Fair point, and I share the maintenance concern. The hand-rolled oracle is a deliberate choice to avoid a third-party generator dependency, but a flatc-driven schema -> (JSON schema + Rust + TS) pipeline with FlatBuffer verification is worth evaluating properly. Since it's non-blocking and pre-dates this PR (Phase 2.5), I've captured it as a design-discussion follow-up rather than acting on it here: #613.

src/core/mxc-sdk/src/dispatch.rs
```diff
@@ -149,27 +149,23 @@ fn spawn_process_container(
...
-    if request.experimental_enabled || version_implies_base_container {
+    // Unlike the executor binaries' run-to-completion path, streaming does NOT
+    // route through `dispatch_with_fallback` — that yields a run-to-completion
+    // `Box<dyn ScriptRunner>` plus a `DaclManager` guard, neither of which fits
+    // the streaming handle (the DACL tier would require the returned
+    // `SandboxProcess` to own the guard so ACE restore outlives the child). So
+    // streaming has no AppContainer-BFS / AppContainer-DACL tiering: it picks
+    // BaseContainer when the API is usable and plain AppContainer otherwise,
+    // using the same `is_base_container_usable()` probe the dispatcher's Tier 1
+    // selection uses, so the two paths agree on the BC-vs-AC decision.
```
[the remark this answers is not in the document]
MGudgin: Good catch -- you're right. `AppContainerScriptRunner::new()` defaults to `FilesystemMode::Bfs`, so the fallback was the AppContainer+BFS tier (still dependent on `bfscfg.exe`), not "plain AppContainer", and the Tier 3 AppContainer-DACL fallback is genuinely unavailable on the streaming path. Fixed in b1085ad: reworded the comment to say streaming offers only BaseContainer (when usable) or AppContainer in BFS mode (the dispatcher's Tier 2, no Tier 3 DACL fallback), and made the BFS choice explicit via `with_filesystem_mode(FilesystemMode::Bfs)` so it can't silently drift if the `new()` default changes.

bbonaby:
just need to fix the merge conflict and we should be good

bbonaby (approved):

MGudgin:
Closing to try to 'jog' license/cla gate.

bbonaby (approved):

PR #581, opened 2026-06-26
Phase 3b: raise the schema support floor to 0.6.0

MGudgin:
This commit raises the minimum accepted configuration schema version to
0.6.0-alpha while retaining support through 0.8.0-alpha.

Details:
- Updates Rust parser and SDK validation to reject 0.4.0-alpha and
  0.5.0-alpha as older than supported.
- Updates the configuration corpus, smoke scripts, tooling, and
  documentation to use supported schema versions.
- Preserves released schema artifacts as immutable files.

Test:
- cargo fmt --all -- --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test -p wxc_common
- npm run build and npm test
- Versioning code-generation and schema checks

Generated-By: Claude Opus 4.8 (claude-opus-4.8)


docs/schema.md
```diff
@@ -155,12 +155,13 @@ The parser compares the config's major.minor against its supported version
...
-| `"0.5.0-alpha"` | >=0.4, <=0.5 | Accepted (0.5 in range) |
-| `"0.6.0"` | >=0.4, <=0.5 | **Rejected** — "newer than supported" |
-| `"1.0.0"` | >=0.4, <=0.5 | **Rejected** — "newer than supported" |
+| absent | >=0.6, <=0.8 | Accepted (assumed compatible) |
+| `"0.5.0-alpha"` | >=0.6, <=0.8 | **Rejected** — "older than supported" |
+| `"0.6.0-alpha"` | >=0.6, <=0.8 | Accepted (0.6 in range) |
+| `"0.7.0-alpha"` | >=0.6, <=0.8 | Accepted (0.7 in range) |
+| `"0.8.0-alpha"` | >=0.6, <=0.8 | Accepted (0.8 in range) |
+| `"0.9.0"` | >=0.6, <=0.8 | **Rejected** — "newer than supported" |
+| `"1.0.0"` | >=0.6, <=0.8 | **Rejected** — "newer than supported" |
```
bbonaby, line 179: question: Anything we can do to no have to update this everytime we release? Perhaps a `--supported-schemas` flag maybe that will print this out to the user?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -2748,13 +2752,17 @@ mod tests {
     }
 
     #[test]
-    fn schema_version_older_accepted() {
+    fn schema_version_0_4_now_rejected() {
```
bbonaby: question: Should we just have 5 tests? Say we had a static min version variable and max version variable (I think we do). Then for tests we'd have:
- one below the min version where we expect it to be rejected
- one at the min version where we expect it to be accepted
- one between the min and the max version where we expect it to be accepted
- one at the max version where we expect it to be accepted
- one above the max version where we expect it to be rejected

Then we'd never need to have these version specific tests right?

MGudgin: Fixed.

docs/schema.md
MGudgin: I'll give it some thought. Seems low-pri though.

sdk/README.md
```diff
@@ -55,9 +55,9 @@ child.on('close', (code) => console.log('exit:', code));
 
 | Version | Status | Schema file |
 | --- | --- | --- |
-| `0.4.0-alpha` | Stable | [`schemas/stable/mxc-config.schema.0.4.0-alpha.json`](https://github.com/microsoft/mxc/blob/main/schemas/stable/mxc-config.schema.0.4.0-alpha.json) |
-| `0.5.0-alpha` | Stable | [`schemas/stable/mxc-config.schema.0.5.0-alpha.json`](https://github.com/microsoft/mxc/blob/main/schemas/stable/mxc-config.schema.0.5.0-alpha.json) |
-| `0.6.0-alpha` | Stable | [`schemas/stable/mxc-config.schema.0.6.0-alpha.json`](https://github.com/microsoft/mxc/blob/main/schemas/stable/mxc-config.schema.0.6.0-alpha.json) |
+| `0.4.0-alpha` | Retired — below the `0.6.0-alpha` floor (no longer accepted) | [`schemas/stable/mxc-config.schema.0.4.0-alpha.json`](https://github.com/microsoft/mxc/blob/main/schemas/stable/mxc-config.schema.0.4.0-alpha.json) |
+| `0.6.0-alpha` | Retired — below the `0.6.0-alpha` floor (no longer accepted) | [`schemas/stable/mxc-config.schema.0.5.0-alpha.json`](https://github.com/microsoft/mxc/blob/main/schemas/stable/mxc-config.schema.0.5.0-alpha.json) |
```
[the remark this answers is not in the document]
MGudgin: Fixed in a63ff7ad.

bbonaby (approved):

PR #585, opened 2026-06-27
Permissive Learning Mode 2/6 filesystem extraction

lilybarkley-msft:
## 📖 Description

**PR 2 of 6** — stacked on PR1. Adds **filesystem extraction** to PLM.

- `EventID=14` (access-failure) decoder:
  - File-path normalization (NT-object / verbatim / DOS-device → DOS form)
  - Post-XPath filters: current-directory, drive-letter, self-access, invalid filename chars
  - Per-event accumulator that pushes `LearningModeAccessEvent` rows (file path + access mask)
- ETW walker (`for_each_event_xml` / `EvtQuery` / `EvtRender`) with bounded peak memory
- Shared `ParseAccumulator` + per-event dispatcher in `event_parser`
- `config.rs` foundations: load/parse, `filesystem.{readwritePaths,readonlyPaths}` merge, `deny_file_set`
- `stop`/`log` wired to the FS-extraction pipeline

Capability extraction, config-generation summaries, and UI policy land in PR3–PR5. `requested_capabilities` is exposed as an always-empty placeholder so call-sites stay stable across the split.

## 🔗 References

- Base: **PR1** (`user/lilybarkley/plm-pr1-audit-skeleton`)
- Next: PR3 — config generation (`Adjusted_<name>.json` write + detection summaries)

## 🔍 Validation

- `cargo test -p plm --target x86_64-pc-windows-msvc` — **37 passed** (12 new over PR1; cover path-normalization, post-XPath filters, accumulator dispatch, ETW XML parsing).

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/585)


src/host/plm/src/main.rs
```diff
@@ -1,493 +1,495 @@
...
-                } else {
-                    cancel_active_plm_trace_from_signal();
-                }
-            }
-            result
-        }
-    }
-}
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
MGudgin: There is some kind of line ending issue with this file as it shows as completely re-written via git diff. Please fix.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,498 @@
...
+    // `EvtRender` writes UTF-16, so the backing buffer is `Vec<u16>`
+    // to guarantee 2-byte alignment (`Vec<u8>` is only 1-byte-aligned
+    // and casting `.as_ptr()` to `*const u16` would be UB even on x86).
+    // Note: `EvtRender`'s `BufferSize` / `BufferUsed` parameters are
+    // BYTE counts, so multiply/divide by `size_of::<u16>()` at the
+    // Win32 boundary.
+    const INITIAL_GUESS_U16: usize = 4 * 1024;
+    unsafe {
+        buf.set_len(0);
+    }
```
MGudgin: This can just be buf.clear(); (with no need for an unsafe call)

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,498 @@
...
+        // Propagate any error AFTER ensuring `buf` is still at len=0
+        // (no uninit u16s exposed to the reused-buffer caller path).
+        second?;
+    }
+
+    // `needed` is bytes written including the terminating NUL.
+    let init_u16 = (needed as usize / std::mem::size_of::<u16>()).min(buf.capacity());
+    unsafe {
+        buf.set_len(init_u16);
+    }
```
MGudgin, line 426: This one *does* need to be a set_len call #Resolved

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,498 @@
...
+                "EvtRender failed (Win32 error {:?})",
+                win_err
+            ));
+        }
+        if needed == 0 {
+            return Err(anyhow::anyhow!("EvtRender returned zero size"));
+        }
+        let needed_u16 = (needed as usize).div_ceil(std::mem::size_of::<u16>());
+        if buf.capacity() < needed_u16 {
+            buf.reserve(needed_u16 - buf.capacity());
```
MGudgin: >buf.reserve(needed_u16 - buf.capacity()); [](http://example.com/codeflow?start=12&length=41)

This won't actually reserve any extra space, because reserve measures from len, not capacity. So given len is zero here, reserve will get needed_u16 - 4096, which, if needed_u16 is less than 8192, will be less than 4096 and will be a no-op. If needed_u16 is greater than 8192, then it will still won't reserve the needed capacity.

Net: This should just be buf.reserve(needed_u16);

build.bat
```diff
@@ -9,6 +9,7 @@ set "WITH_NANVIX=0"
 set "WITH_WSLC=0"
 set "WITH_ISOLATION_SESSION=0"
 set "WITH_HYPERLIGHT=0"
+set "WITH_BFS=0"
```
MGudgin: Please revert the changes to this file.

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,520 @@
...
+/// (sandbox-only paths, deleted files, paths under a virtual mount), so
+/// querying the live filesystem with `Path::is_file()` / `is_dir()` would
+/// silently drop those write findings.
+///
+/// Heuristic: if the final path segment contains a `.` it is treated as a
+/// file and the parent directory is returned (so the directory becomes
+/// writable). Otherwise the path itself is treated as a directory and
+/// returned as-is. This matches the original PowerShell `extract_paths`
+/// behavior and over-grants in the rare directory-with-a-dot case, which
+/// is the safer side to err on.
```
MGudgin: Widening the grant for all 'paths that look like leaves' to the next node up seems bad. Why would we do this when the policy supports specific file grants?

src/host/plm/src/stop.rs
```diff
@@ -127,20 +130,77 @@ pub fn run(opts: StopOptions, exe_dir: &Path) -> Result<()> {
         p
     };
 
-    println!("Trace captured at {}.", trace_file.display());
+    if opts.verbose {
+        println!("Beginning event parsing, this may take several minutes");
+    }
```
MGudgin, line 135: If the code below can take several minutes, it seems unlikely to complete before the 30 second timeout specified at line 218 of plm_launch.rs

src/host/plm/src/config.rs
lilybarkley-msft: When writing to files, downlevel windows APIs will fail if they don't have write access to the entire directory. This was the case when I was testing it last month, @SalahKhan-MSFT do you know if this has been addressed?

lilybarkley-msft: I did some more local testing and it looks like it has. I'll get a fix in tonight

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,463 @@
...
+                rw_existing_norm.push(ev_norm.clone());
+            }
+            if seen_rw.insert(ev_norm) {
+                added_rw.push(ev.file_path.clone());
+            }
+            continue;
+        }
+
+        // Process Read Requests
+        if (ev.access_mask & READ_MASK) != 0 {
```
bbonaby, line 450: question: do we need the `if is_drive_root` clause here for the read as well?

src/host/plm/src/wpr_path.rs
```diff
@@ -82,13 +95,17 @@ fn resolve_wpr_path() -> PathBuf {
 /// would be defence against a strictly higher privilege than the one
 /// we hold. See the module doc for the full rationale.
 ///
-/// Returns `Err` if the resolved path doesn't exist on disk, which
-/// indicates a broken/stripped Windows install (WPT not present) —
-/// something the caller must surface with a clear message rather than
-/// let `CreateProcess` fail cryptically later.
+/// Returns `Err` if `GetSystemDirectoryW` failed outright, or if the
+/// resolved path doesn't exist on disk — both indicate a
+/// broken/stripped Windows install (WPT not present or `System32`
```
bbonaby: note: I think it's suppose to say `WPR not present` right?

src/host/plm/src/access_failure.rs
```diff
@@ -0,0 +1,378 @@
...
+pub(crate) fn normalize_file_path(p: &str) -> String {
+    let mut s = p.to_string();
+    normalize_file_path_in_place(&mut s);
+    s
+}
+
+/// Free-function sibling of `ParseAccumulator::is_skippable` exposed
+/// for unit tests that don't want to build a whole accumulator. Logic
+/// must stay in lock-step with the cached form on `ParseAccumulator`.
+#[cfg(test)]
```
bbonaby: question: rather than have to keep in lock-step, is it too much work to exercise that code path in the tests?

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+        let needed_u16 = (needed as usize).div_ceil(std::mem::size_of::<u16>());
+        if buf.capacity() < needed_u16 {
+            // `Vec::reserve(additional)` measures from `len`, not
+            // `capacity` — since `buf` is empty (cleared above),
+            // `additional == needed_u16` gets us `capacity ≥ needed_u16`.
+            buf.reserve(needed_u16);
+        }
+        let new_cap_u16 = buf.capacity();
+        let new_cap_bytes = new_cap_u16 * std::mem::size_of::<u16>();
+        let second = unsafe {
```
bbonaby, line 403: issue: missing unsafe comments

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// True when the trace produced nothing mergeable into a config.
+    pub fn is_empty(&self) -> bool {
+        self.valid_access_events.is_empty() && self.requested_capabilities.is_empty()
+    }
+}
+
+#[cfg(target_os = "windows")]
+fn to_wide_z(s: &str) -> Vec<u16> {
+    s.encode_utf16().chain(std::iter::once(0)).collect()
+}
```
bbonaby: issue: we can probably use the mxc string_utils crate for this.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+/// propagated rather than silently dropped — silent failure would look
+/// like a successful but short trace and cause PLM to under-grant on
+/// the next run.
+#[cfg(target_os = "windows")]
+fn for_each_event_xml<F>(trace_file: &Path, mut on_xml: F) -> Result<()>
+where
+    F: FnMut(&str) -> Result<()>,
+{
+    let path_w = to_wide_z(&trace_file.to_string_lossy());
+    let query_w = to_wide_z("*[System[EventID=14 or EventID=27]]");
```
bbonaby: note: you can use the `use windows::core::w;` module for this to have it done at compile time with `w!("*[System[EventID=14 or EventID=27]]");`

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+    let mut rendered_count: usize = 0;
+    const BATCH: usize = 256;
+    loop {
+        let mut events: [isize; BATCH] = [0isize; BATCH];
+        let mut returned: u32 = 0;
+        let next_ok = unsafe {
+            EvtNext(
+                h_query.0,
+                &mut events,
+                u32::MAX, // INFINITE
```
bbonaby, line 223: question: do we ever need to worry about hangs if this is indefinite? Also note that we'd want to make sure we have safety comments for unsafe code in this file see: https://std-dev-guide.rust-lang.org/policy/safety-comments.html#safety-comments

src/host/plm/src/access_event.rs
```diff
@@ -0,0 +1,19 @@
+//! Port of the `LearningModeAccessEvent` PowerShell class from
+//! `stop_plm_logging.ps1`. One file-access record from a PLM WPR trace.
+
+use chrono::{DateTime, Utc};
+
+#[derive(Debug, Clone)]
+// Fields kept for parity with the PowerShell version even though the
+// orchestrator currently only uses file_path and access_mask.
+#[allow(dead_code)]
+pub struct LearningModeAccessEvent {
```
bbonaby: question: Are all these needed now that the powershell version is no more? Or are they gonna be used in a future PR?

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+        event_data,
+    })
+}
+
+/// Mutable per-trace accumulator. Fields are `pub(crate)` so the
+/// sibling event-type decoders can write into them directly without an
+/// inflated method surface.
+pub(crate) struct ParseAccumulator<'a> {
+    #[allow(dead_code)]
+    pub(crate) current_directory: Option<&'a str>,
```
bbonaby: issue: This field is the only field that requires a lifetime, but it's marked as dead code. Is it actually needed ? Or will it be used in a future PR?

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+    let trimmed = match buf.iter().position(|&c| c == 0) {
+        Some(n) => &buf[..n],
+        None => &buf[..],
+    };
+    Ok(String::from_utf16_lossy(trimmed))
+}
+
+/// Decoded XML view of a single event's interesting fields.
+pub(crate) struct ParsedEvent {
+    #[allow(dead_code)]
```
bbonaby: note: another marked as dead code. Needed?

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+        .and_then(|n| n.attribute("ThreadID"))
+        .and_then(|s| s.parse::<u32>().ok())
+        .unwrap_or(0);
+
+    let mut event_data = Vec::new();
+    if let Some(ed) = root.children().find(|n| n.has_tag_name("EventData")) {
+        for child in ed.children().filter(|n| n.is_element()) {
+            let tag = child.tag_name().name();
+            if tag == "Data" || tag == "ComplexData" {
+                event_data.push(child.text().unwrap_or("").to_string());
```
bbonaby: question: should we unwraping into an empty string, is it possible that empty string is actually the text for the child? Might be useful to have "mxc-parse-error" instead.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+        let (cwd_lc_trimmed, cwd_lc_prefix) = match current_directory {
+            Some(cwd) => {
+                let trimmed = cwd.trim_end_matches('\\');
+                let is_drive_root = trimmed.len() == 2
+                    && trimmed.chars().nth(1) == Some(':')
+                    && trimmed
+                        .chars()
+                        .next()
+                        .map(|c| c.is_ascii_alphabetic())
+                        .unwrap_or(false);
```
bbonaby: note: we're unwrapping into a false here. Would is_drive_root being false due to an error affect code execution outcomes? I'm wondering if it's worth it erroring out here in general.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+            verbose,
+            valid_access_events: Vec::new(),
+            requested_capabilities: HashSet::new(),
+        }
+    }
+
+    /// Hot-path CWD / drive-letter filter for access events. Uses
+    /// precomputed lowercase forms of `current_directory` to avoid two
+    /// `String` allocs per event.
+    pub(crate) fn is_skippable(&self, file_path: &str) -> bool {
```
bbonaby, line 686: note: This function would probably be a bit easier to parse if we spelt out the variable names instead of abbreviating them.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// and silently swallows malformed records (so a bad event mid-trace
+    /// doesn't abort the rest). EventID=27 (UI violation) is recognized
+    /// by the XPath filter today but contributes no relaxation until the
+    /// UI-policy PR.
+    fn consume(&mut self, xml: &str) {
+        let Some(ev) = parse_event_xml(xml) else {
+            return;
+        };
+        match ev.event_id {
+            27 => {
```
bbonaby: issue: should be a static variable with a description in the file rather than a constant integer. This pattern will help if we add more id's.

src/host/plm/src/access_failure.rs
```diff
@@ -0,0 +1,378 @@
...
+        .unwrap_or(0);
+
+    if acc.verbose {
+        println!("{app_path}");
+        println!("{file_path}");
+    }
+
+    trim_backslashes_in_place(&mut file_path);
+
+    acc.valid_access_events
```
MGudgin, line 110: This pushes every matching event into valid_access_events: Vec<LearningModeAccessEvent> (declared at event_parser.rs:313) via acc.valid_access_events.push(...) here, with no dedup. A trace where an app repeatedly accesses the same file, e.g. opens foo.txt for read 500,000 times over the trace ends up pushing 500,000 near-identical entries into the vector, resulting in excessive and unnecessary memory usage. 

Suggest replacing the Vec with a HashMap<String, LearningModeAccessEvent> keyed on the normalized file_path. On each new event: if the key is absent, insert it as-is; if the key is already present, don't push a second entry but instead update the existing entry's access_mask field to existing.access_mask |= new.access_mask (bitwise OR). That way a file first opened read-only and later opened for write still ends up correctly flagged as read+write, while the map holds at most one entry per unique file path regardless of how many times it was accessed.

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,463 @@
...
+        {
+            continue;
+        }
+
+        // Already-covered short-circuit for readwrite policy.
+        if rw_existing_set.contains(&ev_norm)
+            || path_starts_with_any_norm(&ev_norm, &rw_existing_norm)
+        {
+            continue;
+        }
```
MGudgin, line 410: If an event path is nested under a directory already covered by readwritePaths (e.g. C:\Users\gudge\AppData\Local\SomeApp\Cache\blob_00042.tmp when C:\Users\gudge\AppData\Local is already granted), rw_existing_set.contains(&ev_norm) returns false since the exact nested path was never inserted. It then falls back to path_starts_with_any_norm(&ev_norm, &rw_existing_norm), an O(N) scan over every existing granted prefix. Without event dedup (see the access_failure.rs:99 comment), 100,000 accesses to that same nested file trigger 100,000 identical O(N) scans. This will be less severe assuming the dedup fix from the access_failire.rs comment is implemented (each unique file only reaches this check once), but the fallback itself is still worth hardening independently.

Suggest caching positive prefix matches so repeats of the same path short-circuit in O(1) even before dedup is added upstream:

if rw_existing_set.contains(&ev_norm) { continue; }
if path_starts_with_any_norm(&ev_norm, &rw_existing_norm) {
    rw_existing_set.insert(ev_norm.clone());
    continue;
}

This inserts ev_norm into rw_existing_set the first time a prefix match is found, so any later event for that exact same path hits the cheap contains check instead of re-scanning rw_existing_norm.

src/host/plm/src/access_failure.rs
```diff
@@ -0,0 +1,378 @@
...
+        .get_mut(APP_PATH_INDEX)
+        .map(std::mem::take)
+        .unwrap_or_default();
+    if !app_path.is_empty() {
+        if let Some(tail) = file_path.get(3..) {
+            if !tail.is_empty() && app_path.ends_with(tail) {
+                return;
+            }
+        }
+    }
```
MGudgin, line 74: This self-access filter uses app_path.ends_with(tail), a raw suffix match, where tail = file_path.get(3..), i.e. the accessed event's path with just its drive letter (C:) stripped, not just its filename. Because the comparison is suffix-based rather than an exact match, it produces false positives in at least two ways:

1. Extension-only collision: if file_path is short, e.g. C:\exe, then tail is just "exe", and every .exe process's app_path trivially ends in those same 3 letters. So app_path.ends_with("exe") is true for any audited executable, regardless of its real location, and the access to C:\exe gets silently dropped as "self-access" even though it has nothing to do with the running app.
2. Same-filename-different-location collision: if the app being audited is app_path = "\Device\HarddiskVolume3\Tools\my_app.exe", and an unrelated event reports file_path = "C:\my_app.exe" (a different copy/decoy/backup at the drive root), then tail = "my_app.exe", and app_path.ends_with("my_app.exe") is true, even though C:\my_app.exe is a completely different file from the one actually running.

In both cases a genuinely unrelated file access is silently dropped via the return on line 67, with no counter or warning that anything was filtered, the event just vanishes from the learning-mode output.

Fix: Compare the path portions exactly instead of by suffix. Derive the app's own relative path the same way tail is derived from file_path (strip the volume prefix), then require an exact (case-insensitive) match rather than ends_with:

let app_tail = app_path.splitn(4, '\').nth(3).unwrap_or("");
if !tail.is_empty() && tail.eq_ignore_ascii_case(app_tail) {
    return;
}

This still correctly skips true self-access events (accessed path equals the app's own binary), but stops treating "shares a trailing substring" as equivalent to "is the same file."

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,463 @@
+//! Port of the config-update logic from `stop_plm_logging.ps1`.
```
MGudgin, line 1: >Port of the config-update logic [](http://example.com/codeflow?start=4&length=31)

This module contains over 400 lines of complex path normalization, access-mask bitwise checks, case-insensitive deduplication, deny-path enforcement, and JSON tree mutation (the core policy-merge logic in update_from_access_events, starting at line 336). Despite being pure data manipulation, fully isolated from the OS and the filesystem, it has zero unit tests. Two real bugs were already found in this exact function during review (the missing dedup causing repeated O(N) prefix scans, and the now-fixed parent-widening drop), which is a direct symptom of shipping 400+ lines of branching logic with no tests locking in behavior.

Fix: Add a #[cfg(test)] mod tests block with deterministic test cases. Feed mock LearningModeAccessEvent structures into update_from_access_events and assert on the mutated config JSON output. At minimum, cover: read vs write mask handling, denied-path exact and prefix matches, already-covered short-circuiting for both readwritePaths and readonlyPaths, drive-root write rejection, and duplicate/repeated events for the same file.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+///
+/// `EvtNext` batch size is intentionally large (256) to reduce
+/// user→kernel transitions on traces with tens of thousands of events.
+/// End-of-stream is distinguished from real errors by matching
+/// `ERROR_NO_MORE_ITEMS`; any other `EvtNext` / `EvtRender` failure is
+/// propagated rather than silently dropped — silent failure would look
+/// like a successful but short trace and cause PLM to under-grant on
+/// the next run.
+#[cfg(target_os = "windows")]
+fn for_each_event_xml<F>(trace_file: &Path, mut on_xml: F) -> Result<()>
```
MGudgin: This function, for_each_event_xml, is the real ETW query and render path (EvtQuery, batched EvtNext, EvtRender), including buffer growth, native error handling, and handle cleanup via EvtHandleOwned. It has zero test coverage. Every existing test in this module only feeds synthetic XML strings directly into parse_events_from_xml, bypassing this function entirely. This is the actual Windows code path used in production to walk a real .etl trace, so none of its real behavior (batch boundaries at 256 events, end-of-stream detection via ERROR_NO_MORE_ITEMS, EvtRender buffer resizing, handle leaks on error paths) is exercised by any test.

Fix: Add Windows integration tests that run this function against small real or synthetically-generated .etl trace files, covering multi-batch traces (more than 256 events), oversized rendered XML requiring buffer growth, EvtNext and EvtRender failures other than end-of-stream, and confirming handles are released even when on_xml returns an error partway through a trace.

src/host/plm/readme.md
```diff
@@ -2,28 +2,33 @@
...
 2. **Run** — the operator runs the workload. The OS-side permissive sandbox logs `EventID=14` / `EventID=27` for every access that *would* have been denied.
-3. **Stop** — `plm stop` calls `wpr -stop <trace.etl>` and records the captured trace location.
-4. **Parse / Merge** — *(arrives in later PRs)* the `.etl` is walked with `EvtQuery` / `EvtRender` and findings are merged into a copy of the input config as `Adjusted_<name>.json`.
+3. **Stop** — `plm stop` calls `wpr -stop <trace.etl>` and walks the `.etl` with `EvtQuery` / `EvtRender`.
+4. **Parse** — for each `EventID=14`, the parser pulls the file path / access mask. Capability ACE-blob decoding lands in a later PR; `EventID=27` UI relaxation lands in a later PR.
+5. **Merge** — file paths are added to `filesystem.readwritePaths` / `filesystem.readonlyPaths` on a copy of the input config. The `Adjusted_*.json` writer arrives in the next PR; this PR only prints the per-event summary.
```
MGudgin, line 15: >5.  [](http://example.com/codeflow?start=0&length=3)

The README still describes the blank-config preview as future work in three places, but it is already implemented and active in this PR.

Line 15 says "The Adjusted_*.json writer arrives in the next PR; this PR only prints the per-event summary." Line 61 says "The diff against a blank config preview arrives in later PRs." Both are stale. log.rs (lines 85-108) already merges extracted paths into a blank config and prints the resulting JSON as the actual current output of plm log, not a future feature.

Line 24's table entry for src/log.rs, "Interactive mode: Enter to start, Enter to stop, then diff vs a blank config," is also inconsistent with the surrounding text since it describes the diff as already happening while line 61 says it is still future work.

Fix: Update the README to document the current "blank config after merge" output that log.rs actually produces today, and remove the stale future-work wording in both places. The Adjusted_*.json writer claim in line 15 is a separate, still-accurate future-work statement (see the CLI help finding for main.rs) and should stay, but should not be conflated with the blank-config preview, which is already shipped.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+        false
+    }
+
+    /// Per-event entry point. Decodes the XML, dispatches by event id,
+    /// and silently swallows malformed records (so a bad event mid-trace
+    /// doesn't abort the rest). EventID=27 (UI violation) is recognized
+    /// by the XPath filter today but contributes no relaxation until the
+    /// UI-policy PR.
+    fn consume(&mut self, xml: &str) {
```
MGudgin, line 737: The consume function silently swallows any malformed or unparseable event. If parse_event_xml returns None, the function just returns at line 400 with no counter, no warning, and no record that anything was skipped. parse_events still reports overall success in this case. A corrupt or schema-divergent .etl file, or a future Windows update that changes the rendered XML shape, could silently drop an arbitrary number of real access events and produce a config that looks complete but under-grants, with no visible signal to the user that anything went wrong.

Fix: Track a count of skipped/malformed events (e.g. a malformed_count field on ParseAccumulator, incremented in the None branch before returning) and surface it in ParseResult. Have the caller (stop.rs / log.rs) print a warning when malformed_count is greater than zero, or fail outright if the malformed ratio exceeds a sane threshold, so a corrupt or unexpectedly-shaped trace doesn't produce a silently incomplete policy.

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,463 @@
...
+        }
+    }
+
+    Ok(AddedPaths {
+        readwrite: added_rw,
+        readonly: added_ro,
+    })
+}
+
+pub fn write_added_paths_summary(added: &AddedPaths) {
```
MGudgin: write_added_paths_summary prints every discovered path unconditionally, with no verbose or interactive gate. This output goes to plm.exe's stdout, which under wxc-exec --audit runs as a separate elevated child (wxc-exec itself never elevates, per the invariant documented in plm_launch.rs). Because ShellExecuteExW plus runas cannot use normal pipe inheritance across the elevation boundary, run_plm_elevated redirects the elevated plm.exe child's stdout and stderr to temp files, then unconditionally reads both files into Vec at plm_launch.rs:253-254, regardless of exit code or verbosity. audit.rs only decides whether to replay (print) that buffer afterward, at line 111, gated on verbose or a non-zero exit code.

So on a successful, non-verbose run, the unelevated wxc-exec parent still pays the full cost of buffering every line this function printed: reading it from a temp file into memory, just to discard it. A trace with a large number of distinct newly-granted paths means this function's unconditional output directly drives that wasted temp-file and memory cost on every single audit run.

Fix: Gate write_added_paths_summary's output behind a verbose or interactive flag so a non-verbose plm.exe invocation produces little or no stdout in the first place, removing the need to buffer it at all on the happy path. Alternatively, cap the size of what run_plm_elevated reads back rather than reading the whole file unconditionally.

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,463 @@
...
+        let stripped = part.trim_end_matches(['.', ' ']);
+        // Reject pure traversal segments and other empty segments
+        // (`..` or `.`, or empty after trim — these arise from
+        // malformed inputs or doubled backslashes in non-UNC paths).
+        if part == "." || part == ".." || stripped.is_empty() {
+            return None;
+        }
+        parts.push(stripped.to_string());
+    }
+    Some(parts.join("\\"))
```
MGudgin: normalize_path is called once per event inside the main per-event loop in update_from_access_events. It splits the path on backslash and calls parts.push(stripped.to_string()) for every component (line 193 and further down for each segment), then joins them back together at the end with parts.join. This produces roughly 5 to 10 small heap allocations per path: one String per path component plus the final join allocation. Across a trace with millions of events and no deduplication upstream (see the earlier dedup findings), this adds up to a large number of avoidable allocations purely for path normalization.

Fix: Avoid building an intermediate Vec of owned Strings and then joining. Write directly into a single pre-allocated output String as each segment is validated, instead of collecting into parts and joining afterward, for example:

let mut out = String::with_capacity(trimmed.len());
for (i, part) in trimmed.split('\').enumerate() {
    let stripped = part.trim_end_matches(['.', ' ']);
    if part == "." || part == ".." || stripped.is_empty() { return None; }
    if i > 0 { out.push('\'); }
    out.push_str(stripped);
}

This keeps the same validation and rejection logic but reduces the per-path allocation count to at most one.

src/host/plm/src/event_parser.rs
```diff
@@ -0,0 +1,499 @@
...
+    pub(crate) event_id: u32,
+    pub(crate) time_created: DateTime<Utc>,
+    pub(crate) process_id: u32,
+    pub(crate) thread_id: u32,
+    /// EventData/Data values in document order.
+    pub(crate) event_data: Vec<String>,
+}
+
+pub(crate) fn parse_event_xml(xml: &str) -> Option<ParsedEvent> {
+    let doc = roxmltree::Document::parse(xml).ok()?;
```
MGudgin: >roxmltree::Document::parse(xml).ok()?; [](http://example.com/codeflow?start=14&length=38)

parse_event_xml builds a full roxmltree DOM tree, via roxmltree::Document::parse at line 251, for every single ETW event, just to pull a handful of fields out of it: EventID, TimeCreated, ProcessID, ThreadID, and a small EventData list (line 280-288). Building a full DOM for each event on a trace with potentially millions of events dominates CPU time and produces significant allocation and garbage pressure, especially if the event dedup fix upstream is not yet in place.

Fix: Replace the DOM parse with a streaming pull-parser such as quick-xml::Reader, extracting only the fields actually consumed without materializing the whole tree. 

At some point, we should probably look at switching the ETW render call from EvtRenderEventXml to EvtRenderEventValues so the needed fields (EventID, TimeCreated, ProcessID, ThreadID, EventData values) are obtained directly from the OS as typed values, avoiding XML stringification and parsing entirely.

src/host/plm/src/access_failure.rs
MGudgin: Yeah, my review tool flagged this too

src/host/plm/src/access_failure.rs
```diff
@@ -0,0 +1,378 @@
...
+const LEARNING_MODE_INDEX: usize = 0;
+const RESOURCE_TYPE_INDEX: usize = 1;
+pub(crate) const FILE_PATH_INDEX: usize = 2;
+const APP_PATH_INDEX: usize = 3;
+const ACCESS_MASK_INDEX: usize = 5;
+
+/// Per-event consume helper for `EventID=14`. Applies the post-XPath
+/// filters and pushes a `LearningModeAccessEvent` into
+/// `acc.valid_access_events` on success.
+pub(crate) fn consume_access_failure(acc: &mut ParseAccumulator<'_>, mut ev: ParsedEvent) {
```
MGudgin: consume_access_failure is the dispatcher for EventID=14 records and decides, via mount-point suppression, current-directory and drive-letter filtering, self-access matching, and invalid-filename-character checks, which events actually reach policy generation. The existing tests in this file cover the individual helper functions in isolation (normalization, is_skippable, and so on), but nothing drives consume_access_failure itself end to end through a realistic mix of these branches. There is no test confirming, for example, that a mount-point-manager event is dropped, a too-short or non-drive-letter path is skipped, a genuine self-access is filtered, and a normal valid event still reaches acc.valid_access_events, all through the single dispatcher that actually wires these checks together in order.

Fix: Add tests that build ParsedEvent fixtures (or drive them through parse_events_from_xml) covering each branch of consume_access_failure: a MountPointManager path, a current-directory match, a too-short path, a non-drive-letter path, a true self-access match, a malformed access mask, and a normal valid event, then assert on the resulting acc.valid_access_events contents for each case. This closes the gap between "the helpers are individually correct" and "the dispatcher wires them together correctly."

src/host/plm/src/config.rs
```diff
@@ -0,0 +1,463 @@
...
+    config: &mut Value,
+    bin_path: &str,
+    events: &[LearningModeAccessEvent],
+    deny_set: &HashSet<String>,
+    verbose: bool,
+) -> Result<AddedPaths> {
+    let mut added_rw: Vec<String> = Vec::new();
+    let mut added_ro: Vec<String> = Vec::new();
+    let mut seen_rw: HashSet<String> = HashSet::new();
+    let mut seen_ro: HashSet<String> = HashSet::new();
```
MGudgin: seen_rw and seen_ro are redundant with rw_existing_set / ro_existing_set: their insert(...) calls execute and do mutate the sets, but the false branch (an already-seen path) is unreachable under the loop's current invariant, so they never actually reject an entry. By the time execution reaches the write branch (line 388) or read branch (line 421), the already-covered short-circuit earlier in the loop (lines 380-385 for writes, 422-427 for reads) has already guaranteed ev_norm isn't in rw_existing_set / ro_existing_set, otherwise the loop would have continued before getting here. rw_existing_set (or ro_existing_set) is then updated at line 412 (or 434), immediately before the corresponding seen_rw/seen_ro check, so any later duplicate of the same ev_norm is caught by that earlier short-circuit on its next occurrence, never reaching seen_rw/seen_ro a second time. This holds across write/read interaction too: a later read after an earlier write to the same path is rejected by the readwrite short-circuit before reaching seen_ro, and a later write after an earlier read is a genuine first-time write insertion, so it's still the first seen_rw insertion. So every ev_norm that reaches seen_rw.insert(...) or seen_ro.insert(...) is guaranteed to be a first-time insert, making the sets redundant with the existing-policy sets while still paying for a full extra HashSet of normalized-path allocations across the whole trace.

Fix: Remove seen_rw and seen_ro entirely, including their declarations at lines 309-310, and push into added_rw / added_ro unconditionally once execution reaches lines 412-416 or 434-438, since reachability at that point already guarantees uniqueness

MGudgin (changes requested):
:clock1:

src/host/plm/src/stop.rs
MGudgin: You resolved this but did not change the constant value. Are you resolving as "Won't Fix" or something else?

MGudgin (changes requested):
You need to rebase.

src/host/plm/src/stop.rs
lilybarkley-msft: It should have been bumped up to 10 minutes

MGudgin (approved):

PR #587, opened 2026-06-27
Permissive Learning Mode 4/6 Capability

lilybarkley-msft:
## 📖 Description

**PR 4 of 6** — stacked on PR3. Adds **capability extraction**.

- New `extract_caps` module: decodes the DACL ACE blob attached to each `EventID=14` event and resolves capability SIDs to names via `DeriveCapabilitySidsFromName`.
  - Process-static capability table (built once via `build_capability_table`) so the per-event hot path avoids the syscall + `LocalAlloc`/`LocalFree` round-trip.
  - `CapabilityIndex` for O(1) SID-bytes → name resolution per ACE.
- `event_parser`:
  - `ParsedEvent` gains `complex_data_4_idx` (index of the 5th `<ComplexData>` sibling — the ACE blob) so the largest per-event field is borrowed, not cloned.
  - `ParseAccumulator` gains the capability table/index.
- `access_failure::consume_access_failure` walks the ACE blob through `extract_caps_with_index_into` and accumulates `requested_capabilities`.
- `config::merge_capabilities` — full body replaces the PR3 stub. Resolves the containment sub-object case-insensitively, dedupes case-insensitively, sorts via an allocation-free ASCII compare, and emits a stderr warning when the backend has no `capabilities` array.
- New `plm extract-caps` CLI subcommand for parser debugging on a single hex ACE blob.

UI policy lands in PR5.

## 🔗 References

- Base: **PR3** (`user/lilybarkley/plm-pr3-config-gen`)
- Next: PR5 — UI policy (`EventID=27` decode + `uiPolicy` merge)

## 🔍 Validation

- `cargo test -p plm --target x86_64-pc-windows-msvc` — **52 passed** (15 new: ACE blob decode, capability dedupe/sort, `merge_capabilities` with mixed-case inputs and unsupported-backend warning path).

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/587)

### DIFF
--- a/src/host/plm/src/access_failure.rs
+++ b/src/host/plm/src/access_failure.rs
@@ -8,11 +8,8 @@
 //!     prefixes -> DOS form),
 //!   * the post-XPath filters (current-directory, drive-letter,
 //!     self-access, invalid filename chars),
-//!   * the per-event accumulator helper that pushes the resulting
-//!     access event into `ParseAccumulator`.
-//!
-//! ACE-blob → capability-name extraction lands in a later PR; this PR
-//! only collects file paths and access masks.
+//!   * the per-event accumulator helper that feeds the DACL ACE blob
+//!     through `extract_caps` and pushes the resulting access event.
 //!
 //! `ParseAccumulator` (in `event_parser`) owns the mutable state;
 //! `consume_access_failure` is the only public entry point.
@@ -28,13 +25,31 @@ pub(crate) const FILE_PATH_INDEX: usize = 2;
 const APP_PATH_INDEX: usize = 3;
 const ACCESS_MASK_INDEX: usize = 5;
 
-/// Per-event consume helper for `EventID=14`. Applies the post-XPath
-/// filters and pushes a `LearningModeAccessEvent` into
-/// `acc.valid_access_events` on success.
+/// Per-event consume helper for `EventID=14`. Walks the DACL ACE blob
+/// through `extract_caps`, applies the post-XPath filters, and pushes a
+/// `LearningModeAccessEvent` into `acc.valid_access_events` on success.
 pub(crate) fn consume_access_failure(acc: &mut ParseAccumulator, mut ev: ParsedEvent) {
+    if let Some(idx) = ev.complex_data_4_idx {
+        // Borrow rather than clone — the ACE hex blob was already pushed
+        // by `parse_event_xml`; the other EventData slots taken below
+        // (0/1/3) live at different indices so this is safe.
+        if let Some(blob) = ev.event_data.get(idx) {
+            let blob_str = blob.as_str();
+            if !blob_str.trim().is_empty() {
+                let _ = crate::extract_caps::extract_caps_with_index_into(
+                    blob_str,
+                    &acc.capability_index,
+                    acc.verbose,
+                    &mut acc.requested_capabilities,
+                );
+            }
+        }
+    }
+
     // Pull the file path. Absent paths typically mean capability-only
-    // resource accesses; the capability-extraction PR will use the
-    // DACL ACE blob for those, but this PR drops them.
+    // resource accesses whose capability has already been collected
+    // from the DACL above. Take the slot out via `mem::take` so we can
+    // normalise + trim in place without a second `String` allocation.
     let mut file_path = match ev.event_data.get_mut(FILE_PATH_INDEX) {
         Some(s) if !s.is_empty() => std::mem::take(s),
         _ => return,
@@ -226,7 +241,10 @@ pub(crate) fn is_skippable(
     current_directory: Option<&str>,
     verbose: bool,
 ) -> bool {
-    crate::event_parser::ParseAccumulator::new(current_directory, verbose).is_skippable(file_path)
+    // An empty capability table is fine: `is_skippable` only consults the
+    // cached CWD / drive-letter state, not the capability index.
+    crate::event_parser::ParseAccumulator::new(current_directory, verbose, Vec::new())
+        .is_skippable(file_path)
 }
 
 /// Shared `EventID=14` XML fixture used by tests in this module and
@@ -339,7 +357,7 @@ mod tests {
             make_event_xml("C:\\Users\\test\\foo.txt", "0x1"),
             make_event_xml("C:\\Users\\test\\bar.txt", "0x2"),
         ];
-        let result = parse_events_from_xml(xmls.iter(), None, false);
+        let result = parse_events_from_xml(xmls.iter(), None, false, Vec::new());
         assert_eq!(result.valid_access_events.len(), 2);
         assert_eq!(
             result.valid_access_events[0].file_path,
@@ -364,7 +382,7 @@ mod tests {
             "<not-an-event/>".to_string(),
             valid_b,
         ];
-        let result = parse_events_from_xml(xmls.iter(), None, false);
+        let result = parse_events_from_xml(xmls.iter(), None, false, Vec::new());
         assert_eq!(
             result.valid_access_events.len(),
             2,
@@ -408,7 +426,7 @@ mod tests {
     #[test]
     fn consume_drops_mount_point_manager() {
         let xmls = [make_event_xml("\\Device\\MountPointManager", "0x1")];
-        let result = parse_events_from_xml(xmls.iter(), None, false);
+        let result = parse_events_from_xml(xmls.iter(), None, false, Vec::new());
         assert!(result.valid_access_events.is_empty());
     }
 
@@ -420,7 +438,7 @@ mod tests {
             make_event_xml("C:\\repo\\src\\main.rs", "0x1"),
             make_event_xml("C:\\other\\x.txt", "0x1"),
         ];
-        let result = parse_events_from_xml(xmls.iter(), Some("C:\\repo"), false);
+        let result = parse_events_from_xml(xmls.iter(), Some("C:\\repo"), false, Vec::new());
         assert_eq!(result.valid_access_events.len(), 1);
         assert_eq!(result.valid_access_events[0].file_path, "C:\\other\\x.txt");
     }
@@ -433,7 +451,7 @@ mod tests {
             make_event_xml("abc", "0x1"),
             make_event_xml("\\\\server\\share\\x", "0x1"),
         ];
-        let result = parse_events_from_xml(xmls.iter(), None, false);
+        let result = parse_events_from_xml(xmls.iter(), None, false, Vec::new());
         assert!(result.valid_access_events.is_empty());
     }
 
@@ -442,7 +460,7 @@ mod tests {
     #[test]
     fn consume_drops_invalid_filename_chars() {
         let xmls = [make_event_xml("C:\\foo*bar.txt", "0x1")];
-        let result = parse_events_from_xml(xmls.iter(), None, false);
+        let result = parse_events_from_xml(xmls.iter(), None, false, Vec::new());
         assert!(result.valid_access_events.is_empty());
     }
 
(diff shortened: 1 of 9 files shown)


src/host/plm/src/config.rs
```diff
@@ -651,16 +671,166 @@ pub fn write_requested_capabilities_summary(requested: &HashSet<String>, verbose
...
+        None => {
+            // Backend has no `capabilities` array in its schema (or is
+            // unknown to PLM). Emit a stderr warning so the operator
+            // knows the discovered capabilities are being dropped on
+            // the floor, rather than silently writing a section the
+            // runner will reject.
+            if !requested.is_empty() {
+                eprintln!(
+                    "[plm] warning: containment '{containment_name}' has no `capabilities` \
+                     array in its schema; dropping {} discovered capabilit{}: {}",
```
daniedud-MSFT: Nit: Typo?  capabilit{}

lilybarkley-msft: Formatting for capabilit(ies/y)

src/host/plm/src/event_parser.rs
```diff
@@ -842,10 +921,220 @@ mod tests {
...
+        assert!(
+            result.requested_capabilities.is_empty(),
+            "no 5th <ComplexData> means no capability blob"
+        );
+        assert_eq!(result.valid_access_events.len(), 1);
+        assert_eq!(result.parse_failures, 0);
+    }
+
+    #[test]
+    fn malformed_ace_blob_is_counted_as_a_parse_failure() {
```
MGudgin, line 1060: Could we add a regression test for a trace where every record is malformed and no paths or capabilities are collected? It should verify that `parse_failures > 0`, `ParseResult::is_empty()` remains true, and no `Adjusted_*.json` file is written.

MGudgin (commented):
Adversarial review comment validation probe.

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+        ));
+    }
+
+    Ok(AceSlice {
+        ace_type,
+        ace_flags,
+        access_mask,
+        sub_authority_count,
+        sid_bytes: &buf[sid_offset..sid_offset + sid_size],
+        next_cursor: sid_offset + sid_size,
```
MGudgin, line 719: The walker accepts `ACE_TYPE_ACCESS_ALLOWED_CALLBACK`, but advances only through the SID. Callback ACEs may contain conditional `OpaqueData` after the SID, which would then be interpreted as another ACE header and cause the whole record to be discarded. Can we either advance using the full ACE size from the ETW representation or stop treating callback ACEs as supported until their framing is established?

src/host/plm/src/config.rs
```diff
@@ -853,14 +853,128 @@ pub fn write_requested_capabilities_summary(requested: &HashSet<String>, verbose
...
+            mergeable.len()
+        );
+        return Ok(());
+    };
+
+    // Preserve existing entries verbatim (including any non-string
+    // values, which validation elsewhere owns) and append only the
+    // discovered names that are not already present under a
+    // case-insensitive comparison.
+    let mut merged: Vec<String> = existing
```
MGudgin: This says existing capability entries are preserved verbatim, but non-string values are converted with `to_string()` and then emitted as JSON strings. For example, `true` becomes `"true"` and an object becomes escaped JSON text. Could we retain the existing `Value` nodes and use string projections only for case-insensitive comparison? That preserves malformed input for downstream validation rather than silently rewriting it.

src/host/plm/src/access_failure.rs
```diff
@@ -28,13 +25,61 @@ pub(crate) const FILE_PATH_INDEX: usize = 2;
...
+                // Fail closed. The walker inserts matches as it goes, so
+                // a blob that is valid up to a corrupt tail would
+                // otherwise contribute capabilities from a record we
+                // know is malformed. Since the blob is
+                // attacker-influenceable and the output is a security
+                // policy, stage matches in a scratch set and promote
+                // them only if the entire walk succeeds; on failure the
+                // staged matches are dropped and the record is counted
+                // as data loss.
+                acc.ace_matches.clear();
```
MGudgin: This per-event clear means the output set passed to the walker never saturates across the trace, despite the allocation comment in `extract_caps.rs`. Each event allocates a new `String` for every matched capability before the trace-wide set rejects duplicates. Could the staging/index use static or borrowed capability identifiers so allocation happens only when a capability is first promoted for the trace?

src/host/plm/src/event_parser.rs
```diff
@@ -771,7 +824,31 @@ pub fn parse_events(
     current_directory: Option<&str>,
     verbose: bool,
 ) -> Result<ParseResult> {
-    let mut acc = ParseAccumulator::new(current_directory, verbose);
+    // Build the capability table once. Each entry requires a
+    // `DeriveCapabilitySidsFromName` syscall + LocalAlloc/LocalFree
+    // pair; the table is process-static so doing this per-event would
+    // dominate wall-time on large traces.
+    let built = extract_caps::build_capability_table_with_diagnostics();
```
MGudgin: Could capability discovery be injected into `parse_events` rather than queried directly here? As written, a fixed `.etl` regression fixture can produce different results across Windows versions and cannot be tested deterministically on non-Windows CI. Moving live OS discovery to the CLI boundary would make this integration path snapshot-testable.

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+    /// TODO(golden-fixture): this module's ACE layout (12-byte header,
+    /// SID at `[12..]`) is inherited from `extract_caps.ps1` and differs
+    /// from the packed Win32 `ACCESS_ALLOWED_ACE` (SID at `[8..]`).
+    /// Every fixture here is built with the same layout it asserts, so
+    /// the suite cannot tell the two apart. Capture one real
+    /// `EventID=14` DACL blob, paste its hex here, and assert the
+    /// expected capability set — that single fixture is what makes the
+    /// layout constants authoritative.
+    #[test]
+    #[ignore = "needs a real captured EventID=14 ACE blob; see TODO above"]
```
MGudgin, line 1338: This ignored test leaves the foundational 12-byte ACE layout validated only by fixtures generated with the same offsets they assert. Before landing, can we add one real captured EventID=14 blob with its expected capability set and enable this test? That would distinguish the chosen layout from the packed Win32 `ACCESS_ALLOWED_ACE` layout.

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+    "phoneLineTransportManagement",
+    "userDataAccountsProvider",
+    "userDataSystem",
+    "userSystemId",
+    "cortanaPermissions",
+    "cortanaSpeechAccessory",
+];
+
+#[derive(Debug, Clone)]
+pub struct CapabilityEntry {
```
MGudgin: Do we need `CapabilityEntry`/`CapabilityTable` as an intermediate production representation? The OS builder fills this vector and `CapabilityIndex::from_table` immediately consumes it to build the actual map. Building the index directly, with a test-only insertion seam, would remove an allocation/conversion pass and reduce the module surface.

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+/// Walk every ACE in `buf` and return the case-insensitively-deduped set
+/// of capability names matched along the way. When `verbose` is true, a
+/// per-ACE diagnostic line is emitted to stdout.
+///
+/// Emit a one-shot warning when a capability SID is recognized on an
+/// allow ACE that grants nothing.
+///
+/// Fires at most once per process so a long trace cannot spam the
+/// console. See the call site for why this is worth reporting.
+fn warn_zero_mask_capability_once(name: &str) {
```
MGudgin: A process-global `Once` means only the first parse run can report this condition, and it makes warning behavior order-dependent in tests. Could this warning state live on `ParseAccumulator` (or in parse-local diagnostics) so each trace reports the condition once and tests can assert it deterministically?

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+
+/// Best-effort string form of a SID for diagnostics. Returns `None` if the
+/// bytes aren't a valid SID.
+#[cfg(target_os = "windows")]
+pub fn sid_to_string(sid_bytes: &[u8]) -> Option<String> {
+    // `IsValidSid` dereferences the pointer to read the fixed header and
+    // the declared sub-authority array, so prove the slice is large
+    // enough first. Without this, callers passing a short slice (e.g.
+    // `&[]` or `&[1]`) make Win32 read past the end of the allocation.
+    encoded_sid_len(sid_bytes)?;
```
MGudgin, line 447: These guards are load-bearing because verbose `extract-caps` input is operator-controlled and reaches Win32 SID APIs. Could we keep a focused regression test proving an undersized SID in the verbose path is rejected before `IsValidSid`/`ConvertSidToStringSidW`? That will help prevent a future refactor from bypassing this check.

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+///
+/// **Partial writes on error.** The walk inserts as it goes, so a
+/// buffer that decodes cleanly up to a corrupt tail leaves the
+/// already-matched names in `found` *and* returns `Err`. Callers that
+/// feed a security policy must therefore treat `Err` as fail-closed:
+/// pass a per-blob scratch set and discard it on error rather than
+/// pointing this at their accumulated set. See
+/// `access_failure::consume_access_failure`, which stages into
+/// `ParseAccumulator::ace_matches` and only promotes on `Ok`.
+pub fn invoke_ace_walk_with_index_into(
```
MGudgin: There are six extraction entry-point variants covering combinations of hex/bytes, supplied/rebuilt index, and supplied/local scratch storage, while production appears to need only the one-shot CLI path and this parser hot path. Could the test-only wrappers be removed and tests perform their setup explicitly to keep the API surface smaller?

src/host/plm/src/extract_caps.rs
```diff
@@ -0,0 +1,1323 @@
...
+}
+
+/// Non-Windows stub.
+#[cfg(not(target_os = "windows"))]
+pub fn sid_to_string(_sid_bytes: &[u8]) -> Option<String> {
+    None
+}
+
+/// Result of resolving a SID against the capability table.
+pub enum SidResolution<'a> {
```
MGudgin: `SidResolution` is immediately flattened by its only caller; the variant distinction is used only for verbose text. Would returning the capability name plus an `is_group` flag be simpler than maintaining a dedicated public enum for this debug-only distinction?

src/host/plm/src/event_parser.rs
```diff
@@ -761,6 +813,7 @@ impl ParseAccumulator {
         ParseResult {
             valid_access_events: self.valid_access_events,
             requested_capabilities: self.requested_capabilities,
+            parse_failures: self.parse_failures,
```
MGudgin, line 838: The counter behavior is covered, but I could not find an integration test for the operator-visible warning emitted from `into_result` or its count. Could we add a stderr-capturing test so parse failures cannot become silent during later refactoring?

MGudgin (changes requested):
Comments above? below?

MGudgin (approved):

PR #593, opened 2026-06-30
[Bubblewrap/LXC/WSLC] Add object-based filesystem-policy validation, most restrictive wins

SohamDas2021:
## 📖 Description
Detect when two different policy path strings resolve to the same host filesystem object (via symlink, hard link, or bind mount) but carry conflicting intents — e.g. readwritePaths:["/mnt/storage/data"] and deniedPaths:["/data"] where /data is a symlink to the former. The existing string-level normalizer in config_parser only catches the same path string in multiple lists; it can't see object aliases.

New wxc_common::filesystem_object::normalize_object_conflicts() stats each policy path (following symlinks), groups by object identity ((dev,ino) on unix; volume-serial + 128-bit file-id on Windows) and tightens every alias of a shared object to the strictest intent among them (deny > readonly > readwrite), logging a warning per tightened path. It never rejects the config and skips paths it can't stat (existence is a separate advisory).

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/593)

### DIFF
--- a/src/backends/bubblewrap/common/src/bwrap_runner.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_runner.rs
@@ -101,11 +101,24 @@ impl SandboxBackend for BubblewrapScriptRunner {
         self.validate(request)?;
         // Object-based FS-policy normalization (D6): tighten aliases of the same
         // host object to the strictest intent (deny > ro > rw). Done here, close
-        // to mount, on a clone — config_parser stays string-only and the TOCTOU
-        // window between check and mount is minimized.
-        let mut normalized = request.clone();
-        wxc_common::filesystem_object::normalize_object_conflicts(&mut normalized.policy, logger);
-        let child = self.spawn_bwrap(&normalized, logger, stdio)?;
+        // to mount — config_parser stays string-only and the TOCTOU window
+        // between check and mount is minimized. Only clone the request when an
+        // aliasing conflict actually needs tightening (the common case is none).
+        let normalized;
+        let request = match wxc_common::filesystem_object::normalize_object_conflicts(
+            &request.policy,
+            logger,
+        ) {
+            Some(policy) => {
+                normalized = ExecutionRequest {
+                    policy,
+                    ..request.clone()
+                };
+                &normalized
+            }
+            None => request,
+        };
+        let child = self.spawn_bwrap(request, logger, stdio)?;
         Ok(Box::new(BubblewrapSandboxProcess::new(child)))
     }
 }
--- a/src/backends/bubblewrap/common/src/bwrap_runner.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_runner.rs
@@ -103,20 +103,22 @@ impl SandboxBackend for BubblewrapScriptRunner {
         // host object to the strictest intent (deny > ro > rw). Done here, close
         // to mount — config_parser stays string-only and the TOCTOU window
         // between check and mount is minimized. Only clone the request when an
-        // aliasing conflict actually needs tightening (the common case is none).
+        // aliasing conflict actually needs tightening (the common case is none);
+        // an unresolvable path with deniedPaths present fails closed.
         let normalized;
         let request = match wxc_common::filesystem_object::normalize_object_conflicts(
             &request.policy,
             logger,
         ) {
-            Some(policy) => {
+            Ok(Some(policy)) => {
                 normalized = ExecutionRequest {
                     policy,
                     ..request.clone()
                 };
                 &normalized
             }
-            None => request,
+            Ok(None) => request,
+            Err(msg) => return Err(ScriptResponse::error(&msg)),
         };
         let child = self.spawn_bwrap(request, logger, stdio)?;
         Ok(Box::new(BubblewrapSandboxProcess::new(child)))
--- a/src/backends/bubblewrap/common/src/bwrap_runner.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_runner.rs
@@ -99,7 +99,13 @@ impl SandboxBackend for BubblewrapScriptRunner {
     ) -> Result<Box<dyn SandboxProcess>, ScriptResponse> {
         validate_common(request)?;
         self.validate(request)?;
-        let child = self.spawn_bwrap(request, logger, stdio)?;
+        // Object-based FS-policy normalization (D6): tighten aliases of the same
+        // host object to the strictest intent (deny > ro > rw). Done here, close
+        // to mount, on a clone — config_parser stays string-only and the TOCTOU
+        // window between check and mount is minimized.
+        let mut normalized = request.clone();
+        wxc_common::filesystem_object::normalize_object_conflicts(&mut normalized.policy, logger);
+        let child = self.spawn_bwrap(&normalized, logger, stdio)?;
         Ok(Box::new(BubblewrapSandboxProcess::new(child)))
     }
 }
--- a/src/backends/lxc/common/src/lxc_runner.rs
+++ b/src/backends/lxc/common/src/lxc_runner.rs
@@ -91,10 +91,22 @@ impl LxcScriptRunner {
     fn run_internal(&self, request: &ExecutionRequest, logger: &mut Logger) -> ScriptResponse {
         // Object-based FS-policy normalization (D6): tighten aliases of the same
         // host object to the strictest intent (deny > ro > rw) before building
-        // mounts. See `wxc_common::filesystem_object`.
-        let mut normalized = request.clone();
-        wxc_common::filesystem_object::normalize_object_conflicts(&mut normalized.policy, logger);
-        let request = &normalized;
+        // mounts. See `wxc_common::filesystem_object`. Only clone the request
+        // when an aliasing conflict actually needs tightening.
+        let normalized;
+        let request = match wxc_common::filesystem_object::normalize_object_conflicts(
+            &request.policy,
+            logger,
+        ) {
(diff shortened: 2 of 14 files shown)


src/core/wxc_common/src/filesystem_object.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    // Final intent per entry (defaults to its declared intent).
+    let mut final_intent: Vec<Intent> = entries.iter().map(|(_, it)| *it).collect();
+
+    for members in groups.values() {
+        if members.len() < 2 {
+            continue;
+        }
+        // Strictest intent among the aliases.
+        let target = members.iter().map(|&i| entries[i].1).max().unwrap();
```
huzaifa-d, line 293: What about a case like 
"readwritePaths": ["/data"],         
"deniedPaths":    ["/data/secret.txt"] 

I think it should be handled.

huzaifa-d (changes requested):

src/core/wxc_common/src/filesystem_object.rs
SohamDas2021: /data /data/secret.txt are different filesystem objects in a parent/child relationship, not aliases of the same object. So, this does not fall into this work item where we are specifically addressing paths that lead to the same object.

This specific example falls under "most specific path wins", captured in https://microsoft.visualstudio.com/OS/_workitems/edit/62861414, will be working on it after this.

tests/configs/lxc_filesystem_object.json
```diff
@@ -0,0 +1,19 @@
+{
+  "version": "0.4.0-alpha",
```
MGudgin: Can we update all the configs to the latest schema version?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -99,7 +99,13 @@ impl SandboxBackend for BubblewrapScriptRunner {
     ) -> Result<Box<dyn SandboxProcess>, ScriptResponse> {
         validate_common(request)?;
         self.validate(request)?;
-        let child = self.spawn_bwrap(request, logger, stdio)?;
+        // Object-based FS-policy normalization (D6): tighten aliases of the same
+        // host object to the strictest intent (deny > ro > rw). Done here, close
+        // to mount, on a clone — config_parser stays string-only and the TOCTOU
+        // window between check and mount is minimized.
+        let mut normalized = request.clone();
```
MGudgin: Can we just clone request.policy? Cheaper than cloning the whole request, and we only mutate the policy piece. #Resolved

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -89,6 +89,13 @@ impl LxcScriptRunner {
 
     /// Core execution logic.
     fn run_internal(&self, request: &ExecutionRequest, logger: &mut Logger) -> ScriptResponse {
+        // Object-based FS-policy normalization (D6): tighten aliases of the same
+        // host object to the strictest intent (deny > ro > rw) before building
+        // mounts. See `wxc_common::filesystem_object`.
+        let mut normalized = request.clone();
```
MGudgin: Same comment here about cloning just the policy piece. #Resolved

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -857,6 +857,14 @@ impl WSLContainerRunner {
     ) -> ScriptResponse {
         let _ = writeln!(logger, "[WSLC] Starting WSL Container runner");
 
+        // Object-based FS-policy normalization (D6): tighten aliases of the same
+        // host object to the strictest intent (deny > ro > rw) before mapping to
+        // volume mounts. See `wxc_common::filesystem_object`. (A path moved to
+        // `denied` is simply not mounted by WSLC — unmounted = invisible.)
+        let mut normalized = request.clone();
```
MGudgin: Same comment here about cloning just the policy piece. #Resolved

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: Downstream consumes the whole &ExecutionRequest, so a policy-only clone wouldn't avoid rebuilding the request. Instead normalize_object_conflicts now returns Option<ContainerPolicy> and we clone only when there's an actual conflict (the common no-alias case clones nothing).

Applies to all three comments.

huzaifa-d (approved):

src/core/wxc_common/src/filesystem_object.rs
```diff
@@ -0,0 +1,430 @@
...
+        entries.push((p.clone(), Intent::ReadOnly));
+    }
+    for p in &policy.denied_paths {
+        entries.push((p.clone(), Intent::Denied));
+    }
+
+    // Group entry indices by object identity (skip unstattable paths).
+    let mut groups: HashMap<ObjectId, Vec<usize>> = HashMap::new();
+    for (i, (path, _)) in entries.iter().enumerate() {
+        if let Some(id) = object_id(path) {
```
MGudgin: object_id()  returns  None  for any stat/open failure and this loop silently skips it. If the denied alias is the one that fails to stat, its conflicting RW/RO alias is left untightened and the config proceeds looser than intended, with no diagnostic. For a security boundary this should fail closed (at least when  deniedPaths  are present), and unresolved paths should be logged so "no conflict" is distinguishable from "couldn't examine this path".

src/core/wxc_common/src/filesystem_object.rs
```diff
@@ -0,0 +1,430 @@
...
+
+/// Resolve a path to its filesystem-object identity, following symlinks so two
+/// names for the same target collide. Returns `None` when the path cannot be
+/// stat'd (missing, permission denied, etc.) — such paths are left untouched
+/// (their existence is handled separately as an advisory warning).
+#[cfg(unix)]
+fn object_id(path: &str) -> Option<ObjectId> {
+    use std::os::unix::fs::MetadataExt;
+    // `metadata` follows symlinks, giving the target object's identity.
+    let md = std::fs::metadata(path).ok()?;
```
MGudgin: >metadata [](http://example.com/codeflow?start=22&length=8)

This adds synchronous host I/O ( std::fs::metadata  here;  CreateFileW + GetFileInformationByHandleEx  on the Windows path at line 92) for every policy path directly on the runner launch path ( bwrap_runner.rs:108 ,  lxc_runner.rs:97 ,  wsl_container_runner.rs:866 ), with no timeout or cancellation. A dead autofs/NFS mount, offline UNC share, or blocking special file can halt sandbox startup indefinitely. Consider bounding this (per-path timeout with fail-open logging, or an abandonable worker) or moving it off the critical launch thread. Let's at least file a followup to make this async/canceallable.

src/core/wxc_common/src/filesystem_object.rs
```diff
@@ -0,0 +1,430 @@
...
+///
+/// Run this *after* the string-level normalization in `config_parser`, close to
+/// where the backend builds its mounts.
+pub fn normalize_object_conflicts(
+    policy: &ContainerPolicy,
+    logger: &mut Logger,
+) -> Option<ContainerPolicy> {
+    if policy.readwrite_paths.is_empty()
+        && policy.readonly_paths.is_empty()
+        && policy.denied_paths.is_empty()
```
MGudgin: A conflict needs ≥2 paths, so a policy with ≤1 total path can never tighten. Extending this early-return to  total paths <= 1  would skip all the file I/O in that case (although granted it would only be file I/O on one path...)

src/core/wxc_common/src/filesystem_object.rs
```diff
@@ -0,0 +1,430 @@
...
+        std::fs::hard_link(&target, &alias).unwrap();
+
+        let (t, a) = (target.to_str().unwrap(), alias.to_str().unwrap());
+        let p = policy(&[t, a], &[], &[]);
+        let mut logger = Logger::new(Mode::Buffer);
+        // Same object, same intent: no tightening, so no new policy is produced.
+        assert!(normalize_object_conflicts(&p, &mut logger).is_none());
+    }
+
+    #[cfg(unix)]
```
MGudgin, line 464: >( [](http://example.com/codeflow?start=9&length=1)

Every test that actually verifies tightening ( symlink_rw_and_denied_tightens_to_denied ,  hardlink_rw_and_readonly... ,  denied_wins_over_both... ) is  #[cfg(unix)] ; the only cross-platform test exercises the no-op  None  path. So the Windows  object_id /rebuild path has zero tightening coverage in  cargo test . Windows supports hard links — a  #[cfg(windows)]  tightening test (or the resolver seam from the previous comment) would close this. I know this PR was about Linux, but I worry about this falling through the cracks...

SohamDas2021: Removed the gate, it now runs on Windows too, exercising the real Windows  object_id  (FileIdInfo) grouping + rebuild + tightening path.

src/core/wxc_common/src/filesystem_object.rs
SohamDas2021: Thanks! This is a valid point, I will file a github issue to follow up after merging the PR.

MGudgin (approved):

PR #598, opened 2026-07-01
[Bubblewrap/LXC/WSLC] Add filesystem-policy delegation check

SohamDas2021:
## 📖 Description
Enforce that the container is never granted more filesystem access than the invoking user already holds: every readwritePaths entry requires the user to have read+write access, and every readonlyPaths entry requires read access. deniedPaths are unbounded (denying needs no access) and are not checked. A path the user cannot access is rejected, so a sandboxed process can't reach files the caller themselves couldn't.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/598)

### DIFF
--- a/src/core/wxc_common/src/filesystem_access.rs
+++ b/src/core/wxc_common/src/filesystem_access.rs
@@ -77,9 +77,13 @@ impl AccessMode {
 /// On Windows it probes the caller's effective access with `CreateFileW`
 /// (requesting `GENERIC_READ` / `GENERIC_READ | GENERIC_WRITE`, opened with
 /// `FILE_FLAG_BACKUP_SEMANTICS` so directories are covered too). A successful
-/// open means the access is granted; an `ERROR_ACCESS_DENIED` failure means it
-/// is not; any other failure is treated as undeterminable (`None`) and skipped.
-/// This covers files *and* directories — including the common WSLC
+/// open means the access is granted. Matching the Unix posture, the check
+/// **fails closed**: only a genuinely missing path (`ERROR_FILE_NOT_FOUND` /
+/// `ERROR_PATH_NOT_FOUND` — surfaced separately by the existence warning) is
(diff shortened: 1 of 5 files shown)


src/core/wxc_common/src/filesystem_access.rs
```diff
@@ -0,0 +1,359 @@
...
+    match handle {
+        Ok(h) if !h.is_invalid() => {
+            // SAFETY: `h` is a valid handle returned by CreateFileW.
+            unsafe {
+                let _ = CloseHandle(h);
+            }
+            Some(true)
+        }
+        _ => {
+            // Only an explicit access denial is a delegation failure. Any other
```
huzaifa-d: Copilot tells me - Unix uses a **skip-list** (only genuine "missing" is skipped; the default is reject). Windows uses a **reject-list** (only explicit access-denied is rejected; the default is skip). A `ERROR_SHARING_VIOLATION`, `ERROR_LOCK_VIOLATION`, or transient IO error on a real, existing file hits the Windows default and is **silently skipped** — the exact opposite of what the same situation does on Unix.

So on Unix "non-existent is surfaced separately" is achieved by listing the missing errnos and rejecting everything else; on Windows the author achieves it by skipping everything except one error. Those are not the same posture.

huzaifa-d (changes requested):

src/core/wxc_common/src/filesystem_access.rs
SohamDas2021: Thanks for the catch, seems to be a valid concern, will address.

SohamDas2021: Addressed

huzaifa-d (approved):

PR #603, opened 2026-07-02
Plumb deniedPaths to BaseContainer fs_deny, gated on SANDBOX_CAP_FS_DENY

MGudgin:
This PR plumbs `deniedPaths` into the BaseContainer SandboxSpec `fs_deny` FlatBuffer field so the OS enforces denied paths natively at Tier 1, gated on the OS-advertised `SANDBOX_CAP_FS_DENY` capability. When the capability is absent, a denied-paths policy falls through to a DACL-enforcing AppContainer tier (T2/T3) rather than staying on Tier 1 (where the dispatcher applies no host DACL), so deny is never silently dropped.

Details

* BaseContainerSpecification.fbs / sandbox_spec_generated.rs: add the additive `fs_deny:[string]` field so deniedPaths reach Experimental_CreateProcessInSandbox.
* base_container_runner.rs: emit `fs_deny` from `denied_paths`; `validate` accepts deniedPaths only when the OS advertises SANDBOX_CAP_FS_DENY (via the existing `base_container_supports_deny_paths()` query) and otherwise fails closed -- this runner has no DACL fallback of its own.
* fallback_detector.rs: `detect` keeps a denied-paths policy on Tier 1 only when the capability is present (native fs_deny, no host DACL); otherwise it falls through to the AppContainer DACL tiers. Adds a process-cached, test-seamed `base_container_supports_deny_paths()` wrapper mirroring `is_base_container_usable()`.
* Rename the capability bit SANDBOX_CAP_DENY_PATHS -> SANDBOX_CAP_FS_DENY to match the OS-side capability name; correct the dispatcher comments to state why T1 needs no host DACL (the OS enforces deny natively when the capability is present, and denied-paths policies never reach T1 otherwise).
* error.rs: add DENIED_PATHS_FEATURE_DISABLED_MSG for the capability-absent direct-caller rejection.

Tests

* cargo test -p appcontainer_common -p wxc_common: 129 + 359 passed, including new fallback-detector tests (T1 native-deny when capability present; fall-through when absent) and runner validate tests (accept/reject by capability) using a new MXC_FORCE_DENY_PATHS test seam.
* cargo fmt --all -- --check; cargo clippy -p appcontainer_common -p wxc_common --all-targets -- -D warnings; cargo check --workspace --all-targets: all clean.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/603)

### DIFF
--- a/external/windows-sdk/BaseContainerSpecification.fbs
+++ b/external/windows-sdk/BaseContainerSpecification.fbs
@@ -26,6 +26,12 @@
 // --conform verifies that every field/enum/table in the old schema still exists
 // at the same vtable slot, type, and default in the new schema. Any incompatible
 // change produces an error. Add this as a build step or PR gate to catch regressions.
+//
+//  Steps:
+//      cp SandboxSpec.fbs SandboxSpec.previous.fbs
+//      <make changes>
+//      .\flatc.exe --conform SandboxSpec.previous.fbs SandboxSpec.fbs
+//
 
 namespace SandboxTechSpecLayout;
 
@@ -68,6 +74,9 @@ table SandboxSpec {
 
     // Integrity behavior
     integrity:IntegrityLevel = system_default;
+
+    // File Paths that have been denied access
+    fs_deny:[string];
 }
 
 table proxy_info {
--- a/src/backends/appcontainer/common/src/fallback_detector.rs
+++ b/src/backends/appcontainer/common/src/fallback_detector.rs
@@ -164,16 +164,23 @@ pub fn detect(
 
     // Tier 1 — BaseContainer
     if prefer_base_container && is_base_container_usable() {
-        if denied {
-            ensure_dacl_augmentation_allowed(policy)?;
-            verify_write_dac_all(&policy.denied_paths)?;
+        // Keep deny on Tier 1 only with native fs_deny support
+        // (SANDBOX_CAP_FS_DENY); T1 applies no host DACL, so otherwise
+        // fall through to a DACL-enforcing tier.
+        if !denied || base_container_supports_deny_paths() {
+            return Ok(TierDecision {
+                tier: IsolationTier::BaseContainer,
+                needs_dacl_augmentation: false,
+                bfscfg_path: None,
+                warnings,
+            });
         }
-        return Ok(TierDecision {
-            tier: IsolationTier::BaseContainer,
-            needs_dacl_augmentation: denied,
-            bfscfg_path: None,
-            warnings,
-        });
+        warnings.push(
+            "BaseContainer usable but this OS does not advertise SANDBOX_CAP_FS_DENY; \
+             deniedPaths cannot be enforced natively at Tier 1 — falling back to AppContainer \
+             for deniedPaths enforcement"
+                .to_string(),
+        );
     }
     // Tier 2 — AppContainer + BFS
     //
@@ -186,8 +193,7 @@ pub fn detect(
     // entirely and fall through to Tier 3 (AppContainer + DACL).
     if cfg!(feature = "tier2_bfs") {
         warnings.push(
-            "BaseContainer API not present or not preferred; falling back to AppContainer + BFS"
-                .to_string(),
+            "BaseContainer tier not selected; falling back to AppContainer + BFS".to_string(),
         );
 
         // When the policy has no filesystem rules at all there is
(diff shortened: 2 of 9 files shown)


external/windows-sdk/BaseContainerSpecification.fbs
```diff
@@ -26,6 +26,12 @@
 // --conform verifies that every field/enum/table in the old schema still exists
 // at the same vtable slot, type, and default in the new schema. Any incompatible
 // change produces an error. Add this as a build step or PR gate to catch regressions.
+//
+//  Steps:
+//      cp SandboxSpec.fbs SandboxSpec.previous.fbs
+//      <make changes>
+//      & "$($env:OBJECT_ROOT)\onecore\base\appmodel\execmodel\vcpkg\$($env:_BuildAlt)\vcpkg_packages\flatbuffers_x64-host-windows\tools\flatbuffers\flatc.exe" --conform SandboxSpec.previous.fbs SandboxSpec.fbs
```
bbonaby: note: can probably just use `.\flatc.exe` here instead of the razzle path

src/backends/appcontainer/common/src/fallback_detector.rs
```diff
@@ -164,16 +164,23 @@ pub fn detect(
...
-            tier: IsolationTier::BaseContainer,
-            needs_dacl_augmentation: denied,
-            bfscfg_path: None,
-            warnings,
-        });
+        warnings.push(
+            "BaseContainer usable but this OS does not advertise SANDBOX_CAP_FS_DENY; \
+             deniedPaths cannot be enforced natively at Tier 1 — falling back to AppContainer \
+             for deniedPaths enforcement"
+                .to_string(),
```
bbonaby, line 182: question: does appcontainer enforcement for deny path have any security differences that we need to document?

src/core/wxc_common/src/error.rs
bbonaby, line 52: issue: looks like this is no longer used now?

src/core/wxc_common/src/error.rs
```diff
@@ -54,6 +54,15 @@ pub const DENIED_PATHS_NOT_SUPPORTED_MSG: &str =
      default unless granted via readwritePaths or readonlyPaths. Remove deniedPaths, \
      or narrow readwritePaths/readonlyPaths to exclude the path you wanted to deny.";
 
+#[cfg(target_os = "windows")]
+pub const DENIED_PATHS_FEATURE_DISABLED_MSG: &str =
```
bbonaby, line 59: question:  "rely on the AppContainer DACL tiers (which enforce deniedPaths via DENY ACEs) by not constructing the BaseContainer runner  directly." Should we be telling them what to pass into the schema instead so it's a more direct call to action?

src/backends/appcontainer/common/src/fallback_detector.rs
MGudgin: Documented in schema.md

src/core/wxc_common/src/error.rs
MGudgin: Currently still referenced by appcontainer_runner.rs

src/core/wxc_common/src/error.rs
MGudgin: Fixed.

src/core/wxc_common/src/error.rs
```diff
@@ -50,9 +50,18 @@ impl From<std::io::Error> for WxcError {
...
+     Paths are denied by default unless granted via readwritePaths or readonlyPaths. \
+     Remove deniedPaths, or narrow readwritePaths/readonlyPaths to exclude the path \
+     you wanted to deny.";
+
+#[cfg(target_os = "windows")]
+pub const DENIED_PATHS_FEATURE_DISABLED_MSG: &str =
+    "filesystem.deniedPaths cannot be enforced by the BaseContainer backend on this \
+     OS build: it does not advertise the native deny-paths capability \
+     (SANDBOX_CAP_FS_DENY via Experimental_QuerySandboxSupport). Run on a build with \
+     BaseContainer deny support, or use schema version '0.4.0-alpha' to select the \
```
bbonaby, line 63: note: `BaseContainer` I guess we should do a pass to rename base container to process container. Created issue #610 for this purpose.

bbonaby (approved):

src/core/wxc_common/src/error.rs
MGudgin: No. processcontainer is the platform neutral term. BaseContainer is the Windows specific process container.

PR #621, opened 2026-07-07
Add CodeQL static analysis workflow for Rust and TypeScript

huzaifa-d:

-----

## Summary

Adds `.github/workflows/CodeQL.yml` to enable CodeQL static analysis (SAST) for the repo. Verified against the live repo that CodeQL is the one security feature not yet present (Dependabot and Copilot code review already run); the code-scanning API currently returns "no analysis found".

**What it does**
- Runs on `push`/`pull_request` to `main` plus a weekly `schedule` (Mon 06:00 UTC), so newly-published advisories are caught even without a code change.
- Matrix over two languages, both **buildless** (`build-mode: none`):
  - **`javascript-typescript`** on `ubuntu-latest` (the SDK). CodeQL extracts straight from source; no build needed.
  - **`rust`** on `windows-latest`. CodeQL's Rust extractor is buildless-only, so it runs **no `cargo build`** — it resolves the crate graph via `cargo metadata` and extracts from source. The job still surfaces `src/rust-toolchain.toml` at the repo root (so the pinned toolchain is honored) and appends the public MxcDependencies feed via the existing `./.github/actions/setup-cargo-feed` composite action, so that dependency resolution succeeds (the private feed 401s otherwise).
- `permissions` scoped to `security-events: write` + `contents: read`; `concurrency` cancels superseded runs.

**Note on Rust coverage:** because the Rust extractor is buildless, this scan does not depend on compiling any feature-gated backend (hyperlight/microvm/windows_sandbox), and there are no backend-specific build workarounds to import. The extractor analyzes the workspace source directly. (An earlier revision of this PR used `build-mode: manual` with a `cargo build`; that failed because CodeQL's Rust extractor does not support `manual` — only `none` — so it was switched to the buildless mode. The description and code now both reflect the buildless approach.)

Results will populate the **Security → Code scanning** tab once merged to `main`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/621)

### DIFF
--- a/.github/workflows/CodeQL.yml
+++ b/.github/workflows/CodeQL.yml
@@ -38,7 +38,7 @@ jobs:
           - language: rust
             runner: windows-latest
             target: x86_64-pc-windows-msvc
-            build-mode: manual
+            build-mode: none
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


SohamDas2021:
Is the description correct? From the code it seems, the actual Rust job is build-mode: none (buildless), runs no cargo build, and the inline comment explicitly says, "No cargo build is run."

huzaifa-d:
Good catch, thanks. The description was stale.

bbonaby (approved):

PR #622, opened 2026-07-08
Inject and validate the two learning-mode capabilities

richiemsft:
Recognize `learningModeLogging` (deny-and-record) and `permissiveLearningMode` (audit / allow-all) as distinct, reserved internal capabilities for the AppContainer-family runners. Direct use of either name in `processContainer.capabilities` is rejected case-insensitively.

## 📖 Description

- Adds `LearningModeCapability` classification and effective-mode diagnostics shared by classic AppContainer and BaseContainer. `learningModeLogging` emits an informational deny-and-record note; `permissiveLearningMode` emits an always-visible security warning because it allows access.
- Reserves both learning-mode capability names from direct config use. `processContainer.learningMode: true` is the supported deny-and-record entry point, while `--audit` is the supported permissive entry point for Windows ProcessContainer only.
- `--audit` rejects every resolved backend except Windows ProcessContainer before mutating policy or starting a trace. For ProcessContainer it injects `permissiveLearningMode`, removes `learningModeLogging` if present, and drives the WPR/ETW permissive-learning-mode trace.
- Keeps internally injected learning-mode capabilities available in both debug and release builds so dedicated entry points have consistent runner behavior.
- Documents the three learning-mode flows in `docs/learning-mode/capabilities.md` and updates the README warning to describe the reserved capability names and supported entry points.

## 🔗 References
- #663 (config contract — adds `captureDenials` + its `mode` selector that maps onto these capabilities)
- #661 (consume the learning-mode OS API — the capture engine)

Recommended merge order: #663 (config) → #622 (runtime) → integration.

## 🔍 Validation

- `cargo test -p wxc_common -p appcontainer_common` — 454 + 141 pass (learning-mode classification / diagnostics tests updated)
- `cargo fmt --all -- --check`, `cargo clippy -p wxc_common -p appcontainer_common -p wxc --all-targets -- -D warnings`
- `cargo check --workspace --all-targets` (confirms the `request.audit` field removal has no other construction sites)
- `npm run test:unit` (SDK) — 187 pass
- `node scripts/versioning/{check-schema-codegen,check-sdk-types-codegen,validate-configs}.js`

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/622)


src/core/wxc_common/src/config_parser.rs
```diff
@@ -701,24 +701,23 @@ fn convert_wire_config(
             policy.least_privilege_mode = lp;
         }
 
-        // learningMode handling differs between debug and release.
+        // The learningMode boolean maps to the deny-and-record learning-mode
```
MGudgin, line 761: This is a breaking change and will require a schema and schema version change along with a breaking change note.

docs/learning-mode/capabilities.md
```diff
@@ -0,0 +1,79 @@
...
+while producing a record of what it tried and failed to reach.
+
+### `permissiveLearningMode` — audit / allow-all
+
+The OS records **every** access check and **allows** it. This is an audit mode:
+it answers "what would this workload touch if nothing were blocked?" but it does
+so by **not enforcing deny-by-default** for the duration of the run.
+
+Because it defeats containment, `permissiveLearningMode` emits a **security
+warning** in the run diagnostics. It is nevertheless **permitted in every build,
```
MGudgin: This is not true for basecontainer, permissivelearningmode passes through with no warning, and is not even gated by --audit

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -390,6 +390,74 @@ pub enum FilesystemMode {
...
+    /// (audit / allow-all). Does not enforce deny-by-default.
+    Permissive,
+}
+
+impl LearningModeCapability {
+    /// Classify a policy capability string as a learning-mode capability, or
+    /// `None` if it is an unrelated capability.
+    pub(crate) fn from_capability_name(name: &str) -> Option<Self> {
+        match name {
+            CAP_LEARNING_MODE => Some(Self::Learning),
```
MGudgin: This is a case sensitive match, but windows generates the SIDs in a case-insensitive way. So someone could specify PermissiveLearningMode in their config, it would pass through and no warning would be emitted and --audit would not be needed.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -505,23 +573,29 @@ impl AppContainerScriptRunner {
...
+        // audit) are distinct. Emit per-capability diagnostics for whichever are
+        // present (informational for `learningMode`, a security warning for
+        // `permissiveLearningMode`). See [`learning_mode_capability_diagnostics`].
+        for line in learning_mode_capability_diagnostics(&request.policy.capabilities) {
+            logger.log_line(&line);
         }
 
+        // `permissiveLearningMode` relaxes deny-by-default, so it is gated behind
+        // `--audit` (which sets `request.audit`). The diagnostics above already
+        // emitted the security warning; here we only enforce the gate.
```
MGudgin: Do we have tests for --audit?

docs/learning-mode/capabilities.md
```diff
@@ -0,0 +1,79 @@
...
+including release**, because it is a legitimate auditing tool. Callers are
+responsible for only enabling it against workloads and on hosts where allowing
+all accesses is acceptable.
+
+## How to enable them
```
MGudgin, line 51: >How to enable them [](http://example.com/codeflow?start=3&length=18)

Why is there no mention of --audit in this section?

MGudgin (changes requested):
This PR has several issues, but my main concerns are the weakening of security guarantees, the default behaviour change for appcontainer and the fact that basecontainer now also has differnet behaviour with no warning to the caller. I'm blocking merge.

richiemsft:
@MGudgin addressed your comments, left the permissive learning mode available in retail builds based on the teams chat.

src/core/wxc_common/src/config_parser.rs
richiemsft: learning mode was introduced here: https://github.com/microsoft/mxc/pull/543 at 0.8.0-dev schema.

Since it hasn't been part of any stable schema release, should I increase it to 0.8.1-dev? or keep it as is, just mention the new behavior?

docs/learning-mode/capabilities.md
```diff
@@ -0,0 +1,126 @@
...
+
+Injecting these capabilities makes the OS *emit* learning-mode events. The
+Windows-only `captureDenials` config switch drives collecting those events and
+surfacing the resulting denials to the caller. Its `mode` selects how each
+ungranted access is handled while it is recorded:
+
+- `mode: "block-and-log"` (default) maps onto `learningModeLogging`
+  (deny-and-record) — the app / user-configurable flow.
+- `mode: "allow-and-log"` maps onto `permissiveLearningMode` (allow-and-record)
+  — the fleet-auditing flow.
```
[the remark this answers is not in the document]
richiemsft: upcoming PR that enables this comes right after.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -758,47 +758,38 @@ fn convert_wire_config(
...
+        // Reject direct capability-array use case-insensitively because Windows
+        // derives capability SIDs case-insensitively.
         if let Some(caps) = ac.capabilities {
-            #[cfg(debug_assertions)]
-            if caps.iter().any(|c| c == "permissiveLearningMode") {
-                logger.log(
-                    "[mxc] permissiveLearningMode present in policy capabilities - AppContainer restrictions are NOT enforced\n",
+            if let Some(reserved) = caps.iter().find(|capability| {
+                capability.eq_ignore_ascii_case("learningModeLogging")
+                    || capability.eq_ignore_ascii_case("permissiveLearningMode")
```
MGudgin, line 788: **Comma-injection bypasses the reserved learning-mode capability rejection (BaseContainer path)**

_Status: pre-existing weakness, **not** introduced by this PR — the base-commit release strip (`retain(|cap| cap == "permissiveLearningMode")`) had the same element-wise, single-token blind spot, and `base_container_runner.rs`'s `caps.join(",")` predates this PR. Flagging it here because this PR reworks exactly this validation to reserve the learning-mode capabilities (and already fixed the sibling case-sensitivity gap in the same block), so closing the comma variant is a natural completion of that hardening rather than new scope._

The reservation check matches each array **element** whole-string:

```rust
// config_parser.rs:778
if let Some(reserved) = caps.iter().find(|capability| {
    capability.eq_ignore_ascii_case("learningModeLogging")
        || capability.eq_ignore_ascii_case("permissiveLearningMode")
}) { /* reject */ }
```

But `BaseContainerRunner` flattens the whole array into one comma-separated string that the OS sandbox API splits back apart on commas:

```rust
// base_container_runner.rs:343
Some(builder.create_string(&caps.join(",")))
```

So a caller can hide the reserved capability inside another element: `eq_ignore_ascii_case("internetClient,permissiveLearningMode")` is `false` → passes validation → `caps.join(",")` yields `...,internetClient,permissiveLearningMode,...` → the OS splits on the comma and grants `permissiveLearningMode` (allow-all, deny-by-default disabled) with **no `--audit` and no security warning**. In a release build this reintroduces exactly the bypass the reservation was added to close.

**Config that slips through today:**

```jsonc
{
  "process": { "commandLine": "cmd /c whoami" },
  "containment": "processcontainer",
  "processContainer": {
    "capabilities": ["internetClient,permissiveLearningMode"]
  }
}
```

Expected: rejected at parse time. Actual: accepted; on the BaseContainer tier the process runs allow-all with containment disabled.

**Suggested fix** — reject any comma-bearing element in this same block, before it can reach `caps.join(",")`:

```rust
if let Some(bad) = caps.iter().find(|c| c.contains(',')) {
    let msg = format!(
        "processContainer.capabilities entry '{bad}' must not contain a comma"
    );
    logger.log_line(&msg);
    return Err(WxcError::ConfigParse(msg));
}
```

Better still, split each element on `,` and match every token against the reserved set so `"a,permissiveLearningMode"` is caught explicitly. Please add a regression test asserting the comma-packed element is rejected in **both** debug and release builds.

_Note: the classic AppContainer path derives a SID per array element, so a comma-bearing element is unlikely to resolve there — this bypass is specific to the BaseContainer `caps.join(",")` serialization, which is why the fix belongs at the shared config-parser layer rather than in one runner._

MGudgin (changes requested):

MGudgin (approved):

PR #623, opened 2026-07-08
[Bubblewrap] Mask denied files with a /dev/null bind instead of tmpfs

SohamDas2021:
## 📖 Description
The Bubblewrap backend masked every denied path with `--tmpfs`. Mounting a tmpfs over a **regular file** turns it into an empty *directory* (bwrap creates the mount point as a dir), changing the object's type and breaking tools that expect a regular file.

This PR classifies each denied path at runtime and masks it with the correct primitive:
 - **Directories** keep `--tmpfs` (empty, no leak).
 - **Files** (and other non-directory objects) are masked with a read-only bind of `/dev/null`, preserving a non-directory type with empty content.

This brings Bubblewrap in line with the LXC backend, which already distinguishes files from directories when masking.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/623)


tests/scripts/run_bwrap_denied_masking_test.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
huzaifa-d, line 1: Consider adding a symlink -> dir & symlink -> file test

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -374,6 +408,74 @@ mod tests {
         );
     }
 
+    /// A denied path classified as a **directory** (not in `denied_files`) is
```
huzaifa-d, line 403: Will denied `/data/secret` (dir→tmpfs) + denied `/data/secret/key` (file→/dev/null), work as expected?

SohamDas2021: This should work as expected. resolve_mount_order sorts ascending by path depth, so the parent is always emitted before the child. 

/data/secret will be a writable empty dir. We create the "key" mount point inside the writable tmpfs and bind /dev/null

tests/scripts/run_bwrap_denied_masking_test.sh
SohamDas2021: Thanks, bwrap actually can't create a mask mount point when the destination resolves through a pre-existing host symlink (i.e. the denied symlink's parent is bound in). I am updating the code to rewrite a denied symlink to its canonical target and mask that; I will document this in docs/bwrap-support/bubblewrap-backend.md with symlink->dir/file test coverage.

huzaifa-d (approved):

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -182,9 +208,17 @@ pub fn build_args(request: &ExecutionRequest, proxy_address: Option<&ProxyAddres
             FsIntent::ReadOnly => {
                 args.extend(["--ro-bind".into(), mount.path.clone(), mount.path.clone()]);
             }
-            // Denied: mask with an empty tmpfs so contents are invisible.
+            // Denied: mask so contents are invisible. A directory is masked
+            // with an empty `--tmpfs`; a file must be masked with
+            // `--ro-bind /dev/null` instead, because `--tmpfs` over a file would
+            // replace it with an empty *directory* (wrong type). `denied_files`
+            // holds the paths the runner classified as non-directories.
```
MGudgin: I've seen at least 4 copies of this comment or similar so far. Let's just record the dir vs file stuff once (probably in the backend.md file) and not repeat ourselves. #Resolved

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -486,6 +526,49 @@ fn cleanup_iptables(manager: &mut Option<NetworkIptablesManager>, logger: &mut L
...
+fn resolve_denied_symlinks(
+    policy: &wxc_common::models::ContainerPolicy,
+    logger: &mut Logger,
+) -> Option<Vec<String>> {
+    let mut changed = false;
+    let mut out = Vec::with_capacity(policy.denied_paths.len());
+    for p in &policy.denied_paths {
+        let is_symlink = std::fs::symlink_metadata(p)
+            .map(|md| md.file_type().is_symlink())
+            .unwrap_or(false);
```
MGudgin: I think this only checks if the leaf is a symlink (e.g. \usr\gudge\foo\bar\baz. But if \usr\gudge\foo is a symlink, bwrap will still abort with ENOENT, right? #Resolved

MGudgin (changes requested):
:clock1:

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -486,6 +526,49 @@ fn cleanup_iptables(manager: &mut Option<NetworkIptablesManager>, logger: &mut L
...
+        if is_symlink {
+            if let Ok(target) = std::fs::canonicalize(p) {
+                let target = target.to_string_lossy().into_owned();
```
MGudgin: >target.to_string_lossy().into_owned(); [](http://example.com/codeflow?start=29&length=38)

This will return incorrect strings for file paths that contain none UTF-8 characters. Which will result in a file path that was supposed to be denied, NOT being denied. #Resolved

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -486,6 +526,49 @@ fn cleanup_iptables(manager: &mut Option<NetworkIptablesManager>, logger: &mut L
...
+/// resolve it. Returns `Some(new_denied)` when at least one entry was rewritten,
+/// else `None` (so the caller avoids an unnecessary clone). Dangling/unresolvable
+/// symlinks are left as-is — there is nothing behind them to leak.
+fn resolve_denied_symlinks(
+    policy: &wxc_common::models::ContainerPolicy,
+    logger: &mut Logger,
+) -> Option<Vec<String>> {
+    let mut changed = false;
+    let mut out = Vec::with_capacity(policy.denied_paths.len());
+    for p in &policy.denied_paths {
```
MGudgin, line 551: >denied_paths [](http://example.com/codeflow?start=21&length=12)

Both this and the classification code at ~195 iterate over denied_paths. I suspect that for a variety of reasons, it would be better to do this work and the work at 195+ in a single function. #Resolved

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -486,6 +526,49 @@ fn cleanup_iptables(manager: &mut Option<NetworkIptablesManager>, logger: &mut L
...
+                    logger,
+                    "Bubblewrap: deniedPaths entry '{p}' is a symlink; masking its target \
+                     '{target}' instead."
+                );
+                out.push(target);
+                changed = true;
+                continue;
+            }
+        }
+        out.push(p.clone());
```
MGudgin, line 578: > out.push(p.clone()); [](http://example.com/codeflow?start=7&length=21)

If canonicalize fails, we still run this line, which seems wrong. I expect it will cause a ENOENT abort. #Resolved

src/backends/bubblewrap/common/src/bwrap_command.rs
SohamDas2021: Updated

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: This is valid, I have addressed it. Ty

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -486,6 +512,120 @@ fn cleanup_iptables(manager: &mut Option<NetworkIptablesManager>, logger: &mut L
...
+    })
+}
+
+/// Classify a denied path for masking: `true` = `--ro-bind /dev/null` (file),
+/// `false` = `--tmpfs` (directory). Only a real directory is directory-masked;
+/// files, symlink nodes, and unstattable paths are file-masked.
+fn is_file_mask_target(path: &str) -> bool {
+    std::fs::symlink_metadata(path)
+        .map(|md| !md.file_type().is_dir())
+        .unwrap_or(false)
```
MGudgin, line 592: If we want unstattable paths to be file-masked (per the comment at line 588), this should be unwrap_or(true) #Resolved

SohamDas2021: Thanks for the catch. Fixed the comment to match the code: unstattable paths are directory-masked (--tmpfs ), preserving the original unwrap_or(false) default; switching to unwrap_or(true) would silently regress every non-existent denied path from an empty-dir to a /dev/null file mask.

MGudgin (approved):

PR #627, opened 2026-07-09
User/dahoehna/honoring inbound rule

dhoehna:
## 📖 Description

LXC enforces inbound network policy as a default-deny chain installed **inside
the container's own network namespace** (`nsenter -t <pid> -n`), hooked into
the container's `INPUT` chain.  A packet destined to a container socket
traverses the container's `INPUT`, never the host's, so that is where the deny
is installed.

This branch targets main's `allowLocalNetwork` field.  It does **not** touch
the GA network schema (`ingress` / `hostLoopback` / `egress`), `wire.rs`,
`schemas/dev`, or the generated SDK bindings; schema ownership sits with
MGudgin and Stuart.  This lands the default-deny half of the roadmap's inbound
control (N2, `hostLoopback`); the permissive half stays blocked on the GA
schema and is refused today (see below).

Current behavior:

- `allowLocalNetwork: false` (default): the container `INPUT` chain accepts
  intra-container loopback (`-i lo`) and `ESTABLISHED,RELATED` return traffic,
  then drops new inbound (`--state NEW -j DROP`) and ends with a terminal
  `-j DROP`.  Container-initiated flows and in-container `127.0.0.1` are
  unaffected.
- Inbound is now actually enforced.  On main `allowLocalNetwork` is parsed but
  silently ignored; with this change a firewall-mode LXC container gets a
  default-deny `INPUT` chain in its own netns, so new inbound that previously
  reached it is dropped.
- `allowLocalNetwork: true` (permissive host-loopback inbound) **fails
  closed**.  Scoping inbound to host loopback needs a `loopbackPorts` schema
  field and an MXC-owned host-loopback forwarder that do not exist yet; the
  only rule available today is an unscoped `--state NEW -j ACCEPT` that would
  accept inbound from every interface and source (LAN and WAN).  The firewall
  path returns a hard not-yet-implemented error rather than install that
  over-broad accept.
- The refusal is scoped to the paths that can actually install rules.  It
  fires under `enforcementMode` `"firewall"` and `"both"`, and only when a
  container netns is present.  Under `"capabilities"` (the default) the
  firewall path is not used at all, and the netns-less Bubblewrap shared-net
  path builds an unhooked, inert chain and installs nothing — neither is
  refused.
- The chain is installed and torn down through both `iptables` and `ip6tables`, so IPv6 inbound cannot bypass an IPv4-only deny on a dual-stack container.  The IPv6 chain additionally accepts the ICMPv6 types RFC 4890 requires — Neighbor Discovery (133–136), Multicast Listener Discovery (130–132, 143), and errors 1–4 including packet-too-big for Path MTU Discovery — because a verbatim replay of the IPv4 argv would drop them as `--state NEW`.  Echo request/reply (128/129) and Redirect (137) stay dropped, so ordinary new inbound is still denied.

Fail-closed aborts (previously silent skips): a firewall run whose container
init PID cannot be discovered is aborted rather than run with the inbound
chain inert, and firewall mode with a live host IPv6 stack but an unusable
`ip6tables` is aborted rather than leaving an IPv4-only deny.

Egress (`defaultPolicy`, `allowedHosts`, `blockedHosts`) is a separate control
and is not enforced by this iptables path; egress restriction for Linux
backends is handled out of band (the cooperative proxy).

## ⚠️ Breaking configuration change

A config that runs today on main **fails after this merges** if it sets
`allowLocalNetwork: true` **and** `enforcementMode: "firewall"` or `"both"` on
LXC.  It fails at container provision/start time, not at parse time: the
container is created and started, `apply_firewall_rules` returns `Err`, the
runner destroys the container and returns
`ScriptResponse::error("Network policy error: allowLocalNetwork ... is not yet
implemented ...")`.

Unaffected: `allowLocalNetwork: false` or absent; any non-firewall mode
(including the default `Capabilities`, which returns early); and every backend
other than LXC (seatbelt still emits its own inbound rule).  No pre-existing
config in the repo breaks — no shipping fixture sets `allowLocalNetwork:
true`; the only `true` config is `tests/configs/lxc_local_network_allow.json`,
added by this PR as a negative test.

## 🔗 References

[62864412](https://task.ms/62864412)

## 🔍 Validation

Re-verified in WSL (Ubuntu-24.04) against this branch, `cd src`:

```bash
cargo test -p lxc_common                       # 65 passed, 0 failed, 0 ignored
cargo clippy -p lxc_common --all-targets       # clean
cargo fmt -p lxc_common -- --check             # clean
```

The suite covers default-deny argv ordering, the unconditional `-i lo` accept, the `INPUT` hook appearing only with a container netns, pinned IPv4 and IPv6 argv sequences (so the ICMPv6 additions cannot silently alter IPv4 behavior), and the `allowLocalNetwork` refusal decision table — refused under `"firewall"` and `"both"` with a netns; not refused under `"capabilities"`, without a netns, or when the flag is false — including that a refusal leaves `rules_applied()` false so teardown does not run.

Real `iptables`/`ip6tables` enforcement, netns entry, and the root-gated LXC
behavioral harness (`tests/scripts/run_lxc_local_network_test.sh`) are **not**
exercised by `cargo test` and have not been run here; they require a root
LXC + WSL2 host with a real container.  The full-workspace `cargo test` count
was not re-run in this pass.

## ✅ Checklist

- [X] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [X] Linked to an issue
- [X] Updated documentation (if applicable)
- [X] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes (see [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md))

## 📋 Issue Type

- [ ] Bug fix
- [X] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/627)


SohamDas2021:
Are your changes addressing all the requirements in https://microsoft.visualstudio.com/OS/_workitems/edit/62864412/?view=edit? 

"When ingress.hostLoopback is "allow", accept connections from host loopback only"

My understanding was that we would need the new config json schema changes to go in first and you will have a subsequent PR to address that? 

{
  "network": {
    "egress": {
      "default": "deny",
      "allow": [{ "to": [{ "cidr": "140.82.112.0/20" }], "ports": [{ "protocol": "tcp", "port": 443 }] }],
      "deny": [{ "to": [{ "cidr": "10.0.0.0/8" }] }]
    },
    "ingress": { "hostLoopback": "deny" },
    "proxy": { "http": "127.0.0.1:8080" }
  }
}

dhoehna:
> Are your changes addressing all the requirements in https://microsoft.visualstudio.com/OS/_workitems/edit/62864412/?view=edit?
> 
> "When ingress.hostLoopback is "allow", accept connections from host loopback only"
> 
> My understanding was that we would need the new config json schema changes to go in first and you will have a subsequent PR to address that?
> 
> { "network": { "egress": { "default": "deny", "allow": [{ "to": [{ "cidr": "140.82.112.0/20" }], "ports": [{ "protocol": "tcp", "port": 443 }] }], "deny": [{ "to": [{ "cidr": "10.0.0.0/8" }] }] }, "ingress": { "hostLoopback": "deny" }, "proxy": { "http": "127.0.0.1:8080" } } }

My changes are against the current schema.  I can make a second PR to read `ingress.hostloopback` when the schema change is merged into main.

If I change the code to read `ingress` then I can't test this change until the schema goes in.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -150,9 +150,25 @@ impl NetworkIptablesManager {
...
+
         Self::run_iptables(
-            &["-A", &self.chain_name, "-i", "lo", "-j", "ACCEPT"],
+            &[
+                "-A",
+                &self.chain_name,
+                "-i",
+                "lo",
+                "-j",
+                allow_local_network_verb,
```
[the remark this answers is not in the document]
dhoehna: @SohamDas2021 can you validate this comment?  The roadmap said to add this to the FORWARD chain.  I don't know who is correct.  Your roadmap, or copilot.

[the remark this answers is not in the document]
MGudgin: FWIW my adversarial review had this to say;

 network_iptables.rs  builds a per-container chain  MXC-<id> , then hooks it into the host with (line ~257):

iptables -I FORWARD -o <veth> -j MXC-<id>

The PR's change makes the first rule in that chain conditional:

iptables -A MXC-<id> -i lo -j {ACCEPT | DROP}   # verb from allow_local_network

There are two independent, compounding defects, plus a functional regression risk.

Defect A — the chain never sees loopback packets (unreachable no-op)

Linux routes a packet through exactly one of  INPUT  (locally destined),  OUTPUT  (locally generated), or  FORWARD  (routed through the host). This chain is reached only from  FORWARD , and only for packets whose output interface is  <veth> .

 -i lo  requires the input interface to be the host loopback. A packet cannot both originate on host  lo  and be forwarded out  <veth>  — loopback traffic is locally destined and never enters  FORWARD  at all. So  -i lo -j <anything>  in this chain matches the empty set. Flipping the verb ACCEPT↔DROP changes nothing observable — exactly what the PR's own runbook admits. This is why security, correctness, and intent-coherence all independently rated it High.

Defect B —  -i lo  is the wrong interface for "local network" semantically

Even if the rule were in a reachable chain,  lo  is loopback (127.0.0.0/8) — intra-host / intra-namespace traffic.  allowLocalNetwork  is meant to govern the local network (the LAN / host-adjacent subnet), which for a container flows over  eth0  / the bridge  veth , not  lo . Gating  lo  neither expresses nor enforces the intended policy.

The latent regression (why this isn't harmless scaffolding)

If someone "fixes" Defect A by moving this rule into the container's own netns ( INPUT ) while keeping the  -i lo  target, then  allowLocalNetwork:false  would DROP the container's own  127.0.0.1  traffic — breaking anything that binds/loopback-connects to localhost (local DNS stub resolvers, services talking to a sidecar on 127.0.0.1, health checks, many language runtimes). And it still wouldn't block real LAN access. So the current code isn't just inert — it encodes a shape that becomes actively harmful the moment it's naively made "reachable."

What correct enforcement looks like

Two viable designs:

1. Host-side, in the existing  FORWARD → veth  chain (smallest delta, keeps current architecture):
• Restore the unconditional  -i lo -j ACCEPT  (loopback must always pass).
• To deny inbound, add a NEW-connection drop scoped to the container, e.g.
 -A MXC-<id> -o <veth> -m conntrack --ctstate NEW -j DROP  (deny) / omit for allow — placed after an  ESTABLISHED,RELATED -j ACCEPT  rule so the container's own outbound-initiated return traffic survives. This is where finding #7 (ESTABLISHED must precede any conditional DROP) becomes load-bearing: get the ordering wrong and you kill established flows.
• Match on the container destination ( -d <containerIP>  or the inbound interface), not  -i lo .
2. Inside the container netns  INPUT  path (what the branch name "HonoringInboundRule" implies): apply the NEW-inbound DROP there, and keep  -i lo -j ACCEPT  unconditional in that netns.

Either way the fix must be paired with a behavioral test that traverses the real inbound path (finding #4) and a pure-function CI unit test over the emitted argv (findings #3, testability #12) — because none of this is currently exercised by  cargo test .

Bottom line: the verb-flip is a placeholder in the wrong chain against the wrong interface. It should not be presented as functional enforcement; real inbound control needs a NEW-conntrack rule scoped to the container's forwarded/inbound path, with loopback ACCEPT restored and ESTABLISHED ordering fixed.

docs/lxc-support/lxc-backend.md
```diff
@@ -117,6 +117,8 @@ Network policies are enforced via iptables/nftables rules applied to the contain
 | `defaultPolicy: "allow"` | Default ACCEPT rule on container veth |
 | `allowedHosts` | ACCEPT rules for specific IPs/CIDRs |
 | `blockedHosts` | DROP rules for specific IPs/CIDRs |
+| `allowLocalNetwork: true` | ACCEPT new inbound connections to the container (enables listening servers) |
+| `allowLocalNetwork: false` | DROP new inbound connections to the container |
```
MGudgin: docs promise inbound enforcement that doesn't exist.

The contradiction: three docs, three different stories, none matching the code

The reviewed head ships three authoritative descriptions of  allowLocalNetwork  that disagree with each other and with the implementation:

1.  docs/lxc-support/lxc-backend.md:120-121  — an unqualified functional guarantee:

 allowLocalNetwork: true  → "ACCEPT new inbound connections to the container (enables listening servers)"
 allowLocalNetwork: false  → "DROP new inbound connections to the container"

2.  docs/lxc-support/testing-allowlocalnetwork.md:31-43  (added by the same PR) — the direct refutation:

the rule sits in the host  FORWARD  chain hooked  -o <veth> , which a host→container connection does not traverse… it reports reachable for both configs today.

3.  schemas/stable/mxc-config.schema.0.7.0-alpha.json:149  — a third definition:

"the sandboxed process may  bind()  and  listen()  on local IPs and accept incoming connections… Honored by the  seatbelt  backend; other backends ignore this field today."

So within one tree:  lxc-backend.md  says LXC enforces inbound accept/drop; the runbook says LXC's rule is inert; the stable schema says only Seatbelt honors it, others ignore it. All three cannot be true. In reality the code flips the  -i lo  loopback verb (Finding #1) — which is none of these three semantics.

Why this is High, not a nitpick

This is the finding with direct user-harm potential. A user hardening a sandbox reads  lxc-backend.md , sets  allowLocalNetwork: false , and reasonably believes inbound connections to their container are now dropped. They are not — any listener the container starts remains reachable exactly as before. That is a false isolation guarantee on a security-relevant control, documented as authoritative in the backend's own reference page. Documentation that promises a network boundary the code doesn't enforce is worse than no documentation, because it actively discourages the user from adding real protection.

Two semantic drifts hiding inside it

• Definition drift: the schema frames it as a process capability ("may bind/listen on local IPs") while  lxc-backend.md  frames it as a firewall verdict ("ACCEPT/DROP new inbound connections"). These aren't the same feature — one is about whether the process may open a listening socket, the other about whether packets reach it. The PR never reconciles which one  allowLocalNetwork  actually is.
• Backend-scope drift: the stable schema's "other backends ignore this field today" was accurate before this PR. The PR makes LXC emit a rule for it — so the schema sentence is now stale — yet the rule is inert, so "ignore" is still behaviorally true. The stable schema is a released, immutable artifact (per repo convention stable schemas are never edited), so this drift can't even be corrected in place; it can only be fixed forward in a new schema version. That raises the bar for getting the story straight now.

What defaults do agree on (so you don't over-correct)

Worth noting the defaults are consistent and need no change: schema  default: false  (line 148),  policy.md:345  example  "allowLocalNetwork": false , and code (omitted ⇒ false ⇒ DROP-verb) all line up. The problem is exclusively the enforcement claim, not the default.

Concrete fix

Pick one of two honest states and make all three docs say the same thing:

• If the feature stays non-functional (emission-only): change  lxc-backend.md:120-121  to explicitly mark it not yet enforced — e.g. "Emits an  -i lo  rule; inbound enforcement is not yet active (tracked for the  INPUT -path follow-up)." Align the PR title/body away from "will now DROP or ACCEPT local packets." This matches the runbook and the schema's "ignore" wording.
• If the feature is meant to ship working: implement real enforcement per Finding #1, then keep the  lxc-backend.md  wording but back it with the behavioral + unit tests (Findings #3, #4). Also settle the definition drift: decide whether  allowLocalNetwork  is a listen-capability or a packet-filter and make the schema description and  lxc-backend.md  use one consistent phrasing.

Either way, the runbook and  lxc-backend.md  must stop contradicting each other inside the same PR.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -150,9 +150,25 @@ impl NetworkIptablesManager {
         // Create custom chain
         Self::run_iptables(&["-N", &self.chain_name], logger)?;
 
-        // Always allow loopback and established connections
+        let accept_packet_verb = "ACCEPT";
```
MGudgin: The one behavior this PR adds is:  allow_local_network  → loopback verb ( true ⇒ACCEPT,  false ⇒DROP). Nothing that runs in CI exercises it. Three independent confirmations from the tree:

1. The existing unit tests don't touch it.  network_iptables.rs  already has a CI-run  #[cfg(test)] mod tests  (line 328), but all it covers is the pure  resolve_host  helper — IPv4/IPv6 literal filtering (lines 331–383). Zero tests call  apply_firewall_rules  or assert anything about the loopback verb. The new branch is invisible to  cargo test -p lxc_common .

2. The only new test is the manual harness — not wired into the LXC suite.  run_lxc_all_tests.sh  runs six scripts:

run_lxc_basic_test.sh, run_lxc_filesystem_test.sh, run_lxc_object_test.sh,
run_lxc_network_test.sh, run_lxc_timeout_test.sh, run_lxc_env_cwd_test.sh

 run_lxc_local_network_test.sh  is absent. It's referenced only by its own runbook (Maintainability finding). So even the manual signal is opt-in and orphaned.

3. No CI workflow references it. Grepping  .github/  and  tests/scripts/  for the harness name returns nothing — and the harness is root-gated + needs real LXC + WSL2 bridge/NAT, so it cannot run on a standard CI runner even if it were wired up.

Why this is High

Two compounding consequences:

• Silent regression risk. Because no automated test asserts the verb mapping, the entire feature could be reverted — or the verb accidentally inverted ( true ⇒DROP) — and every CI gate stays green. For a security-relevant network control, a change that can be silently undone is a High-severity coverage hole regardless of whether the current implementation works.
• It compounds Finding #1. The feature is already non-functional. With no test pinning even the emitted-rule text, there's nothing to catch the day someone "fixes" the FORWARD/ -i lo  problem and breaks the verb selection in the process. You'd be repairing an untested path with no safety net.

The fix is cheap — the seam is one refactor away (ties to #12)

The reason this isn't already tested is the testability seam problem: verb selection is buried inside  apply_firewall_rules , which immediately shells out to the real  iptables  binary (needs root, needs a host). Extract the pure argv-generation into a testable function and the CI test becomes trivial:

// returns the iptables argument vectors instead of executing them
fn build_firewall_rules(chain: &str, policy: &NetworkPolicyConfig, veth: Option<&str>)
    -> Vec<Vec<String>>

#[test]
fn allow_local_network_true_accepts_loopback() {
    let rules = build_firewall_rules("MXC-t", &policy(true), Some("veth0"));
    assert!(rules.iter().any(|r| r == &["-A","MXC-t","-i","lo","-j","ACCEPT"]));
}
#[test]
fn allow_local_network_false_drops_loopback() {
    let rules = build_firewall_rules("MXC-t", &policy(false), Some("veth0"));
    assert!(rules.iter().any(|r| r == &["-A","MXC-t","-i","lo","-j","DROP"]));
}

This runs in the existing  mod tests , needs no root, no  iptables , no LXC — and it locks the verb mapping against silent revert. (Note: this pins the emitted text, which is necessary but not sufficient — Finding #4 is the complementary point that even the manual behavioral probe can't tell allow from deny, so a real behavioral test on a traversed path is still needed once the enforcement lands.)

Separately,  run_lxc_local_network_test.sh  should be added to  run_lxc_all_tests.sh  (or its runbook clearly mark it as a standalone root-only harness) so it stops being orphaned.

tests/scripts/run_lxc_local_network_test.sh
```diff
@@ -0,0 +1,122 @@
...
+        rule_fail=$((rule_fail + 1))
+    fi
+
+    # [BEHAVIOR] connect from the host to the sandboxed listener.
+    if [ -n "$ip" ]; then
+        if "$HELPER_BIN" connect --host "$ip" --port "$PORT" --timeout 5; then
+            echo "[BEHAVIOR] reachable  (host reached $ip:$PORT)"
+        else
+            echo "[BEHAVIOR] blocked    (host could not reach $ip:$PORT)"
+        fi
```
MGudgin, line 308: the harness's  [BEHAVIOR]  check gives false confidence.

What the probe actually does

The behavioral layer in  run_lxc_local_network_test.sh  (~lines 97–103):

if [ -n "$ip" ]; then
    if "$HELPER_BIN" connect --host "$ip" --port "$PORT" --timeout 5; then
        echo "[BEHAVIOR] reachable  (host reached $ip:$PORT)"
    else
        echo "[BEHAVIOR] blocked    (host could not reach $ip:$PORT)"
    fi
else
    echo "[BEHAVIOR] SKIP  no container IP discovered for $cid"
fi

Two independent defects make this worse than useless as validation:

Defect A — it never asserts anything

The probe prints  reachable  or  blocked  but never compares against an expected outcome. Both branches  echo  and move on; neither increments  rule_fail , and the script's exit code ( [ "$rule_fail" -eq 0 ] ) is driven solely by the  [RULE]  text-grep layer. So the behavioral result cannot fail the run — it's decorative output. The allow config and the deny config could produce identical behavior (they do) and the harness still exits 0.

Defect B — it probes a path the rule doesn't govern (so both configs read "reachable")

 netcheck connect --host <containerIP>  is a host→container connection over the bridge/veth. Per the PR's own runbook, the emitted  -i lo  rule lives in host  FORWARD -o veth  and is never traversed by that connection (Finding #1). So the probe reports  reachable  for both allow and deny — the  blocked  branch is effectively dead code today. The runbook states this outright: "it will report reachable for both configs."

Why "false confidence" is the right frame (and why it's High)

Put A and B together and picture the reviewer experience: a maintainer runs the harness, sees

[RULE] PASS ... ACCEPT (expected ACCEPT)
[BEHAVIOR] reachable
...
[RULE]     pass=2 fail=0

and concludes the feature is validated end-to-end — the PR body literally says "Verified end-to-end … pass=2 fail=0." But nothing about enforcement was verified:

•  [RULE]  only greps that the rule string exists — and that rule is inert (Finding #1).
•  [BEHAVIOR]  shows  reachable  for the deny case too — which, if anyone read it as validation, is evidence the deny path is broken, yet it's labeled "informational" and silently ignored.

So the harness manufactures a green, authoritative-looking signal for a feature that does nothing. That's strictly more dangerous than having no behavioral test, because it converts "untested" into "appears tested and passing." Combined with Finding #3 (no CI coverage) and Finding #2 (docs claim it works), a reader has three mutually-reinforcing false signals that the feature is real.

The subtle silver lining — and why it doesn't rescue the harness

The probe's traffic path is actually the right one for the intended feature: a host→container connect to a container listener is the inbound path  allowLocalNetwork  is supposed to govern. That's what the runbook means by "starts distinguishing allow/deny automatically once the inbound rule lands." So the harness is aimed correctly — but it's validating against an implementation that puts the rule in the wrong chain. The scaffolding survives; the code under it is what's broken.

Fix

Make the behavioral layer a real, asserting test and point it at a path the (corrected) rule traverses:

1. Pass the expected outcome into  run_case  and assert it:
# allow: expect reachable (exit 0); deny: expect blocked (exit non-0)
if "$HELPER_BIN" connect --host "$ip" --port "$PORT" --timeout 5; then reachable=1; else reachable=0; fi
if [ "$reachable" -ne "$expect_reachable" ]; then
    echo "[BEHAVIOR] FAIL expected=$expect_reachable got=$reachable"; behavior_fail=$((behavior_fail+1))
fi
``` and fold `behavior_fail` into the exit code once enforcement is real.
2. Until the enforcement in Finding #1 lands, don't present the behavioral layer as passing — either gate it behind a "not yet enforced" skip that's loud, or hold the harness. And fix  netcheck connect 's partial-read ( read  →  read_exact(&mut buf[..4]) , correctness Low) so the eventual assert isn't flaky.
3. Stop the PR body from calling this "Verified end-to-end" — it verifies rule-text emission only (ties to Finding #2 / intent-coherence).

The honest end state:  [BEHAVIOR]  asserts opposite outcomes for the two configs and contributes to pass/fail — at which point it becomes the real regression test the feature needs, and it'll go red today (correctly) because the rule is misplaced.

File: docs/schema.md [hunk 5]
```diff
@@ -52,9 +52,10 @@ production configs and the dev schema when working on experimental features:
     "network": {
         "defaultPolicy": "block",          // "allow" or "block"
         "enforcementMode": "firewall",     // "capabilities", "firewall", or "both"
-        "proxy": { "localhost": 8080 }     // Loopback proxy port (processcontainer; bubblewrap)
+        "proxy": { "localhost": 8080 },    // Loopback proxy port (processcontainer; bubblewrap)
                                            // (use { "builtinTestServer": true } for the bundled
                                            //  testing-only proxy; requires --allow-testing-features)
+        "allowLocalNetwork": true          // "true" or "false"
```
MGudgin: >allowLocalNetwork [](http://example.com/codeflow?start=9&length=17)

I'm concerned that this is very LXC specific. It doesn't seem like we have a consistent semantic here across the various platforms and backends.

MGudgin (changes requested):
:clock1:

dhoehna:

Re-ran the test suites locally at branch tip `2cc91e7` on a dev workstation. All green (0 failures).

- `cargo test -p lxc_common` → **39 passed, 0 failed** — the LXC ``iptables`` inbound/local-network enforcement ran on Linux (the original Validation ran the WSL E2E harness; here the crate unit tests ran).

_The end-to-end harness (``tests/scripts/run_lxc_local_network_test.sh``) was not re-run here (it needs LXC + root + a bridge); only the ``lxc_common`` unit tests were executed._

src/backends/lxc/common/src/network_iptables.rs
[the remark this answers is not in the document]
dhoehna: Resolved in b5d2ba8. Both were right about different things, and I implemented your **Design #1**:

- **Location** (roadmap N2 and copilot agree): keep the per-container chain hooked at `-I FORWARD -o <veth>`.
- **Rule** (copilot + your adversarial review): stop flipping `-i lo`; gate NEW inbound instead.

The chain is now:
1. `-N MXC-<id>`
2. `-A MXC-<id> -i lo -j ACCEPT` - **restored unconditional** (kills the localhost-break regression you flagged)
3. `-A MXC-<id> -m state --state ESTABLISHED,RELATED -j ACCEPT`
4. `-A MXC-<id> -m state --state NEW -j {ACCEPT if allowLocalNetwork else DROP}` <- the actual enforcement, after ESTABLISHED so container-initiated return traffic survives
5. DNS / allowed / blocked / default ... then `-I FORWARD -o <veth> -j MXC-<id>`

I used `-m state --state NEW` (not `-m conntrack --ctstate NEW`) to match the existing `-m state --state ESTABLISHED,RELATED` line in the same chain; functionally equivalent.

Honest scoping note: because the chain is `FORWARD -o veth`, this governs **forwarded** inbound (peer container / routed) - the roadmap N2 design. A host->container **direct**-IP connection is OUTPUT->INPUT, not FORWARD, so it isn't governed here; that's the GA `ingress.hostLoopback`/INPUT direction (`docs/sandbox-policy/v2/networking.md`), a separate milestone. See my Thread 6 reply for the matching behavioral-test change.

docs/lxc-support/lxc-backend.md
dhoehna: Resolved in b5d2ba8. Took your "ship it working" branch and made the docs agree with the now-functional code:

- Code enforces inbound for real (see the network_iptables thread) via the `--state NEW` verb, backed by 11 CI unit tests + an asserting behavioral harness.
- `lxc-backend.md:120-121` already said "ACCEPT/DROP new inbound connections" - that is now true, so it stays.
- `testing-allowlocalnetwork.md` rewritten: dropped the "inert / reachable for both" caveat, updated the emitted-rule table to the `--state NEW` rule, added a unit-test section.
- `policy.md:219` reframed from the outbound-sounding "connections to local networks" to "binding/listening on local IPs and accepting inbound connections".

Definition drift: settled on the **packet-filter** framing (accept/drop new inbound), since that's what LXC enforces. Backend-scope drift: agreed the stable 0.6.0/0.7.0 schemas' "other backends ignore" is now stale but immutable - fixable only forward (0.8.0-dev already reads correctly). Defaults untouched (all still false => DROP).

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Resolved in b5d2ba8. Did the extraction you sketched:

- Extracted `build_firewall_rules(chain, policy, veth, allowed_ips, blocked_ips) -> Vec<Vec<String>>` - pure, no root, no `iptables`, no DNS (DNS + logging moved to `resolve_hosts_logged`; `apply_firewall_rules` is now resolve -> build -> run).
- 11 new tests in the existing `mod tests` pin the emitted argv and run under `cargo test -p lxc_common` (all green).

One correction to your example: since Design #1 restores `-i lo` to an **unconditional** ACCEPT and moves the verb-flip to the NEW rule, the tests pin `--state NEW -j ACCEPT|DROP` rather than the `-i lo` verb. They also assert loopback-always-ACCEPT, ESTABLISHED-precedes-NEW, NEW-precedes-default, FORWARD-hook present/absent by veth, DNS, and host rules.

Orphan fixed: `run_lxc_local_network_test.sh` is now registered in `run_lxc_all_tests.sh`.

tests/scripts/run_lxc_local_network_test.sh
dhoehna: Resolved in b5d2ba8. All three points:

1. **Asserts now.** `run_case` takes `expect_reachable`; the behavioral layer compares actual vs expected and folds a `behavior_fail` into the script's exit code (with a loud `INCONCLUSIVE` state when the client container can't start, so infra flakes don't masquerade as a pass).
2. **Governed path.** This is the one place I diverged from your "the probe path is already right" note. A host->container **direct** connect is OUTPUT->INPUT and is *not* traversed by a `FORWARD -o veth` rule - so under the roadmap N2 FORWARD design it would read reachable for both. I switched the client to a **peer container**: container->container traffic is forwarded and hits `-o <server-veth>`, so the `--state NEW` verb actually decides it -> reachable for allow, blocked for deny. (host->container-direct is the GA `ingress.hostLoopback`/INPUT direction - separate milestone.)
3. **netcheck** `read` -> `read_exact(&mut buf[..4])` for the 4-byte PONG.

Also updated `[RULE]` to assert the `--state NEW` verb (+ unconditional `-i lo`) instead of the old `-i lo` verb, and removed "Verified end-to-end" from the PR body.

docs/schema.md
dhoehna: Fair, and I don't want to overclaim. This PR is scoped to making the **LXC** backend honor the legacy `allowLocalNetwork` per roadmap N2 (item 14); it does not unify semantics across backends.

The cross-backend unification is the GA `ingress.hostLoopback` design in `docs/sandbox-policy/v2/networking.md:148`, which defines one semantic for all backends ("host may reach sandbox-local listening sockets over loopback only ... Enforced on all backends: Windows via loopback rules, WSLc/LXC/Bubblewrap via iptables INPUT, Seatbelt via profile").

Worth flagging a real discrepancy that surfaced while doing this: that GA doc says LXC enforces via **INPUT** (loopback-only), whereas roadmap N2 says **FORWARD on the veth** - which is what this PR implements. Different chains, different paths (INPUT = host->sandbox-direct; FORWARD = forwarded/peer inbound). Someone owns reconciling N2 with the GA networking spec before `ingress.hostLoopback` lands, so I'm leaving this thread **open** as the tracking point for that cross-backend decision rather than resolving it.

dhoehna:
@SohamDas2021 @MGudgin — flagging a **roadmap vs GA-doc discrepancy** on the LXC inbound (`ingress.hostLoopback` / `allowLocalNetwork`) enforcement chain, now that this PR implements it.

**GA networking spec** (`docs/sandbox-policy/v2/networking.md`) says every Linux backend enforces ingress via the **iptables INPUT chain in the container's network namespace**:
- L148 — "Enforced on all backends: … WSLc/LXC/Bubblewrap via **iptables INPUT**, …"
- L181 — "`ingress.hostLoopback` is enforced on all backends … WSLc/LXC/Bubblewrap via **iptables INPUT**"
- L289 — "Inbound | Enforced via **iptables INPUT chain**. When `hostLoopback: deny` (default), all host/external inbound blocked."
- L295 (LXC) — "Same model and enforcement as WSLc (… `ingress.hostLoopback` via **INPUT**)."

**LXC roadmap** (`docs/linux-wsl-roadmap-june-2026.md` L81, N2) says the opposite: "add iptables **FORWARD** rules on the container veth — DROP new inbound by default; ACCEPT from host loopback when `ingress.hostLoopback: "allow"`."

They disagree on the chain (host-side FORWARD-on-veth vs container-netns INPUT). A host→container-direct packet is host `OUTPUT` → container `INPUT`, and only hits host `FORWARD` if routed — so the container-netns INPUT chain is where the GA doc places it and where host-loopback inbound is actually caught. (The old FORWARD-on-veth `-d <ip>` rules were effectively dead for this case, since the inbound dest is the container's own IP.)

**This PR implements the GA doc:** an ingress-only chain in the container netns (via `nsenter -t <init-pid> -n iptables`) — `-i lo ACCEPT`, `ESTABLISHED,RELATED ACCEPT`, `--state NEW {ACCEPT|DROP}` keyed on `allowLocalNetwork`, terminal DROP, hooked with `-I INPUT`. Validated end-to-end against real unprivileged LXC containers in WSL (allow → reachable, deny → blocked).

Surfacing so we can correct the roadmap. Does container-netns INPUT look right to you both?

dhoehna:
@SohamDas2021 — separate, **roadmap-internal** inconsistency I hit while doing the above (not GA-doc related):

In `docs/linux-wsl-roadmap-june-2026.md`, the two backends that **share the same `NetworkIptablesManager`** are prescribed **different chains** for the same N2 inbound feature:
- L81 (LXC, N2) — "add iptables **FORWARD** rules on the container veth …"
- L164 (Bubblewrap, N2) — "add **host-side iptables INPUT** rules."
- L484 (cross-cutting) — "Shared `NetworkIptablesManager` … serves LXC and Bwrap."

So one shared manager would have to emit FORWARD for LXC but INPUT for Bwrap — and neither line matches the GA doc, which routes both through the **container-netns INPUT** chain. Should **all the Linux backends (WSLc/LXC/Bubblewrap) standardize on the same chain** (container-netns INPUT, per the GA doc) so the shared manager stays single-behavior? Happy to update the roadmap once you confirm the intended direction.

SohamDas2021:
> @SohamDas2021 — separate, **roadmap-internal** inconsistency I hit while doing the above (not GA-doc related):
> 
> In `docs/linux-wsl-roadmap-june-2026.md`, the two backends that **share the same `NetworkIptablesManager`** are prescribed **different chains** for the same N2 inbound feature:
> 
> * L81 (LXC, N2) — "add iptables **FORWARD** rules on the container veth …"
> * L164 (Bubblewrap, N2) — "add **host-side iptables INPUT** rules."
> * L484 (cross-cutting) — "Shared `NetworkIptablesManager` … serves LXC and Bwrap."
> 
> So one shared manager would have to emit FORWARD for LXC but INPUT for Bwrap — and neither line matches the GA doc, which routes both through the **container-netns INPUT** chain. Should **all the Linux backends (WSLc/LXC/Bubblewrap) standardize on the same chain** (container-netns INPUT, per the GA doc) so the shared manager stays single-behavior? Happy to update the roadmap once you confirm the intended direction.

PR out to address this- https://github.com/microsoft/mxc/pull/675

SohamDas2021:
> @SohamDas2021 @MGudgin — flagging a **roadmap vs GA-doc discrepancy** on the LXC inbound (`ingress.hostLoopback` / `allowLocalNetwork`) enforcement chain, now that this PR implements it.
> 
> **GA networking spec** (`docs/sandbox-policy/v2/networking.md`) says every Linux backend enforces ingress via the **iptables INPUT chain in the container's network namespace**:
> 
> * L148 — "Enforced on all backends: … WSLc/LXC/Bubblewrap via **iptables INPUT**, …"
> * L181 — "`ingress.hostLoopback` is enforced on all backends … WSLc/LXC/Bubblewrap via **iptables INPUT**"
> * L289 — "Inbound | Enforced via **iptables INPUT chain**. When `hostLoopback: deny` (default), all host/external inbound blocked."
> * L295 (LXC) — "Same model and enforcement as WSLc (… `ingress.hostLoopback` via **INPUT**)."
> 
> **LXC roadmap** (`docs/linux-wsl-roadmap-june-2026.md` L81, N2) says the opposite: "add iptables **FORWARD** rules on the container veth — DROP new inbound by default; ACCEPT from host loopback when `ingress.hostLoopback: "allow"`."
> 
> They disagree on the chain (host-side FORWARD-on-veth vs container-netns INPUT). A host→container-direct packet is host `OUTPUT` → container `INPUT`, and only hits host `FORWARD` if routed — so the container-netns INPUT chain is where the GA doc places it and where host-loopback inbound is actually caught. (The old FORWARD-on-veth `-d <ip>` rules were effectively dead for this case, since the inbound dest is the container's own IP.)
> 
> **This PR implements the GA doc:** an ingress-only chain in the container netns (via `nsenter -t <init-pid> -n iptables`) — `-i lo ACCEPT`, `ESTABLISHED,RELATED ACCEPT`, `--state NEW {ACCEPT|DROP}` keyed on `allowLocalNetwork`, terminal DROP, hooked with `-I INPUT`. Validated end-to-end against real unprivileged LXC containers in WSL (allow → reachable, deny → blocked).
> 
> Surfacing so we can correct the roadmap. Does container-netns INPUT look right to you both?

PR out to update the roadmap doc- https://github.com/microsoft/mxc/pull/675

dhoehna:
@MGudgin @SohamDas2021 — surfacing a design blocker on the `hostLoopback: allow` path for the **LXC** backend before I build it out.

**The problem:** honoring `ingress.hostLoopback: allow` on LXC requires MXC to stand up and lifecycle-manage **its own host-side TCP forwarder**. We can't avoid it, and we can't borrow one — and the root cause traces back to the ingress schema.

**Why a forwarder is required at all.** On a netns backend the container's `127.0.0.1` and the host's `127.0.0.1` are two separate, *private* loopback stacks with nothing between them. A host process dialing `127.0.0.1:PORT` never reaches a sandbox process bound to the container's loopback. An iptables `INPUT` rule only *scopes* traffic — it cannot re-originate the connection across the veth. So `allow` fundamentally needs a proxy-loopback/forwarder, not just a firewall hole. (The current `--state NEW -j ACCEPT` is both fail-open *and* a no-op for real host-loopback traffic — those packets never even reach the container's INPUT chain.)

**Why we have to roll our own:**
1. **Classic LXC has no built-in forwarder.** MXC drives classic LXC / liblxc (the `lxc-*` tools), not LXD. The convenient `lxc config device add … proxy listen=… connect=…` is an **LXD** feature (LXD daemon + its `lxc` client). `lxc.container.conf(5)` confirms classic LXC exposes only interface / addressing / route keys + lifecycle hooks — no proxy device. (Verified against the LXD proxy-device docs and the LXC man page.)
2. **WSLC's implementation isn't reusable.** WSLC doesn't forward in-process either — it delegates forwarding to the Windows SDK (`WslcSetContainerSettingsPortMappings`). There is no equivalent SDK on Linux/LXC, so none of that code carries over.

**Root cause — ingress has no port in the schema.** A forwarder needs a port to listen on and a port to connect to, but GA `ingress` is portless (`hostLoopback: Allow | Deny`, no port — see networking.md L161, deliberately no CIDR/port for ingress). Even once the forwarder is written, nothing in policy tells it *which* port(s) to map. WSLC hit exactly this wall and worked around it by adding its own explicit `port_mappings` list bolted on *outside* the GA ingress schema — precisely because forwarders need ports.

**Ask:** before I proceed, what's the intended path?
- Add an **LXC-specific port-mapping surface** (mirroring WSLC's `port_mappings`), or
- Have **ingress gain a port in the GA schema** more broadly?

Either way, `hostLoopback: allow` on LXC can't be completed without (a) a port source in config and (b) an MXC-owned forwarder (I'd lean toward an in-process Rust relay that slots into the existing runner lifecycle).

docs/schema.md
MGudgin: Answering your open question — **yes, container-netns INPUT is the right choke point**, and #675 already brought the roadmap in line, so there's nothing left to reconcile on the doc side. The live question is the `hostLoopback: allow` forwarder (separate thread).

**On INPUT vs FORWARD — already settled in the roadmap.** [PR #675](https://github.com/microsoft/mxc/pull/675) (merged 07-24) rewrote roadmap N2 for both LXC and Bubblewrap to say ingress is enforced in the container's **network-namespace INPUT** chain (`nsenter … -I INPUT`), and it explicitly marks the old host-side **FORWARD-on-veth** design as *superseded*. So the roadmap-vs-GA-spec discrepancy you flagged is resolved — the roadmap now matches both the GA spec (`networking.md` L148/L289) and what this PR implements. No further roadmap change needed there; this thread can close once we agree.

Why INPUT is correct (for the record):

- Any packet that terminates at a socket *inside* the container traverses that netns's `PREROUTING → INPUT`, **regardless of how it arrived** (veth/bridge, routed, peer, host-direct). It's the one universal choke point for "inbound to the sandbox."
- `FORWARD -o <veth>` only sees host-*routed* traffic, so it misses host-originated direct connections (`OUTPUT → … → container INPUT` never hits host `FORWARD`) and is `br_netfilter`-sysctl-dependent. INPUT-in-netns has neither problem.
- Keeps the shared `NetworkIptablesManager` single-behavior across LXC + Bubblewrap (resolves the LXC-FORWARD vs Bwrap-INPUT split #675 also cleaned up).

Two caveats so "INPUT is right" isn't overread:

1. This settles the **deny** direction only. INPUT-in-netns correctly drops external/peer inbound, but it does **not** make `hostLoopback: allow` work for real host-`127.0.0.1` traffic — that never reaches the container's separate loopback stack, so it needs the host-loopback DNAT/forward + port surface from your 07-24 comment. #675's N2 text now says the same (*"a bare `NEW -j ACCEPT` is wrong"*). Let's keep the allow-path decision on its own thread — I'll answer the port-mapping-vs-GA-schema question there.
2. Because it's netns-scoped, it depends on a reliable init-PID. Right now if PID discovery returns `None`, a *deny* policy silently builds an inert chain on the **host** and returns `Ok` — a fail-open (`network_iptables.rs:98-124`). Please fail **closed** (return `Err`, let the runner abort/destroy) when Firewall enforcement is requested and no netns PID is available.

Net: INPUT confirmed; roadmap already correct via #675; only open items are the fail-open above and the `hostLoopback: allow` forwarder.

MGudgin:
> @MGudgin @SohamDas2021 — surfacing a design blocker on the `hostLoopback: allow` path for the **LXC** backend before I build it out.
> 
Agree with your analysis: netns loopback is private, an INPUT rule can only *scope* not *re-originate*, so `allow` genuinely needs an MXC-owned host-loopback→container forwarder, and GA `ingress` being portless means there's no port source to drive it.

**Decision: `ingress` gains a port surface in the GA schema — I'll make that change.** I'd rather fix this once at the spec layer than grow a second per-backend bolt-on (we already have WSLC's `port_mappings` outside GA; a differently-shaped LXC one guarantees drift and two forwarders). Every backend that honors `allow` needs the same two facts — listen port + connect port — so it belongs in `ingress`. I'll extend the GA `ingress` entry so `hostLoopback: "allow"` can carry an explicit port map, and converge WSLC onto it:

```jsonc
"ingress": {
  "hostLoopback": "allow",
  "loopbackPorts": [ { "host": 3000, "sandbox": 3000, "protocol": "tcp" } ]
}
```

Host bind pinned to `127.0.0.1`/`::1` only — let's not repeat the WSLC `windows_address: null` → possibly-`0.0.0.0` gap #675 flagged; bind loopback explicitly. I'll get the schema/wire change up so you have a concrete field to target; build LXC's forwarder against `loopbackPorts`, and I'll update WSLC to read the same field rather than its private `port_mappings`.

**But don't let `allow` block this PR.** #627 is scoped to the **deny** direction (roadmap N2 item 14) — the security-relevant default — and that's correctly enforced by the container-netns INPUT chain. Ship it. `allow` (GA port field + forwarder) is a follow-up milestone.

**One interim ask:** until the forwarder + `loopbackPorts` land, `hostLoopback: "allow"` on LXC should return a clear **not-yet-implemented** error, **not** the current `--state NEW -j ACCEPT` — as you noted that's a fail-open *and* a no-op for real host-loopback traffic. Reject loudly rather than silently pretend to honor it.

On the forwarder: agree with your lean — an in-proc Rust relay slotted into the runner lifecycle (spawn on start, tear down on stop) is the right shape.

Net: I'll add `ingress.loopbackPorts` to GA; build LXC's forwarder + WSLC migration against it; make `allow` fail cleanly meanwhile; don't gate the deny-path PR on any of it.

MGudgin (dismissed):

dhoehna:
Holding this one until the network schema change lands in `main`. Once it's in I'll merge, implement the changes on top of it, and address the outstanding review comments here.

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -204,17 +204,21 @@ impl LxcScriptRunner {
...
-        // Try to discover the container's veth interface for scoped rules
-        if let Some(veth) = NetworkIptablesManager::discover_veth_interface(&container_name) {
-            let _ = writeln!(logger, "Discovered veth interface: {}", veth);
-            fw_manager.set_veth_interface(&veth);
+        // Discover the container's init PID so the INPUT rules are applied
+        // inside the container's network namespace (GA enforces LXC ingress
+        // via the container's iptables INPUT chain). Without a PID the chain
+        // is built but left unhooked — never attaching to the host's own
+        // INPUT — which is the correct fail-safe.
+        if let Some(pid) = container.init_pid() {
```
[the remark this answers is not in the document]
dhoehna: Fixed.  The LXC runner now aborts the run when a firewall mode is requested and the container init PID can't be discovered, instead of building an inert chain and returning Ok(true).  See lxc_runner.rs:229 (the `else if use_firewall` branch) -- it destroys the container and returns an error naming the failure.  Kept it LXC-specific: Bubblewrap shares the host netns and reaches apply_firewall_rules with no PID by design through its own runner, so it's untouched.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -1,361 +1,472 @@
...
+        netns_pid: Option<u32>,
+        args: &[&str],
+        logger: &mut Logger,
+    ) -> Result<bool, String> {
+        let mut command = if let Some(pid) = netns_pid {
+            let mut c = Command::new("nsenter");
+            c.arg("-t").arg(pid.to_string()).arg("-n").arg("iptables");
+            c
+        } else {
+            Command::new("iptables")
```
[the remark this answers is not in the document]
dhoehna: Fixed.  Every rule is now installed and torn down through both iptables and ip6tables, so IPv6 inbound can't bypass an IPv4-only deny on a dual-stack container.  See the IPTABLES_BINARIES const at network_iptables.rs:63, the apply loop at :168 and the teardown loop at :273; run_iptables now takes the binary as a param.  An ip6tables failure is treated the same as an iptables failure (fail-closed) -- a default-deny that can't be installed for a family aborts the run rather than leave that family open.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -1,361 +1,472 @@
...
+        rules.push(argv(&[
+            "-A", chain, "-m", "state", "--state", "ESTABLISHED,RELATED", "-j", accept,
+        ]));
+
+        // hostLoopback toggle: accept or drop NEW inbound connections to the
+        // container's listening sockets.
+        let inbound_verb = if policy.allow_local_network { accept } else { drop };
+        rules.push(argv(&[
+            "-A", chain, "-m", "state", "--state", "NEW", "-j", inbound_verb,
+        ]));
```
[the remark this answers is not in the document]
dhoehna: The permissive path no longer installs that source-agnostic accept at all.  allowLocalNetwork:true now returns a not-yet-implemented error (network_iptables.rs:137), exactly because the only rule available is the unscoped --state NEW -j ACCEPT you're describing, which accepts from every interface and source, not just host loopback.  Default-deny (allowLocalNetwork:false) is unchanged and still enforced.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -1,361 +1,500 @@
...
+            accept,
+        ]));
+
+        // hostLoopback toggle: accept or drop NEW inbound connections to the
+        // container's listening sockets.
+        let inbound_verb = if policy.allow_local_network {
+            accept
+        } else {
+            drop
+        };
```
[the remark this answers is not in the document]
dhoehna: Right, there's no way to scope this to host loopback today -- it needs a loopbackPorts schema field and an MXC-owned host-loopback forwarder that don't exist yet.  So rather than install the over-broad accept, the permissive path now returns a clear not-yet-implemented error (network_iptables.rs:137).  That's the interim MGudgin asked for in comment 5120422486.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -1,361 +1,472 @@
...
+        if !use_firewall {
+            logger.log_line("Network enforcement mode does not use firewall, skipping iptables.");
+            return Ok(true);
+        }
+
+        if self.netns_pid.is_none() {
+            // No container netns to target. Hooking the host's INPUT chain
+            // here would filter the *host's* own inbound traffic, not the
+            // container's, so `build_firewall_rules` omits the hook and the
+            // chain is inert. Bubblewrap shared-net mode reaches this path;
```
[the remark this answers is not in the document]
dhoehna: Not fixed here, and I want to be straight about why.  Restoring egress isn't un-dropping a few lines: a5c3d2b replaced the whole host-side veth/FORWARD egress subsystem (veth discovery, DNS host resolution, -d rules, the FORWARD -o veth hook) with the netns/INPUT ingress model this PR is about.  Bringing egress back is a separate mechanism, not in scope for an inbound PR.  And for Bubblewrap specifically it wouldn't even help: Bubblewrap has no netns, and in shared-net mode no scoped veth, so the old egress path was inert for it too -- the fix for Bubblewrap host lists is the proxy layer, not iptables.  Leaving allowedHosts/blockedHosts egress for a follow-up.

docs/schema.md
dhoehna: The GA ingress schema got reverted, so this branch is back on main's allowLocalNetwork.  I've updated the docs (schema.md and lxc-backend.md) to describe that field's real current semantics: an IPv4+IPv6 inbound default-deny in the container INPUT chain, with allowLocalNetwork:true returning a not-yet-implemented error for now.  Not re-opening the chain-choice discussion -- just making the docs match what the code does.

dhoehna:
**Parked: waiting on the GA network schema.**

Converting to draft.  The inbound work splits into two halves, and only one of them can be expressed against `main` today:

- **Deny path — implemented.**  IPv4 and IPv6 inbound default-deny in the container's network-namespace `INPUT` chain, using `main`'s existing `allowLocalNetwork` field.
- **Allow path — blocked.**  `allowLocalNetwork: true` currently returns a not-yet-implemented error.  Making it work needs `ingress.hostLoopback` and `loopbackPorts` from the GA network schema, which is not in `main`.

The GA networking spec (`docs/sandbox-policy/v2/networking.md`) is in the repository, but the schema types are not.  #676 added them and was reverted by #707 on schema-ownership grounds, so the allow path has nothing well-defined to parse.

Leaving the deny half on the branch rather than splitting it out, since the two halves share the same chain construction and test harness.  I will move this out of draft once the schema owners land the ingress shape.

dhoehna:
Closing this draft.  The work is not lost: the branch `user/dahoehna/HonoringInboundRule` stays on the remote and is archived locally at tag `archive/pr627-head` (`2131c6a`, 26 commits, 15 files, +2097/-266).

This branch is being re-cut because it had drifted 20 commits behind `main` with conflicts in `network_iptables.rs` and `signal_cleanup.rs` — the same drift that forced #632 to be abandoned.  It will come back split along the network GA schema boundary: the container-netns default-deny half needs no GA schema and can ship, while the permissive `hostLoopback: allow` half is blocked on schema fields that do not exist yet.

PR #632, opened 2026-07-10
[LXC] Address network policy gaps - model 2 (deny-all-except-proxy)

dhoehna:
Linked work item: AB#62830341 — [LXC] Address network policy gaps - model 2 (deny-all-except-proxy)

## Summary
Implements the deny-all-except-proxy network model for LXC: when a proxy is configured, the proxy `address:port` is the only destination the container can reach and everything else is dropped, in both `iptables` and `ip6tables`.  The posture is fail-closed by design.

- **Proxy-only egress, both address families.**  In proxy mode the egress chain is the proxy ACCEPT rule followed by a default DROP, emitted in `iptables` and `ip6tables` so a dual-stack container cannot slip out over IPv6 while the v4 chain drops everything.
- **`allowedHosts` and `blockedHosts` do not apply in proxy mode.**  When a proxy is configured the chain contains only the proxy ACCEPT and the default DROP, so blocked hosts stay blocked and hosts named in `allowedHosts` become unreachable.  Reaching an `allowedHosts` destination alongside a proxy is not supported under model 2; the proxy is the single permitted egress.  This is the one behavior change a caller could notice as a loss of reachability, so it is called out explicitly rather than left to be discovered.
- **Hooked on `-i <veth>` in FORWARD and INPUT.**  Container-originated packets arrive on the host-side veth, so egress is matched by input interface.  FORWARD alone leaves host-destined packets routed through INPUT reachable, so a matching INPUT hook filters container→host traffic; a host-local proxy is still permitted by the shared chain's own ACCEPT rule.  DHCP (`udp/67`, `udp/547`) is accepted ahead of the jump so lease renewal survives.
- **The veth requirement lives in `lxc_runner`, not the shared manager.**  The iptables manager is shared with Bubblewrap, which runs in the host network namespace and has no veth.  When no veth is set it builds the policy chain and skips the veth-scoped hooks instead of rejecting, so a Bubblewrap firewall request still starts `bwrap`.  LXC's fail-closed invariant — refuse to start when firewall enforcement is needed but the veth cannot be discovered — lives in `lxc_runner` (`reconcile_veth`).  The residual Bubblewrap gap, where a policy chain exists with nothing hooked to it, is tracked in #755.
- **Proxy pinned host-side; DNS closed in proxy mode.**  The proxy hostname is resolved once on the host and the literal IP is injected into `HTTP(S)_PROXY`, so the firewall rule and the container's proxy setting name the same endpoint.  Because the container never needs a resolver, port 53 stays shut in proxy mode and there is no DNS-tunnel exfil path.  An unresolvable proxy host fails setup rather than yielding a chain that drops everything.
- **A TLS-addressed proxy hostname is rejected rather than silently broken.**  Rewriting the host of an `https://` proxy URL to an IP leaves the client validating the proxy's certificate against an IP with no SNI, which fails unless the certificate carries a matching IP SAN.  Setup refuses with an error naming both remedies: address the proxy by IP, or use an `http://` proxy URL, which still reaches `https` destinations through CONNECT.
- **IPv6 fails closed.**  If `ip6tables` is unusable but the host has a live IPv6 stack and the policy's v6 stance is DROP, setup fails instead of silently handing the container unfiltered IPv6.  A host with no IPv6 stack still enforces the IPv4 policy alone.
- **An unresolvable `blockedHosts` entry fails the start.**  It produces no DROP rule, so the host the policy names stays reachable while setup reports success.  This is fatal under both default policies.  Default-allow is the obvious case, where the chain falls through to ACCEPT.  Default-block is not safe either, for a reason that is easy to miss: outside proxy mode the chain carries an *unscoped* port-53 ACCEPT emitted **after** the block rules, so under first-match-wins a blocked host with no DROP of its own is still reachable on port 53 — the closing DROP covers every other port, not that one.  `allowedHosts` stays best-effort in both cases, because an unresolved allow only ever removes reachability.
- **Deny-wins precedence (outside proxy mode).**  Deny rules are emitted before the DNS carve-out and the allow list by a pure ordered builder, so under first-match-wins a destination named in both lists is denied.
- **Proxy env hygiene.**  LXC shares the cooperative-proxy env module with Bubblewrap and WSLc: inherited `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` / `FTP_PROXY` / `NO_PROXY` (plus lowercase, matched case-insensitively) are scrubbed before the configured values are set, and `NO_PROXY` is forced empty so an inherited exemption cannot steer traffic around the proxy.  With the proxy disabled, `apply_proxy_env` reports that the environment must be cleared, so `--clear-env` is emitted even when the request carried no environment at all.
- **Unreachable proxy endpoints are rejected at parse time.**  For LXC, `network.proxy.builtinTestServer`, `network.proxy.localhost`, and loopback URL literals (`127.0.0.0/8`, `::1`, `localhost`) are rejected because they name the container's own loopback, not the host.  IPv6 proxy endpoints are rejected because the proxy rule is emitted with IPv4 `iptables` only.  A routable IPv4 `network.proxy.url` is required.

## Ownership of installed rules

A partly-failed setup or teardown must leave the record of what exists exactly matching what exists, because that record is what the error path, `Drop`, and the signal watchdog all act on.

- Chain and DHCP ownership flags are set only once their command succeeds, and veth-scoped hook deletes are marked up front, so a partial install cannot leak and cannot flush a colliding live container's chain.
- **Teardown removes only what the record names**, so a rollback never deletes a chain or hook this attempt did not create.
- **Every hook delete is attempted, not just the first.**  Short-circuiting on a failed `-D` left the later hooks installed while the failure was reported once; the deletes are veth-scoped, so attempting all of them cannot touch another container's rules.
- **Absence, not the delete's exit code, clears a hook's flag.**  Ownership is published *before* the insert that creates a hook, so the record routinely names a hook whose `-I` then failed — and `-D` for a rule that was never installed fails too.  Clearing only on a successful delete left such a flag set forever: every retry re-ran a delete that could not succeed, the chain stayed owned but unremovable, and `Drop` never converged.  A failed delete is now followed by an `iptables -C` existence check, and a hook that is not there counts as removed.  An unanswerable check reads as "still installed", because being wrong that way costs a retry while the opposite lets a broken or permission-denied iptables report a live hook as gone.
- **A chain is never flushed while a hook still jumps to it.**  `-F` succeeds regardless of who references the chain, and an emptied user chain returns to its caller instead of reaching its own closing DROP, so flushing a still-hooked chain unfilters a container that may still be running.  The `-X` would have failed anyway, since iptables refuses to delete a referenced chain, so the flush bought nothing and cost the container its filtering.  The whole step — flush included — is gated on the hooks being gone, and the chain stays published so a later pass retries.
- Teardown returns the objects whose removal command itself failed and republishes that residual to the signal-cleanup registry.  Previously teardown published nothing, so the registry kept advertising the full pre-teardown set and a later signal would try to remove objects that no longer existed.  Publishing an empty set instead would be the mirror-image bug: a removal that failed would silently lose its owner.
- Ownership is retained exactly when something survived, so the retry paths run again rather than reporting success over a leaked chain.

## Merge-order constraints with #724 and #633

- This branch and #724 both edit `network_iptables.rs` and conflict.  Whoever merges second must resolve keeping **this PR's** deny-before-allow ordering; resolving toward #724 regresses deny-wins precedence.
- All three branches independently fix the same flush-before-hooks-gone fail-open in this file.  #724's `teardown_chain` is byte-identical to this one, so that hunk resolves to a no-op.  #633 reaches the same outcome by re-reading `FORWARD` instead of trusting the delete's exit code.  Either resolution is correct; neither may be resolved by taking `main`'s unconditional flush.

## Known gaps
- A container is reachable between `container.start()` and the firewall hook installation.  Closing it needs a veth-scoped quarantine chain installed before start and swapped atomically for the real policy, which is a new enforcement stage rather than a reordering.  Tracked in #764.  (#633 installs the firewall before start for the state-aware path; this branch's runner path is unchanged.)
- With `defaultPolicy: allow` plus `blockedHosts`, hosts that *do* resolve remain reachable over IPv6, because host rules on this branch resolve to IPv4 addresses only.  #724 adds A/AAAA resolution and buckets each resolved address into the matching `iptables` or `ip6tables` rule set.
- Two containers whose first 20 sanitized name characters match contend for one chain name and the second fails to start; the rollback does not corrupt the incumbent.  #633 replaces the derivation with a hash-widened one that is collision-resistant within the 28-character netfilter bound.

## Validation
Re-verified on this branch (`10f0599`) on the Windows dev box:
- `cargo test -p lxc_common -p wxc_common --lib` — 683 passed (lxc_common 81, wxc_common 602), 0 failed, 0 ignored.  This is a Windows run, so it includes the Windows-gated tests a Linux run excludes.
- `cargo clippy -p lxc_common -p wxc_common --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.
- Every fail-open repair found in review is mutation-tested: reintroducing the defect fails the corresponding test rather than passing quietly.  Each carries a negative control, so an implementation that never flushes, always bails, or retains ownership unconditionally fails too.
- Unit tests cover rule construction and deny-before-allow ordering (`ordered_egress_rules_put_deny_before_allow`), the pure veth reconciliation decision (`reconcile_veth`), attach-argument / `--clear-env` construction, and IPv6 bracketing.  The live INPUT hook, the DHCP carve-out, `ip6tables` enforcement, real iptables enforcement, real signal delivery, and TLS behavior are not exercised by the unit tests.
- The teardown orderings are asserted through pure seams (`remove_every_hook`, `teardown_chain`) rather than through the real path, and that is a necessity rather than a preference: on Windows every firewall command fails at *spawn*, which is before anything is logged, so the logger observes nothing and a teardown's residual always equals what was created.  A test driving the real path could not observe an ordering at all.  What the seams cannot reach is the wiring that feeds them, or a real `-X` succeeding.
- `tests/scripts/run_bwrap_network_firewall_test.sh` and `tests/scripts/run_lxc_network_proxy_test.sh` require Linux and root; they exit 77 (Automake skip) when root, LXC, `bwrap`, `iptables`, or the built binaries are missing, and the aggregate runners count that as a skip, not a pass.  They are not executed on the Windows dev box.


src/core/wxc_common/src/config_parser.rs
```diff
@@ -787,9 +787,16 @@ fn convert_wire_config(
...
+                let msg = "Network proxy is only supported with the 'processcontainer', \
+                           'bubblewrap', or 'lxc' containment backends";
+                logger.log_line(msg);
+                return Err(WxcError::ConfigParse(msg.to_string()));
+            }
+            if containment == ContainmentBackend::Lxc && proxy_config.builtin_test_server {
+                let msg = "LXC: network.proxy.builtinTestServer is not supported; \
+                           use network.proxy.localhost or network.proxy.url";
                 logger.log_line(msg);
                 return Err(WxcError::ConfigParse(msg.to_string()));
```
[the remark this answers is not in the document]
dhoehna: Fixed in 64c072e. LXC now rejects `network.proxy.localhost` at parse time — it maps to `127.0.0.1`, which inside the container network namespace is the container loopback rather than the host, so the injected `HTTP(S)_PROXY` would be unreachable and the iptables proxy-allow rule would never match. Callers must use `network.proxy.url` with a routable host. The `builtinTestServer` LXC error message was updated accordingly and a note was added to `docs/schema.md`.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -2105,13 +2112,17 @@ mod tests {
     }
 
     #[test]
-    fn proxy_rejected_with_non_processcontainer() {
+    fn proxy_accepted_with_lxc() {
         let json = r#"{"process":{"commandLine":"x"},"containment":"lxc","network":{"proxy":{"localhost":8080}}}"#;
         let encoded = base64_encode(json.as_bytes());
         let mut logger = test_logger();
```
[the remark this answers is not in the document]
dhoehna: Fixed in 64c072e. `proxy_accepted_with_lxc` now uses `network.proxy.url` (`http://proxy.example.com:8080`) and asserts host + port, and a new `proxy_localhost_rejected_with_lxc` test proves the localhost shorthand is rejected for LXC.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -197,55 +343,35 @@ impl NetworkIptablesManager {
...
+            )
         } else {
-            // Without a veth interface, we cannot safely scope rules to the container.
-            // Refuse to apply host-wide rules to avoid affecting all host traffic.
             logger.log_line(
-                "Warning: No veth interface set for container. \
-                 Cannot scope iptables rules. Skipping FORWARD hook.",
+                "Network proxy enabled: allowing proxy egress only and dropping all other outbound traffic.",
             );
+            (Vec::new(), Vec::new())
```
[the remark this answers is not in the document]
dhoehna: Fixed in 64c072e. In proxy mode, DNS (udp/tcp dport 53) is now only opened when the proxy is addressed by a hostname (so the container can resolve it). When the proxy endpoint is an IP literal, DNS stays closed so the sandbox is truly deny-all-except-proxy. Added a `host_is_ip_literal` helper with a unit test, and the log line now reflects whether DNS is allowed.

src/backends/lxc/common/src/lxc_bindings.rs
```diff
@@ -78,16 +78,26 @@ pub fn resolve_default_lxcpath() -> String {
 /// Gated to Linux + test builds because `attach_run` is a Windows stub
 /// that never calls this helper, and the workspace clippy lane on
 /// `windows-latest` would otherwise flag it as dead code.
-#[cfg(any(target_os = "linux", test))]
+#[cfg(test)]
 fn build_attach_args(env: &[String], working_directory: &str, command: &str) -> Vec<String> {
```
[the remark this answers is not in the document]
dhoehna: Fixed in 64c072e. Rewrote the doc comments: `build_attach_args` is now documented as a `#[cfg(test)]`-only convenience wrapper that hardcodes `force_clear_env = false`, and the "Linux + test builds" gating note moved onto `build_attach_args_with_env_control`, which is the helper production Linux code actually calls.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -169,83 +326,90 @@ impl NetworkIptablesManager {
...
+            Self::run_iptables_args(&args, logger)?;
         }
 
+        // Hook the chain into FORWARD for the container's traffic
+        Self::run_iptables(
+            &["-I", "FORWARD", "-o", iface, "-j", &self.chain_name],
+            logger,
+        )?;
+
         self.rules_applied = true;
```
SohamDas2021: Same orphan-chain leak flagged in #631 https://github.com/microsoft/mxc/pull/631/changes#r3582849132

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -129,6 +135,150 @@ impl NetworkIptablesManager {
...
+                    "tcp".to_string(),
+                    "-d".to_string(),
+                    endpoint.ip.clone(),
+                    "--dport".to_string(),
+                    endpoint.port.to_string(),
+                    "-j".to_string(),
+                    "ACCEPT".to_string(),
+                ]);
+            }
+            rules.push(vec![
```
SohamDas2021, line 450: deny-all-except-proxy is IPv4-only; IPv6 egress is unfiltered?

This DROP- the "deny all other outbound" that the whole model-2 posture rests on, is only ever applied to the v4 table: every rule here is run through the `iptables` binary (run_iptables / run_iptables_args), and there's no ip6tables path anywhere in this file. So, on a container with IPv6 connectivity, all IPv6 egress bypasses this chain entirely and the "except-proxy" guarantee doesn't hold.

To be clear this is pre-existing v4-only behavior (resolve_host at :100-119 already drops v6, and run_iptables predates this PR), but this PR is what makes it load-bearing: before, v4-only filtering was a best-effort allow/block list; now it's the enforcement boundary of a deny-all-except-proxy security model, so the IPv6 gap becomes a real bypass rather than a minor omission.

Suggested fix: mirror the chain to ip6tables with a v6 DROP-all?

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -169,83 +326,90 @@ impl NetworkIptablesManager {
...
+                .is_some_and(|addr| !Self::host_is_ip_literal(addr.host()))
+        } else {
+            true
         };
-        logger.log_line(&format!("Default network policy: {}", default_action));
-        Self::run_iptables(&["-A", &self.chain_name, "-j", default_action], logger)?;
 
-        // Hook the chain into FORWARD for the container's traffic
-        if let Some(ref iface) = self.veth_interface {
+        if allow_dns {
```
SohamDas2021: DNS opened to the whole internet in proxy+hostname mode?
                                                                                         
ACCEPTs udp/tcp 53 with no -d, so under "deny-all-except-proxy" the container still has an unrestricted DNS channel to any resolver- a standing DNS-tunnel exfil path. Since the proxy is already resolved host-side, consider injecting the resolved IP into HTTP_PROXY and keeping DNS closed, or scope the DNS ACCEPT to the configured resolver(s).

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -129,6 +135,150 @@ impl NetworkIptablesManager {
...
+
+        let address = policy.network_proxy.address.as_ref().ok_or_else(|| {
+            "Network proxy is enabled but no proxy address is configured".to_string()
+        })?;
+
+        if address.port() == 0 {
+            return Err("Network proxy port must be between 1 and 65535".to_string());
+        }
+
+        let ips = Self::resolve_host(address.host());
```
SohamDas2021, line 730: The ACCEPT rule targets the host-side resolution of the proxy hostname, but the container is handed HTTP_PROXY=http://proxy.example.com:8080 (hostname, via to_url()) and resolves it itself. Under round-robin/split-horizon DNS the container may pick a different IP than the one in the ACCEPT rule → proxy unreachable while everything else is dropped. Can we pin the injected proxy to the resolved IP fixes both this and the DNS-hole above?

SohamDas2021 (changes requested):

SohamDas2021:
Shall we add LXC proxy.url test config json?

dhoehna:

Re-ran the test suites locally at branch tip `775dc87` on a dev workstation. All green (0 failures).

- `cargo test -p wxc_common` → **399 passed, 0 failed**

- `cargo test -p lxc_common` → **44 passed, 0 failed** — the deny-all-except-proxy iptables enforcement ran on Linux (the original Validation only ``cargo check``-compiled it for the Linux target).

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed in 3fd32ed.

`apply_firewall_rules` is now a thin wrapper around a fallible `apply_firewall_rules_inner`; on `Err` it logs, calls a new shared `teardown_chains`, and re-propagates. `teardown_chains` deletes the FORWARD hooks and flushes/deletes the chains in both tables, and every step is best-effort, so it is safe to call when only part of the state was created.

That was the real failure mode: `remove_firewall_rules` short-circuits on `rules_applied == false`, and `rules_applied` was only set after the *last* rule succeeded. So a mid-setup failure left the chain behind and every later attempt failed permanently on `-N` ("chain already exists") until someone cleaned up by hand.

`remove_firewall_rules` now delegates to the same `teardown_chains`, so the rollback path and the normal cleanup path can't drift apart.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed in 3fd32ed. Agreed this was a real bypass rather than a theoretical one: with no v6 chain, a dual-stack container reached the internet over IPv6 while the v4 chain dropped everything.

The chain is now created and hooked in `ip6tables` too. Every per-destination rule stays IPv4 (`resolve_host` keeps only A records, and the proxy endpoint is a v4 literal), so the v6 chain carries the base rules — loopback and `ESTABLISHED,RELATED`, both family-agnostic — plus the same closing stance. In proxy mode that stance is DROP, which is what actually closes the hole.

Two things worth flagging:

- It sits behind an `ip6tables -S` probe, which fails both when the binary is missing and when the kernel has IPv6 disabled. In either case we warn and still enforce the v4 policy rather than failing a policy that worked before dual-stack support existed — and such a host has no v6 egress to leak. In proxy mode the warning states explicitly that the rule set is IPv4-only.
- Residual gap I deliberately did **not** close here: with `defaultPolicy: allow` plus `blockedHosts`, those hosts stay reachable over IPv6. Blocking them needs AAAA resolution, which is #631's territory, so I mirrored the v4 default instead of silently changing `allow` semantics. Happy to file that as a follow-up work item if you'd rather track it.

Covered by `base_chain_rules_are_family_agnostic` (the shared argv must not name an address family) and `proxy_mode_default_action_is_drop_regardless_of_default_policy`.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed in 3fd32ed. DNS is now closed unconditionally in proxy mode — `allow_dns = !proxy_enabled`.

The old condition opened port 53 to the entire internet whenever the proxy was addressed by hostname. In a posture whose whole point is that the proxy is the only reachable destination, that left a standing DNS-tunnel exfil path.

What makes closing it possible is your suggestion on the other thread: the proxy hostname is resolved once on the host and the resulting literal is injected into `HTTP(S)_PROXY`, so the container never needs a resolver at all. Details in the reply there.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed in 3fd32ed, taking your suggestion.

`NetworkIptablesManager::pin_proxy_to_resolved_ip` resolves the proxy host once, host-side, and returns a `ProxyConfig` whose address is that resolved literal. `lxc_runner` calls it before anything consumes the policy and hands the same pinned policy to both `apply_firewall_rules` and `apply_proxy_env`, so the ACCEPT rule and the container's `HTTP(S)_PROXY` can no longer disagree. That was the bug you spotted: the container re-resolved the name itself and, under round-robin or split-horizon DNS, could pick an address the firewall never allowed.

Details:

- `ProxyAddress::pinned_to_ip` swaps only the host, preserving scheme, credentials, port and path, and doesn't add a trailing slash the caller didn't configure. It has to set `original_url`, since `to_url()` otherwise falls back to a hardcoded `127.0.0.1`.
- No-op for a disabled proxy, the built-in test server, or an address that is already an IP literal.
- Multi-A-record hosts collapse to the first address and log that they did. That is the intent — both sides then agree on one IP instead of racing DNS.
- An unresolvable proxy host now **fails setup** rather than silently producing a chain that drops everything.
- This is also what lets DNS stay closed, per the other thread.

Covered by `pin_proxy_rewrites_hostname_to_resolved_ip`, `pin_proxy_leaves_ip_literals_untouched`, `pin_proxy_is_a_noop_when_disabled`, `pin_proxy_errors_when_hostname_does_not_resolve`, plus four `pinned_to_ip` URL-shape tests in `wxc_common`.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -114,21 +120,314 @@ impl NetworkIptablesManager {
...
+            vec![
+                "-A".to_string(),
+                chain_name.to_string(),
+                "-m".to_string(),
+                "state".to_string(),
+                "--state".to_string(),
+                "ESTABLISHED,RELATED".to_string(),
+                "-j".to_string(),
+                "ACCEPT".to_string(),
+            ],
```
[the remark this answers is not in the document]
dhoehna: Fixed in 4c71e2e — you're right, and the reasoning about direction is the key part.

The chain is only ever reached from a hook scoped to `-i <veth>`, so reply packets (which arrive `-o <veth>`) never traverse it. The `ESTABLISHED,RELATED` accept therefore did nothing for return traffic; its only effect was to let flows the container opened before the chain was installed keep running straight through the deny-all policy.

Removing it doesn't break permitted traffic, because every remaining rule matches on **destination** (`-d <proxy_ip> --dport <port>`, `-d <allowed_ip>`), which holds for every packet of a flow rather than only the SYN — no conntrack exemption is needed.

I removed the `-i lo` accept in the same change: the input interface is `<veth>` by construction, so that rule could never match either.

New test `egress_chain_has_no_conntrack_or_loopback_accept` asserts no rule in the chain carries `--state` / `--ctstate` / `ESTABLISHED,RELATED` / `lo`.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,136 +438,215 @@ impl NetworkIptablesManager {
...
-        } else {
-            // Without a veth interface, we cannot safely scope rules to the container.
-            // Refuse to apply host-wide rules to avoid affecting all host traffic.
+        } else if proxy_enabled {
             logger.log_line(
-                "Warning: No veth interface set for container. \
-                 Cannot scope iptables rules. Skipping FORWARD hook.",
+                "Warning: ip6tables unavailable, so the deny-all-except-proxy rule set is \
+                 IPv4-only; IPv6 egress is unfiltered if the host has IPv6 connectivity.",
             );
```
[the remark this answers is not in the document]
dhoehna: Fixed in 4c71e2e. You're right that the probe conflated the two cases — that's my bug: I added the blanket skip to resolve a review comment about ip6tables being a hard dependency on IPv4-only hosts, and in doing so made it fail-open on hosts that do have IPv6.

The two are now separated. `ip6tables_available()` still answers "can I run it", and a new `host_has_ipv6()` answers "is there IPv6 to filter" by reading `/proc/net/if_inet6` — that file exists only when the kernel's IPv6 support is present, and is empty when IPv6 is administratively disabled (`net.ipv6.conf.all.disable_ipv6=1`). Unreadable is treated as "IPv6 is live" so the ambiguous case fails closed.

Behavior now:
- ip6tables usable → unchanged, v6 chain installed.
- ip6tables unusable, no live IPv6 stack → skip v6, enforce IPv4, log it. This is the pure-IPv4 host the earlier comment was about, and it has no v6 egress to leak.
- ip6tables unusable, IPv6 live, and the policy's v6 stance is DROP (proxy mode, or `defaultPolicy: block`) → **setup fails** with an actionable message rather than handing the container unfiltered IPv6.
- ip6tables unusable, IPv6 live, v6 stance is ACCEPT → log that IPv4-only enforcement is in effect; nothing is being silently dropped, since the v6 chain would have accepted anyway.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,136 +438,215 @@ impl NetworkIptablesManager {
...
-            return Ok(());
+        // Hook the chains into FORWARD for the container's egress traffic.
+        // Packets originating in the container arrive at the host on the
+        // host-side veth, so they match FORWARD by input interface (`-i`);
+        // `-o` would instead match traffic flowing toward the container and
+        // leave container egress — the thing this policy exists to restrict —
+        // entirely unfiltered.
+        Self::run_iptables(
+            &["-I", "FORWARD", "-i", iface, "-j", &self.chain_name],
+            logger,
```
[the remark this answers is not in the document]
dhoehna: Fixed in 4c71e2e. Confirmed — netfilter delivers locally-destined packets via `PREROUTING → INPUT` and never through `FORWARD`, so a FORWARD-only hook left the bridge gateway and every service on the host reachable from inside the container. That defeats the model-2 guarantee regardless of how complete the v4/v6 chains are.

The same chain is now hooked into **INPUT** as well, scoped identically to `-i <veth>`, in both address families. Reusing the chain gives the right semantics for free: a host-local proxy is still permitted by its own ACCEPT rule, while everything else on the host falls through to the chain's DROP.

One carve-out was needed. `lxc-net` runs dnsmasq on the bridge, so a DHCPREQUEST at lease-renewal time is addressed to the host and would have hit the new DROP, costing the container its IP mid-run. `udp/67` (and `udp/547` for DHCPv6) is inserted ahead of the chain jump — `-I` twice, jump first — so renewal survives. It's a link-local exchange with the bridge, not an egress path, so it doesn't weaken the posture.

Worth flagging on validation: this is covered by unit tests at the rule-construction level, but I don't have an LXC host to exercise the live INPUT path against, so the DHCP carve-out in particular hasn't been observed working end-to-end.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,136 +438,215 @@ impl NetworkIptablesManager {
...
+                // Roll back whatever was created before the failure. Without
+                // this, `remove_firewall_rules` short-circuits on
+                // `rules_applied == false` and the orphan chain(s) survive, so
+                // the next attempt fails permanently on `-N` ("chain already
+                // exists") until someone cleans up by hand.
+                logger.log_line(&format!(
+                    "Firewall setup failed: {}. Cleaning up partial iptables state.",
+                    e
+                ));
+                self.teardown_chains(logger);
```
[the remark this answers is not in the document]
dhoehna: Fixed in 4c71e2e — and this one was mine: I added the unconditional rollback to resolve an earlier comment about orphan chains, without accounting for the truncated-name collision you've identified.

Apply now records what it actually created in a `CreatedFirewallState`, and the rollback removes only those objects. The ordering rules matter and are commented in the code:

- **Chain flags are set only after `-N` succeeds.** `-F`/`-X` are name-scoped and would hit a colliding container, so a failed `-N` now leaves `v4_chain`/`v6_chain` false and the rollback is a no-op — the pre-existing chain is untouched, and its FORWARD/INPUT hooks keep pointing at a populated chain.
- **Hook and DHCP flags are set before the insert.** Every hook delete is scoped to *this* container's veth (`-D FORWARD -i <our-veth> -j <chain>`), so a `-D` for a rule never inserted is a harmless no-op that cannot affect another container — whereas marking afterwards would leak the FORWARD hook if the INPUT insert failed.

`teardown_chains` (post-success cleanup and `force_cleanup`) is now `teardown_created(CreatedFirewallState::all())`, so the two paths can't drift. Covered by `rollback_state_starts_empty_so_a_failed_chain_create_touches_nothing` and `full_teardown_state_covers_every_created_object`.

Note the collision itself still exists — two containers whose first 20 sanitized name characters match will contend for one chain name. This change stops the rollback from corrupting the incumbent, but the second container still fails to start. Making the chain name collision-free is a separate change; happy to file it as a work item.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed.  Rollback is now scoped to what this specific apply created.  `CreatedFirewallState` (network_iptables.rs:30-52) records each object as it is made, and the error path in `apply_firewall_rules` calls `teardown_created(created, ...)` (network_iptables.rs:588, def at :803), so it removes only this call's own chain, hooks, and DHCP rules.  The chain flags (`v4_chain`/`v6_chain`) are set only after `-N` succeeds (:651, :654), so a failed `-N` on a truncated name that collides with a live container's chain never triggers a flush or delete of that incumbent.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed.  The chain is mirrored to ip6tables now: the v6 `-N` is at network_iptables.rs:653, the closing DROP stance is mirrored in at :680-695, and FORWARD/INPUT are hooked over ip6tables at :762-771.  On top of that, if ip6tables is unusable but the host has a live IPv6 stack and the v6 stance is DROP, setup fails closed instead of handing the container unfiltered v6 (:618-651, `host_has_ipv6` at :204).  So a dual-stack container can no longer get out over IPv6 while the v4 chain drops everything.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed.  `build_chain_rules` returns early in proxy mode (network_iptables.rs:314) before the port-53 ACCEPT is ever emitted, so under deny-all-except-proxy there is no DNS channel at all.  The DNS carve-out at :328-347 only runs outside proxy mode, where the allow/block lists actually need the container to resolve names.  The proxy itself is resolved host-side and pinned (see the proxy-pin thread), so the container never needs a resolver.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed.  The proxy hostname is resolved once on the host and pinned to that literal before anything consumes it.  `pin_proxy_to_resolved_ip` (network_iptables.rs:440-490) rewrites the address to the resolved IP, and lxc_runner calls it at lxc_runner.rs:219 before the firewall ACCEPT and the injected HTTP(S)_PROXY are built (:301).  Both sides name the same IP now, so round-robin or split-horizon DNS cannot hand the container a different address than the ACCEPT rule.  Because the container no longer resolves anything, DNS stays closed, which is what closes the DNS hole in the sibling thread too.

tests/configs/lxc_network_proxy.json
```diff
@@ -0,0 +1,20 @@
+{
+  "version": "0.6.0-alpha",
```
SohamDas2021: 0.8?

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -207,18 +257,69 @@ impl LxcScriptRunner {
...
+        // DNS stays closed under deny-all-except-proxy.
+        let mut effective_policy = request.policy.clone();
+        match NetworkIptablesManager::pin_proxy_to_resolved_ip(
+            &effective_policy.network_proxy,
+            logger,
+        ) {
+            Ok(pinned) => effective_policy.network_proxy = pinned,
+            Err(e) => {
+                if self.destroy_on_exit || container_created {
+                    let _ = container.destroy();
```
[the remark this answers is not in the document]
SohamDas2021: Agree with bot. Shall we fix this?

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -192,13 +192,14 @@ impl LxcScriptRunner {
...
         let needs_network = matches!(
             request.policy.network_enforcement_mode,
             NetworkEnforcementMode::Firewall | NetworkEnforcementMode::Both
         ) || !request.policy.allowed_hosts.is_empty()
-            || !request.policy.blocked_hosts.is_empty();
+            || !request.policy.blocked_hosts.is_empty()
+            || request.policy.network_proxy.is_enabled();
 
         if needs_network {
             Self::wait_for_network(&container_name, Duration::from_secs(10), logger);
```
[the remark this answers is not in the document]
SohamDas2021: Shall we address this?

src/core/wxc_common/src/models.rs
```diff
@@ -420,6 +420,68 @@ impl ProxyAddress {
...
+            _ => std::borrow::Cow::Borrowed(ip),
+        }
+    }
+
+    /// Replace the host of `raw` with `ip`, preserving scheme, credentials,
+    /// port and path. Returns `None` when `raw` is not a parseable URL or the
+    /// host cannot be replaced, letting the caller fall back.
+    fn rewrite_url_host(raw: &str, ip: &str) -> Option<String> {
+        let ip_host = Self::bracket_if_ipv6(ip);
+        let mut parsed = url::Url::parse(raw).ok()?;
```
SohamDas2021, line 466: Rewriting the proxy URL host to the resolved IP means an https:// -scheme proxy is contacted at an IP literal, so the client's SNI/cert validation fails unless the proxy cert has an IP SAN. Narrow (http CONNECT tunneling and the destination server's SNI are unaffected) and documented as a known gap. Consider an /etc/hosts pin mapping the hostname→IP instead of mutating the URL, so TLS identity is preserved

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -114,22 +154,403 @@ impl NetworkIptablesManager {
...
+
+        rules
+    }
+
+    fn resolve_policy_hosts(hosts: &[String], action: &str, logger: &mut Logger) -> Vec<String> {
+        let mut resolved = Vec::new();
+
+        for host in hosts {
+            let ips = Self::resolve_host(host);
+            if ips.is_empty() {
```
SohamDas2021, line 505: On a resolution failure this logs a warning and continues, so under defaultPolicy:allow a transient DNS failure silently omits the DROP for a blockedHosts entry → that host becomes reachable (fail-open). The IPv6-address-dropped half is a documented gap deferred to #724, but the resolution-failure fail-open is undocumented. Consider failing closed (error out) when a  blockedHosts entry can't be resolved.

SohamDas2021 (changes requested):

docs/lxc-support/lxc-backend.md
```diff
@@ -109,18 +109,53 @@ Filesystem policies are enforced via bind mounts in the container configuration:
 
 ## Network Policy
 
-Network policies are enforced via iptables/nftables rules applied to the container's virtual ethernet (veth) interface:
+Network policies are enforced via iptables/ip6tables rules in a per-container
```
bbonaby, line 112: question: Does setting the iptables require root access? I'm thinking about the GitHub copilot usecase. I don't think they'd want to be root when trying this but I could be wrong. @Caarlos0, in GitHub copilot, for linux do you know if the cli will need root access to perform these iptable actions? Not sure what it's doing on Linux today when these are passed in. Figured I would circle back to this now that self hosting is underway.

SohamDas2021: lxc needs root access by default, to create containers for example. 

That is the reason ghcp switched to bubblewrap. Bubblewrap creates the container unprivilege, root is only required for the optional iptables enforcement path, not by default.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed.  `pin_proxy_to_resolved_ip` now resolves the hostname once on the host, rewrites the effective proxy address to that IP, and `lxc_runner.rs` does this before `container.start()`.  The same `effective_policy` is then used for both the firewall and `HTTP(S)_PROXY`, so the container cannot resolve a different endpoint.  This has not been exercised against live LXC or iptables.

src/core/wxc_common/src/proxy_env.rs
```diff
@@ -124,9 +127,36 @@ pub fn apply_cooperative_proxy_env(caller_env: &[String], proxy_url: &str) -> Ve
...
+    if let Some(address) = &proxy.address {
+        *env = apply_cooperative_proxy_env(env, &address.to_url());
+        return true;
+    }
+
+    // Proxy disabled: still strip inherited proxy vars so a caller cannot point
+    // the sandbox at an egress path the policy never authorized.
+    let original_len = env.len();
+    env.retain(|entry| !is_managed_proxy_key(env_key(entry)));
+    env.len() != original_len
```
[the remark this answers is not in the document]
dhoehna: Fixed.  `apply_proxy_env` now returns `true` even when the proxy is disabled and the request environment is empty, and `lxc_runner` passes that value to `attach_run`.  `build_attach_args_with_env_control` therefore emits `--clear-env`, so inherited host proxy variables are not retained.  This is covered by argument-construction tests, not a live LXC run.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,136 +560,308 @@ impl NetworkIptablesManager {
...
+        // The hooks are the point where LXC's and Bubblewrap's scoping
+        // contracts diverge. LXC always has a veth and its enforcement is
+        // veth-scoped, so the hooks below match `-i <veth>`. Bubblewrap runs in
+        // the host network namespace and has no veth: there is no per-container
+        // interface to scope to, and hooking an unscoped rule into FORWARD or
+        // INPUT would filter the *host's* own traffic. So the no-veth path
+        // builds the policy chain (above) and installs no hooks, which is the
+        // behavior Bubblewrap had before veth-scoping was introduced. An LXC
+        // container that reaches here without a veth is a bug caught up-stream
+        // in `lxc_runner`, which fails closed before calling this.
```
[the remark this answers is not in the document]
dhoehna: The claim is correct.  Bubblewrap leaves the host network namespace shared when host rules are present, and `NetworkIptablesManager` returns success after building an unhooked chain because Bubblewrap has no veth.  This is the pre-existing Bubblewrap enforcement gap tracked by #755, so I am keeping the implementation out of this LXC PR; that follow-up should reject firewall host-rule requests until Bubblewrap has an enforceable per-sandbox scope.  This has not been exercised against live iptables.

src/core/wxc_common/src/models_proxy_url_spec_tests.rs
```diff
@@ -0,0 +1,350 @@
...
+    assert!(
+        !out.contains("proxy.example.com"),
+        "input={input:?} — old host must not appear in output {out:?}"
+    );
+    assert!(
+        out.contains(":3128"),
+        "input={input:?} — port must survive; output was {out:?}"
+    );
+    // Strongest assertion: the output must be a parseable URL with [::1] as host.
+    let parsed = url::Url::parse(&out).expect(&format!("output {out:?} must be parseable"));
```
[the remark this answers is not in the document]
dhoehna: Fixed.  All three `expect(&format!(...))` calls in `models_proxy_url_spec_tests.rs` now use `unwrap_or_else(|_| panic!(...))`, so they no longer trigger `clippy::expect_fun_call`.  I verified the source at lines 251, 278, and 307; I did not run the Linux Clippy lane on this Windows host.

src/backends/lxc/common/src/lxc_runner.rs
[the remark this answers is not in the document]
dhoehna: Fixed.  `SetupUndo` now distinguishes destroy, stop, and leave, and `setup_undo_action` returns `Stop` when this invocation started a pre-existing container that must be preserved.  `container_started` is now the result of the start `if` expression, and all six setup failure exits go through `undo_container_setup`.  This is source and unit-test coverage only; it has not been exercised against live LXC.

tests/configs/lxc_network_proxy.json
dhoehna: Fixed.  `tests/configs/lxc_network_proxy.json` now references the existing `0.8.0-alpha` schema version.  This changes only the test configuration reference; it does not change the schema.

src/core/wxc_common/src/models.rs
dhoehna: Fixed by failing closed rather than rewriting a TLS proxy hostname.  `pin_proxy_to_resolved_ip` leaves an IPv4 literal unchanged, but rejects a hostname-backed `https://` proxy before any rewrite or container start because replacing the host would break SNI and certificate validation.  Plain `http://` proxy URLs can still be pinned for CONNECT tunneling.  This is source and unit-test coverage only; I have not exercised it against a live TLS proxy or LXC host.

docs/lxc-support/lxc-backend.md
dhoehna: Root specifically is not the exact requirement.  The LXC manager invokes host `iptables` and `ip6tables` directly, without `sudo` or an elevation helper, so the MXC process needs `CAP_NET_ADMIN` in that network namespace; running as root is one way to provide it.  LXC creation and start are also invoked directly, and this repo documents root or a configured unprivileged LXC setup.  Bubblewrap's configured cooperative-proxy path skips iptables and is intended to remain unprivileged, but this repo does not establish which path GitHub Copilot CLI deploys.

src/backends/lxc/common/src/lxc_runner.rs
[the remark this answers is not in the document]
dhoehna: Confirmed, and it is not fixed in this PR.  `container.start()` is at `lxc_runner.rs:330`, the optional address wait at `:358`, veth discovery at `:361`, and `apply_firewall_rules` only at `:408`, so anything already inside the container has unfiltered egress for that whole span.  Moving proxy pinning ahead of start removed host DNS work from the window but did not close it.

Closing it properly needs a veth-scoped quarantine chain installed the moment the interface exists and then atomically swapped for the real policy, which is a new enforcement stage with its own rollback and teardown semantics rather than a reordering.  I filed #764 for it rather than growing this PR further.  The window is established by reading the call ordering, not by observing a packet escape.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,136 +560,308 @@ impl NetworkIptablesManager {
...
-        Ok(true)
+        Ok(())
     }
 
-    /// Remove all iptables rules created by this manager.
+    /// Best-effort removal of every hook and chain this manager could have
+    /// created, in both tables. Used by the post-apply teardown paths, where
+    /// the manager owns all of it and iptables is the source of truth.
+    fn teardown_chains(&self, logger: &mut Logger) {
+        self.teardown_created(CreatedFirewallState::all(), logger);
```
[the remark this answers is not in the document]
dhoehna: Fixed in c3af5f3.  You're right that the signal path was ownership-blind: the watchdog called `force_cleanup`, which reached `teardown_chains` with `CreatedFirewallState::all()`, so a signal arriving before or during apply tore down whatever answered to the chain name.  Since chain names truncate to 20 characters and can collide, that teardown could belong to a different live container and would fail it open.

`CreatedFirewallState` is now published incrementally -- the moment each resource starts existing, not after apply succeeds -- and `force_cleanup` returns early on an empty record, so a signal arriving before anything was created runs no iptables command at all.  Mutation-tested: deleting the `is_empty` guard makes the test observe a foreign container's teardown.

This has not been exercised against live iptables or real signals; the oracle is the logger, not rule state.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed in c3af5f3, though scoped more narrowly than "always fail".  An unresolved `blockedHosts` entry was logged and skipped, so under `defaultPolicy: allow` the chain fell through to ACCEPT and a host the policy explicitly names as blocked stayed reachable while start reported success.

Resolution failure is now fatal under default-allow only.  Under default-block the closing DROP already covers the host, so failing there would be gratuitously strict and would break a config naming a temporarily-down host without closing any hole; those stay logged skips.  Allowed hosts remain best-effort, since an unresolved allow only ever removes reachability, never adds it.  Mutation-tested: disabling the `Err` return makes an unresolvable blocked host return `Ok([])`.

Not exercised against live iptables or real DNS failure.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,136 +602,308 @@ impl NetworkIptablesManager {
...
+        // Create custom chains. A failure here means the name is already taken
+        // (chain names are truncated, so a collision with another live
+        // container is possible), which is why `created` is only marked after
+        // the command succeeds — the rollback must not touch a chain we did
+        // not make.
+        Self::run_iptables(&["-N", &self.chain_name], logger)?;
+        created.v4_chain = true;
+        if ipv6_enabled {
+            Self::run_ip6tables(&["-N", &self.chain_name], logger)?;
+            created.v6_chain = true;
```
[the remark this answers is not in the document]
dhoehna: Confirmed and fixed in 6756a3b.  `teardown_created` removed objects but published nothing, so the registry kept advertising the pre-teardown set exactly as you describe.

I did not publish `CreatedFirewallState::default()`, though, because that is the mirror-image bug: a removal command that failed would silently lose its owner, and nothing afterward would know the object was ours.  Instead `teardown_created` now tracks which removals actually succeeded, publishes the residual, and returns it; both callers use it to keep `rules_applied` open exactly when something survived, so `Drop` still retries a teardown whose commands failed.

Honest limit on the test: on this Windows host every removal command fails, so the residual always equals the input.  The new test proves republication happens -- mutation-tested, removing the publish call fails it -- but it cannot distinguish a correctly computed residual from the pre-teardown record.  That distinction needs a Linux host with real iptables.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Correcting this again, in the other direction: I was wrong to scope the fix to `defaultPolicy: allow`, and the reasoning I gave for default-block does not survive contact with the chain the builder actually emits.

`build_chain_rules` emits an unscoped port-53 ACCEPT outside proxy mode — `-p udp --dport 53 -j ACCEPT` and the tcp equivalent, with no `-d` at all — and it emits it **after** the blocked-host DROPs.  iptables is first-match-wins, so a blocked host that got no DROP rule of its own is still reachable on port 53.  The closing DROP covers every other port, not that one.  That is a DNS-tunnel path to a host the operator explicitly named as blocked, which is the exfil route the deny posture exists to close.

So your original comment was right and my narrowing was wrong.  An unresolvable `blockedHosts` entry is now fatal under **both** default policies, and the parameter that selected between them is gone (`6c52d93`).  `allowedHosts` stays best-effort in both cases, since an unresolved allow only ever removes reachability.

I also pinned the reasoning with a test instead of a comment.  `the_port_53_accept_follows_the_block_rules_and_names_no_destination` asserts both facts the argument rests on — that the DNS accept follows the block rules, and that it names no destination — so if either changes the test fails rather than the comment quietly going stale.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,149 +667,418 @@ impl NetworkIptablesManager {
...
         }
 
-        // Flush and delete the chain
-        let _ = Self::run_iptables(&["-F", &self.chain_name], logger);
-        let _ = Self::run_iptables(&["-X", &self.chain_name], logger);
+        // Flush and delete the chains. `-X` is the command that actually
+        // relinquishes the chain, so ownership is only cleared when it succeeds.
+        if created.v4_chain {
+            let _ = Self::run_iptables(&["-F", &self.chain_name], logger);
+            if Self::run_iptables(&["-X", &self.chain_name], logger).is_ok() {
```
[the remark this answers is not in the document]
dhoehna: Correct, and fixed in 10f0599.  -X failing on a referenced chain is harmless, but -F succeeds no matter who points at the chain, and an empty user chain returns to the caller instead of reaching its closing DROP.  A container whose hook survived teardown kept running with the jump intact and nothing behind it, which is fail-open for as long as it lives.

Both commands are now gated on the hooks being confirmed gone, so the chain stays populated and stays published and a later pass retries.  I extracted teardown_chain to make the ordering testable without iptables -- the previous shape had no test at all, and a mutant that flushed unconditionally survived the entire suite.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,155 +710,421 @@ impl NetworkIptablesManager {
...
+                residual.v4_dhcp = false;
+            }
+            if created.v4_hooks {
+                let removed = remove_every_hook(HOOKS, |hook| {
+                    Self::run_iptables(&["-D", hook, "-i", iface, "-j", &self.chain_name], logger)
+                        .is_ok()
+                });
+                if removed {
+                    residual.v4_hooks = false;
+                }
```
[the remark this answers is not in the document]
dhoehna: The failure mode is real and is fixed in 10f0599, but not the way you suggested, and the difference matters.

Per-hook flags do not fix it on their own.  Ownership is published before the -I that installs a hook, deliberately, to close the window where a signal would find the rule unpublished.  So a per-hook flag can be published ahead of an -I that then fails, and -D for a rule that was never installed fails too -- the flag sticks exactly as the shared one did.  Splitting the record changes its granularity, not the question being asked of iptables.

What is wrong is treating a failed delete as evidence the hook is still there.  The postcondition that matters is absence.  A failed -D is now followed by -C, and a hook that is not present counts as removed.

-C answers 0 for present and 1 for absent; any other exit, or a failure to run the binary, means the question was not answered, and an unanswered question reads as still-installed.  The opposite default would let a broken or permission-denied iptables report a live hook as gone and let the caller flush a chain something still jumps to.  Being wrong in this direction only costs a retry.

Mutation-verified: reverting to delete-must-succeed fails the new test.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,149 +667,418 @@ impl NetworkIptablesManager {
...
+        // This process created nothing, so there is nothing of ours to remove.
+        // Anything answering to this chain name belongs to a different, live
+        // container, and tearing it down would silently fail that container
+        // open.
+        if created.is_empty() {
+            return;
+        }
         let mut mgr = Self::new(container_name);
         if let Some(v) = veth_interface {
             mgr.set_veth_interface(v);
```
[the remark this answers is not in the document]
dhoehna: I checked this one against the branch and it no longer holds.  force_cleanup sets mgr.created = created -- the published record verbatim -- and remove_firewall_rules tears down self.created.  Nothing on that path expands to CreatedFirewallState::all().

The expansion you are describing was the original bug, and it is what this PR fixed.  all() is now #[cfg(test)] (network_iptables.rs:79), so a production path that called it would not compile.  It survives only as the stale superset a test hands to teardown to prove teardown replaces it.

The partial-record case you raise is the right thing to worry about and it is handled by construction: a signal after IPv4 -N but before IPv6 -N finds a record with v4_chain set and v6_chain clear, and teardown acts on exactly that.  It issues no ip6tables command at all, so there is nothing for a colliding container to lose.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -139,149 +667,418 @@ impl NetworkIptablesManager {
...
-            NetworkPolicy::Block => "DROP",
-            NetworkPolicy::Allow => "ACCEPT",
+        // Create custom chains. A failure here means the name is already taken
+        // (chain names are truncated, so a collision with another live
+        // container is possible), which is why `created` is only marked after
+        // the command succeeds — the rollback must not touch a chain we did
+        // not make.
+        Self::run_iptables(&["-N", &self.chain_name], logger)?;
+        created.v4_chain = true;
+        Self::publish_created(created);
```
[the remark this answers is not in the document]
dhoehna: Confirmed, still open, and I am recording it as a known gap rather than attempting a blind fix.

The window is real and it is narrow: -N succeeds, and the very next statement publishes ownership.  A signal in between leaves a chain that exists and is unpublished, so the watchdog skips it.  I cannot close it by reordering, because publishing before -N is worse -- a process whose -N lost the race to a concurrent start would then own a chain it did not create and delete the winner's firewall on a signal.  Getting this right needs the publish and the command to be atomic with respect to the watchdog, which means either blocking signals around the pair or a cleanup design that is safe to run against a chain someone else owns.

Neither is a small change, and neither can be validated on this workstation: reproducing it needs Linux with real signal delivery, and every iptables command fails at spawn here.  Landing a fix I cannot test for a race I cannot reproduce is how the wrong invariant gets shipped with a green suite behind it.

What the outcome actually costs is a leaked chain -- host state with nobody to remove it -- not a container running unfiltered.  That is a leak, not a fail-open.  I have written it into the PR description as a known gap so it survives this thread.

dhoehna:
Closing this PR.  The model-2 work is being re-cut from `main` as a series of small, independently reviewable PRs rather than continued on this branch.

Two things drove that, and the review backlog is the bigger one.  This branch carries 63 review comments across 33 commits, and **52 of them are substantive design or correctness objections** rather than nits.  At 3,593 insertions across 17 files, each individual correctness fix became hard to verify in isolation, which is the same failure that ended #633.  Separately, the branch no longer merges with `main` — 4 conflicting files and 22 hunks, most of them from the `CreatedResources` / `CreatedFirewallState` rename — but that is the smaller problem and would not on its own have justified abandoning the branch.

The review comments here are the **input** to the re-cut, not discarded.  Six substantive themes were extracted and are being carried forward as design constraints on the new work:

| Theme | Comments | Constraint on the re-cut |
|---|---:|---|
| IPv6 security and bypass | 5 | The deny-all floor was IPv4-only, so egress over IPv6 was unfiltered and the entire model-2 posture was evadable (#discussion_r3583026859).  `ip6tables` parity is now a launch requirement of the enforcement slice, not a follow-up. |
| Firewall teardown, orphan chains, watchdog races | 13 | Teardown removes exactly what was created, with ownership tracked (#discussion_r3582991747). |
| Firewall rule scope and egress isolation | 9 | No blanket DNS hole in proxy mode — port 53 was unconditionally allowed, so the sandbox could reach arbitrary DNS servers directly (#discussion_r3562432927). |
| Proxy handling, hostname resolution, TLS SNI | 8 | Do not rewrite a TLS proxy URL's host to a resolved IP; it breaks SNI and certificate validation (#discussion_r3724788051). |
| Fail-open behavior and error handling | 7 | Fail **closed**.  A resolution failure previously logged a warning and continued, silently dropping a `blockedHosts` DROP rule (#discussion_r3724803457). |
| Container lifecycle and environment isolation | 10 | The firewall must be installed before the container can emit a packet; there was a window of up to 10 seconds where it could not have been (#discussion_r3715977786). |

The first slice is already up: https://github.com/microsoft/mxc/pull/788 covers the proxy environment-variable hygiene (roadmap requirement 4) — scrubbing every caller-supplied proxy variable so a workload cannot redirect or disable the cooperative proxy through its own environment.  It is deliberately small and carries no firewall code.

Still to be re-cut, each as its own PR: the proxy URL and address model, loopback proxy validation, the `iptables` **and** `ip6tables` enforcement core, the runner and signal-cleanup lifecycle integration, and the docs plus integration scripts.

The branch itself is preserved and nothing is lost.

PR #634, opened 2026-07-10
[Bubblewrap/LXC] Address network policy gaps - schema

dhoehna:
Linked work item: AB#62830582 - [Bubblewrap/LXC] Address network policy gaps - schema

## Summary

Migrates LXC and Bubblewrap network policy from the current schema (`allowedHosts` / `blockedHosts` / `defaultPolicy`) to the GA egress/ingress schema, with a shared parser and SDK types.  This is roadmap item 18 "(N7) Schema migration".

- New wire shape under `network`: `egress.{default, allow[], deny[]}` (each rule `to: [{ cidr }]`, `ports: [{ protocol, port }]`) and `ingress.hostLoopback`.
- `egress.default: "allow" | "deny"` maps to the internal default outbound policy; omitted means deny (fail-closed).
- `ports` is optional; omitted means all ports and protocols.  `protocol: "icmp"` rejects a `port`, since ICMP has none.
- `(protocol, port)` selectors are grouped by protocol, so `tcp/80` + `udp/53` does not widen into `tcp/53`.
- Destinations are CIDR-only, validated with `std::net`; DNS hostnames are rejected with a clear parse error.
- `ingress.hostLoopback` maps to `allow_local_network`.
- Legacy `allowedHosts` / `blockedHosts` / `defaultPolicy` continue to work.  GA `network.egress` supersedes the legacy list fields; `ingress.hostLoopback` overrides `allowLocalNetwork`.
- An external `network.proxy` combined with `network.egress` is rejected for all backends.
- Adds `EgressRule`, `Protocol`, and `RuleAction` to `wxc_common::models`, the GA SDK policy types, and regenerated JSON schema and TS wire types.

## Status — blocked pending network-schema owners

This PR is parked and should not be merged as-is.  Network-schema changes in this repository are owned by the schema owners (MGudgin and Stuart), and this PR introduces GA schema types that do not exist in `main` today — `NetworkEgress`, `EgressDefault`, `EgressRuleWire`, `NetworkProtocol`, `NetworkIngress`, `HostLoopbackPolicy`, and `loopbackPorts`, 39 references across `policy.rs`, `config_parser.rs`, and `wire.rs`.  It awaits the schema owners' review and direction.  A prior GA network-schema PR, #676, was reverted by #707 on the same ownership grounds.

To unblock: the schema owners land or approve the GA network schema shape, after which this parser and its SDK types are reconciled against the owned schema.

This is the only one of the author's open PRs with a real dependency on the GA schema.  The sibling PRs (#724, #633, #632, and #627) ship against the current `main` schema and are independent of this one.

## Notes

- Proxy shape is unchanged: this PR keeps `network.proxy.http`, which shares the same URL parser as `network.proxy.url` and so inherits non-HTTP scheme rejection and credential redaction.  It does not introduce `runtimeConfig.networkProxy`, since `runtimeConfig` is not part of the current schema; relocating the proxy is tracked separately.
- The internal `EgressRule` / `Protocol` / `RuleAction` types are also added by the enforcement PRs (AB#62830559, AB#62830341).  This PR owns the JSON parser; those PRs consume the model field, and merge-time reconciliation is expected.

## Validation

Linux (WSL, Ubuntu 24.04) at `bdded01`:

- `cargo build -p wxc_common -p mxc_engine` — builds clean.
- `cargo test -p wxc_common -p mxc_engine` — `wxc_common` 498 passed, `mxc_engine` 14 passed.
- `cargo clippy -p wxc_common -p mxc_engine --all-targets -- -D warnings` — clean.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/634)


src/core/wxc_common/src/config_parser.rs
```diff
@@ -411,40 +413,166 @@ fn convert_wire_proxy(proxy: wire::Proxy) -> Result<ProxyConfig, WxcError> {
...
+                wire::NetworkProtocol::Icmp => Protocol::Icmp,
+            });
+        }
+
+        out.push(EgressRule {
+            destinations,
+            ports,
+            protocols,
+            action: action.clone(),
         });
```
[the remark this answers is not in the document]
dhoehna: Thanks — addressed in 269909d.

`convert_wire_egress_rules` now groups the GA `(protocol, port)` selectors by protocol and emits one internal `EgressRule` per protocol group, so `protocols x ports` is no longer widened into a cartesian product. For example, a rule with `tcp/80`, `tcp/443`, `udp/53` now maps to two rules — `{ tcp: [80, 443] }` and `{ udp: [53] }` — and no longer permits `tcp/53` or `udp/80`. Ports are grouped in first-seen order and deduped per protocol.

Added regression test `ga_egress_groups_mixed_selectors_by_protocol_without_widening`; `cargo fmt`/`clippy -p wxc_common -D warnings` are clean and `cargo test -p wxc_common` is green (404 passed).

src/core/wxc_common/src/wire.rs
```diff
@@ -261,20 +261,98 @@ pub struct Fallback {
...
+/// GA outbound port selector.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema-gen", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct EgressPortWire {
+    /// Transport protocol.
+    pub protocol: NetworkProtocol,
+    /// Destination port.
+    #[cfg_attr(feature = "schema-gen", schemars(range(min = 1, max = 65535)))]
+    pub port: u16,
```
SohamDas2021: You are making port mandatory, but ICMP has no ports. The GA schema (see here- https://microsoft-my.sharepoint-df.com/:w:/p/bbonaby/cQpR4CPfeKqgSLuQGG_a9QA2EgUCrPdXr5J7b-jWip1_VeYFUA) makes ports optional and omits it for icmp. Please make ports optional and reject port when protocol is icmp.

src/core/wxc_common/src/wire.rs
```diff
@@ -261,20 +261,98 @@ pub struct Fallback {
...
+    pub ingress: Option<NetworkIngress>,
+    /// Proxy configuration (legacy localhost / builtinTestServer / url, or GA http).
     pub proxy: Option<Proxy>,
 }
 
+/// GA outbound policy rule set.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema-gen", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct NetworkEgress {
```
SohamDas2021, line 354: The GA schema also has egress.default: deny|allow. Without it there's no way to express model‑1 "allow everything except this deny‑list" (D4's canonical case). Combined with the parser dropping legacy defaultPolicy when egress is present, a user who writes defaultPolicy:"allow" + egress is silently downgraded to default‑deny. Please add egress.default .

src/core/wxc_common/src/config_parser.rs
```diff
@@ -390,42 +413,186 @@ fn convert_wire_proxy(proxy: wire::Proxy) -> Result<ProxyConfig, WxcError> {
...
+    path: &str,
+    out: &mut Vec<EgressRule>,
+) -> Result<(), WxcError> {
+    for (rule_index, rule) in rules.into_iter().enumerate() {
+        if rule.to.is_empty() {
+            return Err(WxcError::ConfigParse(format!(
+                "{path}[{rule_index}].to must include at least one cidr entry"
+            )));
+        }
+        if rule.ports.is_empty() {
```
SohamDas2021: Requiring at least one ports entry per rule means "allow all ports to this CIDR" is inexpressible. The GA schema treats ports as optional (omitted = all ports/protocols). Please allow an empty/absent ports to mean match‑all.

src/core/wxc_common/src/wire.rs
```diff
@@ -309,6 +387,8 @@ pub struct Proxy {
     pub builtin_test_server: Option<bool>,
     /// Proxy URL (parsed into host:port).
     pub url: Option<String>,
+    /// GA HTTP proxy URL (parsed into host:port).
+    pub http: Option<String>,
```
SohamDas2021, line 478: The GA spec puts the proxy in runtimeConfig.networkProxy (a URL, outside the network policy block), not network.proxy.http, right?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -821,10 +1017,11 @@ fn convert_wire_config(
             && !policy.network_proxy.builtin_test_server
             && (!policy.allowed_hosts.is_empty()
                 || !policy.blocked_hosts.is_empty()
+                || !policy.egress_rules.is_empty()
                 || policy.default_network_policy == NetworkPolicy::Block)
         {
-            let msg = "Bubblewrap: an external network.proxy (url/localhost) cannot be \
-                       combined with allowedHosts, blockedHosts, or defaultPolicy='block'. \
+            let msg = "Bubblewrap: an external network.proxy (url/localhost/http) cannot be \
```
SohamDas2021, line 1352: This rejects an external proxy combined with egress for Bubblewrap only. The GA doc says egress.allow/deny should be rejected under proxy mode generally; on LXC and other backends both can currently be set at once. Should this consistency check be backend‑agnostic?

SohamDas2021 (changes requested):

src/core/wxc_common/src/wire.rs
dhoehna: Good catch — fixed in 70c94d6. `EgressPortWire.port` is now `Option<u16>` (optional in the schema), and the parser rejects a `port` paired with `protocol: "icmp"` with a clear message (`… specifies a port for protocol 'icmp', which has no ports; omit 'port' for icmp`). An ICMP selector is now written as `{ "protocol": "icmp" }`. Covered by the new `ga_egress_rejects_port_with_icmp` test and the updated `ga_egress_accepts_ipv4_ipv6_cidr_and_bare_ip` test.

src/core/wxc_common/src/wire.rs
dhoehna: Added in 70c94d6. `network.egress` now accepts `default: "allow" | "deny"` (wire `EgressDefault`). It maps to the internal `default_network_policy` (`allow` → Allow; `deny` or omitted → Block, i.e. fail-closed), so the "allow everything except this deny-list" model is expressible and a legacy `defaultPolicy: "allow"` is no longer silently downgraded when GA egress is present. New tests: `ga_egress_default_allow_sets_default_network_policy` and `ga_egress_default_deny_and_omitted_both_fail_closed`.

src/core/wxc_common/src/config_parser.rs
dhoehna: Fixed in 70c94d6. `EgressRuleWire.ports` is now `#[serde(default)]` (optional). An omitted/empty `ports` list means match-all — all ports and all protocols to the listed destinations (internally an `EgressRule` with empty `ports` and empty `protocols`). A portless `tcp`/`udp` selector such as `{ "protocol": "tcp" }` matches all ports for that protocol. New tests: `ga_egress_omitted_ports_matches_all_protocols_and_ports` and `ga_egress_portless_tcp_selector_matches_all_ports`.

src/core/wxc_common/src/wire.rs
dhoehna: Thanks — I'd like to defer the `runtimeConfig.networkProxy` relocation to a follow-up rather than fold it into this PR. Two reasons:

1. **Scope / risk.** There is no `runtimeConfig` section anywhere in the wire model today, so this introduces a brand-new top-level config surface that is cross-cutting — schema, SDK public types, generated wire types, and every backend's proxy handling. That is meaningfully broader than this PR's "network policy schema" scope.
2. **Spec fidelity.** Doing it correctly needs the canonical GA `runtimeConfig.networkProxy` spec (exact shape, precedence vs. the existing `network.proxy`, and which backends honor it). I don't want to guess the shape and bake in something we then have to break.

This PR intentionally leaves the existing `network.proxy.http` surface unchanged. I'll open a tracking issue for the `runtimeConfig.networkProxy` migration and link it here. If you have a link to the GA spec you'd like followed, I'm happy to do the relocation in a dedicated follow-up PR.

src/core/wxc_common/src/config_parser.rs
dhoehna: Fixed in 70c94d6. The external-proxy + egress rejection is now backend-agnostic: it runs in `convert_wire_config` for all backends, guarded by `network_proxy.is_enabled() && !builtin_test_server && !egress_rules.is_empty()`, with a message that `network.egress` can't be combined with an external `network.proxy` because MXC does not forward egress rules to the proxy. The Bubblewrap-specific check no longer mentions egress (that responsibility moved up). New test `external_proxy_with_egress_is_rejected_backend_agnostic` exercises this using `processcontainer` and `bubblewrap`.

src/core/wxc_common/src/wire.rs
dhoehna: Tracking issue filed: #655 — I'll do the `runtimeConfig.networkProxy` relocation there in a dedicated follow-up (it needs the canonical GA spec for the exact shape and precedence vs. the existing `network.proxy`, plus it touches a new top-level `runtimeConfig` surface across schema, SDK types, generated wire types, and every backend). This PR intentionally leaves `network.proxy.http` unchanged. Leaving this thread open as the pointer to #655.

dhoehna:

Re-ran the test suites locally at branch tip `66c65d6` on a dev workstation. All green (0 failures).

- `cargo test -p wxc_common` → **409 passed, 0 failed**
- `cargo test -p mxc-sdk` → **11 passed, 0 failed** (+4 runtime-ignored — require an elevated, host-prepped Windows host per `docs/host-prep.md`)

- `cargo test -p lxc_common` → **39 passed, 0 failed** — the workspace Linux crate also builds and passes.

_``lxc_common`` is unchanged by this PR; this PR's schema/parser changes are exercised by the ``wxc_common`` + ``mxc-sdk`` suites above._

dhoehna:
@SohamDas2021 — every item from your review has been addressed and pushed; requesting a re-review.

**Resolved** (in `70c94d6`; the Copilot selector item in `269909d`):
- ICMP has no ports — `EgressPortWire.port` is now `Option<u16>` and a `port` paired with `protocol: "icmp"` is rejected with a clear message (thread r3581216215).
- Missing `egress.default` — `network.egress` now accepts `default: allow|deny`, mapped to the internal `default_network_policy` (fail-closed when omitted), so `defaultPolicy: "allow"` is no longer silently downgraded (thread r3581235209).
- `ports` should be optional (match-all) — omitted/empty `ports` now means all ports/protocols to the destination; a portless `tcp`/`udp` selector matches all ports for that protocol (thread r3581245979).
- Proxy + egress rejection should be backend-agnostic — the check moved into `convert_wire_config` and now runs for all backends, not just Bubblewrap (thread r3581281785).
- Egress selectors cartesian-widened (Copilot) — selectors are now grouped into one internal `EgressRule` per protocol, so `tcp/80 + udp/53` no longer permits `tcp/53` (thread r3562427462).

**Deferred (tracked in #655):**
- Relocate proxy to `runtimeConfig.networkProxy` — filed as #655. It needs the canonical GA spec and introduces a new top-level `runtimeConfig` surface (schema, SDK types, generated wire types, every backend), so it is out of scope for this schema-migration PR. Thread r3581267529 is left open as the pointer to #655.

**Validation @ `66c65d6`:** CI green (20/20 checks). Re-verified locally on this branch tip: `cargo test -p wxc_common` -> 409 passed, 0 failed; `cargo test -p mxc-sdk` -> green, 0 failed (streaming test ignored — needs an elevated, host-prepped Windows host).

src/core/wxc_common/src/wire.rs
SohamDas2021: I would block on this. Let's get the schema finalized at once and not check in temporary incorrect shapes.

dhoehna: Agreed on not checking in temporary incorrect shapes — that's the same conclusion I reached, which is exactly why I'd rather not add `runtimeConfig` in this PR.

`runtimeConfig` does not exist anywhere in `main` today. If this PR introduces `runtimeConfig.networkProxy`, its shape is defined *by this PR* rather than by a landed canonical schema, and the real schema then has to land on top of whatever I guessed. That is the temporary-incorrect-shape risk, not the mitigation for it.

I did build the relocation out to see how large it is: it reaches the schema, the generated wire types, the SDK public types, the parser, and every backend's proxy handling. It's ready to replay, but I'd rather replay it against the finalized schema than land it ahead of one.

So — two ways to sequence this, and I'm happy with either:

1. **Hold #634** until the GA schema is in `main`, then land the egress work and the proxy relocation together. Matches "get the schema finalized at once"; cost is that the reviewed egress fixes sit unmerged.
2. **Land #634's egress work now** — the `(protocol, port)` selectors, `egress.default`, optional/`match-all` ports, ICMP port rejection, and the backend-agnostic proxy+egress check are all reviewed and addressed in the threads above — and take the proxy relocation in #655 once the schema is available.

Which do you prefer? This PR is currently conflicting with `main`, so if we're holding it I'll leave it conflicting rather than keep rebasing it, and pick it back up when the schema lands.

dhoehna: Agreed — no temporary shapes. To be explicit about the commitment: **this code will be changed to match the GA spec once the schema change is merged into `main`.** I'm not going to land a `runtimeConfig` shape that this PR invents.

Where `main` actually stands today (checked at `839ace4`):

- The **prose spec is in**: `docs/sandbox-policy/v2/networking.md` (added by #615) documents `runtimeConfig.networkProxy` and the GA restriction that only `localhost:<port>`, `127.0.0.1:<port>`, and `[::1]:<port>` are accepted.
- The **machine-readable schema is not**: `schemas/dev/mxc-config.schema.0.8.0-dev.json` has zero occurrences of `runtimeConfig`, `networkProxy`, `egress`, or `ingress`. Its `Network` block is still the legacy `allowedHosts` / `blockedHosts` / `allowLocalNetwork` shape.

So there is a documented target but nothing landed at the wire/schema level to conform to yet. That's the gap this thread is really about, and it's why this PR deliberately leaves `network.proxy.http` alone instead of adding a parallel surface.

When the schema change lands I'll do the relocation against it — including enforcing the loopback-only GA restriction on the proxy URL — rather than re-deriving the shape. I already built the relocation out once end to end (schema, generated wire types, SDK public types, parser, and each backend's proxy handling) and validated it locally, so it's parked and ready to replay against the real schema; this isn't an open-ended deferral. Tracked in #655.

Holding this PR in the meantime, so I'm leaving it conflicting with `main` rather than repeatedly rebasing a branch that's waiting on someone else's merge. Happy to pick it straight back up the moment the schema is in.

dhoehna: Correction to my last message: I said I'd leave this conflicting with `main`. I've merged instead, because it turned out to matter.

While re-verifying this branch I found a test that only ever passed on Windows: `build_request` resolves the containment backend from the host platform, so `build_request_maps_http_proxy_without_egress` ran under Bubblewrap on Linux and hit the guard rejecting an external proxy combined with the implicit `defaultPolicy: block`. That guard is correct — this branch adds the `http` proxy form, which doesn't exist in `main`, and widening the guard to cover it was the right call — so the test was the thing that was wrong. Fixed, plus the regression test that the `http` form trips the guard the same way `url` and `localhost` already do.

Pushing that fix showed the second problem: while the PR was conflicting, GitHub couldn't build the merge commit, so none of the workflows ran — the fix landed with only `license/cla` reporting. Leaving a branch conflicting is fine when nothing is changing on it, but not when I've just pushed code that CI can't check.

So I merged `main` (one conflicted file, `config_parser.rs`). Worth flagging one thing in that resolution: `main` added a non-HTTP scheme rejection with credential redaction inline in the `url` branch, while this branch had refactored URL parsing into a shared helper. I folded the new guard into the shared helper so `network.proxy.http` inherits it too — otherwise the GA proxy field would have been a way around a check that had just been added.

Now green on both platforms: Windows `wxc_common` 570 / `mxc_engine` 17 / `mxc-sdk` 18, Linux `wxc_common` 494 / `mxc_engine` 14, fmt and clippy clean on both.

None of this changes the answer above — the proxy stays at `network.proxy.http` and moves to `runtimeConfig.networkProxy` when the schema lands. The PR is just now in a reviewable state while it waits.

dhoehna:
**Parked: waiting on the GA network schema.**

Converting to draft.  This PR cannot be completed until the GA network schema lands, because it *is* the schema migration (roadmap item 18, N7).  It introduces GA types that do not exist in `main` today — `NetworkEgress`, `EgressDefault`, `EgressRuleWire`, `NetworkProtocol`, `NetworkIngress`, `HostLoopbackPolicy`, and `loopbackPorts` — across 39 references in `policy.rs`, `config_parser.rs`, and `wire.rs`.

Network-schema changes in this repository are owned by the schema owners.  The prior GA network-schema PR, #676, was reverted by #707 on those grounds, so this branch is not something I should land unilaterally.

Draft status here means "blocked on an external decision," not "abandoned."  The review feedback on this PR is still being worked — the `egress.default: allow|deny` addition requested in review landed in 70c94d6, and the remaining items (optional `ports` for ICMP, `runtimeConfig.networkProxy` placement, and making the proxy/egress consistency check backend-agnostic) are tracked on their threads.

To unblock: the schema owners land or approve the GA network schema shape, after which this parser and its SDK types get reconciled against the owned schema.  I will move it out of draft at that point.

PR #641, opened 2026-07-13
Phase 3d: add actionable configuration parse errors

MGudgin:
This PR adds actionable configuration parse errors: the parser reports
malformed JSON separately from typed policy-shape errors and includes the full
policy path plus whole-file source location in every diagnostic, without leaking
secrets or letting untrusted text corrupt diagnostics or the state-aware stdout
response envelope.

## Details

* New `config_deserialize` module: a path-aware deserialize layer that
  distinguishes JSON syntax errors from typed policy errors, attaching the full
  JSON path and whole-file line/column, escaping control and invisible
  formatting characters (incl. U+2028/U+2029 and bidi overrides), and redacting
  secret-bearing values. Format-character detection uses the
  `unicode-general-category` crate (Format | LineSeparator | ParagraphSeparator),
  so the Cf/Zl/Zp set tracks the crate's Unicode tables.
* Secret redaction is path-driven: a value is redacted when its policy path
  carries a secret marker. Descriptive markers (`token`, `secret`, `password`,
  ...) match anywhere within a path segment (so `wamToken` / `clientSecret`
  redact), while `user` is matched as a whole path segment so the credential
  bundle (`user: { upn, wamToken }`) redacts even when a scalar is supplied
  where the object is expected, without colliding with unrelated fields like
  `username`.
* State-aware per-backend per-phase configs deserialize positionally from the
  retained request text, so typed errors carry whole-file coordinates matching
  base-config errors; navigation uses owned-key maps so escaped sibling keys
  never silently drop the location, with graceful fallback to the value-based
  path.
* Request loading borrows `RawValue` to discriminate one-shot vs state-aware
  without building a full untyped AST, masking the `experimental` subtree to
  preserve base-config source locations.
* Diagnostic routing keeps stdout reserved for the state-aware envelope and
  script output; parse and dispatch errors go to auxiliary sinks (log file /
  diagnostic pipe) only and are emitted exactly once. Filesystem-path and
  schema-version diagnostics escape untrusted text via the shared
  `escape_diagnostic_text` helper.
* Exec dispatch drops the parsed request (and its retained source text) before
  the blocking child run and stdio relay.
* Docs updated: `docs/versioning.md`, the state-aware API doc, and the
  copilot-instructions config-flow description.

## Tests

* Added parser coverage: syntax-vs-typed classification, whole-file line/column
  (computed, not hardcoded), escaped-sibling-key location preservation, exact
  column, CRLF line stability, positional secret redaction, source-present
  locator fallback, malformed-telemetry single auxiliary emission with an empty
  primary buffer, a pinned `serde_json` positioned-error display-suffix
  contract, format-character escaping (U+202E/U+200B/U+2028/U+2029), and
  segment-aware secret paths (`user` / nested `user` / `user.upn` / `[i].user`
  redact; `username` / `userProfile` do not).
* `cargo fmt --all -- --check`: passed.
* `cargo clippy -p wxc_common -p mxc_engine -p wxc --all-targets -- -D warnings`:
  clean.
* `cargo test -p wxc_common -p wxc`: 538 + 28 passed.
* `node scripts/versioning/check-schema-codegen.js`,
  `check-sdk-types-codegen.js`, `check-schema-versions.js`: passed.

### DIFF
--- a/src/core/wxc_common/src/config_deserialize.rs
+++ b/src/core/wxc_common/src/config_deserialize.rs
@@ -1,0 +1,748 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::fmt;
+
+use serde::{de::DeserializeOwned, Deserialize, Deserializer};
+use serde_json::{error::Category, Value};
+use unicode_general_category::{get_general_category, GeneralCategory};
+
+/// Field-name substrings that mark a value as secret-bearing. Matched anywhere
+/// within a single path segment, so `token` catches `wamToken`, `secret`
+/// catches `clientSecret`, and so on.
+const SECRET_PATH_MARKERS: &[&str] = &[
+    "token",
+    "password",
+    "secret",
+    "credential",
+    "apikey",
+    "accesskey",
+    "privatekey",
+    "passphrase",
+    "pwd",
+    "bearer",
+    "saskey",
+    "connectionstring",
+];
+
+/// Whole path segments that mark a value as secret-bearing. Matched as a
+/// complete field name (not a substring) so they never collide with unrelated
+/// fields such as `username`. `user` is the credential bundle
+/// (`user: { upn, wamToken }`): a shape error where the object is expected is
+/// pathed at `user`, not at `wamToken`, so the whole subtree must be marked.
+/// Over-redaction fails safe — it only replaces an error value with a location,
+/// never leaks one.
+const SECRET_PATH_SEGMENTS: &[&str] = &["user"];
+
+/// A JSON deserialization failure with the path at which typed policy parsing
+/// failed. Syntax errors have no meaningful policy path.
+#[derive(Debug)]
+pub(crate) struct ConfigDeserializeError {
+    path: Option<String>,
+    source: serde_json::Error,
+    /// Whole-file `(line, column)` that overrides the location baked into
+    /// `source` when the error was produced from a sub-slice of a larger
+    /// request (e.g. a state-aware `experimental.<backend>.<phase>` fragment).
+    /// `None` leaves `source`'s own location untouched.
+    location_override: Option<(usize, usize)>,
+}
+
+impl ConfigDeserializeError {
+    fn from_path_error(error: serde_path_to_error::Error<serde_json::Error>) -> Self {
+        let path = error.path().to_string();
+        let path = (path != ".").then_some(path);
+        Self {
+            path,
+            source: error.into_inner(),
+            location_override: None,
+        }
+    }
+
+    /// Override the source location rendered by `Display` with whole-file
+    /// coordinates. Used when translating a fragment-local serde location back
+    /// to its position in the complete request text.
+    pub(crate) fn with_source_location(mut self, line: usize, column: usize) -> Self {
+        self.location_override = Some((line, column));
+        self
+    }
+
+    /// The `(line, column)` serde recorded for this error, or `None` when serde
+    /// could not attribute a position (it reports line 0 in that case).
+    pub(crate) fn source_line_column(&self) -> Option<(usize, usize)> {
+        let line = self.source.line();
+        (line > 0).then(|| (line, self.source.column()))
+    }
+
+    /// Prefix a path produced while deserializing a JSON subtree with its path
+    /// in the complete request.
+    pub(crate) fn with_prefix(mut self, prefix: &str) -> Self {
+        self.path = Some(match self.path.take() {
+            None => prefix.to_string(),
+            Some(path) if path.starts_with('[') => format!("{prefix}{path}"),
+            Some(path) => format!("{prefix}.{path}"),
+        });
+        self
+    }
+
+    fn path_contains_secret(&self) -> bool {
+        self.path.as_deref().is_some_and(|path| {
+            path.to_ascii_lowercase().split('.').any(|segment| {
+                // Match on the field name only, dropping any array-index suffix
+                // so `field[0]` matches on `field`.
+                let field = segment.split('[').next().unwrap_or(segment);
+                SECRET_PATH_SEGMENTS.contains(&field)
+                    || SECRET_PATH_MARKERS
+                        .iter()
+                        .any(|marker| field.contains(marker))
+            })
+        })
+    }
+}
+
+impl fmt::Display for ConfigDeserializeError {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let source = if self.path_contains_secret() {
+            redact_secret_value(&self.source)
+        } else {
+            self.source.to_string()
+        };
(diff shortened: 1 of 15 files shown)


src/core/wxc_common/src/config_parser.rs
```diff
@@ -135,67 +179,70 @@ pub fn load_mxc_request_with_options(
...
-            WxcError::ConfigParse(msg.to_string())
+            WxcError::ConfigParse("Base64 decoded content is not valid UTF-8".to_string())
         })
     } else {
         if !std::path::Path::new(input).exists() {
-            let _ = write!(logger, "Configuration file not found: {}", input);
             return Err(WxcError::ConfigParse(format!(
-                "Configuration file not found: {}",
-                input
+                "Configuration file not found: {input}"
```
huzaifa-d: This includes the raw config-file path in a diagnostic. On Linux and macOS a filename can contain newlines or terminal control characters, so a missing file can inject forged log output even though this PR otherwise escapes untrusted diagnostic text.

src/core/wxc_common/src/state_aware_request.rs
```diff
@@ -109,15 +110,11 @@ impl ParsedStateAwareRequest {
         let Some(phase_value) = backend_obj.get(phase_name) else {
             return Ok(None);
         };
-        // Deserialize directly from the borrowed `&Value` — no need to clone
-        // the (potentially large) backend config subtree first.
-        <C as serde::Deserialize>::deserialize(phase_value)
+        config_deserialize::from_value_ref(phase_value)
```
huzaifa-d, line 162: Deserializing backend phase config from `serde_json::Value` has already discarded its source span, so these state-aware errors cannot include line and column information. This leaves the stated source-location improvement incomplete for backend-specific configuration.

src/core/wxc_common/src/config_parser.rs
MGudgin: escape_diagnostic_text now applied to the path in the "not found"/"failed to read" diagnostics

src/core/wxc_common/src/config_deserialize.rs
```diff
@@ -0,0 +1,714 @@
...
+    pub(crate) fn with_prefix(mut self, prefix: &str) -> Self {
+        self.path = Some(match self.path.take() {
+            None => prefix.to_string(),
+            Some(path) if path.starts_with('[') => format!("{prefix}{path}"),
+            Some(path) => format!("{prefix}.{path}"),
+        });
+        self
+    }
+
+    #[cfg(test)]
```
bbonaby: question (non-blocking): Any reason to have this function as a test?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1207,9 +1250,13 @@ fn convert_wire_state_aware(
...
             serde_json::from_value(telemetry_val.clone()).map_err(|e| {
-                let msg = format!("invalid experimental.telemetry: {e}");
-                logger.log_line(&msg);
-                WxcError::ConfigParse(msg)
+                // Do not log here: state-aware parse errors are routed centrally
+                // and exactly once by the outer `load_mxc_request*` wrapper via
+                // `log_error(..., ErrorOutput::DiagnosticOnly)`. Logging here as
+                // well produced a duplicate auxiliary diagnostic (finding F2).
+                // Returning the error keeps stdout clean (envelope-owned) and
+                // yields a single auxiliary-sink line.
```
bbonaby, line 1331: note: this comment still refers to other iterations I believe (F2).

src/core/wxc_common/src/config_deserialize.rs
```diff
@@ -0,0 +1,714 @@
...
+    "passphrase",
+    "pwd",
+    "bearer",
+    "saskey",
+    "connectionstring",
+    // Credential-bearing container: a shape error at the `user` node (e.g. a
+    // token supplied where the object is expected) is pathed at `user`, not at
+    // `wamToken`, so mark the whole subtree sensitive. Over-redaction here fails
+    // safe — it only replaces an error value with a location, never leaks one.
+    "user",
```
bbonaby: note: This here would make us redact anything that starts with `user` right?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1428,6 +1544,44 @@ mod tests {
         assert!(matches!(r, Err(ParseError::StateAware(_))), "got {:?}", r);
     }
 
+    #[test]
+    fn state_aware_malformed_telemetry_logs_once_and_keeps_primary_clean() {
+        // F2 regression: the malformed-telemetry error must reach the auxiliary
```
bbonaby: note: Note about comment about a previous iteration/PR maybe

src/core/wxc_common/src/config_deserialize.rs
```diff
@@ -0,0 +1,714 @@
...
+/// Invisible Unicode formatting characters (Unicode general category `Cf` and a
+/// few related invisibles) that `char::is_control()` does **not** cover. This
+/// table is a deliberate security control, not incidental hardening: escaping
+/// bidirectional overrides/isolates (U+202A–U+202E, U+2066–U+2069) defends
+/// against "Trojan Source" (CVE-2021-42574) visual-spoofing of diagnostics,
+/// escaping the line/paragraph separators (U+2028/U+2029) prevents forging hard
+/// line breaks that some terminals/log viewers honor, and escaping zero-width /
+/// joiner / interlinear characters prevents concealing or forging log and
+/// error-envelope content rendered in a terminal or editor.
+fn is_diagnostic_format_character(character: char) -> bool {
```
bbonaby, line 287: thought: I do wonder if there is a library we can use for this sort of thing.  I asked this one to copilot and interestingly it gave me this crate https://crates.io/crates/unicode-general-category where we check by category. What are your thoughts on this? We could test this with your tests below.

```rust
use unicode_general_category::{get_general_category, GeneralCategory as Gc};
fn is_diagnostic_format_character(character: char) -> bool {
    matches!(
        get_general_category(character),
        Gc::Format | Gc::LineSeparator | Gc::ParagraphSeparator
    )
}
```

MGudgin: LGTM. Adopted.

src/core/wxc_common/src/config_parser.rs
MGudgin: Fixed.

src/core/wxc_common/src/config_deserialize.rs
MGudgin: inlined.

src/core/wxc_common/src/config_deserialize.rs
MGudgin: Yes, probably over-redacts in some cases. I'm working on something slightly less aggressive.

MGudgin: Fixed.

bbonaby (approved):

PR #647, opened 2026-07-14
Enable one-click crates.io release for mxc-sdk

dhoehna:
## Summary

Adds a one-click Azure DevOps pipeline that publishes the `mxc-sdk` crate and its first-party dependency closure to crates.io, plus the crate-metadata changes required to make that closure publishable.

This branch contains **two logical commits**:

1. **Make the mxc-sdk crate closure publishable** (11 `Cargo.toml` files) — registry-agnostic metadata: workspace `repository`, per-crate `description`/`license`/`repository`, dual `{ path, version }` on internal path deps, and making `sandbox_spec` publishable. Required to publish from anywhere.
2. **Add the release pipeline** (2 files) — `.azure-pipelines/1ES.Crate.Release.yml` and `.azure-pipelines/scripts/publish_crates_to_cratesio.py`.

## How it works

- `Run pipeline` with **dryRun = true** (the default) previews every crate via `cargo package --list` and publishes nothing.
- With **dryRun = false**, a Publish stage waits for manual approval on the `MXC-CratesIo-Production` environment, then publishes leaf-first, waiting for index propagation between crates. Re-runs are idempotent (a version already on crates.io is skipped).

## Secrets stay out of the repo

`CARGO_REGISTRY_TOKEN` comes from the ADO variable group `MXC-CratesIo-Publish` and is mapped to the publish step's env only. Nothing secret is committed.

## One-time setup (not codeable in the repo)

Variable group, environment + approval, crates.io egress allowlist, crate-name reservation / OSPO sign-off, and registering the pipeline. See the checklist at the bottom of `1ES.Crate.Release.yml`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/647)


.azure-pipelines/1ES.Crate.Release.yml
```diff
@@ -0,0 +1,137 @@
...
+# One-time ADO setup required before the first real publish (cannot live in the
+# repo) -- see the checklist at the bottom of this file.
+#
+# Unlike the build/lint jobs, the publish job deliberately does NOT run
+# Cargo.Setup.Public/Private.yml: those redirect crates.io to the internal
+# Azure Artifacts mirror. Publishing must talk to the real crates.io, so the
+# only cargo config in effect is the repo-root .cargo/config.toml (Windows
+# rustflags only -- no source replacement, and inert on Linux).
+
+trigger: none
```
bbonaby: issue: To publish to crates.io using the official Microsoft open source account we will need to publish it through the ESRP release system. See: https://eng.ms/docs/microsoft-security/identity/trust-and-security-services/tss-release-distribute/tss-release-esrp-parent/oss-publishing/releasing-open-source/cratesio . You will need to follow that and do something similar to what we do in the 1ES.Release.yml. To be honest, we should be able to reuse that same  yaml file to release to either Npm, Crates.io or both.

At the very least that one should be the entry point to all releases, but we can still have template yml files that we point to to keep it clean. Everything in terms of set up on the back end for ESRP should already be set up for us. So, let me know if you have any issues with that.

In the ESRP docs they say the azure rust sdk folks have this working, where they publish all their crate dependencies first then publish their main crate. I believe the yml file they do this in is this one https://github.com/Azure/azure-sdk-for-rust/blob/95eb8c3878cb2c7d3239b5f584b8179ee0483c31/eng/pipelines/templates/stages/archetype-rust-release.yml . We can probably do something similar.

bbonaby (changes requested):

.azure-pipelines/1ES.Crate.Release.yml
dhoehna: Thanks Brandon -- followed this. Reworked the crate release to go through **ESRP** and folded it into `1ES.Release.yml` so one pipeline releases npm, crates.io, or both:

- `1ES.Release.yml` now takes `publishNpm` / `publishCrates` boolean params. The existing NPM stage is gated on `publishNpm`, and a new `Publish_to_CratesIo` stage runs the ESRP path.
- Following the azure-sdk-for-rust approach, the official build packages the closure **leaf-first in a single `cargo package` invocation** (all `--package` flags together, so intra-closure deps resolve locally) into the `mxc-crates-package` artifact. The release job then runs one `EsrpRelease@12` task per crate (`contenttype: Rust`, account `microsoft-oss-releases`), leaf-first, verifying each on the crates.io sparse index before the next.
- Deleted the old token-based `1ES.Crate.Release.yml` + `publish_crates_to_cratesio.py` -- no more `CARGO_REGISTRY_TOKEN`. New templates: `templates/Package.Crates.Job.yml` (packaging) and `templates/Publish.CratesIo.Job.yml` (ESRP release loop).

This also obsoletes the two Copilot comments above (they were on the now-deleted script/pipeline).

One heads-up before we flip publishing on -- `publishCrates` defaults to **false** for now. Packaging the true `--all-features` closure of `mxc-sdk` surfaced that it now routes through the `mxc_engine` hub crate and expands to **17** first-party crates. Before a first real publish we still need to: (1) add a `version` to the 7 path deps in `src/Cargo.toml` that currently only carry `path`; (2) resolve `isolation_session_bindings` being `publish = false`; and (3) do the OSPO OSS-release registration + reserve the crate names. Happy to take that as a follow-up -- let me know if you'd rather have it in this PR or a separate one.

.azure-pipelines/templates/1ES.Build.Stages.yml
```diff
@@ -82,6 +82,16 @@ stages:
 
   - template: Mxc.Binary.Packaging.Job.yml
 
+- stage: Package_Crates
```
bbonaby: issue: we should probably add a conditional here to only do this packaging when the `isOfficialBuild` parameter is true.

.azure-pipelines/templates/Publish.CratesIo.Job.yml
```diff
@@ -0,0 +1,89 @@
...
+                    approvers: ${{ parameters.ESRPInfo.ApproversEmail }}
+                    waitforreleasecompletion: true
+                    serviceendpointurl: 'https://api.esrp.microsoft.com'
+                    mainpublisher: 'ESRPRELPACMAN'
+                    domaintenantid: ${{ parameters.ESRPInfo.tenantId }}
+
+                - script: |
+                    set -euo pipefail
+                    python3 "$(script)" wait --order-file "$(orderFile)" --crate "${{ crate }}"
+                  displayName: "Wait for ${{ crate }} on crates.io index"
```
bbonaby: note: the ESRP task above will actually keep polling until the package is published so we probably don't need this. It's done when the `waitforreleasecompletion` input to the task is set to true. That said, I'm not against the extra wait task.

.azure-pipelines/1ES.Release.yml
```diff
@@ -10,14 +44,45 @@ resources:
...
     source: 'MXC-Official-Build'
     trigger: none
 
 
 # Parameters for ESRP release info will be passed from the ADO UI.
 parameters:
+  - name: publishNpm
+    displayName: "Publish the npm SDK to npmjs.com"
+    type: boolean
+    default: true
```
bbonaby: issue: I like the parameters, lets make both false so it's clear in the UI that we expect the person initiating the pipeline run to select  one or both.

.azure-pipelines/scripts/crates_release.py
```diff
@@ -0,0 +1,321 @@
...
+    "sandbox_spec",
+    "wxc_common",
+    "lxc_common",
+    "seatbelt_common",
+    "appcontainer_common",
+    "bwrap_common",
+    "mxc-sdk",
+]
+
+CRATES_IO_SPARSE_INDEX = "https://index.crates.io"
```
bbonaby: issue: So this won't worked in the 1ES pipeline due to network isolation. You'd need to use the private feed's sparse index. Should be this one https://github.com/microsoft/mxc/blob/3fd07f5c1a81dfe8f9cc2c06e2346c78de47d5b2/.azure-pipelines/.cargo/config.toml#L8

So it would look something like this:
```python
AZURE_FEED_REGISTRY = "Mxc-Azure-Feed"
AZURE_FEED_SPARSE_INDEX = (
    "https://microsoft.pkgs.visualstudio.com/"
    "Dart/_packaging/Mxc-Azure-Feed/Cargo/index"
)
SYSTEM_ACCESS_TOKEN_ENV = "SYSTEM_ACCESSTOKEN"
```

then you can have a helper function in here somewhere:
```python
def _feed_request(url: str) -> urllib.request.Request:
    """Create an Azure Artifacts request authenticated by the pipeline token."""
    token = os.environ.get(SYSTEM_ACCESS_TOKEN_ENV)
    if not token:
        raise RuntimeError(
            f"{SYSTEM_ACCESS_TOKEN_ENV} is not set. Map $(System.AccessToken) "
            f"to that environment variable in the pipeline task."
        )

    return urllib.request.Request(
        url,
        headers={"Authorization": f"Bearer {token}"},
    )
```

Then your `_published_versions` function below can look similar to:
```python
def _published_versions(crate: str) -> set[str]:
    """Return the versions of `crate` already in the Azure Artifacts feed."""
    url = f"{AZURE_FEED_SPARSE_INDEX}/{_sparse_index_path(crate)}"
    with urllib.request.urlopen(_feed_request(url), timeout=30) as response:
        body = response.read().decode("utf-8")
    versions: set[str] = set()
    for line in body.splitlines():
        line = line.strip()
        if line:
            versions.add(json.loads(line)["vers"])
    return versions
```

You also have have references to crates.io in the comments of this file as well. You'd want to update those to just say `Azure Artifacts Feed` or something like that.

.azure-pipelines/scripts/crates_release.py
```diff
@@ -0,0 +1,321 @@
...
+# in the session plan / PR #647 discussion):
+#   1. 7 path deps in src/Cargo.toml [workspace.dependencies] lack a `version`
+#      (add `version = "0.7.0"` to match their 8 siblings).
+#   2. `isolation_session_bindings` is `publish = false` (a Windows WinRT bindings
+#      crate) -- it must be made publishable or dropped from mxc_engine's published
+#      manifest, since crates.io validates even optional/feature-gated deps.
+#   3. This list + the pipeline's `crateOrder` parameter must be reconciled to the
+#      final closure and OSPO OSS-release setup / crate-name reservation completed.
+# Until then the pipeline ships with `publishCrates=false` (default) and the
+# `verify-order` guard fails safely if `crateOrder` and this list disagree.
```
bbonaby, line 77: issue: We should update this to include all the crates that will be published as part of the mxc-sdk. That said, The crate names themselves should actually be prefixed with `mxc-` so they are easily discoverable in crates.io.  

I think we also need a follow up PR to add a GitHub PR workflow/action to check that all crates are publishable.

.azure-pipelines/1ES.Release.yml
```diff
@@ -1,5 +1,39 @@
...
+#    lxc_common, seatbelt_common, appcontainer_common, bwrap_common, mxc-sdk.
+# 3. Have the ESRP Release team enable the `Rust` content type for this pipeline's
+#    ESRP client id (see the "Tenant does not have access to the given Main
+#    Publisher" note in the ESRP crates.io docs). The same App-Registration
+#    identity used for npm/signing (ESRPInfo below) is reused.
+# 4. Allow read egress to the crates.io index (index.crates.io) on the release
+#    pool so the between-crates propagation check can confirm each publish.
+# No CARGO_REGISTRY_TOKEN, variable group, or custom environment is required --
+# ESRP owns the crates.io credentials, and the 1ES production releaseJob provides
+# the manual release gate.
```
bbonaby: issue: most of this comment can be removed or reduced now I believe.

.azure-pipelines/1ES.Release.yml
```diff
@@ -10,14 +44,45 @@ resources:
...
-  # Reference pipeline that created the signed npm artifacts so we can consume them later
+  # Reference pipeline that created the signed release artifacts so we can consume them later
   - pipeline: MXC 
     source: 'MXC-Official-Build'
     trigger: none
 
 
 # Parameters for ESRP release info will be passed from the ADO UI.
 parameters:
+  - name: publishNpm
```
bbonaby, line 34: Note: for your testing, you can add a `dry-run` boolean in here too, that when set won't go into the `EsrpRelease@10` tasks. You can actually test this out if you create an in repo branch (noticed this PR is coming from your fork). But yea if you create an in repo branch e.g u`ser/dhoehna/test-rust-publishing` with these same changes, you should be able to use the branch to run the official pipeline  via https://microsoft.visualstudio.com/Dart/_build?definitionId=190018 and then see any errors that pop up so you can fix them in this PR. Let me know if you don't have access.

.azure-pipelines/templates/1ES.Build.Stages.yml
dhoehna: Done in 287305b -- gated the `Package_Crates` stage on `isOfficialBuild` in `1ES.Build.Stages.yml`, so crate packaging only runs on the official signed build (not PR/CI).

.azure-pipelines/templates/Publish.CratesIo.Job.yml
dhoehna: Good point -- `waitforreleasecompletion: true` already blocks until each ESRP publish finishes. I kept the `wait` step as a lightweight guard that confirms the crate is resolvable on the feed index before the next dependent publishes, and it only emits a warning (never fails the release) on timeout. It is also skipped in the new `cratesDryRun` mode. Happy to drop it entirely if you would prefer -- your call. (Also fixed it to query the private feed per your other comment.)

.azure-pipelines/1ES.Release.yml
dhoehna: Done in 287305b -- `publishNpm` now defaults to `false` too, so the run dialog forces the operator to pick npm and/or crates.

.azure-pipelines/scripts/crates_release.py
dhoehna: Done in cfaaf18 -- switched the version check from public `index.crates.io` to the private `Mxc-Azure-Feed` sparse index (`AZURE_FEED_SPARSE_INDEX`, matching `.azure-pipelines/.cargo/config.toml`), added a `_feed_request()` Bearer-auth helper that reads `SYSTEM_ACCESSTOKEN`, and rewrote `_published_versions()` to hit the feed. Mapped `$(System.AccessToken)` -> `SYSTEM_ACCESSTOKEN` on the wait step and relabeled the crates.io comments to "Azure Artifacts feed". One assumption: the build identity has read access to `Mxc-Azure-Feed` -- let me know if that needs onboarding.

.azure-pipelines/scripts/crates_release.py
dhoehna: Agreed on all three -- two need a decision/setup before I wire them in, so flagging rather than guessing:

- **Full closure**: the real `mxc-sdk --all-features` closure is 17 crates (the `mxc_engine` hub was added by a later refactor). Before expanding `CRATES`/`crateOrder`, two blockers need resolving -- 7 path deps in `src/Cargo.toml` lack a `version`, and `isolation_session_bindings` is `publish = false`. Want those in this PR, or a follow-up?
- **`mxc-` prefix**: renaming published crate names touches each `[package] name` and needs the names reserved on crates.io. Can you confirm the scheme (e.g. `wxc_common` -> `mxc-wxc-common`)? I will apply it once we agree.
- **Publishability check action**: agree this is the follow-up PR you mentioned; I can open it separately.

The list is a clearly-marked placeholder and `publishCrates` defaults false, so nothing ships until this is settled.

.azure-pipelines/1ES.Release.yml
dhoehna: Done in 287305b -- trimmed the header to the essentials and dropped the stale "egress to index.crates.io" note (superseded by the private-feed fix).

.azure-pipelines/1ES.Release.yml
dhoehna: Added a `cratesDryRun` boolean in `1ES.Release.yml` (passed through to `Publish.CratesIo.Job.yml`) in 287305b: when set, each crate still runs `stage` + `verify-order` but skips the `EsrpRelease` task and the feed wait, so the plumbing runs without publishing.

On the in-repo branch: I do not have push access to `microsoft/mxc` from here (this PR is from my fork), so I cannot create `user/dhoehna/test-rust-publishing` myself. Could you create it from these changes (or grant access)? Then I will run pipeline 190018 against it and fix whatever surfaces.

dhoehna:
@bbonaby quick process question. I now have **write access to `microsoft/mxc`**, so in-repo branches are possible.

As you noted earlier, this fork-based PR can't actually exercise the release pipeline end-to-end: the official build and `1ES.Release` are both `trigger: none` and depend on the `MXC-ESRP-Signing` variable group / ESRP service connection, none of which is exposed to a fork PR (and a manual queue can't target a fork ref).

Would you like me to **abandon this PR and re-create it from an in-repo branch** (e.g. `user/dhoehna/test-rust-publishing`) so we can queue `MXC-Official-Build` and then `1ES.Release` with `publishCrates: true` + `cratesDryRun: true` (stage/verify only, nothing published)? Or would you rather keep this PR for the code review and I open a separate in-repo branch purely for the test run? Either works for me — whichever you prefer.

bbonaby:
> @bbonaby quick process question. I now have **write access to `microsoft/mxc`**, so in-repo branches are possible.
> 
> As you noted earlier, this fork-based PR can't actually exercise the release pipeline end-to-end: the official build and `1ES.Release` are both `trigger: none` and depend on the `MXC-ESRP-Signing` variable group / ESRP service connection, none of which is exposed to a fork PR (and a manual queue can't target a fork ref).
> 
> Would you like me to **abandon this PR and re-create it from an in-repo branch** (e.g. `user/dhoehna/test-rust-publishing`) so we can queue `MXC-Official-Build` and then `1ES.Release` with `publishCrates: true` + `cratesDryRun: true` (stage/verify only, nothing published)? Or would you rather keep this PR for the code review and I open a separate in-repo branch purely for the test run? Either works for me — whichever you prefer.

Yes let's abandon this one and make a new in repo branch for it. You can do the testing for your branch before you create a PR too. That's what I usually do so I can try things out fast without having Github send PR emails to reviewers about new commits.

dhoehna:
Closing in favor of an **in-repo branch**. Per discussion with @bbonaby above, I've moved these changes to [`user/dhoehna/test-rust-publishing`](https://github.com/microsoft/mxc/tree/user/dhoehna/test-rust-publishing) on `microsoft/mxc` (head `237abe6`, full history + all review feedback preserved).

A fork PR can't exercise the release pipeline (the official build and `1ES.Release` are `trigger: none` and depend on the `MXC-ESRP-Signing` variable group / ESRP service connection, none of which is exposed to forks). From the in-repo branch I'll queue `MXC-Official-Build` and then `1ES.Release` with `publishCrates: true` + `cratesDryRun: true` (stage/verify only, nothing published), and open a fresh PR from that branch once the dry run is green. Thanks for the review here!

PR #650, opened 2026-07-14
[WSLC] Reject nested deniedPaths overlaps in WSLC policy mapping

SohamDas2021:
## 📖 Description
WSLC mounts host paths as flat WSL volumes and has no overlay primitive, so a deniedPaths entry nested under a mounted
(readwrite/readonly) parent cannot be masked and would remain accessible through the parent mount. Add validate_denied_path_overlap, which rejects such configs at policy-mapping time (component-wise, case- and separator-insensitive, drive-letter aware) instead of silently leaving the subtree exposed.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/650)

### DIFF
--- a/src/backends/wslc/common/src/policy_mapping.rs
+++ b/src/backends/wslc/common/src/policy_mapping.rs
@@ -118,10 +118,15 @@ pub fn build_volume_mounts(
 /// separator does not skew the result), and lowercases each component because
 /// Windows paths are case-insensitive (`C:\Project` and `c:\project` are the
 /// same object). `C:\Project` and `C:/project/` both yield `["c:", "project"]`.
+///
+/// Uses full-Unicode [`str::to_lowercase`] (not ASCII-only) so non-ASCII case
+/// variants (`C:\Ä` vs `c:\ä`) fold together. Still best-effort: NTFS does not
+/// fold Unicode normalization forms, so precomposed vs decomposed spellings
+/// stay distinct here (as on disk) and are left to D6 object identity.
 fn normalize_path_components(path: &str) -> Vec<String> {
     path.split(['/', '\\'])
         .filter(|component| !component.is_empty())
-        .map(|component| component.to_ascii_lowercase())
+        .map(|component| component.to_lowercase())
         .collect()
 }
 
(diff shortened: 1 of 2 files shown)


src/backends/wslc/common/src/policy_mapping.rs
```diff
@@ -111,6 +111,108 @@ pub fn build_volume_mounts(
...
+/// components and does not canonicalize `..`/symlinks, so a traversal spelling
+/// (e.g. `C:\proj\sub\..\..\secrets`) will not match a mounted `C:\proj` here.
+/// It folds case (full Unicode) but not Unicode normalization forms.
+/// Object-level enforcement — an alias reaching a denied object via symlink,
+/// hard link, `..`, or an unfolded spelling — is the job of the D6
+/// object-identity pass (`normalize_object_conflicts`), which runs before this
+/// check and fails closed on unresolvable paths when `deniedPaths` are present.
+/// Treat this function as defense-in-depth for the flat-mount overlay gap, not
+/// a complete traversal-safe deny on its own.
+pub fn validate_denied_path_overlap(
```
MGudgin, line 232: the compare is textual, not path-aware

THE BUG

normalize_path_components lowercases + splits on / and \; is_strict_descendant does a
component-wise prefix compare. Nothing is canonicalized, so spellings that resolve into
a mounted tree pass the check:

1. `..` traversal (main one). Mount C:\project\sub\.. (resolves to C:\project) while
   denying C:\project\secret. Textually mount = ["c:","project","sub",".."], not a prefix
   of ["c:","project","secret"] -> accepted, but secret is exposed. Same for denying
   C:\outside\..\project\secrets against a C:\project mount.

2. Bare root mount (line 197). \ or / normalizes to an empty vec, so
   `if mounted_components.is_empty() { continue; }` skips that mount entirely; anything
   denied under it is accepted. (C:\ -> ["c:"] is fine/tested; only the drive-less root
   spelling leaks.)

3. Drive-relative (line 127). C:secrets splits into one fused token ["c:secrets"], which
   doesn't prefix-match a C:\ mount ["c:"] -> escapes.

4. Aliases: \\?\ prefixes, 8.3 short names, junctions/symlinks, trailing dots/spaces all
   differ textually but resolve into the mounted tree.

D6 does NOT cover these: normalize_object_conflicts only reconciles paths resolving to the
SAME object and only fails closed on UNRESOLVABLE paths. A mount vs a denied child are
different (parent/child) objects, and a resolvable `..` spelling isn't unresolvable - so
neither layer catches it. The doc-comment claiming D6 handles this is wrong and should be
corrected.

Bonus false-positive (Medium): stripping leading separators makes \project\x (absolute)
and project\x (relative) identical, so a valid config can be wrongly rejected.

THE FIX (one change closes cases 1-3 + the false positive)

1. Parse structurally instead of splitting strings: std::path::Path::components() on
   Windows (or typed-path for cross-build parsing) so Prefix and RootDir stay distinct -
   fixes C:secrets vs C:\ and \ vs project.

2. Lexically fold . and .. on BOTH the mounted and denied sides before comparing, so
   C:\project\sub\.. becomes C:\project. Pure lexical, no disk access.

3. Where paths exist, canonicalize (std::fs::canonicalize / GetFinalPathNameByHandle) to
   resolve 8.3/junctions/symlinks. If deniedPaths are present and a path can't be
   resolved unambiguously, FAIL CLOSED rather than falling back to the textual compare.

4. Use the same canonical form for the actual SDK mount, so you don't validate one path
   and mount another.

5. Drop the empty-components `continue`: after structural parsing a root mount is an
   ancestor of everything on that drive - treat it as such (or reject when denies present).

TESTS TO ADD

- `..` on the denied side and on the mounted side -> rejected.
- bare \ / root mount with a nested denied path -> rejected.
- drive-relative C:secrets under C:\ -> rejected.
- absolute \project\x vs relative project\x -> not a false rejection.
- integration test at the runner call site (wsl_container_runner.rs:899): runner returns
  the overlap error and starts no container. The 11 unit tests only cover the pure
  function; a wiring/ordering regression would pass them all.

SohamDas2021: Thanks for the catch!

I am fixing the lexical/structural half in this branch, will do the full alias canonicalization in a follow-up right after, hope that is alright?

MGudgin (approved):

PR #635, opened 2026-07-12
User/shschaefer/rust lib

shschaefer:
## 📖 Description
This PR begins the work necessary to create SDKs for multiple languages.  We refactor MXC so the executor binaries become thin CLI shells over a single execution library, then prove that boundary by adding a C ABI and a C# SDK that binds to it.

Previously each executor ( wxc-exec ,  lxc-exec ,  mxc-exec-mac ) carried its own  match request.containment  backend-selection logic, and the public Rust SDK ( mxc-sdk ) was a provisional streaming-only wrapper that duplicated dispatch. This PR consolidates all dispatch into one engine and layers the public surfaces on top.

New crates

•  core/mxc_engine  (internal) — the single home for "given an  ExecutionRequest , run it": run-to-completion selection over all backends (incl. the Windows ProcessContainer BaseContainer/AppContainer BFS/DACL fallback tiers and every experimental backend), streaming  spawn , state-aware lifecycle dispatch, host probing, and policy/config building. Absorbs the logic previously duplicated across the executors and the old  mxc-sdk  internals.
•  ffi/mxc_ffi  ( cdylib / staticlib ) — a flat, panic-safe C ABI over  mxc-sdk .  mxc_run(policyJson, command, out)  runs a sandbox to completion and returns captured stdout/stderr + exit outcome via a  #[repr(C)] MxcRunResult . Every entry point is  catch_unwind -wrapped; JSON in, bytes+status out. A  build.rs  runs csbindgen to generate the C# P/Invoke layer.

Refactors

•  mxc-sdk  is now a thin public facade re-exporting the engine, and gains  run()  (run-to-completion capture) alongside  spawn_sandbox()  (live streaming).
•  wxc-exec  /  lxc-exec  /  mxc-exec-mac  reduced to arg-parse + config-load + maintenance modes + delegate to  mxc_engine ; no backend  match  remains in any binary. The  ContainmentBackend → CommandLineContext  mapping moved into  wxc_common::cmdline  as the single source of truth.

New C# SDK ( csharp/ )

•  Microsoft.Mxc.Sdk  —  MxcSandbox.Run/RunAsync  →  RunResult , policy POCOs mirroring the Rust wire shape,  MxcException / ErrorCode , and a dev/NuGet native-library resolver. Plus a console sample and xUnit tests, wired into a solution.

## 🔗 References
[](https://github.com/microsoft/mxc/issues/484)
Follow phase will add streaming and state-awareness over the FFI and through the C# SDK.

## 🔍 Validation
[...]

PR #651, opened 2026-07-15
Add streaming and state-aware lifecycle to the C ABI and .NET SDK

shschaefer:
## 📖 Description
Extends the `mxc_ffi` C ABI and the `Microsoft.Mxc.Sdk` .NET binding beyond run-to-completion to cover **live streaming** and the **state-aware sandbox lifecycle** — the deferred "Phase 8" of the Rust-SDK/FFI work (follow-up to #635 and #645). The lower layers (`mxc-sdk` streaming, `mxc_engine` state-aware dispatch) already existed; this surfaces them through the C ABI and the .NET SDK.

### Streaming
- **`mxc_ffi`**: opaque-handle C ABI — `mxc_spawn` returns an `MxcSandbox*`; separate `MxcReadStream*` / `MxcWriteStream*` handles (`take_stdin`/`stdout`/`stderr`) allow reading stdout, reading stderr, and writing stdin concurrently; `mxc_stream_read`/`write`/`flush`; `mxc_sandbox_id`/`try_wait`/`wait`/`kill`/`free`.
- **.NET**: `MxcSandbox.Spawn(policy, command)` → `MxcSandboxProcess` with `Stream`-based `StandardInput`/`Output`/`Error`, `Wait`/`WaitAsync`/`WaitBlocking`/`WaitForExitWithOutputAsync`, `Kill`.

### State-aware lifecycle
- **`mxc-sdk`**: `run_state_aware_json` (envelope phases) + `exec_sandbox` (live exec → `Sandbox`).
- **`mxc_ffi`**: `mxc_state_aware` (provision/start/stop/deprovision) + `mxc_state_aware_exec` (live exec, reusing the streaming handle).
- **.NET**: `MxcLifecycle` — `ProvisionSandbox`/`StartSandbox`/`ExecInSandbox`/`ExecInSandboxAsync`/`StopSandbox`/`DeprovisionSandbox`, with a typed `SandboxId`.
- Backend-agnostic plumbing: `wxc_common::state_aware_dispatch::dispatch_state_aware_exec` returns the backend's `ExecHandle` instead of relaying to the executor's stdio, and `exec_stream::ExecSandboxProcess` adapts it to the streaming trait (duplicating pipe handles so it never shares ownership with the backend).

The generated C# P/Invoke (`NativeMethods.g.cs`) is regenerated from the three FFI source files; the `MXC_STATUS_*` space already reserved the state-aware phase codes, so no renumbering.

## Scope / known limitations
- The only in-tree state-aware backend, **IsolationSession**, is Windows-only, experimental, and needs its OS-side service. It still **relays exec internally** (returns null pipe handles), so live-exec streaming for *that* backend is a deliberate follow-up; the FFI/SDK/.NET plumbing is complete and backend-agnostic (validated with a fake backend over real OS pipes).
- The C ABI remains **not a stable external contract** — native + binding are co-versioned and generated together.

## 🔍 Validation
- Rust: `cargo build --workspace`, `cargo fmt`/`clippy -D warnings`; unit + integration tests (FFI, `exec_stream` incl. real `std::io::pipe` round trips, engine, `mxc-sdk` state-aware). Real streaming round-trips run via BaseContainer locally.
- .NET: **30/30** tests pass (host-independent, and with `MXC_E2E_HOST_PREPPED=1` for the real streaming/lifecycle paths — kill, stderr, WaitAsync cancellation, and a >64KB-both-streams deadlock guard, plus a dispose-during-read UAF regression).
- All CI gates green: `check-dotnet-bindings-codegen`, `check-dotnet-errorcode-parity`, `check-version-sync`, `check-schema-*`, `validate-configs`.

> ⚠️ Real streaming/lifecycle behavior on **Windows** runs only on a host-prepped lane (`MXC_E2E_HOST_PREPPED=1`), matching the existing `#[ignore]` ProcessContainer tests; the macOS Seatbelt lane runs its streaming suite un-gated. CI YAML/MSBuild changes are verified by inspection until the first pipeline run.

## ✅ Checklist
- [ ] Linked to an issue

## 📋 Issue Type
- [ ] Bug fix
- [x] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/651)


src/ffi/mxc_ffi/src/streaming.rs
```diff
@@ -0,0 +1,844 @@
...
+    cap: usize,
+    out_read: *mut usize,
+) -> i32 {
+    if stream.is_null() || buf.is_null() || out_read.is_null() {
+        return MXC_STATUS_NULL_ARGUMENT;
+    }
+    let status = catch_unwind(AssertUnwindSafe(|| {
+        // SAFETY: `stream` non-null live handle; `buf`/`cap` describe a valid
+        // writable region per the caller contract.
+        let s = unsafe { &mut *stream };
```
MGudgin, line 314: >let s = unsafe { &mut *stream }; [](http://example.com/codeflow?start=8&length=32)

This line takes an exclusive Rust borrow of the stream (&mut *stream), and the
same pattern is in mxc_stream_write (line 348) and mxc_stream_flush (line 373).
Rust requires a &mut to be exclusive for its whole lifetime, so if two calls into
this function run against the same stream handle at once, there are two live &mut
to one MxcReadStream, which is undefined behaviour (not merely interleaved bytes).

The C# binding does not enforce that exclusivity. MxcReadPipeStream.Read
(MxcSandboxProcess.cs:447) and MxcStdinStream.Write (MxcSandboxProcess.cs:518)
guard the native call only with SafeHandle.DangerousAddRef / DangerousRelease.
DangerousAddRef is a refcount that stops the handle from being freed while a call
is in flight; it is not a mutex and gives no mutual exclusion. Two threads can
both AddRef and both be inside mxc_stream_read on the same handle simultaneously.
The earlier use-after-free fix closed the free-vs-read race but left the
read-vs-read / write-vs-write race open.

This is reachable two ways:
1. A caller reads the same stdout/stderr stream from two threads. Here that is undefined behaviour, not just a logic bug.
2. Without any caller mistake: Wait()/WaitAsync starts an internal drain task on
   stdout/stderr and stores that stream, and a later StandardOutput / StandardError
   access hands the same stream back to the caller. The internal drain and the
   caller's read then run on one handle concurrently.

Also note the module doc above claims the C# binding enforces the single-owner
contract by refcounting the handle via SafeHandle. That is true for free ordering
only, not for concurrent read/write, so the documented safety contract is not
actually upheld.

Suggested fix:
- Add a per-handle lock around the native read/write/flush call on the C# side,
  held only across the P/Invoke (no lock held during unrelated work, so no
  deadlock). This makes the boundary sound even under caller misuse.
- Fix the stream state machine so a stream being drained internally by
  Wait()/WaitAsync is never handed back out via StandardOutput / StandardError
  (track untaken vs caller-owned vs internally-draining). This removes the
  SDK-induced concurrent-access path.
- Update the streaming.rs module doc so the stated single-owner contract matches
  what the binding actually guarantees.

Refs: streaming.rs:311/348/373 (the &mut *stream borrows);
MxcSandboxProcess.cs:447 (MxcReadPipeStream.Read), :518 (MxcStdinStream.Write),
and the Wait()/WaitAsync drain-then-hand-out path.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -0,0 +1,581 @@
...
+
+            int status;
+            int exit = 0;
+            int running = 1;
+            lock (_controlLock)
+            {
+                ThrowIfDisposed();
+                unsafe
+                {
+                    status = NativeMethods.mxc_sandbox_try_wait(_handle.Ptr, &exit, &running);
```
MGudgin, line 221: The comment on lines 199-203 assumes the policy timeout is enforced natively
("killing the tree, which try_wait then sees as an exit"). That isn't true for
the spawn_sandbox path: nothing kills the child at the deadline unless
mxc_sandbox_wait is called, and WaitCore only ever calls mxc_sandbox_try_wait
(line 186). So Wait()/WaitAsync()/WaitForExitWithOutputAsync() never enforce
SandboxPolicy.TimeoutMs and can hang indefinitely.

The deadline+kill live only in SandboxProcess::wait (WaitForSingleObject(process,
timeout_ms) then kill on WAIT_TIMEOUT: appcontainer_runner.rs:1405-1416,
base_container_runner.rs:1422-1433; wait_with_timeout on bwrap/seatbelt:
bwrap_runner.rs:445). try_wait is a pure exited/still-running check with no
deadline (appcontainer_runner.rs:1378, base_container_runner.rs:1389), and
spawn_sandbox starts no watchdog (mxc-sdk/src/lib.rs:96) -- so on the polling
path the deadline is never consulted.

Worked example:
  var policy = new SandboxPolicy { TimeoutMs = 5000 };
  using var proc = MxcSandbox.Spawn(policy, "ping -n 61 127.0.0.1"); // ~60s
  var result = proc.Wait();   // caller expects a return in ~5s
WaitCore loops on mxc_sandbox_try_wait, which reports "still running" the whole
time; at t=5s nothing kills the child, so Wait() returns only when the command
exits on its own (~60s) with TimedOut=false, or never returns if it never exits.
The sandbox tree stays alive the entire time.

The same policy is honored by WaitBlocking() (mxc_sandbox_wait -> kill at 5s ->
TimedOut=true), but that holds _controlLock for the whole blocking wait so it
can't be interrupted by a concurrent Kill() (lines 213-215). So no wait API today
gives both timeout enforcement and kill-responsiveness, and Wait() contradicts
its own summary on lines 153-155 ("honouring the policy's TimeoutMs").

Suggested fix: keep the try_wait poll loop for kill-responsiveness, but enforce
the deadline in managed code -- thread TimeoutMs into MxcSandboxProcess and, once
the deadline passes in the WaitCore loop, call mxc_sandbox_kill and return. That
stays lock-free across the native call, so it doesn't reintroduce the
WaitBlocking kill-deadlock. Update the comment on lines 199-203 and the Wait()
summary on lines 153-155 accordingly.

Refs: MxcSandboxProcess.cs:159 (Wait) / :166 (WaitAsync) / :169 (WaitCore) / :186
(try_wait poll) / :196 (TimedOut=false) / :199-203 / :217-236 (WaitBlocking);
backends appcontainer_runner.rs:1378,1405-1416,
base_container_runner.rs:1389,1422-1433, bwrap_runner.rs:445; spawn_sandbox
mxc-sdk/src/lib.rs:96.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -0,0 +1,581 @@
...
+            var stderr = await stderrTask.ConfigureAwait(false);
+            return (result, stdout, stderr);
+        }
+        catch
+        {
+            // On cancellation/failure, still observe the reader tasks so they do
+            // not run detached. The stream handles are refcounted, so this is
+            // safe regardless of when the caller disposes.
+            await Task.WhenAll(SwallowAsync(stdoutTask), SwallowAsync(stderrTask))
+                .ConfigureAwait(false);
```
MGudgin, line 312: On cancellation this catch awaits the two reader tasks before rethrowing, but
those reads can never complete, so the wait deadlocks and the child is never
killed.

The readers (stdoutTask/stderrTask, started at lines 253-254) are parked in the
blocking mxc_stream_read on the child's stdout/stderr. MxcReadPipeStream has no
cancellable read -- ReadToEndAsync uses CopyToAsync(ms, ct) (line 288), but the
token only gates between reads; a read already in flight in native code is not
interruptible. For a still-running child that hasn't closed its pipes, those
reads never return, so Task.WhenAll on line 267 blocks forever.

The only thing that would unblock them is disposing/killing the child (frees the
handle -> kills the tree -> EOFs the pipes). But in ExecInSandboxAsync the dispose
is in the finally (MxcLifecycle.cs:173), which is unreachable because this catch
never returns. So the code blocks awaiting the readers before it can run the one
step that would release them -- a permanent deadlock. Net effect:
ExecInSandboxAsync hangs, and the native sandbox container + handles leak (never
disposed or deprovisioned).

The comment on lines 264-266 ("the stream handles are refcounted, so this is
safe regardless of when the caller disposes") is about a different property:
refcounting prevents a use-after-free crash, not the deadlock. The code is
memory-safe and permanently hung at the same time.

Worked example:
  using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
  var r = await MxcLifecycle.ExecInSandboxAsync(id, "ping -n 600 127.0.0.1",
                                                cts.Token);
  // t=5s: WaitAsync throws OCE -> this catch awaits the parked reads -> hangs
  // forever; ExecInSandboxAsync's finally/Dispose never runs; sandbox leaks.

Suggested fix: on cancellation/failure, kill (or dispose) the child before
awaiting the readers, so the pipes EOF and the reads return. e.g. in this catch,
call Kill() (mxc_sandbox_kill) first, then await Task.WhenAll(...), then throw;
or restructure ExecInSandboxAsync to Kill/Dispose on cancellation before the
await unwinds. Optionally also give MxcReadPipeStream a genuinely cancellable
read so the token can abandon a parked read without killing. Update the comment
on lines 264-266 to speak to liveness, not just refcount safety.

Refs: MxcSandboxProcess.cs:244 (WaitForExitWithOutputAsync) / :253-254 (reader
tasks) / :262-269 (catch) / :285-290 (ReadToEndAsync, non-cancellable native
read); MxcLifecycle.cs:148 (ExecInSandboxAsync) / :160 (await) / :171-174
(finally/Dispose).

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -0,0 +1,270 @@
...
+
+    /// <summary>Start a provisioned sandbox.</summary>
+    /// <exception cref="MxcException">Starting failed.</exception>
+    public static void StartSandbox(SandboxId id, StartSandboxOptions? options = null)
+    {
+        var envelope = NewEnvelope("start");
+        envelope["sandboxId"] = id.Value;
+        if (options?.Size is { } size)
+        {
+            SetBackendConfig(envelope, "start", "size", size);
```
MGudgin: >SetBackendConfig(envelope, "start", "size", size); [](http://example.com/codeflow?start=12&length=50)

This emits the size under experimental.isolationSession.start.size, but the Rust
wire model never reads that key, so the value is silently dropped.

The start phase deserializes into IsolationSessionPhase (wire.rs), whose only
fields are configuration_id ("configurationId") and user -- there is no "size"
field and no #[serde(alias = "size")]. The experimental block is intentionally
permissive (no deny_unknown_fields), so "size" is an unknown field that serde
silently ignores; configuration_id stays None and the backend falls back to its
default (Composable). The one-shot path and every other phase use configurationId,
so "start" is the only place using the wrong key.

Note the field name is misleading: configurationId is not an opaque id -- it is
typed as the sizing-profile enum IsolationConfigurationId
(small/medium/large/composable), the same value space StartSandboxOptions.Size
carries. So value-wise Size maps directly onto configurationId; only the wire key
differs. The configurationId-vs-Size naming mismatch is almost certainly what led
to emitting "size" here.

Effect: every StartSandboxOptions.Size value, including valid ones like "medium"
or "large", is discarded. The size profile is never applied and no error or
warning is raised.

Also, the XML doc on StartSandboxOptions.Size (StateAwareTypes.cs:37-40) says
"Unknown values are warned and downgraded to composable on the native side."
That does not happen: the value never reaches the native validator (wrong key),
and configurationId is a typed enum, so an unknown value under the correct key
would be a deserialization error, not a warn-and-downgrade. The doc describes
behavior the code doesn't implement.

Fix: emit "configurationId" instead of "size" (SetBackendConfig(envelope,
"start", "configurationId", size)), and add a test asserting a non-default size
reaches IsolationSessionConfig.configuration_id (e.g. start with Size="large" and
assert the parsed phase config is Large, not the Composable default). To make the
intuitive key also work and prevent a recurrence, consider adding
#[serde(alias = "size")] to configuration_id, or renaming the wire field to
size/sizeProfile so the C# Size and the wire key line up. Either way, reconcile
the StartSandboxOptions.Size doc with the actual enum-validation behavior.

Refs: MxcLifecycle.cs:88-90 (emits "size"); wire.rs IsolationSessionPhase
(configuration_id / "configurationId", typed as IsolationConfigurationId
size enum; no size field; experimental block has no deny_unknown_fields);
StateAwareTypes.cs:37-41 (Size doc + string type).

src/ffi/mxc_ffi/src/streaming.rs
shschaefer: Fixed.  Add per-handle lock and tri-state draining guard.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
shschaefer: Fixec with managed deadline -> native blocking wait.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
shschaefer: Fixed.  Kill before awaiting readers.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
shschaefer: Fixed.  Emit configurationId.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -0,0 +1,682 @@
...
+            {
+                return new SandboxWaitResult { ExitCode = exit, TimedOut = false };
+            }
+
+            // try_wait only reports exited / still-running, never a timeout, and
+            // spawn starts no native watchdog — so once the policy deadline passes
+            // we hand off to the native blocking wait, which enforces the same
+            // deadline by killing the tree and reporting the timeout distinctly.
+            // We are already at/past the deadline, so it returns promptly.
+            if (_timeoutMs is { } timeoutMs &&
```
MGudgin, line 239: Low/Medium: the new timeout enforcement doesn't define what TimeoutMs == 0 means
on the .NET side, and the current handling turns it into a permanent Kill() hang.

Impedance mismatch: SandboxPolicy.TimeoutMs is documented as "null = no timeout"
(SandboxPolicy.cs:31), and 0 is not a documented .NET value. But 0 is the
wait-forever sentinel everywhere below the SDK: the domain default is 0
(config_parser.rs:657, process.timeout.unwrap_or(0)), the trait contract says
"0 means wait forever" (sandbox_process.rs:105-106), Windows maps it to
u32::MAX via get_timeout_milliseconds (script_runner.rs:52-56), and Unix maps it
to None (bwrap_runner.rs:307). So a caller porting from config/JSON semantics, or
one who reads 0 as "no limit", can reasonably pass 0 -- and nothing validates it
(TimeoutMs is an unguarded uint?).

When _timeoutMs == 0, the guard on lines 239-240:

    if (_timeoutMs is { } timeoutMs &&
        Stopwatch.GetElapsedTime(_startTimestamp).TotalMilliseconds >= timeoutMs)
        return WaitBlocking();

is true on the first poll iteration (elapsed >= 0), so WaitCore hands off to
WaitBlocking() immediately. WaitBlocking() holds _controlLock for the whole
native blocking wait (line 261), and with scriptTimeout == 0 that wait is
unbounded -- so a concurrent Kill() (which also needs _controlLock) can never
acquire it and hangs. A conforming caller using null is unaffected (the guard is
skipped), so this only bites the undocumented 0 input.

Fix -- decide what .NET TimeoutMs == 0 means and make it consistent:
- If 0 means "wait forever" (matching the wire): guard with timeoutMs > 0 so the
  loop just keeps polling try_wait (Kill-responsive), and document 0 = wait
  forever on SandboxPolicy.TimeoutMs.
- If 0 is invalid on the .NET surface: reject it at the entry point (Spawn /
  policy validation) with a clear ArgumentException, so it never reaches this
  path.
  Either way, add a test: spawn a long-running child with TimeoutMs = 0 and assert
Kill() from another thread unblocks Wait() promptly (the existing
Wait_EnforcesPolicyTimeout uses 1000, so 0 is uncovered).

Refs: MxcSandboxProcess.cs:239-242 (deadline guard) / :258-261 (WaitBlocking holds
_controlLock) / :87-90 (_timeoutMs, no clamp); SandboxPolicy.cs:31 (doc: null = no
timeout, 0 undocumented); config_parser.rs:657, sandbox_process.rs:105-106,
script_runner.rs:52-56, bwrap_runner.rs:307 (0 = wait forever below the SDK).

MGudgin (approved):

PR #657, opened 2026-07-16
[WSLC] Add on-disk alias canonicalization to denied-path overlap check

SohamDas2021:
## 📖 Description
The WSLC denied-path overlap validator was lexical-only: it folded case and `.`/`..` but could not see on-disk aliases (symlinks, junctions, 8.3 short names, `\\?\` prefixes), so a deny that resolved into a mounted tree via an alias slipped through and stayed accessible.

Added a shared `wxc_common::filesystem_canonical` primitive that resolves a path to its canonical DOS form via GetFinalPathNameByHandleW, and wire it into `validate_denied_path_overlap` as a second tier: after the lexical pre-check, canonicalize every mount and deny, fail closed on any path that exists but cannot be resolved, then re-run the structural
containment compares on the resolved forms.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.


src/core/wxc_common/src/filesystem_canonical.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// Like [`canonicalize_path`] but tolerates a not-yet-created leaf: when the
+/// full path is missing it resolves the deepest existing ancestor (collapsing
+/// its aliases) and re-appends the missing components. This lets callers compare
+/// a denied path that does not exist yet but whose parent is an alias
+/// (symlink/junction) into a mounted tree. Mirrors the bubblewrap runner's
+/// `resolve_through_symlinks`. Returns [`PathCanonical::Absent`] only when no
+/// ancestor resolves.
+#[cfg(windows)]
+pub fn canonicalize_allowing_absent_tail(path: &str) -> PathCanonical {
```
MGudgin, line 135: The helper handles nonexistent paths by finding the deepest existing ancestor and appending the missing components. While walking backward,  file_name()  returns nothing for  .. , so the algorithm can forget that a preceding component should be cancelled.

For example:

<existing>\X\Y\..\Z

should normalize to:

<existing>\X\Z

but can be reconstructed as:

<existing>\X\Y\Z

This matters when  <existing>  is a junction or other alias. Tier 1 folds  ..  correctly but cannot resolve the alias; Tier 2 resolves the alias but then reconstructs the missing tail incorrectly. It can consequently compare the wrong path and miss a real denied-path overlap.

The fix is to normalize  .  and  ..  before searching for an existing ancestor, or explicitly track parent-component cancellations while building the missing tail. A regression test should combine an aliased ancestor, multiple nonexistent components and  .. .

src/backends/wslc/common/src/policy_mapping.rs
```diff
@@ -215,45 +218,129 @@ impl NormalizedPath {
...
+/// Tier 2 does not fold Unicode normalization forms and compares path endpoints,
+/// not object identity: a hard link inside a mounted tree pointing at a denied
+/// file resolves to its own in-tree name (not the denied one), so it is not
+/// caught here or by D6 (the two are distinct objects). Creating such a link
+/// requires write access to the denied target the guest does not have, so this
+/// is out of scope under the trusted-author threat model. A residual TOCTOU
+/// window also remains between canonicalization and the SDK mount (an alias
+/// could be swapped in between); fully closing it needs handle-based mounting
+/// the WSLC SDK does not expose, so it is likewise accepted.
 pub fn validate_denied_path_overlap(
```
MGudgin, line 265: This PR canonicalizes only the paths explicitly listed in the policy. It catches aliases in the mount or deny strings themselves, but not aliases located inside a mounted directory.

Example:

Mounted: C:\project
Denied:  C:\secrets
Junction: C:\project\link -> C:\secrets

Canonicalizing the two policy entries produces  C:\project  and  C:\secrets , which do not overlap, so validation succeeds. If WSLC follows the junction inside the mounted tree, the guest can still reach the denied directory through  C:\project\link .

The immediate concern is that the new documentation says canonicalization is “closing the alias gap,” while this gap remains. At minimum, the limitation should be documented precisely. Fully enforcing it would require controlling traversal beneath the mount or inspecting reparse points, which is harder, potentially expensive and still subject to races.

MGudgin (dismissed):

src/backends/wslc/common/src/policy_mapping.rs
SohamDas2021: Thanks for the catch! I will document this accordingly.

src/core/wxc_common/src/filesystem_canonical.rs
SohamDas2021: Thanks, fixing!

SohamDas2021:

src/backends/wslc/common/src/policy_mapping.rs
```diff
@@ -215,45 +218,140 @@ impl NormalizedPath {
...
+        .iter()
+        .map(|(m, list)| (*m, *list, resolve(m)))
+        .collect();
+
+    for (path, canon) in denied_canon
+        .iter()
+        .map(|(p, c)| (*p, c))
+        .chain(mount_canon.iter().map(|(p, _, c)| (*p, c)))
+    {
+        if matches!(canon, PathCanonical::Unknown) {
```
huzaifa-d, line 342: This rejects the whole config if any denied path is Unknown. A mount on a different drive that exists but is not openable by the wxc-exec user (access denied resolves to Unknown) would block a config whose deniedPaths target a completely separate location. Expected?

src/backends/wslc/common/src/policy_mapping.rs
```diff
@@ -215,45 +218,140 @@ impl NormalizedPath {
...
+    for (denied, denied_canon) in &denied_canon {
+        let PathCanonical::Canonical(denied_resolved) = denied_canon else {
+            continue; // Absent: no on-disk ancestor resolved, so nothing to alias.
+        };
+        let denied_norm = NormalizedPath::parse(denied_resolved);
+        for (mounted, list_name, mount_canon) in &mount_canon {
+            let PathCanonical::Canonical(mount_resolved) = mount_canon else {
+                continue;
+            };
+            if NormalizedPath::parse(mount_resolved).contains_strictly(&denied_norm) {
```
huzaifa-d, line 361: An alias that resolves to exactly the mount root is not caught here, because contains_strictly requires the child to be strictly deeper than the parent.

src/core/wxc_common/src/filesystem_canonical.rs
```diff
@@ -0,0 +1,339 @@
...
+            FILE_FLAG_BACKUP_SEMANTICS,
+            None,
+        )
+    };
+    let handle = match handle {
+        Ok(h) if !h.is_invalid() => OwnedHandle(h),
+        _ => {
+            // Distinguish a cleanly-missing path from an unexaminable one.
+            // SAFETY: reads the thread-local last error set by the failed call.
+            let err = unsafe { GetLastError() };
```
huzaifa-d: After CreateFileW fails, the code calls GetLastError again instead of just reading the code from the returned windows error.

SohamDas2021: Fixing.

src/backends/wslc/common/src/policy_mapping.rs
SohamDas2021: Correct, the contains_strictly returns false on an exact match, that's intentional, and the exact-object-alias case is already closed upstream, so it isn't a gap

src/backends/wslc/common/src/policy_mapping.rs
SohamDas2021: Intended. Unknown means "exists/maybe-exists but its real location couldn't be determined", so the drive anchor can't be trusted to prove non-overlap: a subst drive, junction, or symlink can make a D:\…  mount actually point into a denied C:\…  subtree. Narrowing the fail-closed by drive/prefix would reopen exactly the alias bypass this tier exists to close. It also mirrors D6's own fail-closed, keeping the two layers consistent.

huzaifa-d (approved):

huzaifa-d (approved):

PR #658, opened 2026-07-17
Harden the Windows Sandbox state-aware daemon and lifecycle

MGudgin:
This PR hardens the Windows Sandbox state-aware daemon and lifecycle: it makes
daemon diagnostics observable, bounds a frozen guest with a cancellation-safe
exec watchdog, reports VM-slot contention coherently, and versions the
client/daemon IPC wire protocol behind a bilateral, daemon-enforced handshake.
Each change is backend-only, with no SDK or wire-schema impact.

**Details**

* **Daemon diagnostics are captured.** The daemon emits ~40 `[wsb-daemon]`
  diagnostics, and the VM-owning detached process is exactly where boot /
  teardown / orphan-reclaim failures need to be observable. `spawn_daemon`
  redirects the daemon's stderr to `daemon.log` in the per-sandbox owner-only
  record directory (append), and the start-timeout error names the concrete log
  path.
* **The streaming exec path has a cancellation-safe host-side watchdog.** A
  frozen-but-alive guest (the VM crash watchdog only fires once the VM processes
  are *gone*) would otherwise wedge the single-flight slot indefinitely. The
  watchdog deadline lives inside `stream_exec_on_guest`'s `select!` loop and is
  honoured only at frame boundaries, so a timeout can never truncate a data
  frame and desync the client's decoder. On expiry the daemon marks the guest
  slot Unusable and sends the client a clean synthetic timeout exit frame (the
  client reads with no deadline once streaming starts, so it is freed too).
* **VM-slot contention is reported coherently.** `HostVmLock::try_acquire`
  distinguishes genuine contention (the wait timed out) from a real create/wait
  failure: on contention the daemon exits with a distinct
  `DAEMON_EXIT_VM_SLOT_BUSY` code that `start` maps to `backend_unavailable`,
  while a real mutex failure propagates as an opaque backend error (with
  daemon.log diagnostics) rather than being mis-reported as busy. (`start` does
  not hold the slot itself — the daemon must own it for the VM's lifetime, so
  holding it across the spawn would deadlock; the distinct-exit-code mapping
  achieves the same observable outcome without that hazard.)
* **The IPC wire protocol is versioned behind a bilateral, daemon-enforced
  handshake.** The client declares its version as the trailing token of the
  `EXEC` line (`EXEC <nonce> <proto>`); the daemon refuses a mismatched or
  malformed version with `ERR protocol` before any frame I/O, so a
  differently-framed request is never parsed. An absent token maps to
  `LEGACY_IPC_PROTOCOL_VERSION`, so a future daemon cleanly refuses a legacy
  client instead of misframing it. `DaemonRecord` also carries a
  `protocol_version` (added with `#[serde(default)]` so a pre-versioning record
  reads back as incompatible) as a client-side fast-path (`ensure_daemon_protocol`).
  An orphaned daemon from a prior install authenticates (its nonce is in the
  record) but is refused when its wire version differs. `stop`/`deprovision`
  keep using the stable single-line `STOP` command (explicit `OK` ack) so an
  orphaned VM can always be reclaimed.
* Docs: `docs/windows-sandbox/windows-sandbox-reference.md` covers the
  daemon.log sink, the exec watchdog, the VM-slot busy mapping, and the
  wire-protocol handshake.

**Tests**

* Unit tests: `protocol_compatible` and a serde-default case (a pre-versioning
  record deserialises and reads back protocol-incompatible), `parse_client_protocol`,
  `ensure_daemon_protocol` accept/refuse, the `vm_slot_busy` exit-code
  classifier, a two-thread `named_mutex_try_acquire` test asserting a contended
  mutex returns `Ok(None)` (busy) rather than `Err`, the daemon `EXEC`
  matching/mismatched/malformed-version handshake, the `ERR protocol` status
  mapping to `backend_unavailable`, and a watchdog frame-boundary regression
  asserting no partial data frame precedes the terminal frame.
* `cargo fmt --all -- --check` and `cargo clippy` (affected crates,
  `-D warnings`): clean.
* `cargo test` (aarch64-pc-windows-msvc): windows_sandbox_lifecycle 180,
  wxc_windows_sandbox_daemon 27 — all pass.

### DIFF
--- a/src/backends/windows_sandbox/daemon/src/control_server.rs
+++ b/src/backends/windows_sandbox/daemon/src/control_server.rs
@@ -11,13 +11,15 @@ use anyhow::{Context, Result};
 use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt, BufReader};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::{Mutex, MutexGuard, Notify, Semaphore};
-use tokio::time::timeout;
+use tokio::time::{timeout, Instant};
 
 use windows_sandbox_lifecycle::bridge::{
-    reconnect_data_streams, stream_exec_on_guest, write_exit_frame, GuestConnection,
+    host_watchdog_deadline, reconnect_data_streams, stream_exec_on_guest, write_exit_frame,
+    GuestConnection, HOST_WATCHDOG_GRACE,
 };
 use windows_sandbox_lifecycle::control_plane::{
-    IPC_ERR, IPC_ERR_BUSY, IPC_ERR_NOT_READY, IPC_EXEC, IPC_PING, IPC_STOP,
+    parse_client_protocol, IPC_ERR, IPC_ERR_BUSY, IPC_ERR_NOT_READY, IPC_ERR_PROTOCOL, IPC_EXEC,
+    IPC_PING, IPC_PROTOCOL_VERSION, IPC_STOP,
 };
 use windows_sandbox_lifecycle::ipc_exec;
 
@@ -219,9 +221,13 @@ async fn handle_client(
     };
 
     let trimmed = line.trim();
-    let mut parts = trimmed.splitn(2, ' ');
+    // `EXEC` may carry a trailing IPC-protocol-version token: split into up to
+    // three parts (verb, nonce, optional version). The nonce is a space-free
+    // uuid, so it is always exactly the second token.
+    let mut parts = trimmed.splitn(3, ' ');
     let verb = parts.next().unwrap_or("");
     let supplied = parts.next().unwrap_or("");
+    let proto_token = parts.next().unwrap_or("");
 
     if supplied != nonce {
         writer.write_all(b"ERR auth\n").await.ok();
@@ -239,7 +245,7 @@ async fn handle_client(
         shutdown.notify_one();
         Ok(())
     } else if verb == IPC_EXEC {
-        handle_exec(reader, writer, guest, guest_nonce, permit).await
+        handle_exec(reader, writer, guest, guest_nonce, permit, proto_token).await
     } else {
         writer.write_all(b"ERR unknown command\n").await.ok();
         Ok(())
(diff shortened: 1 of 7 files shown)


src/backends/windows_sandbox/daemon/src/main.rs
```diff
@@ -99,8 +105,23 @@ async fn main() -> Result<()> {
...
+            // nothing to clean up before exiting.
+            eprintln!(
+                "[wsb-daemon] host Windows Sandbox VM slot is busy (held by another VM owner); \
+                 exiting {} after waiting {:?}: {e:#}",
+                control_plane::DAEMON_EXIT_VM_SLOT_BUSY,
+                HOST_VM_LOCK_TIMEOUT
+            );
+            std::process::exit(control_plane::DAEMON_EXIT_VM_SLOT_BUSY);
+        }
+    };
```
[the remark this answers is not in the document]
MGudgin: Good catch — fixed in afa7b0e9. Added `HostVmLock::try_acquire`, backed by a new `named_mutex_try_acquire` that distinguishes the three outcomes: `Ok(Some)` acquired, `Ok(None)` when `WaitForSingleObject` returns `WAIT_TIMEOUT` (genuine contention), and `Err` for a `CreateMutexW` failure or an unexpected wait result (e.g. `WAIT_FAILED`). The daemon now maps only `Ok(None)` to `DAEMON_EXIT_VM_SLOT_BUSY`; a real failure propagates via `?` and surfaces as an opaque backend error with daemon.log diagnostics, exactly as you suggested. Covered by a new two-thread unit test asserting a contended mutex yields `Ok(None)`, not `Err`.

src/backends/windows_sandbox/lifecycle/src/control_plane.rs
```diff
@@ -93,11 +105,24 @@ pub struct DaemonRecord {
     pub active_sandbox_id: String,
     /// `false` while the daemon is booting; `true` once the VM is ready.
     pub ready: bool,
+    /// IPC wire-protocol version this daemon speaks (see [`IPC_PROTOCOL_VERSION`]).
+    /// Absent (pre-versioning) records default to `0`, i.e. incompatible.
+    #[serde(default)]
```
huzaifa-d, line 124: The new protocol_version check only protects new clients. An older MXC binary ignores this unknown field, still accepts schema_version 1, and can send v1 frames to a future v2 daemon.

src/backends/windows_sandbox/daemon/src/control_server.rs
```diff
@@ -331,7 +332,39 @@ async fn handle_exec(
...
     let exec_id = format!("exec-{}", EXEC_COUNTER.fetch_add(1, Ordering::Relaxed));
-    match stream_exec_on_guest(&mut conn, &exec_id, &req, &mut reader, &mut writer).await {
+
+    // Host-side watchdog: a frozen-but-alive guest (never sends Exit, never
+    // closes its streams) would otherwise wedge the single-flight slot forever.
+    // `None` == infinite budget (`timeout_ms == u32::MAX`), as in one-shot.
+    let watchdog = host_watchdog_deadline(req.timeout_ms, HOST_WATCHDOG_GRACE);
+    let stream_fut = stream_exec_on_guest(&mut conn, &exec_id, &req, &mut reader, &mut writer);
+    let stream_result = match watchdog {
+        Some(deadline) => match timeout(deadline, stream_fut).await {
```
huzaifa-d: The outer timeout can cancel stream_exec_on_guest after write_data_frame_split has written a frame header but before its payload completes. Writing an Exit frame afterward appends it to a partial frame, so the client can receive corrupted output or a framing error instead of the watchdog timeout.

huzaifa-d (approved):

PR #661, opened 2026-07-20
Consume the Windows Learning Mode trace API: FFI loader + capture lifecycle

richiemsft:
## 📖 Description

Adds the Windows runtime FFI and RAII lifecycle for **consuming the new OS Learning Mode
trace API** exported by `processmodel.dll` — the foundation for the upcoming
`captureDenials` feature (capture a sandboxed process's access denials to an ETL for
later analysis).

This PR is **foundation only**; it does not yet wire any runner. It introduces a new
crate, `learning_mode_windows` (`src/backends/learning_mode/windows`):

- **Trace FFI** (`ffi.rs`) — `LearningModeApi` loads/resolves `StartLearningModeTrace` /
  `StopLearningModeTrace` via `LoadLibraryExW(SYSTEM32)` + `GetProcAddress`, with a
  `is_learning_mode_api_available()` capability probe. Mirrors the existing
  `Experimental_CreateProcessInSandbox` loader pattern.
- **2-phase security-environment FFI** (`secenv.rs`) — `SecurityEnvironmentApi`
  (`CreateProcessSecurityEnvironment` / `CreateProcessAsUserInsideSecurityEnvironment` /
  `CloseProcessSecurityEnvironment`) that produces the `HPROCESS_SECURITY_ENVIRONMENT`
  handle `StartLearningModeTrace` keys on, plus a per-export diagnostic probe with an
  `Experimental_`-prefixed fallback.
- **RAII lifecycle** (`lifecycle.rs`) — `CaptureSession` sequences create-env →
  start-trace → *[caller launches child]* → stop-trace → close-env in the required order,
  with best-effort `Drop` teardown (discard trace + close env) on the early-exit/unwind
  path.
- **Validation examples** — `lm_probe` (export-resolution probe) and `lm_capture` (drives
  the entire lifecycle against a real child and asserts an ETL is produced).

The API surface is resolved at runtime and degrades cleanly on OS builds that lack it
(the probe returns `false`, `load()` returns a typed error). The crate compiles on all
platforms (non-Windows is a stub).

**Not in this PR (follow-ups):** the `captureDenials` config, the production runner path
in `appcontainer_common`, and ETL parsing. The runner wiring will reuse BaseContainer's
FlatBuffer `sandbox_spec` builder and swap the one-shot launch for `CaptureSession` +
`launch_fn`; it depends on the `captureDenials` config landing first.

## 🔗 References

Part of the Windows learning-mode / `captureDenials` effort. Follow-up PRs will add the
[...]

PR #663, opened 2026-07-20
Add processContainer.captureDenials config for Windows denial capture

richiemsft:

-----

## Summary

Adds a `captureDenials` object under `processContainer` that will enable the
Windows learning-mode trace. This is the **config-contract** half of the
learning-mode denial-capture feature — pure plumbing, no runtime behavior yet.

**Shape** — `processContainer.captureDenials: { mode?: "block" | "allow", outputPath?: string }`:
- Presence of the object enables capture.
- `mode` (optional) selects how each ungranted access check is handled while it
  is recorded. Both modes log every access the policy does not grant; the mode
  only decides whether that access is blocked or allowed:
  - `block` (**default**) — the access stays **denied** and the denial
    is recorded. Deny-by-default containment is preserved; the safe default.
  - `allow` — the access is **allowed** and recorded (audit mode). This
    relaxes deny-by-default for the run, so it is security-sensitive and the
    runner emits a security warning.
  - Omitting `mode` defaults to `block`. An unknown value is rejected at
    parse time with an error naming both valid modes.
- `outputPath` (optional) names where the ETL trace is sealed. The OS model is
  caller-controlled (`StopLearningModeTrace(trace, lpOutputPath)`), so the
  caller names the path; it is opened under the caller''s own identity. When
  omitted, the runner will fall back to a managed per-run temp file.
- The path is validated at parse time (must be absolute with an existing parent
  directory) with actionable errors.

**What''s in scope (this PR):**
- Wire model (`wire::CaptureDenials` + `wire::CaptureDenialsMode` on `wire::ProcessContainer`).
- Domain model (`CaptureDenialsConfig` + `CaptureDenialsMode` on `ContainerPolicy.capture_denials`).
- Parser mapping + validation in `config_parser.rs`.
- Regenerated dev schema (`schemas/dev/mxc-config.schema.0.8.0-dev.json`) and SDK
  wire types (`sdk/node/src/generated/wire.ts`) via `mxc_schema_gen` (codegen —
  not hand-edited).
- Docs (`docs/schema.md`) + example (`tests/examples/29_capture_denials.json`).
- Unit tests (parse/validate/mode) + all schema/version/codegen gates green.

**What''s intentionally NOT here (follow-up integration PR):**
- Auto-injection of the `learningModeLogging` capability when `captureDenials`
  is set — that reuses the capability recognition from the capabilities PR.
- The runner trace lifecycle (Start -> run -> Stop -> write ETL) and the
  block-vs-allow enforcement behind each `mode` — that consumes the
  learning-mode OS API.

Both land in the integration PR stacked on top of the capabilities + consume-API
PRs plus this one.

### Related Issues

- #622 (learning-mode capabilities — provides `learningModeLogging` recognition)
- #661 (consume the learning-mode OS API — the capture engine)

### Validation

- `cargo test -p wxc_common` — 466 pass (9 `captureDenials` tests, incl. 4 `mode` tests)
- `cargo fmt --all -- --check`, `cargo clippy -p wxc_common --all-targets -- -D warnings`
- `cargo check --workspace --all-targets`
- `npm run test:unit` (SDK) — 201 pass
- `node scripts/versioning/{validate-configs,check-schema-codegen,check-sdk-types-codegen,check-schema-versions}.js`

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/663)

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -61,8 +61,9 @@ production configs and the dev schema when working on experimental features:
         "leastPrivilege": false,
         "capabilities": ["internetClient"],
         "captureDenials": {                // Windows-only: record the process's access
-            "outputPath": "C:\\logs\\denials.etl"  // attempts to a learning-mode ETL trace.
-        }                                  // Omit outputPath for a managed per-run temp file.
+            "outputPath": "C:\\logs\\denials.etl"  // attempts to a learning-mode ETL trace. The
+        }                                  // parent dir must already exist; omit outputPath
+                                           // for a managed per-run temp file.
     },
 
     "lxc": {                               // LXC-specific
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -61,6 +61,10 @@ production configs and the dev schema when working on experimental features:
         "leastPrivilege": false,
         "capabilities": ["internetClient"],
         "captureDenials": {                // Windows-only: record the process's access
+            "mode": "block-and-log",       // "block-and-log" (default): access stays denied and
+                                           // is logged (deny-by-default preserved). "allow-and-log":
+                                           // access is allowed and logged (audit; relaxes
+                                           // deny-by-default, emits a security warning).
             "outputPath": "C:\\logs\\denials.etl"  // attempts to a learning-mode ETL trace. The
         }                                  // parent dir must already exist; omit outputPath
                                            // for a managed per-run temp file.
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -59,7 +59,10 @@ production configs and the dev schema when working on experimental features:
 
     "processContainer": {                  // Process-based container-specific
         "leastPrivilege": false,
-        "capabilities": ["internetClient"]
+        "capabilities": ["internetClient"],
+        "captureDenials": {                // Windows-only: record the process's access
+            "outputPath": "C:\\logs\\denials.etl"  // attempts to a learning-mode ETL trace.
+        }                                  // Omit outputPath for a managed per-run temp file.
     },
 
     "lxc": {                               // LXC-specific
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -61,8 +61,8 @@ production configs and the dev schema when working on experimental features:
         "leastPrivilege": false,
         "capabilities": ["internetClient"],
         "captureDenials": {                // Windows-only: record the process's access
-            "mode": "block-and-log",       // "block-and-log" (default): access stays denied and
-                                           // is logged (deny-by-default preserved). "allow-and-log":
+            "mode": "block",               // "block" (default): access stays denied and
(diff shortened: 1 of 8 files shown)


docs/schema.md
```diff
@@ -59,7 +59,15 @@ production configs and the dev schema when working on experimental features:
 
     "processContainer": {                  // Process-based container-specific
         "leastPrivilege": false,
-        "capabilities": ["internetClient"]
+        "capabilities": ["internetClient"],
+        "captureDenials": {                // Windows-only: record the process's access
+            "mode": "block-and-log",       // "block-and-log" (default): access stays denied and
```
bbonaby: question: is it always `and-log`? if so we should just do allow and block  instead of the `and-log` if we'll always log stuff.

docs/schema.md
```diff
@@ -59,7 +59,15 @@ production configs and the dev schema when working on experimental features:
...
+                                           // is logged (deny-by-default preserved). "allow-and-log":
+                                           // access is allowed and logged (audit; relaxes
+                                           // deny-by-default, emits a security warning).
+            "outputPath": "C:\\logs\\denials.etl"  // attempts to a learning-mode ETL trace. The
```
bbonaby, line 68: note: when you stop and etl trace you can provide a path, I believe it creates the folder and the file in the folder as well. Should we think about having the same behavior where the consumer doesn't have to create it themselves?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -592,6 +592,51 @@ fn map_wire_containment(c: Option<&wire::Containment>) -> ContainmentBackend {
...
+    match candidate.parent() {
+        // A filesystem root ("/", "C:\\") has either no parent (`None`) or an
+        // empty parent, and cannot name a trace file.
+        None => {
+            let msg = format!(
+                "processContainer.captureDenials.outputPath must name a file, not a \
+                 directory root: '{path}'"
+            );
+            logger.log_line(&msg);
+            Err(WxcError::ConfigParse(msg))
```
bbonaby, line 618: note: fyi on this Gudge is updating how logging and errors work in this PR #641  so depending on if that goes in first you'd need to update these.

docs/schema.md
richiemsft: yeah, it's always and-log. I was going for over-describing. going back to just 1 word.

docs/schema.md
richiemsft: if they dont provide a path, it will be hard for them to know where to look for the files. In the scenario where they want the app to be aware of denials and generate UX to ask the user, they would need to know where to look.

bbonaby (approved):

PR #669, opened 2026-07-22
[WSLC] Generate WSLC SDK FFI bindings with bindgen

SohamDas2021:
## 📖 Description
Replaces the hand-written WSLC SDK FFI bindings with **bindgen-generated** bindings so any ABI drift between the SDK headers and our Rust declarations becomes a **compile error** instead of silent undefined behavior at runtime.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/669)


.github/copilot-instructions.md
```diff
@@ -163,6 +163,7 @@ Per-backend guides:
 - `docs/macos-support/seatbelt-backend.md` — macOS Seatbelt backend
 - `docs/windows-sandbox/windows-sandbox.md` / `docs/windows-sandbox/windows-sandbox-reference.md` — Windows Sandbox backend
 - `docs/wsl/wsl-container-getting-started.md` / `docs/wsl/wsl-container-support-plan.md` — WSL Container (WSLC SDK)
+- `docs/wsl/wslc-sdk-bindings.md` — WSLC SDK FFI bindings: `src/backends/wslc/common/src/wslcsdk_sys.rs` is **generated** by bindgen from `wslcsdk.h` (do NOT hand-edit); `wslc_bindings.rs` is a thin façade over it. On every SDK version bump, regenerate via `scripts/generate-wslc-bindings.ps1` (needs libclang + `bindgen-cli`, required only on the regen machine — normal/CI builds need neither) and commit the regenerated file with the `WSLC_SDK_VERSION` + hash change. See the doc for the full runbook.
```
bbonaby: note: facade has a weird `c` in it `façade`

docs/wsl/wslc-sdk-bindings.md
```diff
@@ -0,0 +1,126 @@
...
+
+bindgen is deliberately kept out of `Cargo.toml` and `build.rs`. Generating at
+build time would force **every** builder of the `wslc` feature to install
+libclang and pull `bindgen`/`clang-sys` into the dependency graph (and into the
+`MxcDependencies` feed). Committing the generated file keeps normal builds
+dependency-free and deterministic; only SDK bumps touch the toolchain.
+
+## Regeneration prerequisites
+
+Needed **only** when regenerating bindings (i.e., on an SDK version bump):
```
bbonaby: suggestion: we should say WSLC SDK version bump instead of SDK alone so folks don't think this means Mxc version bump

docs/wsl/wslc-sdk-bindings.md
```diff
@@ -0,0 +1,126 @@
...
+
+Do all of the following in a **single commit** so the checked-in bindings always
+match the pinned SDK.
+
+1. **Bump the pinned version.** Update `WSLC_SDK_VERSION` in
+   `src/backends/wslc/common/build.rs`.
+
+2. **Pin the new package hash.** Add a match arm for the new version in
+   `expected_sha256()` in `build.rs` with the new `.nupkg`'s lowercase-hex
+   SHA-256. Without a pinned hash the feature build refuses the package
```
bbonaby, line 61: note: we can probably make it case agnostic in the future if it's not already

docs/wsl/wslc-sdk-bindings.md
```diff
@@ -0,0 +1,126 @@
...
+   ```powershell
+   cd src
+   cargo test -p wslc_common      # includes ABI size/enum asserts
+   cargo fmt -p wslc_common -- --check
+   cargo clippy -p wslc_common --all-targets -- -D warnings
+   ```
+
+7. **Run the WSLC E2E suite on a live WSL host** (cannot run in GitHub Actions —
+   needs WSL2 + the WSLC runtime). Build with the feature so `wslcsdk.dll` is
+   staged next to `wxc-exec.exe`, then:
```
bbonaby, line 108: note: we should make sure we onboard these to Elliots testing solution in the future.

docs/wsl/wslc-sdk-bindings.md
```diff
@@ -0,0 +1,126 @@
...
+specific SDK ABI. A header change you didn't notice — a struct that grew, a
+renamed export, a reordered field — produced **silent memory corruption or a
+load failure at runtime**. For example, the 2.8.1 → 2.9.3 bump changed struct
+sizes, renamed `WslcCanRun` → `WslcGetMissingComponents`, renamed
+`CurrentDirectory` → `WorkingDirectory`, and added a `registryAuth` field — all
+of which had to be found and fixed by hand.
+
+With bindgen, the same class of change **cannot compile** until a human
+regenerates the file (reading the new header through bindgen) and fixes the
+callers. Drift is converted from a runtime hazard into a build-time error.
```
bbonaby: note: we can probably remove this section since we're moving to Bindgen. It's more for the PR description really.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -59,7 +59,7 @@ impl Drop for IoCtxRawGuard {
 /// The `Arc` is kept alive in `run_internal` via `IoCtxRawGuard` (which reclaims it
 /// on drop), so the pointer remains valid for the duration of all callbacks.
 /// The SDK guarantees `data` is valid for `data_size` bytes during the callback.
-unsafe extern "system" fn io_callback(
+unsafe extern "C" fn io_callback(
```
bbonaby, line 62: issue: it's recommended to use extern "system" for these types of unsafe functions. See: https://doc.rust-lang.org/reference/items/external-blocks.html#r-items.extern.abi.system

scripts/generate-wslc-bindings.ps1
```diff
@@ -0,0 +1,155 @@
...
+.PREREQUISITES
+    - LLVM / libclang   (winget install LLVM.LLVM)  — provides libclang for bindgen
+    - bindgen CLI       (cargo install bindgen-cli)
+    - Visual Studio 2022 with the MSVC toolchain + a Windows 10/11 SDK
+
+.NOTES
+    The header is extracted from the vendored NuGet package checked into
+    `external/wslc-sdk/` so the generated bindings always match the pinned SDK
+    version, independent of any local build cache.
+#>
```
bbonaby, line 26: thought: don't we download this now as well as a first attempt?

scripts/generate-wslc-bindings.ps1
```diff
@@ -0,0 +1,155 @@
...
+
+# --- 3. Locate the MSVC toolchain include (via vswhere) -------------------------
+$vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+if (-not (Test-Path $vswhere)) { Fail "vswhere.exe not found; a Visual Studio install is required." }
+$vsPath = & $vswhere -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
+if (-not $vsPath) { Fail "No Visual Studio install with the MSVC toolchain was found." }
+$msvcRoot = Join-Path $vsPath "VC\Tools\MSVC"
+$msvcVer  = Get-ChildItem $msvcRoot -Directory | Sort-Object Name -Descending | Select-Object -First 1
+$msvcInc  = Join-Path $msvcVer.FullName "include"
+if (-not (Test-Path $msvcInc)) { Fail "MSVC include directory not found at '$msvcInc'." }
```
bbonaby, line 89: note: might have to check if this all works on arm64 machines too.

scripts/generate-wslc-bindings.ps1
```diff
@@ -0,0 +1,155 @@
...
+        '// @generated by scripts/generate-wslc-bindings.ps1 from the WSLC SDK header',
+        '// (Microsoft.WSL.Containers wslcsdk.h). DO NOT EDIT BY HAND -- re-run the',
+        '// script after bumping WSLC_SDK_VERSION in build.rs and commit the result.',
+        '#![allow(non_snake_case, non_camel_case_types, non_upper_case_globals, dead_code)]',
+        '#![allow(clippy::all)]'
+    )
+
+    # --- 7. Run bindgen ---------------------------------------------------------
+    $bitfieldEnums = 'Wslc(ContainerFlags|SessionFeatureFlags|ContainerStartFlags|DeleteContainerFlags|ComponentFlags|VhdRequirementsFlags)'
+    $newtypeEnums  = 'Wslc(PortProtocol|Signal|ProcessIOHandle|ContainerNetworkingMode|VhdType|SessionTerminationReason|ContainerState|ProcessState|ImageProgressStatus)'
```
bbonaby, line 157: note: might need a comment here for future folks to know why we're doing this

scripts/generate-wslc-bindings.ps1
```diff
@@ -0,0 +1,155 @@
...
+        "--allowlist-file", ".*wslcsdk\.h",
+        "--bitfield-enum", $bitfieldEnums,
+        "--newtype-enum", $newtypeEnums
+        # Layout size/offset asserts are KEPT (default) for ABI-drift safety.
+    )
+    foreach ($l in $rawLines) { $bindgenArgs += @("--raw-line", $l) }
+    $bindgenArgs += @(
+        "-o", $outFile,
+        "--",
+        "--target=x86_64-pc-windows-msvc",
```
bbonaby: note: due to arm64 machines we might need to make this one configurable. Or have a variable that we update based on the system running the script.

src/backends/wslc/common/src/wslc_bindings.rs
```diff
@@ -1,213 +1,67 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-//! Rust FFI bindings for the WSLC SDK C API (`wslcsdk.h`).
```
bbonaby, line 4: suggestion: Stuart added a `mxc_ffi` module here https://github.com/microsoft/mxc/tree/main/src/ffi/mxc_ffi. I'm thinking the unsafe `extern` functions should go in there and we have safe wrappers for them also in there so we can start to move towards a future where everything outside that module is considered safe rust. Then, the unsafe code can all be contained in a single module and auditable. Doesn't have to be this PR.

docs/wsl/wslc-sdk-bindings.md
SohamDas2021: Will do

docs/wsl/wslc-sdk-bindings.md
SohamDas2021: Yes, I have a future work-item to sync with Elliot on this.

scripts/generate-wslc-bindings.ps1
SohamDas2021: No, this script never downloads. It reads the header purely from the vendored NuGet package checked into external/wslc-sdk/. This is deliberate- pinning regen to the committed .nupkg  guarantees the generated bindings always match the exact SDK version in the repo, independent of any feed state or local build cache. 

The download-first/feed-restore logic lives in build.rs (and the SDK-update workflow) for building; binding regeneration stays offline-deterministic against the vendored copy. I've noted this in the .NOTES block of the script.

src/backends/wslc/common/src/wsl_container_runner.rs
SohamDas2021: extern "C" is actually required here rather than a style choice- bindgen generates the callback typedef as extern "C" to match how the SDK declares it in wslcsdk.h:

pub type WslcStdIOCallback = ::std::option::Option<unsafe extern "C" fn(...)>;

io_callback is assigned into WslcProcessCallbacks {onStdOut: Some(io_callback), ...}, and Rust treats extern "system" and extern "C" fn items as distinct, incompatible types. So, switching to extern "system" fails to compile against the generated struct.

src/backends/wslc/common/src/wslc_bindings.rs
SohamDas2021: https://github.com/microsoft/mxc/issues/686 to track.

src/backends/wslc/common/src/wsl_container_runner.rs
bbonaby: ah it's a bindgen thing. If you don't believe in your testing it'll cause an issue then I'm all good. For what it's worth looks like we can override what bindgen outputs via the `override_abi` parameter to its builder. That said I think from reading through the docs it's only useful if we expect to run things on x86 systems.

bbonaby (approved):

PR #682, opened 2026-07-25
fix(isolation-session): refuse network policy the backend cannot enforce; require the unrestricted-network acknowledgment

adpa-ms:
## 📖 Description

The IsolationSession backend has no primitive to filter or deny the container's network (outbound is open; a process inside can listen on a localhost-reachable port). It nonetheless silently **accepted** a default-deny / `Block` policy — plus host allowlists and proxies — it could not enforce, giving callers a false guarantee. This makes the backend honest:

- **Provision and one-shot** accept **only** the canonical unrestricted-network acknowledgment — `network.defaultPolicy = allow` and `network.allowLocalNetwork = true`, no host lists, no proxy. Anything else is refused with a clear `policy_validation` error.
- **Post-provision phases** reject any *supplied* network policy as immutable (new `ExecutionRequest.network_specified` flag distinguishes an explicit `block` from an absent policy).
- The Node SDK's `IsolationSessionProvisionConfig.network` is now a **required** literal, so callers acknowledge the unrestricted network at compile time.

**⚠️ Breaking change** — intentionally rejects IsolationSession configs that were previously accepted; targets `feature/isolation-session-internal`, which already carries breaking changes for this backend.

## 🔍 Validation

- `cargo fmt` / `cargo clippy --all-targets --all-features -D warnings` clean; `cargo test` green with `isolation_session` ON and OFF.
- `policy.rs` validators restructured with a full unit matrix (canonical accept, every refusal shape, proxy refusal, post-provision immutability, precedence).
- Node SDK `npm test` and `validate-configs.js` green.
- **VM end-to-end on a real host:** 72/72 automated (one-shot 14, state-aware 52, Node E2E 6, incl. new network-refusal tests) + 3/3 manual TTY smokes.

## ✅ Checklist

## 📋 Issue Type

- [x] Bug fix

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/682)


jeffstall (approved):

docs/isolation-session/state-aware-rust.md
```diff
@@ -111,19 +111,28 @@ described in [Idempotence](#idempotence-per-phase).
...
 | Field | provision | start | exec | stop | deprovision |
 |---|---|---|---|---|---|
 | `policy.filesystem.{readwritePaths,readonlyPaths}` | rejected | rejected | rejected | rejected | rejected |
 | `policy.filesystem.deniedPaths` | rejected | rejected | rejected | rejected | rejected |
-| `policy.network.{allowedHosts,blockedHosts,defaultPolicy}` | rejected | rejected | rejected | rejected | rejected |
-| `policy.network.proxy` | rejected | rejected | rejected | rejected | rejected |
-| `policy.ui` | rejected | rejected | rejected | rejected | rejected |
+| `policy.network` — canonical `allow` acknowledgment (`defaultPolicy=allow` + `allowLocalNetwork=true`, no host rules, no proxy, default enforcement) | **required** | rejected | rejected | rejected | rejected |
+| `policy.network` — any other value (incl. absent → `block`, host rules, proxy) | rejected | rejected | rejected | rejected | rejected |
+| `policy.ui` — see the known gap below | ignored | ignored | ignored | ignored | ignored |
```
bbonaby, line 135: note: we should probably warn folks with a message in this case.

src/backends/isolation_session/common/src/policy.rs
```diff
@@ -63,18 +82,45 @@ pub(super) fn validate_isolation_session_user(user: &IsolationSessionUser) -> Re
...
-    if !request.policy.allowed_hosts.is_empty()
-        || !request.policy.blocked_hosts.is_empty()
-        || request.policy.default_network_policy != NetworkPolicy::Block
-    {
+    let policy = &request.policy;
+    let is_canonical_allow = policy.default_network_policy == NetworkPolicy::Allow
+        && policy.allow_local_network
+        && policy.allowed_hosts.is_empty()
+        && policy.blocked_hosts.is_empty()
+        && policy.network_enforcement_mode == NetworkEnforcementMode::Capabilities;
```
bbonaby, line 117: note: you can remove this line `policy.network_enforcement_mode == NetworkEnforcementMode::Capabilities` I'm going to remove `NetworkEnforcementMode::Capabilities` from top level network schema and move it to the process container one since that's the only one where it actually  makes sense.

src/core/wxc_common/src/models.rs
```diff
@@ -655,6 +655,13 @@ pub struct ExecutionRequest {
     pub lifecycle: LifecycleConfig,
     /// ProcessContainer-specific policy (used when containment == ProcessContainer).
     pub policy: ContainerPolicy,
+    /// Whether the caller supplied a `network` block on the wire (any field
+    /// present), captured at parse time. Distinguishes an absent network policy
+    /// from an explicit one whose values equal the defaults — the domain
+    /// `policy` alone cannot, since `default_network_policy` defaults to
+    /// `Block`. Used by backends (e.g. IsolationSession) that must reject a
+    /// network policy supplied on a phase where it is immutable.
+    pub network_specified: bool,
```
bbonaby: issue: Seems kind of weird that we have a separate single call out for something policy related in the overall request when we have a bunch of other fragments here related to the execution in general. Perhaps the `ContainerPolicy` is better.

That said, I'm still on the fence about whether this boolean is really needed when the default stance for iso session is to allow network. In my head I would have assumed it was ok to return an error in all state aware scenarios for a config that doesn't follow that. Saying all of that though, I could be interpreting this all wrong so let me know.

.github/copilot-instructions.md
```diff
@@ -113,7 +113,7 @@ The Rust workspace (`src/`) implements multiple sandboxing backends behind the `
...
...nd`) modes. Rejects all filesystem policy (`readwritePaths`/`readonlyPaths`/`deniedPaths`) at every phase with `policy_validation` — the backend has no host-folder-sharing primitive. The container's network is unrestricted (outbound open; a process inside can listen on a localhost-reachable port) and MXC has no primitive to filter or deny it, so provision (and one-shot) accept ONLY the canonical unrestricted-network acknowledgment — `network.defaultPolicy=allow` + `network.allowLocalNetwork=true`, no host rules, no proxy, default enforcement — and refuse anything else (including an absent policy, which defaults to the unenforceable deny) with `policy_validation`; post-provision phases reject any supplied network policy (fixed at provision, tracked via `ExecutionRequest.network_specified`) and inherit an absent one. State-aware additionally accepts an optional `user` bundle (`upn`, `wamToken`) at provision and start to provision Entra cloud-agent sandboxes; one-shot rejects the bundle, and hosts that don't support Entra agents surface `backend_unavailable`. Streams stdout/stderr, forwards stdin, and switches to ConPTY mode when wxc-exec's stdout is a TTY for `spawnSandbox` parity. |
```
bbonaby, line 116: thought: Thinking about this some more, can't we just add this PR to main instead of the feature branch?

ionvasilian (approved):

docs/isolation-session/state-aware-rust.md
adpa-ms: You're right — it's a real gap. There's no `UiPolicy` / `clipboard` / `injection` reference anywhere under `backends/isolation_session/`. This PR documents it; the note says rejection is the intended end state.

I'd rather not fold the fix in here, though:

- A warning isn't deliverable today — no `validate_<phase>` hook takes a logger, and on the one-shot path `wxc-exec` uses `Mode::Buffer` unless `--debug` and never flushes on success. The existing windows_sandbox "ignored fields" warning in `mxc_engine/src/run.rs` lands in that same dropped buffer.
- For policy we don't honor at all, the convention looks like reject rather than warn (`config_parser.rs`: _"reject them explicitly rather than silently discarding a policy the caller believes is in effect"_).

That makes it a phase-matrix decision plus a second breaking change on top of the network one, so I'd like to give it its own pass. Happy to file it.

src/backends/isolation_session/common/src/policy.rs
adpa-ms: I'd like to keep this one if you're OK with it — it's still load-bearing. `{"defaultPolicy":"allow","allowLocalNetwork":true,"enforcementMode":"firewall"}` passes the other four conjuncts and fails only this one, so dropping it would silently accept a filtering request the backend can't honor.

One thing worth flagging for your refactor: `enforcementMode` isn't process-container-only today. `config_parser.rs` rejects `firewall`/`both` for Seatbelt and Bubblewrap when combined with a proxy, and accepts `capabilities` for Bubblewrap (_"Capabilities mode never invokes iptables"_). Those guards would need to move too.

Either way it shouldn't block you — once the field leaves top-level `network`, iso configs can't set it and the conjunct just becomes vacuously true.

src/core/wxc_common/src/models.rs
adpa-ms: **(a)** Agreed, good call — moved onto `ContainerPolicy` next to the other parse-derived member (`#[serde(skip)] network_proxy`). 3 non-test lines, no behavior change.

**(b)** I think it's still needed. `NetworkPolicy` is `{ Allow, #[default] Block }`, so absent and explicit `block` are literally the same value. The rule I'd draw: presence-tracking is needed when the domain default _is_ the value that must be refused. Compare-to-default — what `run.rs` does for the windows_sandbox ignored-field warning — would read `{"defaultPolicy":"block"}` on `exec` as absent and silently accept a deny against an allow-all sandbox.

`post_provision_policy_rejects_specified_network_even_when_block` and `post_provision_policy_accepts_absent_network` have identical `policy` values and opposite outcomes.

If the boolean still bothers you, the cleaner end state is `default_network_policy: Option<NetworkPolicy>` so absence is representable in the type. I'd want that as its own PR — it touches 18 files across 9 backend crates plus the SDK/FFI `build_request` path, and needs a per-backend call on what `None` means.

.github/copilot-instructions.md
adpa-ms: Fair question. The blocker is that retargeting isn't really a rebase here:

- 3 of the 4 docs this PR edits don't exist on `main` — `oneshot.md`, `state-aware-rust.md`, and `state-aware-typescript.md` were renamed in b2c438a.
- The change would be factually wrong on `main`: it says iso rejects all filesystem policy, but `main` still honors `readwritePaths`/`readonlyPaths` via `share_folders`. This branch is what removed `folder_sharing.rs`.
- `policy.rs` differs materially — `main`'s `validate_provision_policy` still has a `denied_paths` branch, and its tests are named `provision_policy_accepts_readwrite_paths`.

There's a caller-impact angle too: this needed 17 `tests/configs` updates, since every iso config now needs a `network` block. Going to `main` now would break iso callers twice — once for network, again when folder sharing goes. Batching them costs ~2 weeks, which we think is the better trade-off.

bbonaby (approved):

Issue #611, opened 2026-07-07
[Seatbelt] If the requested cwd is not allowed by policy, start the process from a sandbox-allowed fallback directory


SLdragon:
### Relevant area(s)

macOS

### Brief description of your issue

When a sandboxed command is launched on macOS Seatbelt with a process working directory that is not allowed by the sandbox policy, the child shell can print noisy `getcwd` warnings during startup.

The command itself may still succeed, especially when it uses an absolute path that is allowed by the policy. The issue is confusing stderr noise, not a command failure .

## Example

Command:

```sh
cat /Users/tester/Downloads/coding-agent-test/ro/protected.txt
```

Output:

```text
shell-init: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
shell-init: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
job-working-directory: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
original read-only content
```

The file read succeeds, but the startup warnings make the result look like something failed.

Related code:

https://github.com/microsoft/mxc/blob/main/src/backends/seatbelt/common/src/seatbelt_runner.rs#L164-L170

## Expected Behavior

mxc should avoid or reduce noisy `getcwd` startup warnings when the requested process working directory is not allowed by the sandbox policy.

Possible approaches:

- If the requested working directory is not allowed by policy, start the process from a sandbox-allowed fallback directory.
- Or provide a way for callers to distinguish the logical/session cwd from the process launch cwd.

### Steps to reproduce

As above

### Expected behavior

As above

### Actual behavior

As above

theelliotm:
This mainly stems from a misconfiguration issue, and a lot of the noise-related concerns were addressed in prior PRs. Approach A sounds like a viable feature request.

theelliotm:
This will remain a configuration issue. No plans to implement a fallback directory.

PR #688, opened 2026-07-27
Seatbelt: fall back to policy-allowed cwd to avoid getcwd warnings

richiemsft:
## 📖 Description
When an explicit process.cwd (or the inherited host cwd) is not readable under the deny-by-default Seatbelt profile, the child shell's startup getcwd() walk fails and leaks noisy "cannot access parent directories" warnings from bash's shell-init / job-working-directory onto stderr.

resolve_working_directory now honors an explicit working directory only when it is readable under the filesystem policy (within a readwrite/readonly path and not within a denied path, matched component-wise). Otherwise it launches from a policy-allowed directory (first readwrite, else first readonly, else /) and logs an informational diagnostic to the mxc log. The command still runs unchanged; only the launch directory changes, never the granted filesystem access.

Adds unit tests for the allow/deny/fallback/tilde/component-boundary cases and an end-to-end characterization test asserting an out-of-policy cwd produces no getcwd noise while the command still succeeds. Updates the Seatbelt backend doc.

## 🔗 References
Resolves the Mac issue from #611 

## 🔍 Validation
Validated locally on a Mac. All green

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/688)

### DIFF
--- a/src/backends/seatbelt/common/src/seatbelt_runner.rs
+++ b/src/backends/seatbelt/common/src/seatbelt_runner.rs
@@ -697,7 +697,13 @@ fn spawn_error(error: &std::io::Error) -> String {
 /// it never grants additional filesystem access.
 fn resolve_working_directory(request: &ExecutionRequest, logger: &mut Logger) -> String {
     if !request.working_directory.is_empty() {
-        if is_working_directory_allowed(request, &request.working_directory) {
+        // A raw `profileOverride` replaces the generated profile entirely, so the
+        // readwrite/readonly/denied policy fields are not applied and cannot be
+        // used to predict readability. Honor the explicit cwd unchanged in that
+        // case rather than second-guessing an opaque profile.
+        if has_profile_override(request)
+            || is_working_directory_allowed(request, &request.working_directory)
+        {
             return request.working_directory.clone();
         }
         let fallback = policy_fallback_directory(request);
@@ -728,6 +734,16 @@ fn policy_fallback_directory(request: &ExecutionRequest) -> String {
     crate::profile_builder::expand_tilde(&default).unwrap_or(default)
 }
 
+/// Whether a raw Seatbelt `profileOverride` is set, in which case the generated
+/// filesystem policy fields are ignored by the profile builder.
+fn has_profile_override(request: &ExecutionRequest) -> bool {
+    request
+        .seatbelt
+        .as_ref()
+        .and_then(|c| c.profile_override.as_ref())
+        .is_some()
+}
+
 /// Whether `dir` is readable under the sandbox's filesystem policy — i.e. a
 /// process launched there can `getcwd()` without the profile denying the walk.
 ///
@@ -738,13 +754,19 @@ fn policy_fallback_directory(request: &ExecutionRequest) -> String {
 /// expands them so the comparison sees the same absolute paths the profile
 /// grants. Matching is component-wise, so `/data` never matches `/database`.
 fn is_working_directory_allowed(request: &ExecutionRequest, dir: &str) -> bool {
-    let dir = crate::profile_builder::expand_tilde(dir).unwrap_or_else(|_| dir.to_string());
+    // Expand `~` exactly as the profile builder does, then lexically fold `.` /
+    // `..` so the containment test compares the path the kernel actually
+    // resolves. Without this, `/work/../private` would spuriously match an
+    // allowed `/work` even though it resolves outside it.
+    let dir = normalize_path(
+        &crate::profile_builder::expand_tilde(dir).unwrap_or_else(|_| dir.to_string()),
+    );
     let policy = &request.policy;
 
     // Deny wins: if the directory is within any denied subpath it is unreadable.
     for denied in &policy.denied_paths {
         if let Ok(denied) = crate::profile_builder::expand_tilde(denied) {
-            if path_within(&dir, &denied) {
+            if path_within(&dir, &normalize_path(&denied)) {
                 return false;
             }
         }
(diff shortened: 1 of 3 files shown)


src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -708,6 +734,103 @@ fn resolve_working_directory(request: &ExecutionRequest) -> String {
...
+/// Whether `dir` is readable under the sandbox's filesystem policy — i.e. a
+/// process launched there can `getcwd()` without the profile denying the walk.
+///
+/// A directory is considered readable when it is within (a subpath of, or
+/// equal to) some `readwritePaths` / `readonlyPaths` entry and is not within
+/// any `deniedPaths` entry (deny overrides allow, matching the profile's rule
+/// ordering). Policy paths are tilde-expanded exactly as the profile builder
+/// expands them so the comparison sees the same absolute paths the profile
+/// grants. Matching is component-wise, so `/data` never matches `/database`.
+fn is_working_directory_allowed(request: &ExecutionRequest, dir: &str) -> bool {
```
MGudgin, line 756: ### Relative working directories are incorrectly replaced by the fallback

`process.cwd` currently accepts relative paths, which `Command::current_dir` resolves against the executor's host cwd. This check instead compares the unresolved relative string directly with absolute policy paths, so it always considers the path unreadable.

For example:

```json
{
  "process": {
    "cwd": "src"
  },
  "filesystem": {
    "readwritePaths": ["/repo"]
  }
}
```

If the executor is running from `/repo`, the intended child cwd is `/repo/src`. The new check compares `src` with `/repo`, fails, and launches the child from the fallback `/repo`. Commands using cwd-relative files then run from the wrong directory.

Similarly, `cwd: "."` can be replaced by the first read-write path even when the inherited cwd is already inside an allowed read-only path.

Could we resolve relative cwd values against `std::env::current_dir()` before normalization and containment checking? Please also add regression tests for `cwd: "."` and `cwd: "src"` when their resolved absolute paths are policy-readable.

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -708,6 +734,103 @@ fn resolve_working_directory(request: &ExecutionRequest) -> String {
...
+    // Expand `~` exactly as the profile builder does, then lexically fold `.` /
+    // `..` so the containment test compares the path the kernel actually
```
MGudgin, line 758: ### The cwd readability check does not match the generated Seatbelt profile

`is_working_directory_allowed` only considers `readwritePaths` and `readonlyPaths`, but the generated Seatbelt profile also grants read access to several system locations unconditionally, including:

- `/bin`
- `/usr/bin`
- `/usr/lib`
- `/usr/share`
- `/System`
- `/Library`
- selected paths under `/private`

This means an explicit cwd such as `/usr/share` or `/Library/Application Support` may be readable under the actual profile, but this check still rejects it and silently launches from the first configured read-write or read-only path instead.

That changes command behavior unnecessarily—for example, `pwd`, relative arguments, and tools that discover configuration from the cwd will see a different directory even though the requested directory was already accessible.

Could we derive this decision from the same baseline path data used by `profile_builder` rather than maintaining a narrower second definition of readability? A regression test using a baseline-readable cwd such as `/usr/share` would also help prevent the two implementations from drifting.

To be fair, filesystem policy being deny-by-default means that we really need a way to 'include' the directories that are accessible by default on a given container platform, so that a) policy authors don't have to list them explicitly and b) the policy is still accurate. The policy 'spec' does allude to this, but we've not implemented it yet.

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -708,6 +734,103 @@ fn resolve_working_directory(request: &ExecutionRequest) -> String {
...
+    // resolves. Without this, `/work/../private` would spuriously match an
```
MGudgin, line 759: ### Escape the requested cwd before writing it to the diagnostic log

The fallback message interpolates `request.working_directory` directly into a line-oriented log entry. Unix paths can contain control characters, including newlines, so a value such as:

```text
/outside-policy
ERROR: sandbox initialization failed
```

would create what appears to be a separate, forged log entry. This does not cross the sandbox boundary because the cwd comes from the caller's configuration, but it can make diagnostics misleading or difficult to parse reliably.

Could we log the path using escaped/debug formatting, such as `{:?}` or `escape_default()`, so each fallback remains a single well-formed log line?

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -708,6 +734,103 @@ fn resolve_working_directory(request: &ExecutionRequest) -> String {
...
+    // allowed `/work` even though it resolves outside it.
```
MGudgin, line 760: ### Add coverage for the new fallback diagnostic

The new fallback behavior includes an operator-facing log message explaining why the requested cwd was replaced, but the tests only verify the returned directory. They do not verify that the diagnostic was emitted or that it identifies the requested and fallback paths.

Since these tests already use a buffered `Logger`, could we add assertions that:

- A rejected cwd logs both the requested directory and the selected fallback.
- An allowed cwd does not emit a fallback diagnostic.
- The logged paths are escaped so control characters cannot create additional log lines.

This would protect the primary diagnostic users have when a command unexpectedly starts in a different directory.

MGudgin (changes requested):

richiemsft:
@theelliotm is looking at it.

PR #689, opened 2026-07-27
Enable one-click crates.io release for mxc-sdk (ESRP) [in-repo, supersedes #647]

dhoehna:
Supersedes #647 (fork PR could not exercise the ESRP release pipeline). Moved to this in-repo branch per @bbonaby's guidance; full history and review feedback preserved.

## What this does
Adds a single ADO pipeline that publishes the `mxc-sdk` crate closure to **public crates.io via ESRP** (Microsoft OSS release path -- no registry tokens in-repo), while third-party build dependencies are pulled from the internal `Mxc-Azure-Feed` Azure Artifacts feed.

- `.azure-pipelines/scripts/crates_release.py` -- ESRP package/release helper
- `.azure-pipelines/templates/Publish.CratesIo.Job.yml` -- leaf-first `EsrpRelease@12` per crate
- `.azure-pipelines/.cargo/config.toml` -- redirects crates.io -> internal feed for the build

Ships with `publishCrates=false` by default; the `verify-order` guard fails safe if the crate list and release order disagree.

## Status -- open design question (needs sign-off)
See the comment below. There is a dependency-ordering issue (cargo [#17196](https://github.com/rust-lang/cargo/issues/17196)) and a leaf-first fix I need **@bbonaby** and **@MGudgin** to approve before anything is published to crates.io. crates.io publishes are permanent, so I want a thumbs-up on the approach **and** the exact crate names first.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/689)


dhoehna:
@bbonaby @MGudgin -- moving this to an in-repo branch (supersedes #647). Before I publish anything to crates.io, there's a dependency-ordering issue plus a fix I'd like your thumbs-up on.

### Issue -- the closure must be published in dependency order
Publishing an interdependent set of crates isn't atomic. When `cargo package` processes a crate whose dependency is another first-party crate in this repo (e.g. `wxc_common` -> `mxc_telemetry`), it strips the local `path` and looks the dependency up **in the registry**, so that dependency must already be published. Our build redirects the crates.io source to the internal `Mxc-Azure-Feed`, so the lookup hits the feed and fails -- the first-party crate was never published there. Net: a dependent can't be packaged/published until every crate it depends on already exists on crates.io. (Upstream: [cargo #17196](https://github.com/rust-lang/cargo/issues/17196) -- batch co-resolution breaks under source replacement.)

### Solution -- leaf-first bootstrap
Publish the closure one crate at a time, **leaf-first**: each crate reaches crates.io before anything that depends on it is packaged. Once a dependency is live (and mirrored into the feed), its dependents build cleanly. This is the standard pattern (cargo-release / release-plz work this way) and sidesteps the bug. Mechanically: a one-off leaf-first ESRP publish to seed the closure, after which the normal pipeline runs unblocked.

### Names that would land on crates.io (permanent)
Full first-party closure of `mxc-sdk` under `--all-features` -- currently ~17 crates:

`mxc-sdk`, `mxc_engine`, `wxc_common`, `mxc_telemetry`, `nanvix_common`, `nanvix_runner`, `mxc_pty`, `sandbox_spec`, `lxc_common`, `bwrap_common`, `seatbelt_common`, `appcontainer_common`, `windows_sandbox_common`, `isolation_session_common`, `isolation_session_bindings`, `wslc_common`, `hyperlight_common`

(The pipeline is presently exercising an older 10-crate subset.) Since crates.io names are permanent, this is also the moment to confirm final naming -- e.g. whether we adopt an `mxc-` prefix across the closure.

### Ask
A thumbs-up on **(1)** the leaf-first bootstrap approach and **(2)** the crate names / closure above, before I publish anything to crates.io. Happy to rehearse the exact flow against the private feed (reversible) first to prove it end-to-end.

.azure-pipelines/templates/Publish.CratesIo.Job.yml
```diff
@@ -0,0 +1,104 @@
...
+      inputs:
+        - input: pipelineArtifact
+          pipeline: MXC
+          artifactName: ${{ parameters.cratesArtifactName }}
+          targetPath: $(Pipeline.Workspace)/crates
+    variables:
+      cratesDir: $(Pipeline.Workspace)/crates
+      esrpDir: $(Pipeline.Workspace)/esrp-one
+      orderFile: $(Pipeline.Workspace)/crates/release-order.json
+      script: $(Pipeline.Workspace)/crates/crates_release.py
```
MGudgin, line 51: The production release job executes `crates_release.py` directly from the consumed build artifact. That makes the artifact both the data being released and the code that verifies/stages it; a compromised or wrongly selected artifact can replace `verify-order` and then run in a production release job that has the ESRP connection (and later receives `System.AccessToken`).

Please execute the helper from a pinned, reviewed repository ref and treat the artifact as data only. The build artifact should carry an independently verified manifest containing the source SHA and hashes for `release-order.json` and every `.crate` file.

.azure-pipelines/templates/Publish.CratesIo.Job.yml
```diff
@@ -0,0 +1,104 @@
...
+    displayName: Publish crates to crates.io (ESRP)
+    pool:
+      name: Azure-Pipelines-1ESPT-ExDShared
+      image: ubuntu-latest
+      os: linux
+    templateContext:
+      type: releaseJob
+      isProduction: true
+      inputs:
+        - input: pipelineArtifact
```
MGudgin, line 43: The new crates.io release path consumes the `MXC-Official-Build` pipeline artifact, but the pipeline resource in `1ES.Release.yml` is not constrained by branch/tag and the artifact does not record a source commit or hashes. That means an irreversible public release can be queued against an artifact that does not correspond to the reviewed `main` revision.

Please constrain the resource to an approved release source and verify the artifact's source SHA and package hashes before entering the ESRP loop.

.azure-pipelines/scripts/crates_release.py
```diff
@@ -0,0 +1,355 @@
...
+    )
+    return json.loads(result.stdout)
+
+
+def _run(args: list[str], cwd: str | None = None) -> int:
+    print("+ " + " ".join(args), flush=True)
+    return subprocess.run(args, cwd=cwd).returncode
+
+
+def _load_order(order_file: str) -> list[dict]:
```
MGudgin, line 169: This adds the full package/verify-order/stage/wait safety path without automated tests, and the implementation hardwires subprocesses, HTTP, environment lookup, filesystem mutation, monotonic time, and sleeping. The most important failure modes—cargo failure, missing crate output, malformed order JSON, destructive staging, 401/404/feed timeout, malformed sparse-index data, and partial publication—therefore cannot be exercised deterministically without broad monkeypatching.

Please separate pure release planning/parsing from injected side effects and add unit tests for each command's success and failure exit paths before this becomes part of every official build.

.azure-pipelines/1ES.Release.yml
```diff
@@ -10,14 +23,49 @@ resources:
...
     trigger: none
 
 
 # Parameters for ESRP release info will be passed from the ADO UI.
 parameters:
+  - name: publishNpm
+    displayName: "Publish the npm SDK to npmjs.com"
+    type: boolean
+    default: false
+  - name: publishCrates
```
MGudgin, line 38: This changes the existing release pipeline's default behavior: `publishNpm` is now false, while `publishCrates` is also false. A default queue operation that previously published npm now selects no publication target (and may leave the expanded `stages` list empty, depending on template validation).

Please either preserve the npm default or add an unconditional validation stage that fails clearly when neither target is selected, so the default run cannot silently become a no-op.

.azure-pipelines/scripts/crates_release.py
```diff
@@ -0,0 +1,355 @@
...
+    #                  .cargo/config.toml) leaves the tree in a state cargo
+    #                  considers dirty.
+    #   --registry crates-io: the pipeline's .cargo/config.toml replaces the
+    #                  crates-io source with the internal Mxc-Azure-Feed, so
+    #                  cargo refuses to `package` without an explicit target
+    #                  registry ("crates-io is replaced with remote registry
+    #                  Mxc-Azure-Feed; include --registry ..."). Package for
+    #                  crates-io -- the real publish target -- so intra-closure
+    #                  sibling deps are recorded as plain crates.io deps;
+    #                  packaging for the feed would embed a registry-index that
```
MGudgin, line 216: `--allow-dirty` disables Cargo's provenance guard for the entire source tree, not just the intended Cargo configuration mutation. Any unrelated agent-side change under `src/` can therefore be baked into the public `.crate` files, and `--no-verify` removes the second chance to detect it.

Please keep the feed configuration outside the checkout or explicitly assert that `git status --porcelain -- src/` is empty before packaging, then remove `--allow-dirty`.

src/Cargo.toml
```diff
@@ -101,10 +102,10 @@ base64 = "0.22"
 clap = { version = "4", features = ["derive"] }
 chrono = { version = "0.4", default-features = false, features = ["std", "clock"] }
 quick-xml = "0.41"
-wxc_common = { path = "core/wxc_common" }
+wxc_common = { path = "core/wxc_common", version = "0.7.0" }
```
MGudgin, line 105: These explicit `version = "0.7.0"` requirements are necessary for publication, but they create a second manually maintained copy of the workspace version across every first-party path dependency. A future workspace version bump can easily leave one requirement stale and produce unusable published manifests.

Please extend the repository's version-sync/version-bump tooling to validate and update all publishable workspace path-dependency versions in the same change.

.azure-pipelines/scripts/crates_release.py
```diff
@@ -0,0 +1,355 @@
...
+    "nanvix_common",
+    "mxc_pty",
+    "sandbox_spec",
+    "wxc_common",
+    "lxc_common",
+    "seatbelt_common",
+    "appcontainer_common",
+    "bwrap_common",
+    "mxc-sdk",
+]
```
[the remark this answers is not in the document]
MGudgin: One additional consequence of the duplicated Python/YAML closure is build-artifact compatibility. `CRATES` and `release-order.json` are fixed at build time, while `crateOrder` is expanded from the release pipeline revision at release time. After the closure changes, the current release YAML will reject older otherwise-valid artifacts, so historical builds become unreleasable unless the release tooling is manually checked out at a matching revision.

Deriving the order once is preferable; the artifact should carry a versioned/signed order and the release job should consume it dynamically (or explicitly version the release tooling with the artifact).

.azure-pipelines/templates/1ES.Build.Stages.yml
```diff
@@ -82,6 +82,19 @@ stages:
 
   - template: Mxc.Binary.Packaging.Job.yml
 
+# Crate packaging only runs on the official (signed) build -- PR/CI builds have
+# no reason to produce the crates.io release artifact.
+- ${{ if parameters.isOfficialBuild }}:
+  - stage: Package_Crates
```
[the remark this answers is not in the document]
MGudgin: Even after the immediate packaging failure is fixed, this stage still packages and uploads the full crate closure on every official build despite `publishCrates` defaulting to false. Since the closure is currently known-incomplete, that adds recurring build cost and artifact churn before the release path is usable. Please gate artifact production behind a separate default-off build parameter, at least until crates publishing is enabled end-to-end.

dhoehna:
Closing in favor of #716.

The crates.io work continued there and #716 is well ahead of this branch: the release runs as its own pipeline separate from the npm release, the ref is selected by tag and gated so a non-release tag is rejected, the publish order is computed from cargo rather than hand-maintained, and there is a dry-run path.  None of that is on this branch.

@MGudgin your 1 Aug review carries over.  I am tracking all eight items against #716 rather than dropping them.  Apologies for the split -- #716 was started in a session that did not pick up this PR, which is on me.

PR #696, opened 2026-07-28
Wire captureDenials to Windows Learning Mode runtime

richiemsft:
## 📖 Description

Wires the merged `processContainer.captureDenials` contract to the Windows Learning Mode runtime for the BaseContainer tier.

- Maps `mode: "block"` to `learningModeLogging` and `mode: "allow"` to `permissiveLearningMode` after rejecting user-supplied reserved capability names.
- Enforces one effective mode: `allow` removes deny-and-record before injecting permissive mode.
- Creates a process security environment, starts capture before launch, attaches the environment with `PROC_THREAD_ATTRIBUTE_SECURITY_ENVIRONMENT`, and launches through normal `CreateProcessW`.
- Restricts inherited handles with `PROC_THREAD_ATTRIBUTE_HANDLE_LIST`.
- Supplies an explicit clean user environment to `CreateProcessW`, preventing
  `wxc-exec` process variables and secrets from being inherited.
- Uses the caller-provided `outputPath` or a managed per-run temporary ETL path, then seals the trace after process-tree teardown.
- Rejects unsupported AppContainer fallback directly in the dispatcher before backend construction or DACL mutation.
- Keeps native `try_wait()` non-blocking; blocking waits and managed `Wait`/`WaitAsync` complete process-tree and capture finalization.
- Preserves recovery tracking whenever the security environment may remain live.
- Adds an injectable capture lifecycle seam with deterministic begin/finalization failure tests.

## 🔗 References

- #622
- #661
- #663

## 🔍 Validation

- `cargo test -p learning_mode_windows -p appcontainer_common -- --test-threads=1`
- `cargo clippy -p learning_mode_windows -p appcontainer_common --all-targets -- -D warnings`
- `dotnet build Microsoft.Mxc.Sdk/Microsoft.Mxc.Sdk.csproj --no-restore`
- Feature-enabled V1 Windows VM `172.17.7.153` (build `26657.1002`):
  - parent environment sentinel was absent in the sandbox (`CLEAN`, exit 0)
  - standalone attribute-based capture: exit 0, non-empty ETL
  - `block`: exit 0, 512-byte ETL
  - `allow`: exit 0, 3,614-byte ETL, always-visible permissive security warning

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.

### DIFF
--- a/src/backends/appcontainer/common/src/appcontainer_runner.rs
+++ b/src/backends/appcontainer/common/src/appcontainer_runner.rs
@@ -48,6 +48,9 @@ use wxc_common::sandbox_process::{
 use wxc_common::script_runner::get_timeout_milliseconds;
 use wxc_common::{string_util, ui_policy};
 
+pub(crate) const CAPTURE_DENIALS_FALLBACK_UNSUPPORTED_MSG: &str =
+    "captureDenials requires the BaseContainer learning-mode APIs and is not supported by the AppContainer fallback tier";
+
 /// `UpdateProcThreadAttribute` value for
 /// `PROC_THREAD_ATTRIBUTE_ALL_APPLICATION_PACKAGES_POLICY` that opts the
 /// process out of inheriting `ALL APPLICATION PACKAGES` grants. This
@@ -1306,11 +1309,9 @@ impl AppContainerScriptRunner {
 impl SandboxBackend for AppContainerScriptRunner {
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
         if request.policy.capture_denials.is_some() {
-            let msg = "captureDenials requires the BaseContainer learning-mode APIs and is not \
-                       supported by the AppContainer fallback tier";
             return Err(ScriptResponse {
                 failure_phase: FailurePhase::BackendUnavailable,
-                ..ScriptResponse::error(msg)
+                ..ScriptResponse::error(CAPTURE_DENIALS_FALLBACK_UNSUPPORTED_MSG)
             });
         }
         if !request.policy.denied_paths.is_empty() && self.filesystem_mode != FilesystemMode::Dacl {
@@ -1833,7 +1834,9 @@ mod tests {
 
     // ---- validate_runner: unsupported policy fields surface as errors. ----
 
-    use super::{AppContainerScriptRunner, FilesystemMode};
+    use super::{
+        AppContainerScriptRunner, FilesystemMode, CAPTURE_DENIALS_FALLBACK_UNSUPPORTED_MSG,
+    };
     use wxc_common::models::{ExecutionRequest, FailurePhase};
     use wxc_common::sandbox_process::SandboxBackend;
 
@@ -1906,6 +1909,9 @@ mod tests {
             .validate(&request)
             .expect_err("AppContainer fallback must not silently ignore captureDenials");
         assert_eq!(error.failure_phase, FailurePhase::BackendUnavailable);
-        assert!(error.error_message.contains("requires the BaseContainer"));
+        assert_eq!(
+            error.error_message,
+            CAPTURE_DENIALS_FALLBACK_UNSUPPORTED_MSG
+        );
     }
 }
--- a/src/backends/appcontainer/common/src/appcontainer_runner.rs
+++ b/src/backends/appcontainer/common/src/appcontainer_runner.rs
@@ -34,7 +34,9 @@ use crate::job_object::UiJobObject;
 use crate::process_mitigation;
 use wxc_common::error::WxcError;
 use wxc_common::logger::Logger;
-use wxc_common::models::{ExecutionRequest, NetworkEnforcementMode, NetworkPolicy, ScriptResponse};
+use wxc_common::models::{
+    ExecutionRequest, FailurePhase, NetworkEnforcementMode, NetworkPolicy, ScriptResponse,
+};
 use wxc_common::process_util::{
     create_std_pipes, InterruptiblePipeReader, OwnedHandle, PipeReadCanceller, PipeWriter,
     SendOwnedHandle, SidAndAttributes,
@@ -1303,6 +1305,14 @@ impl AppContainerScriptRunner {
 
 impl SandboxBackend for AppContainerScriptRunner {
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
+        if request.policy.capture_denials.is_some() {
+            let msg = "captureDenials requires the BaseContainer learning-mode APIs and is not \
+                       supported by the AppContainer fallback tier";
+            return Err(ScriptResponse {
+                failure_phase: FailurePhase::BackendUnavailable,
+                ..ScriptResponse::error(msg)
+            });
+        }
         if !request.policy.denied_paths.is_empty() && self.filesystem_mode != FilesystemMode::Dacl {
             return Err(ScriptResponse::error(
                 wxc_common::error::DENIED_PATHS_NOT_SUPPORTED_MSG,
@@ -1824,7 +1834,7 @@ mod tests {
     // ---- validate_runner: unsupported policy fields surface as errors. ----
 
     use super::{AppContainerScriptRunner, FilesystemMode};
-    use wxc_common::models::ExecutionRequest;
+    use wxc_common::models::{ExecutionRequest, FailurePhase};
     use wxc_common::sandbox_process::SandboxBackend;
 
     #[test]
@@ -1884,5 +1894,18 @@ mod tests {
         let runner = AppContainerScriptRunner::new();
         let request = ExecutionRequest::default();
         assert!(runner.validate(&request).is_ok());
+    }
+
+    #[test]
+    fn validate_runner_rejects_capture_denials_on_appcontainer_fallback() {
+        let runner = AppContainerScriptRunner::new();
+        let mut request = ExecutionRequest::default();
+        request.policy.capture_denials = Some(Default::default());
+
+        let error = runner
+            .validate(&request)
+            .expect_err("AppContainer fallback must not silently ignore captureDenials");
+        assert_eq!(error.failure_phase, FailurePhase::BackendUnavailable);
+        assert!(error.error_message.contains("requires the BaseContainer"));
     }
 }
--- a/src/backends/appcontainer/common/src/appcontainer_runner.rs
+++ b/src/backends/appcontainer/common/src/appcontainer_runner.rs
@@ -64,7 +64,7 @@ const PROXY_VAR_NAMES: &[&str] = &["HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "ALL
 /// Serialize `KEY=VALUE` pairs into a double-null-terminated UTF-16 environment block.
 ///
 /// Entries are sorted case-insensitively by key as required by `CreateProcessW`.
-fn encode_env_block(entries: &[(String, String)]) -> Vec<u16> {
+pub(crate) fn encode_env_block(entries: &[(String, String)]) -> Vec<u16> {
     let mut sorted: Vec<&(String, String)> = entries.iter().collect();
     sorted.sort_by(|(a, _), (b, _)| a.to_ascii_uppercase().cmp(&b.to_ascii_uppercase()));
 
@@ -85,7 +85,7 @@ fn encode_env_block(entries: &[(String, String)]) -> Vec<u16> {
 /// Calls `CreateEnvironmentBlock` with `bInherit = FALSE` so that only the
 /// system/user profile variables are included (no process-level vars leak in).
 /// Returns the entries as `(key, value)` pairs.
-fn create_default_env_entries() -> Result<Vec<(String, String)>, WxcError> {
+pub(crate) fn create_default_env_entries() -> Result<Vec<(String, String)>, WxcError> {
     unsafe {
         let mut token = HANDLE::default();
         OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &mut token)
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -1138,13 +1138,22 @@ impl BaseContainerRunner {
         let mut pi: PROCESS_INFORMATION = unsafe { std::mem::zeroed() };
 
         // Environment block for the sandboxed child.
-        // If the caller specified explicit env vars, use only those.
-        // Otherwise, pass NULL to let the OS provide the default environment
-        // for the sandbox (CreateProcessInSandbox handles this internally).
+        // Explicit variables are always isolated from the parent environment.
+        // The one-shot API supplies its own default when this is NULL, but the
+        // attribute-based CreateProcessW capture path must receive an explicit
+        // clean block or it would inherit all wxc-exec process variables.
         let env_block: Option<Vec<u16>> = if request.env.is_empty() {
-            // TODO: consider calling CreateEnvironmentBlock(NULL, FALSE) here
-            // for a cleansed default env if the OS API doesn't do it for us.
-            None
+            if capture_denials.is_some() {
+                let entries =
+                    crate::appcontainer_runner::create_default_env_entries().map_err(|error| {
+                        ScriptResponse::error(&format!(
+                            "captureDenials failed to create a clean child environment: {error}"
+                        ))
+                    })?;
+                Some(crate::appcontainer_runner::encode_env_block(&entries))
+            } else {
+                None
+            }
         } else {
             Some(encode_env_block(&request.env))
         };
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -140,6 +140,13 @@ fn learning_mode_api_not_implemented(error: &learning_mode_windows::LearningMode
     }
 }
 
+fn learning_mode_cleanup_failed(error: &learning_mode_windows::LearningModeError) -> bool {
+    matches!(
+        error,
+        learning_mode_windows::LearningModeError::CleanupFailed { .. }
+    )
+}
+
 /// Script runner that uses `Experimental_CreateProcessInSandbox` API
 /// to launch a sandboxed process.
 #[derive(Default)]
@@ -1072,6 +1079,26 @@ impl BaseContainerRunner {
                     } else {
                         FailurePhase::LaunchFailed
                     };
+                    if request.lifecycle.destroy_on_exit {
+                        if learning_mode_cleanup_failed(&e) {
+                            sandbox_tracking::mark_cleanup_deferred(
+                                &sid_string,
+                                "capture initialization failed and the process security environment could not be closed",
+                                logger,
+                            );
+                        } else {
+                            // No process was launched and the security
+                            // environment is closed, so real profile/tracking
+                            // cleanup is safe.
+                            sandbox_tracking::cleanup_unlaunched_sandbox(
+                                &identity,
+                                &sid_string,
+                                logger,
+                            );
+                        }
+                        sandbox_tracking::unregister_ctrl_c_cleanup();
+                    }
+                    self.proxy_coordinator.stop(logger);
                     return Err(ScriptResponse {
                         exit_code: -1,
                         error_message: msg.clone(),
@@ -1810,6 +1837,29 @@ mod tests {
             code: 87,
         };
         assert!(!learning_mode_api_not_implemented(&ordinary));
+    }
+
+    #[test]
+    fn learning_mode_cleanup_failure_is_detected() {
+        use learning_mode_windows::LearningModeError;
+
+        let error = LearningModeError::CleanupFailed {
+            primary: Box::new(LearningModeError::ApiCall {
+                function: "StartLearningModeTrace",
+                code: 87,
+            }),
+            cleanup: Box::new(LearningModeError::ApiCall {
+                function: "CloseProcessSecurityEnvironment",
+                code: 5,
+            }),
+        };
+        assert!(learning_mode_cleanup_failed(&error));
+
+        let ordinary = LearningModeError::ApiCall {
+            function: "StartLearningModeTrace",
+            code: 87,
+        };
+        assert!(!learning_mode_cleanup_failed(&ordinary));
     }
 
     #[test]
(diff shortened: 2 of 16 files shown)


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1456,8 +1637,8 @@ impl SandboxProcess for BaseContainerSandboxProcess {
         }
         cancel_and_join_discard(stdout_thread, &self.stdout_canceller);
         cancel_and_join_discard(stderr_thread, &self.stderr_canceller);
-        self.run_teardown();
-        result
+        let teardown_result = self.run_teardown();
```
MGudgin, line 1879: **Reliability (High): the SDK''s polling `Wait()` path can return success before the denials ETL is written.**

`run_teardown()` — the only code that calls `CaptureSession::finish()` and seals/writes the ETL — is invoked here from `wait()` and from `Drop`, but **not** from `try_wait()` (`base_container_runner.rs:1569`, which only does `WaitForSingleObject(..., 0)` + `GetExitCodeProcess`).

The .NET SDK''s `MxcSandboxProcess.Wait()` polls `mxc_sandbox_try_wait` and returns as soon as it observes the child exit, rather than calling the blocking native `wait()`. So a caller can successfully "wait" for a `captureDenials` run, observe exit code `0`, and then find **no output file** — the trace isn''t sealed until the handle is later dropped/freed. For the feature''s headline use case (run sandbox → read the denials file) that''s a race yielding a missing/empty deliverable.

`Drop` does still seal, so this isn''t permanent data loss — it''s a "the output isn''t there yet when `Wait()` returns" ordering problem.

**Suggested fix:** make a terminal `try_wait()` observation perform the same finalization as `wait()` — once `try_wait` sees the child has exited, run (and propagate) `run_teardown()` exactly once; or have the SDK poll loop call `mxc_sandbox_wait` after `try_wait` reports exit. A single "child exited → seal once" path shared by `wait` and `try_wait` avoids double-sealing.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1303,6 +1305,14 @@ impl AppContainerScriptRunner {
 
 impl SandboxBackend for AppContainerScriptRunner {
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
+        if request.policy.capture_denials.is_some() {
+            let msg = "captureDenials requires the BaseContainer learning-mode APIs and is not \
+                       supported by the AppContainer fallback tier";
```
bbonaby: thought: would probably be better to put this in constant variable then we can reuse it in the test instead of having to just check if the error message contains a fragment of the real error message.

src/backends/appcontainer/common/src/sandbox_tracking.rs
```diff
@@ -193,6 +193,23 @@ pub fn cleanup_sandbox(identity: &str, sid_string: &str, logger: &mut Logger) {
...
+/// The tracking entry is removed only after profile deletion succeeds. If
+/// deletion fails, the entry is retained and marked deferred so a later
+/// maintenance pass still has the identity and SID needed for recovery.
+pub fn cleanup_unlaunched_sandbox(identity: &str, sid_string: &str, logger: &mut Logger) {
+    if crate::appcontainer_runner::delete_app_container_profile(identity, logger) {
+        delete_tracking_key(sid_string, logger);
+    } else {
+        mark_cleanup_deferred(
+            sid_string,
+            "capture initialization failed and the AppContainer profile could not be deleted",
```
bbonaby: note: same thing here about constant variable usage

src/core/wxc/src/main.rs
```diff
@@ -200,16 +200,22 @@ fn apply_permissive_learning_mode(capabilities: &mut Vec<String>) -> bool {
...
              resolved containment is '{}'",
-            containment.wire_name()
-        ))
+            request.containment.wire_name()
+        ));
+    }
+    if request.policy.capture_denials.is_some() {
+        return Err(
+            "--audit cannot be combined with processContainer.captureDenials; \
+             use captureDenials.mode: \"allow\" for permissive capture"
```
bbonaby: note: same thing about constant variable usage. I know there are some others that aren't using constants but this would probably be the way we should move towards.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -863,6 +859,36 @@ enforced; access denials are recorded for diagnostics.\n",
...
+            // which requires the corresponding learning-mode capability on the
+            // child token so the OS emits the access-check records the capture
+            // path collects. Inject it additively (preserving the workload's real
+            // capabilities). `block` keeps deny-by-default via
+            // `learningModeLogging`; `allow` replaces deny-and-record with
+            // `permissiveLearningMode` (the runner surfaces the security warning).
+            let capture_capability = match mode {
+                CaptureDenialsMode::Block => {
+                    policy.capabilities.retain(|capability| {
+                        !capability.eq_ignore_ascii_case("permissiveLearningMode")
```
bbonaby, line 876: question: should this be `contains` instead of equals? Was thinking more to prevent something like "`internetClient, permissiveLearningMode`" from staying in the list. That said, the OS would  probably return an error if you provide an invalid capability.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -661,6 +685,64 @@ impl BaseContainerRunner {
 
         Self::log_sandbox_spec(&spec_bytes, logger);
 
+        // Resolve the learning-mode capture APIs when captureDenials is
+        // configured. The mode-mapped learning-mode capability was already
+        // injected into policy.capabilities (and hence the spec) by the config
+        // parser; here we obtain the OS handles that drive the ETL trace. Both
```
bbonaby: note: we can probably remove this part
```rust
 // Resolve the learning-mode capture APIs when captureDenials is
// configured. The mode-mapped learning-mode capability was already
// injected into policy.capabilities (and hence the spec) by the config
// parser;
```
The rest does look valuable to keep though

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -661,6 +685,64 @@ impl BaseContainerRunner {
...
+        // degrade.
+        let capture_denials = request.policy.capture_denials.clone();
+        let mut capture_apis: Option<(SecurityEnvironmentApi, LearningModeApi)> = None;
+        if capture_denials.is_some() {
+            let _ = writeln!(logger, "{EMOJI_SECTION} SECTION: captureDenials");
+            match (SecurityEnvironmentApi::load(), LearningModeApi::load()) {
+                (Ok(se), Ok(lm)) => {
+                    let _ = writeln!(
+                        logger,
+                        "captureDenials: learning-mode trace API available (processmodel.dll)"
```
bbonaby: note: see my comments about constant variables for string literals for this and below.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -661,6 +685,64 @@ impl BaseContainerRunner {
...
+        // exports live in the same processmodel.dll the BaseContainer launch API
+        // is loaded from, and only exist on supported feature-enabled Windows
+        // builds, so a failed load is a clean, actionable "backend unavailable"
+        // degrade.
+        let capture_denials = request.policy.capture_denials.clone();
+        let mut capture_apis: Option<(SecurityEnvironmentApi, LearningModeApi)> = None;
+        if capture_denials.is_some() {
+            let _ = writeln!(logger, "{EMOJI_SECTION} SECTION: captureDenials");
+            match (SecurityEnvironmentApi::load(), LearningModeApi::load()) {
+                (Ok(se), Ok(lm)) => {
```
bbonaby: thought: might be better for quick readability to spell these out via
```rust
(Ok(security_environment_api), Ok(learning_mode_api))
```

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -961,71 +1043,173 @@ impl BaseContainerRunner {
...
-
-            if result != 0 {
-                break (result, None);
+            if ok != 0 {
+                (ok, None)
+            } else {
+                (0, Some(unsafe { GetLastError() }))
             }
+        } else {
+            loop {
```
bbonaby: question: I think this part of the function would benefit from being refactored out into its own function so we can explain why we need the loop. Was a bit confused when I first saw it.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -961,71 +1043,173 @@ impl BaseContainerRunner {
...
+                            // cleanup is safe.
+                            sandbox_tracking::cleanup_unlaunched_sandbox(
+                                &identity,
+                                &sid_string,
+                                logger,
+                            );
+                        }
+                        sandbox_tracking::unregister_ctrl_c_cleanup();
+                    }
+                    self.proxy_coordinator.stop(logger);
```
bbonaby: question: I don't want to put more work on you, but thinking about this, should we be doing this in the base_container's drop method? Then here you can really just return the ScriptResponse without having to worry about cleaning up when we go out of scope.

richiemsft: I can move that there and see what happens.

src/core/wxc_common/src/config_parser.rs
richiemsft: the comma separated values are not allowed.

richiemsft:
Made a few mistakes with the stack tracking: 
https://github.com/microsoft/mxc/pull/709 is the old #699 
https://github.com/microsoft/mxc/pull/710 is the old #701

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -12,8 +12,13 @@
 use std::ffi::c_void;
 use std::fmt::Write;
 use std::io::IsTerminal;
+use std::path::PathBuf;
 use std::ptr;
 
+use learning_mode_windows::{
+    CaptureSession, LearningModeApi, SecurityEnvironmentApi, PROCESS_SECURITY_ENVIRONMENT_FLAG_NONE,
```
bbonaby: issue: So I spoke to @jsidewhite and turns out you should not be calling CreateProcessAsUserInsideSecurityEnvironment and we shouldn't have references to it anywhere in the code base. 

https://github.com/microsoft/mxc/blob/839ace45b0cebd7399ca73bf3a25589be3f6089b/src/backends/learning_mode/windows/src/secenv.rs#L87

In actuality the 2 phase call for creating a process in a security environment goes like:

1. Call `CreateProcessSecurityEnvironment` to get a handle 
2. Call [UpdateProcThreadAttribute](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute) with the enum value 35 to represent `ProcThreadAttributeSecurityEnvironment`. I believe the handle we get in step 1 should be passed into this call as the value for the attribute.
3. Call `CreateProcess`  like you normally would without any of the security environment stuff. Under the hood this should  actually call into the internal `CreateProcessAsUserInSecurityEnvironment` with the handle to the security environment you added to the proc thread attribute in step 2. 

@jsidewhite  let me know if I have the steps wrong here.

richiemsft: Updated in a09ad489. Removed all references to `CreateProcessAsUserInsideSecurityEnvironment`; capture now attaches the environment through `PROC_THREAD_ATTRIBUTE_SECURITY_ENVIRONMENT` on `STARTUPINFOEXW` and launches with `CreateProcessW`. The attribute list also includes `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` to restrict inheritance to the configured stdio handles. Validated on VM: standalone capture and integrated block/allow modes all exited 0 and produced non-empty ETLs.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -957,75 +1144,131 @@ impl BaseContainerRunner {
...
+                vec![h_stdin, h_stdout, h_stderr]
+            } else {
+                Vec::new()
+            };
+            let extended_startup = match SecurityEnvironmentStartupInfo::new(
+                si,
+                session.environment(),
+                &inherited_handles,
+            ) {
+                Ok(startup) => startup,
```
MGudgin, line 1275: ### Preserve recovery tracking until the live capture session is explicitly torn down

If `CaptureSession::begin` succeeds but `SecurityEnvironmentStartupInfo::new` fails, this branch calls `cleanup_capture_prelaunch_failure` before the live `CaptureSession` is dropped.

For an ordinary startup-info error, `cleanup_capture_prelaunch_failure` removes the sandbox tracking entry. The function then returns, and `CaptureSession::Drop` performs best-effort trace/environment cleanup while discarding any errors.

This creates the following failure sequence:

1. The process security environment and trace are created successfully.
2. Attaching the environment to the startup attributes fails.
3. Recovery tracking is removed.
4. Dropping the session fails to close the security environment.
5. The broker-side environment is left orphaned without a tracking entry for later cleanup.

Could we explicitly take and finish the capture session before changing the tracking state? If `finish(None)` reports a stop or close failure, the tracking entry should be retained and marked for deferred cleanup, with that failure included in the returned launch error.

src/backends/appcontainer/common/src/base_container_runner.rs
MGudgin, line 1815: ### `captureDenials` changes both the timing and process-tree semantics of `try_wait()`

`try_wait()` is documented as a non-blocking exit check, but when capture is enabled, the first poll that observes the root process exit does substantially more work before returning:

1. `terminate_and_reap()` kills any surviving descendants.
2. `CaptureSession::finish()` stops and flushes the ETW trace.
3. Later stack changes also decode the ETL and write the JSON output synchronously.

Without capture enabled, the same poll immediately returns the root exit code without killing descendants or running teardown.

This means an observability option changes two externally visible behaviors:

- A supposedly non-blocking poll can stall on ETW flushing and output generation.
- Background descendants are killed immediately only when capture is enabled.

Could we define one process-tree lifecycle independent of `captureDenials` and preserve non-blocking polling semantics? For example, `try_wait()` could report that the root exited while finalization remains pending, with `wait()` or a background task completing capture teardown before the final result becomes available.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1341,16 +1600,41 @@ impl BaseContainerSandboxProcess {
...
+
+        // Seal the learning-mode ETL trace now that the child has exited and
+        // been reaped (both `wait` and `Drop` kill + reap before calling this).
+        // `finish` stops the trace to the resolved output path, then closes the
+        // security environment. The output-file path is surfaced on stderr so a
+        // caller can locate the denials (the full NDJSON contract is finalized
+        // in the output/consume stage).
+        let capture_result = if let Some(session) = self.capture_session.take() {
+            let output_path = self.capture_output_path.take();
+            match session.finish(output_path.as_deref()) {
```
MGudgin, line 1734: ### Treat a standalone security-environment close failure as a deferred-cleanup condition

`CaptureSession::finish()` returns a plain `LearningModeError::ApiCall` when stopping the trace succeeds but `CloseProcessSecurityEnvironment` fails. However, `learning_mode_cleanup_failed()` only recognizes the `CleanupFailed` variant used when both teardown operations fail.

A close-only failure is particularly important because it means the process security environment may still be live. Treating it like an ordinary capture error allows the normal cleanup/unregistration path to continue without recording that broker-side recovery may still be required.

Could we classify an `ApiCall` from `CloseProcessSecurityEnvironment` as a cleanup failure too, or return an explicit teardown outcome indicating whether the environment was successfully closed? That would let callers retain or mark the tracking entry for deferred cleanup whenever the environment may still be live.

src/backends/appcontainer/common/src/dispatcher.rs
```diff
@@ -330,6 +330,21 @@ fn select_backend_with_fallback(
     DispatchError,
 > {
     let decision = fallback_detector::detect(&request.policy, /*prefer_bc=*/ true)?;
+    if request.policy.capture_denials.is_some() && decision.tier != IsolationTier::BaseContainer {
```
MGudgin, line 342: ### Reject unsupported `captureDenials` fallback directly in the dispatcher

At this point the dispatcher already knows that `captureDenials` was requested and that the selected fallback tier is AppContainer, which cannot support it. Instead of returning an error here, it constructs an AppContainer runner specifically so that `AppContainerScriptRunner::validate()` will reject the request later with `BackendUnavailable`.

The current behavior is safe because the DACL guard is deliberately not created, but it introduces hidden coupling: this path remains correct only while the AppContainer runner continues performing that validation with the expected failure classification. A future dispatcher or runner refactor could accidentally bypass the rejection or mutate host DACLs first.

Could we represent this as an explicit dispatcher error and map it directly to `BackendUnavailable`? That would keep the unsupported-feature decision at the point where the fallback tier is selected and make the “reject before backend construction or DACL mutation” invariant explicit.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -957,75 +1144,131 @@ impl BaseContainerRunner {
...
+
+        // When captureDenials is active, launch inside a process security
+        // environment that already has a learning-mode trace started against it,
+        // instead of the one-shot CreateProcessInSandbox. `begin` creates the
+        // environment and starts the trace *before* the child launches (so no
+        // early denials are missed); the environment handle is attached to a
+        // normal CreateProcessW call via PROC_THREAD_ATTRIBUTE_SECURITY_ENVIRONMENT.
+        // On any early return below, `capture_session` drops and its Drop
+        // discards the trace and closes the environment (no broker leak).
+        let mut capture_session: Option<CaptureSession> = None;
```
MGudgin: ### Add an injectable seam for the capture lifecycle state machine

The capture path directly constructs the real `CaptureSession`, `SecurityEnvironmentApi`, and `LearningModeApi` implementations. This means ordinary unit tests cannot exercise the runner’s most failure-sensitive transitions without a feature-enabled Windows host.

In particular, there is no deterministic way to test:

- Capture starts successfully but startup-attribute attachment fails.
- Process launch fails after a capture session exists.
- Trace sealing or environment closure fails during `try_wait()` or `wait()`.
- Repeated terminal observations reuse the cached teardown result.
- `Drop` encounters a capture-finalization failure.

The existing tests cover helper functions and error classifiers, but not the state machine that decides when capture teardown runs and how its errors combine with process results.

Could we introduce a small capture-session factory/trait with a test implementation? A fake session that can independently fail `begin`, `finish`, or environment closure would allow these lifecycle branches to be covered in fast, deterministic tests.

MGudgin (changes requested):
### Update the PR description to match the final launch mechanism

The PR description says the child is launched through `CreateProcessAsUserInsideSecurityEnvironment`, but the final implementation no longer uses that export.

The current code attaches the process security environment using `PROC_THREAD_ATTRIBUTE_SECURITY_ENVIRONMENT` and launches the child with `CreateProcessW` plus `EXTENDED_STARTUPINFO_PRESENT`.

Could you update the description so other reviewers and future readers are evaluating the mechanism that will actually merge?

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in a59c0627. A standalone `CloseProcessSecurityEnvironment` `ApiCall` is now classified as a cleanup failure, recursively through combined errors. Normal teardown retains and marks recovery tracking whenever the environment may still be live.

src/backends/appcontainer/common/src/dispatcher.rs
richiemsft: Fixed in a59c0627. The dispatcher now returns an explicit `DispatchError::CaptureDenialsUnsupported` immediately when fallback selects an AppContainer tier, before constructing a backend or applying DACL changes. Existing dispatch error mapping surfaces it as `BackendUnavailable`.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in a59c0627. Added injectable `CaptureSessionFactory` / `CaptureSessionOps` seams, with fake begin and finish failures. The runner now uses the abstraction for prelaunch and teardown transitions, with deterministic tests for begin failure, close-only finalization failure, and combined prelaunch/cleanup errors.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in a59c0627. Native `try_wait()` is now a non-blocking root-exit poll for every configuration and no longer kills descendants or finalizes capture. `wait()`/`Drop` own process-tree teardown; managed `Wait`/`WaitAsync` hand off to native blocking wait after a terminal poll so capture finalization completes before those managed APIs return.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in a59c0627. Startup-attribute failure now explicitly takes and `finish(None)`s the live session before tracking changes. A teardown failure is combined with the attribute error, retains/marks deferred recovery tracking, and is returned in the launch error.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -957,78 +1202,143 @@ impl BaseContainerRunner {
...
-                    spec_bytes.len() as u32, // sandboxSpecificationSize
-                    &mut pi,                 // processInformation
+                CreateProcessW(
+                    PCWSTR::null(),
+                    Some(PWSTR(cmd_wide.as_mut_ptr())),
+                    None,
+                    None,
+                    !inherited_handles.is_empty(),
+                    PROCESS_CREATION_FLAGS(current_creation_flags | EXTENDED_STARTUPINFO_PRESENT.0),
+                    environment,
```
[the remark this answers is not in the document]
richiemsft: Fixed in 4ca8575e. The attribute-based `CreateProcessW` path now always receives an explicit environment block: caller-provided entries when configured, otherwise the shared `CreateEnvironmentBlock(..., bInherit=false)` clean user environment. V1 VM validation set `MXC_HOST_SECRET=super-secret-value` in the parent; the sandbox reported `CLEAN`, exited 0, and produced a non-empty ETL.

MGudgin (approved):

PR #708, opened 2026-07-29
feat(iso): surface structured error fields on the state-aware wire envelope

adpa-ms:
## 📖 Description

IsolationSession failures arrive as one concatenated string in `error.message` — operation, HRESULT, API message and remediation hint flattened together — so callers have to parse prose. This promotes them to discrete fields on the state-aware error envelope: `operation`, `nativeCode` and `remediation`. `message` becomes the bare text, which for a semantic API failure is the API's own message passed through verbatim.

The three fields live in a boxed `ApiFailure` on `MxcError`. Grouping them makes the invariant the easy path — `nativeCode` and `remediation` sit beside `operation` rather than as independent optionals, so the normal construction path cannot produce a status without the call it came from — and keeps `MxcError` under clippy's `result_large_err` threshold. Classification is extracted into pure functions so the rules are unit-testable; `IsoSessionError` is WinRT-activated and cannot be constructed in a test.

The TypeScript `MxcError` gains a constructor overload taking a flat object mirroring the wire shape; the positional signature is retained and declared last, so existing callers and `ConstructorParameters` are unaffected.

Out of scope by design: the one-shot path keeps composing its full string, having no structured envelope to read fields from. Windows Sandbox has no semantic error channel analogous to `IsoSessionError`, so the fields would be uniformly absent — the reference doc now says so explicitly rather than implying the fields are universal. A follow-on can unify the one-shot envelope and carry the fields to the C# SDK.

## ⚠️ Observable changes

The envelope shape is backward-compatible — every new field is optional and absent unless an API call was in flight. Two **content** changes are worth checking before you merge:

- **`message` no longer embeds the operation and HRESULT** for a failure raised by the IsolationSession API; they moved to their own fields. Anything that greps the message string for an HRESULT needs to read `nativeCode` instead. `MxcError`'s `Display` re-attaches them (`code: message [operation nativeCode]`) so log-only consumers keep the detail, but the wire `message` is bare.
- **Provision-time `ERROR_NOT_FOUND` now returns `backend_error`, not `stale_id`.** The promotion applied to every operation, including provision, which cannot produce a stale id because it mints the sandbox — both the doc comment and `docs/isolation-session/state-aware-rust.md` already specified non-provision-only, so this aligns the code with its stated contract. Retry logic branching on `stale_id` during provision behaves differently. The promotion stays semantic-path only by design: a transport HRESULT of the same value lacks the provenance that gives it that meaning, so promoting it would emit a false `stale_id` and tell the caller to destroy a healthy sandbox.

## 🔍 Validation

Tests were added for every decision above, so a later rebase that silently undoes one fails loudly — field population, the invariant across all variants, the camelCase serde rename, both halves of the `stale_id` rule, the legacy positional constructor, and the full chain through the SDK.

Windows x64: fmt, clippy `-D warnings`, build and test with `isolation_session` ON and OFF, `wxc_host_prep` elevated, the versioning and .NET parity gates, SDK build/unit/integration. Isolation-capable host (OS build 26662.1004): one-shot, state-aware and SDK Node E2E suites all green (73 passed, 0 failed), plus the manual interactive TTY tests. No agent users leaked, verified as a diff against a pre-run snapshot of the full local-account set.

The four host-independent policy-validation cases moved out of the runtime-probe-gated suite. They are rejections MXC raises before any IsolationSession API call, so they need `wxc-exec` built with the feature but not a host that can run isolation sessions — they now execute in normal CI, where the whole suite previously skipped.

Observed from the live API on the `stale_id` path:

```json
{"error":{"code":"stale_id","message":"The provision was not found.","operation":"IsoSessionOps.StopSessionAsync","nativeCode":"0x80070490"}}
```

## ✅ Checklist

## 📋 Issue Type

- [x] Feature

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/708)

### DIFF
--- a/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
+++ b/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
@@ -578,7 +578,8 @@ The wire contract is a typed envelope, JSON-serialised, that flows from the SDK 
 executor (`wxc-exec` on Windows, `lxc-exec` on Linux) via the existing `--config-base64`
 CLI argument. Both ends agree on the same shape: the SDK serialises a TypeScript value,
 the executor parses the same value into a Rust struct (§9.1). The only open content in
-the envelope is at the leaves of `ErrorEnvelope.details`.
+the envelope is at the leaves of `ErrorEnvelope.details`; every other field, including
+the error envelope's named structured fields, is statically typed.
 
 ### 7.1 Request envelope
 
@@ -750,6 +751,9 @@ output to a file and leaves stderr as pure script content.
 interface ErrorEnvelope {
   code: ErrorCode;
   message: string;
+  operation?: string;
+  nativeCode?: string;
+  remediation?: string;
   details?: Record<string, unknown>;
 }
 
@@ -778,10 +782,36 @@ Because MXC diagnostic output is routed to `stderr` in state-aware mode, this
 stdout-based discrimination has no false positives or negatives — the content is always
 either pure envelope or pure script output.
 
+`code` and `message` are always present. `code` is the machine-readable category a
+consumer branches on; `message` is the human-readable description, and for a failure
+raised by an underlying platform API it is that API's own message, passed through
+verbatim rather than concatenated with the other fields.
+
+The three optional named fields describe a failure that originated in an underlying
+platform API:
+
+| Field | Meaning |
+|---|---|
+| `operation` | The API call that failed, namespaced by its interface — e.g. `IsoSessionOps.RunProcessWithOptionsAsync`. Low-cardinality and free of call parameters, so it is safe to aggregate on in telemetry. |
+| `nativeCode` | The underlying platform status as a string. An HRESULT such as `0x80070490` on Windows; the field is platform-neutral, so another backend can carry an errno or equivalent. |
+| `remediation` | The API's actionable "how to fix it" hint, when it supplies one. |
+
+**Invariant:** `nativeCode` implies `operation`, and `remediation` implies `operation`.
+`operation` marks that an API operation was in flight; the other two refine it, and
+neither ever appears alone. A failure MXC raises before or outside any API call — a
+malformed request or id, a policy rejection, or an internal failure of MXC's own
+machinery — carries only `code` and `message`.
+
+**Which fields earn a place here.** A named top-level field is for a **backend-neutral**
+concept: `operation`, `nativeCode` and `remediation` all apply equally to a Windows
+HRESULT, a Linux errno, or any other backend's failure. **Backend-specific** structured
+data belongs in `details` instead. That is what keeps `details` from becoming vestigial
+as named fields are added — it remains the designated home for anything without a
+cross-backend meaning.
+
 `ErrorEnvelope.details` is the only `Record<string, unknown>` in the contract. It's the
-escape hatch backends use to convey structured failure information that's
-per-error-code (a backend's native HRESULT, partial output captured before a timeout,
-etc.). Each backend's plan doc (§11) specifies what `details` contains for which error
+escape hatch backends use to convey structured failure information that has no dedicated
+field. Each backend's plan doc (§11) specifies what `details` contains for which error
 codes.
 
 ### 7.4 Worked example: IsolationSession end-to-end
--- a/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
+++ b/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
@@ -792,9 +792,11 @@ platform API:
 
 | Field | Meaning |
 |---|---|
-| `operation` | The API call that failed, namespaced by its interface — e.g. `IsoSessionOps.RunProcessWithOptionsAsync`. Low-cardinality and free of call parameters, so it is safe to aggregate on in telemetry. |
+| `operation` | The API call that failed, namespaced by its interface — e.g. `IsoSessionOps.RunProcessWithOptionsAsync`. Low-cardinality and free of call parameters, so it is safe to aggregate on in telemetry. **Best-effort diagnostic, not a versioned contract** — see below. |
 | `nativeCode` | The underlying platform status as a string. An HRESULT such as `0x80070490` on Windows; the field is platform-neutral, so another backend can carry an errno or equivalent. |
 | `remediation` | The API's actionable "how to fix it" hint, when it supplies one. |
+
+**Stability.** Unlike `code`, which is a closed and versioned enum, the *values* of `operation` and `nativeCode` are **best-effort diagnostics and may change without a schema version bump**. They are derived from the underlying platform API — for IsolationSession, from the projected WinRT class and method names — which MXC does not own and cannot version. Consumers should aggregate on them for telemetry and log them for diagnosis, but branch program logic on `code`, and should not treat a particular `operation` value as a guarantee. (MXC's own end-to-end tests do pin exact values; that is deliberate — they verify MXC's mapping, and move with it in the same change.)
 
 **Invariant:** `nativeCode` implies `operation`, and `remediation` implies `operation`.
 `operation` marks that an API operation was in flight; the other two refine it, and
--- a/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
+++ b/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
@@ -796,6 +796,13 @@ platform API:
 | `nativeCode` | The underlying platform status as a string. An HRESULT such as `0x80070490` on Windows; the field is platform-neutral, so another backend can carry an errno or equivalent. |
 | `remediation` | The API's actionable "how to fix it" hint, when it supplies one. |
 
+**Availability.** These fields are currently populated only by **IsolationSession
+state-aware** operations. Windows Sandbox has no semantic error channel to derive them
+from, and the one-shot surface composes its full detail into `message` instead, so all
+three are uniformly absent there. Other backends may adopt them as they grow an
+equivalent channel — treat all three as optional on every backend, and branch program
+logic on `code` first.
+
 **Stability.** Unlike `code`, which is a closed and versioned enum, the *values* of `operation` and `nativeCode` are **best-effort diagnostics and may change without a schema version bump**. They are derived from the underlying platform API — for IsolationSession, from the projected WinRT class and method names — which MXC does not own and cannot version. Consumers should aggregate on them for telemetry and log them for diagnosis, but branch program logic on `code`, and should not treat a particular `operation` value as a guarantee. (MXC's own end-to-end tests do pin exact values; that is deliberate — they verify MXC's mapping, and move with it in the same change.)
 
 **Invariant:** `nativeCode` implies `operation`, and `remediation` implies `operation`.
--- a/sdk/node/src/errors.ts
+++ b/sdk/node/src/errors.ts
@@ -92,9 +92,19 @@ export class MxcError extends Error {
     message?: string,
     details?: Record<string, unknown>,
   ) {
+    // The overload discriminates on `typeof === 'string'`, so anything else
+    // — including `null` and `undefined` from an `any` cast or an absent wire
+    // envelope — would take the object branch and fail inside `super()` with
+    // a `TypeError` naming `message`, which says nothing about the real
+    // mistake. Reject it here instead, where the message can.
+    if (codeOrFields === null || codeOrFields === undefined) {
+      throw new TypeError(
+        `MxcError: expected an error code string or a field object, got ${String(codeOrFields)}`,
+      );
+    }
     const fields: MxcErrorFields =
       typeof codeOrFields === 'string'
-        ? { code: codeOrFields, message: message as string, details }
+        ? { code: codeOrFields, message: message ?? '', details }
         : codeOrFields;
     super(fields.message);
     this.code = fields.code;
--- a/sdk/node/src/errors.ts
+++ b/sdk/node/src/errors.ts
@@ -5,8 +5,8 @@
  * Closed set of MXC wire-format error codes. Mirrors `MxcErrorCode` on the
  * Rust side one-for-one and serialises as the same snake_case strings on
  * the wire. Backend-specific failures that don't fit one of these codes
- * surface as `backend_error`, with structured information carried in
- * `details`.
+ * surface as `backend_error`, with structured information carried in the
+ * named fields below (or in `details`).
  */
 export type ErrorCode =
   | 'malformed_request'
@@ -23,6 +23,51 @@ export type ErrorCode =
   | 'backend_error';
 
 /**
+ * Every field an `MxcError` can carry, in the same flat shape as the wire
+ * error envelope — `operation`, `nativeCode` and `remediation` sit alongside
+ * `code` and `message`, not nested inside `details`.
+ *
+ * **Invariant:** `nativeCode` implies `operation`, and `remediation` implies
+ * `operation`. `operation` marks that an underlying API call was in flight;
+ * the other two refine it, and neither appears on its own. A failure MXC
+ * raises before or outside any API call carries only `code` and `message`.
+ */
+export interface MxcErrorFields {
+  /** Machine-readable category. Branch on this first. */
+  code: ErrorCode;
+  /** Human-readable description of the failure. */
+  message: string;
+  /**
+   * The underlying API call that failed, namespaced by its interface — e.g.
+   * `IsoSessionOps.RunProcessWithOptionsAsync`. Low-cardinality and free of
+   * call parameters, so it is safe to group on in telemetry.
+   */
+  operation?: string;
+  /**
+   * The underlying platform status as a string — an HRESULT such as
+   * `0x80070490` on Windows, an errno or equivalent elsewhere.
+   */
+  nativeCode?: string;
+  /** The API's actionable "how to fix it" hint, when it supplied one. */
+  remediation?: string;
+  /**
+   * Open extension point for backend-specific structured data that has no
+   * dedicated field. Named fields are reserved for backend-neutral concepts.
+   */
+  details?: Record<string, unknown>;
+}
+
+/**
+ * The `error` arm of a wire response envelope, as received from the
+ * executor. Identical to {@link MxcErrorFields} except that `code` is an
+ * open `string`: an unrecognised code is passed through verbatim rather than
+ * being coerced or dropped.
+ */
+export interface WireError extends Omit<MxcErrorFields, 'code'> {
+  code: string;
+}
+
+/**
  * Typed error thrown by the MXC SDK in response to a wire-format error
  * envelope. Discriminate by comparing `.code` to a wire-format error code
  * string; the TypeScript string-literal union gives the same IDE
@@ -30,12 +75,33 @@ export type ErrorCode =
  */
 export class MxcError extends Error {
   readonly code: ErrorCode;
+  readonly operation?: string;
+  readonly nativeCode?: string;
+  readonly remediation?: string;
   readonly details?: Record<string, unknown>;
 
-  constructor(code: ErrorCode, message: string, details?: Record<string, unknown>) {
-    super(message);
-    this.code = code;
-    this.details = details;
+  /** Canonical form: pass the full field set as one object. */
+  constructor(fields: MxcErrorFields);
+  /**
+   * Positional form, retained for compatibility. Declared last so that
+   * `ConstructorParameters<typeof MxcError>` keeps resolving to this shape.
+   */
+  constructor(code: ErrorCode, message: string, details?: Record<string, unknown>);
+  constructor(
+    codeOrFields: ErrorCode | MxcErrorFields,
+    message?: string,
+    details?: Record<string, unknown>,
+  ) {
+    const fields: MxcErrorFields =
+      typeof codeOrFields === 'string'
+        ? { code: codeOrFields, message: message as string, details }
+        : codeOrFields;
+    super(fields.message);
+    this.code = fields.code;
+    this.operation = fields.operation;
+    this.nativeCode = fields.nativeCode;
+    this.remediation = fields.remediation;
+    this.details = fields.details;
     // Restore the prototype chain so `instanceof MxcError` keeps working
     // after the TypeScript ES2020 → ES5-compatible class downlevelling.
     Object.setPrototypeOf(this, new.target.prototype);
@@ -48,6 +114,9 @@ export class MxcError extends Error {
  * `string` so callers parsing a wire envelope don't need to narrow first;
  * unknown codes still produce an `MxcError` with `.code` set to whatever
  * was on the wire.
+ *
+ * For a complete wire envelope prefer {@link mxcErrorFromEnvelope}, which
+ * carries the structured fields too.
  */
 export function mxcErrorFromCode(
   code: string,
@@ -55,4 +124,22 @@ export function mxcErrorFromCode(
   details?: Record<string, unknown>,
 ): MxcError {
   return new MxcError(code as ErrorCode, message, details);
+}
+
+/**
+ * Constructs an `MxcError` from the `error` arm of a wire response envelope.
+ *
+ * This is the single place the wire's open `code` string is widened to the
+ * closed `ErrorCode` union, so unknown-code passthrough behaves identically
+ * everywhere the SDK parses an envelope.
+ */
+export function mxcErrorFromEnvelope(error: WireError): MxcError {
+  return new MxcError({
+    code: error.code as ErrorCode,
+    message: error.message,
+    operation: error.operation,
+    nativeCode: error.nativeCode,
+    remediation: error.remediation,
+    details: error.details,
+  });
 }
--- a/sdk/node/tests/integration/isolation-session-state-aware.test.ts
+++ b/sdk/node/tests/integration/isolation-session-state-aware.test.ts
@@ -28,9 +28,14 @@ import {
 } from '@microsoft/mxc-sdk';
 import { probeStateAwareRuntime, safeDeprovision, sandboxSkipReason } from './test-helpers.js';
 
-const skipReason = os.platform() !== 'win32'
-  ? 'IsolationSession is Windows-only'
-  : sandboxSkipReason ?? await probeStateAwareRuntime('isolation_session');
+const platformSkipReason =
+  os.platform() !== 'win32' ? 'IsolationSession is Windows-only' : undefined;
+
+// Host-dependent gate: adds the runtime probe (and the CI opt-out) on top of
+// the platform check. `??` short-circuits, so the probe is never spawned on a
+// non-Windows host.
+const skipReason =
+  platformSkipReason ?? sandboxSkipReason ?? (await probeStateAwareRuntime('isolation_session'));
 
 describe('IsolationSession state-aware lifecycle E2E', { skip: skipReason }, () => {
   it('runs full lifecycle: provision -> start -> exec -> stop -> deprovision', async () => {
@@ -159,13 +164,26 @@ describe('IsolationSession state-aware lifecycle E2E', { skip: skipReason }, () 
       await safeDeprovision(sandboxId);
     }
   });
+});
 
-  // --- Runtime backend guard for the network acknowledgment ----------------
+// Policy rejections are raised by MXC's own validation, before any
+// IsolationSession API call: the dispatcher runs `validate_provision` ahead of
+// `provision`, and `IsolationSessionRunner` is a stateless marker whose
+// construction touches no WinRT. So these need a `wxc-exec.exe` built with
+// `--features isolation_session` (which CI builds) but *not* a host that can
+// actually run isolation sessions.
+//
+// Keeping them out of the probe-gated suite above is deliberate. That suite is
+// additionally gated on `sandboxSkipReason`, and both CI systems set
+// `MXC_SKIP_OS_BUILD_DEPENDENT_TESTS=1`, so anything inside it never runs in
+// CI. These assertions cover the full chain this feature depends on — Rust
+// envelope serialisation → dispatcher → SDK parse → typed `MxcError` — which
+// is exactly the path where drift would otherwise go unnoticed.
+describe('IsolationSession state-aware policy validation', { skip: platformSkipReason }, () => {
   // The TypeScript type makes `network` required (and pins its value) at
   // provision, but a plain-JS caller can bypass that. These assert the backend
   // itself refuses a missing or non-canonical network acknowledgment — the
   // "respect or refuse" guarantee must not rest on the compile-time type alone.
-  // Validation runs before the OS service is touched, so no cleanup is needed.
   type UntypedProvision = (
     containment: 'isolation_session',
     config: unknown,
--- a/sdk/node/tests/integration/isolation-session-state-aware.test.ts
+++ b/sdk/node/tests/integration/isolation-session-state-aware.test.ts
@@ -20,6 +20,7 @@ import path from 'node:path';
 import os from 'os';
 import {
   execInSandboxAsync,
+  IsolationSessionUserConfig,
   MxcError,
   provisionSandbox,
   startSandbox,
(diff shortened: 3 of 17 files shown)


src/core/wxc_common/src/mxc_error.rs
```diff
@@ -57,17 +57,77 @@ impl std::fmt::Display for MxcErrorCode {
...
+///
+/// A failure MXC raises itself — a malformed request, a policy rejection, or
+/// an internal failure with no API call in flight — leaves it unset and so
+/// carries only `code` and `message`.
+///
+/// A new *backend-neutral* concept earns a field on `ApiFailure`;
+/// *backend-specific* structured data belongs in `details`, which stays open
+/// for that purpose.
 #[derive(Debug, Clone, PartialEq, Eq, Error)]
 #[error("{code}: {message}")]
```
MGudgin: **[High · backward-compatibility] `message` loses the operation and HRESULT for every log-only consumer**

The PR moves `operation` and `nativeCode` out of the concatenated string into discrete
fields, and `message` becomes "the bare text". But the `Display` impl is unchanged:

```rust
#[error("{code}: {message}")]   // <- still only code + message
pub struct MxcError {
    ...
    pub api_failure: Option<Box<ApiFailure>>,   // never rendered
}
```

So a caller that *destructures* the envelope gains the new fields, but a caller that
just logs the error — `error!("{e}")`, `e.to_string()`, or on the TS side anything
reading `Error.message` — silently loses the operation name and the HRESULT that used
to be right there in the string. That is a diagnostic regression on the exact path
most on-call tooling takes, and it ships with no version signal: the PR is described
as "not a breaking change", and the shape *is* compatible, but the **content** of an
existing field changed.

Concretely, a failure that previously logged as:

```
backend_error: IsoSessionOps.StopSessionAsync failed: 0x80070490: The provision was not found.
```

now logs as:

```
stale_id: The provision was not found.
```

The HRESULT is still on the wire — but not in any string a logger will emit.

Suggested fix: render `api_failure` in `Display` when present, which recovers the old
diagnostic content without touching the wire contract.

```rust
impl fmt::Display for MxcError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{}: {}", self.code, self.message)?;
        if let Some(api) = &self.api_failure {
            write!(f, " [{}", api.operation)?;
            if let Some(nc) = &api.native_code {
                write!(f, " {nc}")?;
            }
            write!(f, "]")?;
        }
        Ok(())
    }
}
```

(That means dropping the derived `#[error("...")]` for a manual impl, or keeping the
derive and adding the detail via a helper — either is fine.) The TS side wants the
equivalent treatment when building the `Error` message.

If you would rather keep `Display` minimal, then this is a deliberate, observable
behaviour change to `message` and should be called out in the PR description and
release notes rather than filed under "not a breaking change" — otherwise consumers
discover it from a gap in their telemetry.

src/core/wxc_common/src/mxc_error.rs
```diff
@@ -134,14 +225,27 @@ impl MxcError {
...
 /// Wire shape of the `error` arm. `code` is a closed `MxcErrorCode` that
-/// serialises to its snake_case wire string; `details` is omitted from JSON
-/// when absent.
+/// serialises to its snake_case wire string; every optional field is omitted
+/// from JSON when absent.
+///
+/// See [`MxcError`] for the meaning of the structured failure fields and the
+/// invariant relating them.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ErrorEnvelope {
```
MGudgin, line 265: **[Medium · backward-compatibility] New public fields on exhaustive structs — worth `#[non_exhaustive]` now**

This PR adds fields to two public structs in `wxc_common`: `api_failure` on `MxcError`,
and `operation` / `native_code` / `remediation` here on `ErrorEnvelope`. Neither struct
carries `#[non_exhaustive]`, so any struct-literal construction or exhaustive
destructuring outside this module breaks on each addition.

**Caveat, so this is not overstated:** `wxc_common` is an internal workspace crate with
no crates.io release — every consumer is in-tree and updated in the same commit, which
is why your build is green and why this is not a live break. I would have called it
High if the crate were published; as-is it is forward-looking hygiene.

The reason to do it anyway: this is the third field added to this envelope in as many
PRs, the PR itself notes a follow-on will carry the fields to the C# SDK and unify the
one-shot path, and `mxc_ffi` / `mxc-sdk` are explicitly positioned as binding surfaces.
`#[non_exhaustive]` costs one line each and makes every future field a non-event:

```rust
#[non_exhaustive]
pub struct MxcError { .. }

#[non_exhaustive]
pub struct ErrorEnvelope { .. }
```

In-tree construction keeps working via `MxcError::new(..)` / the `with_*` builders,
which is already the documented path.

Related, and the part I would actually gate on: the PR description says "not a breaking
change", which is true of the *shape* but not the *content* — see my comment on the
`Display` impl above.

src/core/wxc_common/src/mxc_error.rs
```diff
@@ -57,17 +57,77 @@ impl std::fmt::Display for MxcErrorCode {
...
+pub struct ApiFailure {
+    /// The API call that failed, namespaced by its interface — e.g.
+    /// `IsoSessionOps.RunProcessWithOptionsAsync`. Kept low-cardinality and
+    /// free of call parameters so it can be grouped in telemetry.
+    pub operation: String,
+    /// The underlying platform status as a string, e.g. `0x80070490`.
+    pub native_code: Option<String>,
+    /// The API's actionable "how to fix it" hint, when it supplies one.
+    pub remediation: Option<String>,
+}
```
MGudgin: **[Medium · maintainability + correctness] The "unrepresentable to violate" invariant is representable, in three ways**

*(Surfaced independently by two reviewer axes on two different models — the
highest-confidence design finding in this PR.)*

The doc comment states the design goal plainly: grouping the fields makes it so
"`native_code` and `remediation` cannot exist without `operation`". Grouping gets you
*co-location*, not *enforcement*. Three ways to violate it as written:

1. `ApiFailure::default()` — the derived `Default` yields `operation: ""`, so
   `ApiFailure::default().with_native_code("0x80070490")` is a native code with no
   operation. The builders make this the path of least resistance.
2. Direct struct literal — fields are `pub`, so
   `ApiFailure { operation: String::new(), native_code: Some(..), remediation: None }`
   compiles anywhere in the workspace.
3. **`ErrorEnvelope` does not group them at all** — it carries `operation`,
   `native_code`, `remediation` as three independent `Option<String>` siblings, each
   `#[serde(default)]`. So this deserializes cleanly today:

   ```json
   { "code": "backend_error", "message": "boom", "nativeCode": "0x80070490" }
   ```

   giving `operation: None` alongside `native_code: Some(..)` — the exact state the type
   was meant to make impossible, reachable from untrusted input with no `unsafe` and no
   misuse.

Since `MxcError` is the shared error type for every backend and all three SDKs, whatever
this type permits is what future backends will do. Worth closing before anyone else
adopts the pattern.

Suggested fix — make the constructor the only door:

```rust
#[derive(Debug, Clone, PartialEq, Eq)]   // no Default
pub struct ApiFailure {
    operation: String,                    // private
    native_code: Option<String>,
    remediation: Option<String>,
}

impl ApiFailure {
    /// Panics/errors on an empty operation — the invariant's anchor.
    pub fn new(operation: impl Into<String>) -> Self {
        let operation = operation.into();
        debug_assert!(!operation.is_empty(), "ApiFailure requires an operation");
        Self { operation, native_code: None, remediation: None }
    }
    pub fn operation(&self) -> &str { &self.operation }
    pub fn native_code(&self) -> Option<&str> { self.native_code.as_deref() }
    pub fn remediation(&self) -> Option<&str> { self.remediation.as_deref() }
}
```

For `ErrorEnvelope`, either nest the three fields behind a
`#[serde(flatten)] Option<ApiFailure>` so serde reconstructs the group as a unit, or add
post-deserialize validation that rejects `native_code` / `remediation` without
`operation`. Then the doc comment's claim becomes true.

If enforcement is deliberately out of scope, the comment should say "the grouping
documents the invariant" rather than "unrepresentable to violate" — right now the type
promises more than it delivers.

sdk/node/src/errors.ts
```diff
@@ -30,12 +75,33 @@ export type ErrorCode =
...
+  constructor(
+    codeOrFields: ErrorCode | MxcErrorFields,
+    message?: string,
+    details?: Record<string, unknown>,
+  ) {
+    const fields: MxcErrorFields =
+      typeof codeOrFields === 'string'
+        ? { code: codeOrFields, message: message as string, details }
+        : codeOrFields;
+    super(fields.message);
```
MGudgin, line 109: **[High · correctness] Constructor throws `TypeError` on nullish input instead of a usable error**

The overload discriminates on `typeof === 'string'`, so anything that is not a string
takes the object branch — including `null` and `undefined`. `fields` is then nullish and
`super(fields.message)` throws:

```
TypeError: Cannot read properties of undefined (reading 'message')
```

That matters because `MxcError` is on the error path by construction — it is what runs
when something has already gone wrong. A nullish arg reaching it (an `any` cast, a
truncated or absent wire envelope, a `catch` that forwards an unparsed body) converts a
diagnosable backend failure into an opaque `TypeError` from inside the SDK, losing the
original context entirely. The old positional-only signature had the same shape of
hazard, but the object overload widens the set of callers who can hit it.

Two cheap options — fail loudly, but with a message that names the real problem:

```ts
if (codeOrFields == null) {
  throw new TypeError('MxcError: expected an error code or field object, got ' + codeOrFields);
}
```

or degrade to a valid error rather than throwing from a constructor:

```ts
const fields: MxcErrorFields =
  typeof codeOrFields === 'string'
    ? { code: codeOrFields, message: message ?? '', details }
    : (codeOrFields ?? { code: 'backend_error', message: 'Unknown error' });
```

Either way, a unit test pinning `new (MxcError as any)(undefined)` and
`new (MxcError as any)(null)` would lock the behaviour in — the existing tests cover both
valid forms but not the nullish edge.

Also worth guarding `message` in the string branch: `message as string` is an unchecked
assertion, so `new MxcError('stale_id')` (message omitted) yields `message: undefined`
and an `Error` whose `.message` is the string `"undefined"`.

sdk/node/src/errors.ts
```diff
@@ -22,6 +22,51 @@ export type ErrorCode =
...
+
+/**
+ * The `error` arm of a wire response envelope, as received from the
+ * executor. Identical to {@link MxcErrorFields} except that `code` is an
+ * open `string`: an unrecognised code is passed through verbatim rather than
+ * being coerced or dropped.
+ */
+export interface WireError extends Omit<MxcErrorFields, 'code'> {
+  code: string;
+}
```
MGudgin, line 68: **[Medium · proportionality] Internal envelope parsing became permanent public API**

The feature needs the new readonly properties on `MxcError`. It does not obviously need
`MxcErrorFields` and `WireError` exported, nor the object-form constructor overload, nor
`mxcErrorFromEnvelope` — these serve three internal parsing sites, but they are
re-exported from `sdk/node/src/index.ts`, which makes them supported surface the SDK has
to keep working indefinitely.

Public API is the hardest thing to walk back: once these ship, a consumer can build
against `WireError` and any later change to the wire-parsing internals becomes a
breaking change to them, for no benefit either side asked for.

Suggested: keep `MxcErrorFields`, `WireError` and `mxcErrorFromEnvelope` module-internal
(drop them from `index.ts`), preserve the existing public constructor signature, and
export only the new readonly `operation` / `nativeCode` / `remediation` properties. If
the object-form constructor turns out to be wanted externally, adding it later is
additive and safe — removing it is not.

src/backends/isolation_session/common/src/error.rs
```diff
@@ -47,106 +215,236 @@ impl From<IsolationSessionError> for ScriptResponse {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum StalePromotion {
+    /// Non-provision operations address an existing agent user, so
+    /// `ERROR_NOT_FOUND` means that user is gone.
+    Eligible,
+    /// Provision *mints* the agent user. There is no `sandboxId` yet, so
+    /// reporting `stale_id` — whose remediation is "re-provision; treat the
+    /// id as dead" — would be incoherent.
+    NotEligible,
+}
```
MGudgin, line 292: **[Medium · proportionality] + [Low · backward-compatibility] The `stale_id` narrowing is a separate, wire-visible change**

Two axes landed on this same code, from different directions — recording both here.

**Proportionality:** `StalePromotion` changes provision-time `ERROR_NOT_FOUND` from
`stale_id` to `backend_error`. That is an independently observable classification
change, not plumbing required by the structured-fields feature. It rides along in a PR
whose stated purpose is promoting error components to discrete fields, which means the
two cannot be reviewed, reverted, or bisected independently. Suggest splitting it (and
its tests) into its own PR — it is a clean, self-contained fix that would land fast on
its own merits.

**Backward-compatibility:** it is a wire-visible change to the emitted `code` for an
existing failure path. The old behaviour was incoherent — the doc comment here makes
that case well, and I agree with the direction — but any caller with retry logic
branching on `stale_id` during provision behaves differently after this merges. That
deserves an explicit line in the PR description / release notes, not just a code
comment, so downstream consumers can check their handling.

No change requested to the logic itself; the `Eligible` / `NotEligible` split and the
"semantic-path only" reasoning in the doc comment below are both right, and the
"Do not 'fix' the asymmetry" note is a good thing to have written down.

src/backends/isolation_session/common/src/error.rs
```diff
@@ -3,40 +3,208 @@
...
-    ServiceUnavailable(String),
+    /// The in-proc IsolationSession runtime API is not available on this
+    /// host (not registered, or the OS feature gate is off). This is a real
+    /// COM activation failure, so it does carry the operation and the
+    /// HRESULT.
+    ServiceUnavailable(IsoApiFailure),
     /// A lifecycle step (provision / start / exec / stop / deprovision)
-    /// returned a failure from the OS API.
-    Lifecycle(String),
+    /// failed.
```
MGudgin, line 176: **[Medium · testability] Fallback classification is stranded outside the pure seam**

The extraction of `classify_api_failure` into a pure function is a genuine win — it is
the right call given `IsoSessionError` is WinRT-activated and cannot be constructed in a
unit test, and the PR deserves credit for it.

The gap: the fallback path that handles a *failed* `err.Code()` getter still lives
inside `format_iso_error`, which takes `&IsoSessionError`. So the behaviour on that path
— appending the read error to the message and producing an `IsoApiFailure` with no code
— is unreachable from `#[cfg(test)]`. It is precisely the kind of rarely-exercised
degenerate branch that unit tests exist to pin, and it is the one branch the new seam
does not cover.

Suggested: have `format_iso_error` do only the WinRT reads, lowering them to plain Rust
types (`Result<HRESULT, _>`, `Option<String>`, `Option<String>`), then hand those to a
pure extraction function that owns the whole decision tree including the fallback. That
moves the last piece of logic behind the testable boundary and leaves `format_iso_error`
as a thin, obviously-correct adapter.

src/backends/isolation_session/common/src/error.rs
```diff
@@ -47,106 +215,236 @@ impl From<IsolationSessionError> for ScriptResponse {
...
+/// on, while the finer detail rides in the message.
+pub(super) fn transport_err(
+    operation: &str,
+    step: &str,
+    err: &windows_core::Error,
+) -> IsolationSessionError {
+    IsolationSessionError::Lifecycle(LifecycleFailure::Api(IsoApiFailure::new(
+        operation,
+        Some(err.code().0 as u32),
+        Some(format!("{}: {}", step, err.message())),
```
MGudgin: **[Low · correctness] `transport_err` emits a dangling `"step: "`, defeating the `NO_API_MESSAGE` guard**

The message is unconditionally formatted as `format!("{}: {}", step, err.message())`. If
the underlying `windows_core::Error` carries no message text, this produces a string like
`"wait failed: "` — non-empty, so the `is_empty()` check inside `IsoApiFailure::new`
never fires and the `NO_API_MESSAGE` fallback is skipped. The caller gets a trailing
colon and no explanation.

Worth fixing because it lands exactly on what the second commit
("never emit an empty error message") set out to prevent: the guard is in place, but
this path routes around it by producing a technically-non-empty string.

```rust
let message = if err.message().is_empty() {
    step.to_string()
} else {
    format!("{}: {}", step, err.message())
};
```

A unit test with an empty-message `windows_core::Error` would lock in both halves —
that the fallback fires, and that a populated message is still prefixed with `step`.

sdk/node/tests/integration/isolation-session-state-aware.test.ts
```diff
@@ -200,4 +201,35 @@ describe('IsolationSession state-aware lifecycle E2E', { skip: skipReason }, ()
...
+        },
+        { experimental: true },
+      ),
+      (err: unknown) => {
+        assert.ok(err instanceof MxcError, `expected MxcError, got ${String(err)}`);
+        assert.strictEqual(err.code, 'policy_validation');
+        assert.match(err.message, /upn/i, `expected the message to name upn: ${err.message}`);
+        assert.strictEqual(err.operation, undefined, 'operation must be absent');
+        assert.strictEqual(err.nativeCode, undefined, 'nativeCode must be absent');
+        assert.strictEqual(err.remediation, undefined, 'remediation must be absent');
```
MGudgin, line 248: **[Medium · test coverage] These structured-field assertions do not run in default CI**

This is a well-constructed negative case — asserting the fields reach the caller *absent*
rather than empty is exactly the right test for the invariant, and the comment explaining
why the canonical network acknowledgment is supplied is genuinely helpful.

The problem is that it never executes on a normal CI run. The enclosing `describe` is
gated at line 31-35 by `skipReason`, which resolves via `sandboxSkipReason` /
`probeStateAwareRuntime('isolation_session')`; the default SDK integration workflow sets
`MXC_SKIP_OS_BUILD_DEPENDENT_TESTS=1`, so the whole suite skips. The result is that the
full chain this PR is really changing — Rust envelope serialization to dispatcher to SDK
parse to typed `MxcError` — has no automated regression gate. The unit tests cover the
mapping in isolation and the wire shape in isolation, but nothing joins them, and
serialization drift between the two is the most likely way this feature silently breaks.

Note this particular case does not actually need an isolation-capable host: it is a
policy rejection raised by MXC's own validation before any IsolationSession API call, as
the comment above says. So it could run anywhere `wxc-exec` is built with the feature.

Suggested: lift the host-independent cases (this one, and the `policy_validation` case
above) into a `describe` that is not gated on `probeStateAwareRuntime`, leaving only the
genuinely host-dependent lifecycle tests behind the skip. That converts the end-to-end
claim into something CI can actually defend.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -778,10 +782,38 @@ Because MXC diagnostic output is routed to `stderr` in state-aware mode, this
 stdout-based discrimination has no false positives or negatives — the content is always
 either pure envelope or pure script output.
 
+`code` and `message` are always present. `code` is the machine-readable category a
+consumer branches on; `message` is the human-readable description, and for a failure
+raised by an underlying platform API it is that API's own message, passed through
+verbatim rather than concatenated with the other fields.
+
+The three optional named fields describe a failure that originated in an underlying
```
MGudgin, line 790: **[Medium · documentation-drift] Cross-backend contract presents the fields as universally available**

This section is the cross-backend state-aware contract, and it describes the structured
failure fields as applying equally to any backend. As implemented, they are populated
only by IsolationSession state-aware operations: the one-shot path still composes its
full string (deliberately, per the PR description), and Windows Sandbox has no semantic
error channel analogous to `IsoSessionError`, so they will be uniformly absent there.

A consumer reading this document — which is the cross-backend reference, not the
IsolationSession guide — would reasonably write code that reads `nativeCode` from any
state-aware backend and find it always `undefined` everywhere except one. Since the
fields are optional, that failure is silent.

Suggested: add an availability note right where the fields are introduced, e.g.
"Currently populated only by IsolationSession state-aware operations. Absent for Windows
Sandbox and for the one-shot surface; other backends may adopt them, so treat all three
as optional and always branch on `code` first."

That also sets the expectation correctly for the follow-on work the PR describes, rather
than implying the gap is a bug.

MGudgin (changes requested):
I think this is pretty much as we discussed. Some comments from findings by my review skill.

src/core/wxc_common/src/mxc_error.rs
adpa-ms: You're right, and the doc comment is corrected in `d453262`.

Verified all three sub-claims: `Default` was derived, so `ApiFailure::default()` yields an empty `operation`; the fields are `pub`, so a literal can pair an empty `operation` with a `native_code`; and `operation()` returns it unfiltered, so `to_envelope()` would emit `"operation": ""` alongside `"nativeCode"`. "Unrepresentable" was the wrong word.

The comment now claims only what's true — grouping makes the invariant the easy path and keeps the normal construction path from producing a status without its call, while noting the fields are `pub` for destructuring so a hand-rolled literal can still violate it. I also dropped the `Default` derive, which was unused and was the main way to reach an empty operation without writing a deliberate literal.

For calibration: I checked reachability before deciding how far to go. `ApiFailure::default()` has zero call sites workspace-wide, there are no `ApiFailure { .. }` literals, and Rust only ever deserializes `ErrorEnvelope` inside one round-trip test — production is serialize-only, with the TS SDK on the consuming end. So there's no live path to a violating value today. That's why I fixed the claim and removed the derive rather than making the fields private, which would have cost destructuring for a hazard nothing currently reaches. Happy to go further if you'd rather have the type enforce it outright.

src/core/wxc_common/src/mxc_error.rs
adpa-ms: Agreed, and your suggested fix is the right shape — it recovers the diagnostic content without touching the wire contract, which is what makes it safe to take.

Confirmed the regression is real: the base branch's `map_lifecycle_error` did `let message = err.to_string()`, so the operation and HRESULT were genuinely in the string a logger emits. They aren't any more.

Implemented in `d453262`: dropped the `thiserror` derive for an explicit `Display` plus a manual `std::error::Error` impl, rendering `code: message [operation nativeCode]`, with the bracketed section omitted entirely when there's no `api_failure`.

One deliberate constraint on top of your sketch: I added a test asserting the enrichment does **not** leak into `to_envelope()`. The bare `message` on the wire is a documented property of this design, so `Display` and the envelope needed to be pinned apart — otherwise a later "simplification" could reasonably fold one into the other.

Tests: `display_appends_operation_and_native_code`, `display_renders_operation_without_native_code`, `display_without_api_failure_is_code_and_message_only`, and `display_enrichment_does_not_leak_into_the_envelope`.

On dropping the `thiserror` derive: `impl std::error::Error for MxcError {}` is intentionally empty. The derive generated no `source()` either — the struct has no nested error — so this is behavior-preserving, not a silent loss of error chaining.

On the framing point you and the `#[non_exhaustive]` comment both raised: you're right that "not a breaking change" was true of the shape but not the content. The PR description now has an explicit **Observable changes** section covering both this and the `stale_id` narrowing.

Small correction for the record: the illustrative before/after strings aren't literal — the base rendering also carried an `"Isolation Session lifecycle error: "` prefix, and the `backend_error` → `stale_id` shift in the example is a separate axis. The substance is unaffected; I verified the underlying claim directly against the base branch.

src/core/wxc_common/src/mxc_error.rs
adpa-ms: Taken, and your caveat is accurate — `wxc_common` has no `publish` key, so this was forward-looking hygiene rather than a live break.

I checked the cost before applying: there are zero `ErrorEnvelope { .. }` and zero `MxcError { .. }` struct literals anywhere outside the defining module (the only literals are in `mxc_error.rs` itself — one in `to_envelope` and four in tests), and `#[non_exhaustive]` doesn't restrict the defining crate. So it's genuinely free today, which is the best time to add it.

Both attributes are in `d453262`.

To be explicit about the one risk here: `#[non_exhaustive]` does **not** break the `Serialize`/`Deserialize` derives on `ErrorEnvelope`. Serde's generated code lives in the defining crate, where the attribute doesn't apply, so construction during deserialization is unaffected. Verified by the full workspace test run rather than by inspection — the existing `ErrorEnvelope` round-trip test still passes.

src/backends/isolation_session/common/src/error.rs
adpa-ms: Correct. `format_iso_error` takes `&IsoSessionError`, which is WinRT-activated, so nothing inside it is reachable from a `#[cfg(test)]` build — and the `Err(read_err)` arm had real logic in it: the note composition, the merge with an existing message, and the decision to leave the code `None`.

Extracted as `unreadable_code_failure(operation, message, remediation, read_err)` in `d453262`. `format_iso_error` is now purely the WinRT adapter, matching the split you already endorsed for `classify_api_failure`.

Two tests: `unreadable_code_keeps_the_message_and_omits_native_code` pins the merged message and the `operation`-present / `nativeCode`-absent shape, and `unreadable_code_never_promotes_to_stale` asserts this path can never promote — the code is precisely what couldn't be read, so there's nothing to classify on.

src/backends/isolation_session/common/src/error.rs
adpa-ms: Confirmed, and reproduced. This is the most consequential finding in the review, and I'd rate it above Low: it defeats the guard that the immediately preceding commit existed to add.

I probed `windows_core::Error::message()` across a range of HRESULTs:

```
0x00000000 → "The operation completed successfully."
0x00000001 → "Incorrect function."
0x80004005 → "Unspecified error"
0xDEADBEEF → ""        ← empty
```

An HRESULT with no OS message-table entry returns an empty string, so the join produced exactly the `"wait failed: "` you predicted — non-empty, therefore straight past the `is_empty()` check in `IsoApiFailure::new`. Any custom facility code the service invents lands in the same bucket, so this is reachable rather than theoretical.

Fixed in `d453262` with the fallback you suggested, extracted as `compose_transport_message` so it's directly testable. Worth noting the leverage: 34 production call sites route through that single join, so the one fix covers all of them.

Regression gates, by name: `transport_message_without_platform_text_is_the_step_alone`, `transport_message_with_platform_text_keeps_the_step_prefix`, and `transport_failure_with_unmapped_hresult_has_no_dangling_separator`. The third asserts `err.message().is_empty()` on `0xDEADBEEF` *before* asserting the outcome, so if that HRESULT ever gains a message-table entry the test fails rather than passing vacuously.

src/backends/isolation_session/common/src/error.rs
adpa-ms: Taking these in reverse order, because I'm accepting one and pushing back on the other.

**On the release-note point — accepted.** You're right that a code comment isn't where a consumer looks. The PR description now has an **Observable changes** section stating plainly that provision-time `ERROR_NOT_FOUND` returns `backend_error` instead of `stale_id`, and that retry logic branching on `stale_id` during provision behaves differently.

**On splitting it into its own PR — I'd like to push back**, on the grounds that it isn't actually an independent behavior change. Both descriptions of that code on the base branch already specified the narrow behavior:

- the base code comment: *"**Every non-provision lifecycle op** (start / exec / stop / deprovision) surfaces this HRESULT… we promote it to `Stale`"*
- `docs/isolation-session/state-aware-rust.md:227` on the base branch: *"After `deprovision`, **every non-provision op** against the dead `sandboxId` triggers this."*

Only the code disagreed — it promoted unconditionally. So this aligns the code with a contract that was already written down in two places, and the rewrite is what forced the discrepancy into the open.

Both quotations are verifiable at the merge base, `86ceab7f5acb014e4af0b55fdfaa820718476bfb` ("fix(isolation-session): refuse network policy the backend cannot enforce… (#682)"):

```
git show 86ceab7:src/backends/isolation_session/common/src/error.rs
git show 86ceab7:docs/isolation-session/state-aware-rust.md
```

In the first, the promotion is `if code == ERROR_NOT_FOUND_HRESULT { IsolationSessionError::Stale(formatted) }` — no operation filter — sitting directly beneath the comment claiming non-provision-only. The doc quote is line 227 of the second.

I also searched every `stale_id` reference outside Rust: `stale_id|StaleId` across `sdk/`, `tests/` and `docs/`, 43 matches. The only behavioral consumers are `tests/scripts/run_isolation_session_state_aware_tests.ps1:634` (stop on a deprovisioned sandbox), `docs/isolation-session/state-aware-rust.md:192` (the deprovision row) and `sdk/node/README.md:277` (generic re-provision guidance) — all non-provision. Nothing in-tree branches on provision-time `stale_id`, and no doc ever described it as reachable.

Splitting is also awkward in both directions. Backward means reimplementing the narrowing against the pre-rewrite `format_iso_error` that this PR deletes, then rebasing through it. Forward means shipping `Eligible` everywhere and leaving the code contradicting two docs that are themselves in this PR.

Given you've endorsed the logic and the behavior is now called out explicitly, I'd rather keep it here than pay that churn. Happy to be overruled if you still think the bisect story justifies it.

sdk/node/src/errors.ts
adpa-ms: Accepted for two of the three, with a specific reason for keeping the third.

Your premise checks out exactly: the base branch exported `ErrorCode`, `MxcError`, `mxcErrorFromCode`, and this PR added exactly three more for three internal call sites. `WireError` and `mxcErrorFromEnvelope` are now un-exported from `index.ts` in `d453262`. All three consumers import from `./errors.js` directly rather than through the barrel, so it was zero-churn. The SDK's job is to parse envelopes, so a consumer shouldn't need the helper.

**`MxcErrorFields` I've kept exported**, because un-exporting it doesn't reduce the public surface — it's the parameter type of a *public* constructor overload on a *public* class. I built a standalone reproduction to check rather than reason about it. With the interface un-exported and `declaration: true`, it compiles (exit 0) and emits:

```ts
interface MxcErrorFields { ... }          // declared, not exported
export declare class MxcError extends Error {
    constructor(fields: MxcErrorFields);  // public ctor still references it
}
export {};
```

A consumer can then still call `new MxcError({ code, message, operation })` — structural typing — but `import type { MxcErrorFields }` fails with `TS2459: Module declares 'MxcErrorFields' locally, but it is not exported`. So we'd carry the identical commitment while making it impossible to name in a helper signature or variable annotation.

Reproduction details so this is checkable: TypeScript 5.9.3, `"declaration": true`, `"strict": true`, `"module"`/`"moduleResolution": "NodeNext"` — i.e. the SDK's own `tsconfig.json` settings. The consumer-side failure is on `import type { MxcErrorFields }`; the `new MxcError({ ... })` call on the line above it compiles fine, which is precisely the asymmetry that makes un-exporting counterproductive.

Also relevant to the overload's value: `errors.test.ts` has `keeps ConstructorParameters resolving to the positional form`, pinning `ConstructorParameters<typeof MxcError>` to `['stale_id', 'boom']`. The overload ordering is load-bearing and already guarded, so the object form is additive at the type level too.

The only way to genuinely remove `MxcErrorFields` from the public surface is to drop the object-form overload, which is a larger change than the one proposed and would give up the non-breaking migration path the overload exists to provide. If you think that's the right trade I'm open to discussing it separately, but it seemed out of proportion to the concern as written.

sdk/node/src/errors.ts
adpa-ms: The primary finding is correct and now fixed; the secondary one I think is inaccurate, with evidence below.

**Nullish input — confirmed.** Probed against a fresh build: `undefined` and `null` both produce exactly `TypeError: Cannot read properties of undefined (reading 'message')`. `d453262` adds an explicit guard throwing `MxcError: expected an error code string or a field object, got <value>`, with tests pinning both `null` and `undefined`.

I chose to fail loudly rather than degrade to a valid error. Degrading would manufacture a wire-shaped `backend_error` that never came from a backend, and a consumer branching on `.code` would then act on a fiction — a clear TypeError naming the real mistake seemed more useful on a path that exists to explain failures.

**On `new MxcError('stale_id')` yielding `.message === "undefined"`** — this one doesn't reproduce. The observed value is the empty string:

```
'stale_id' (message omitted) → code="stale_id" message=""
```

The mechanism, not just the observation: per ECMA-262's `Error` constructor, the own `message` property is created **only** when the argument is not `undefined`. With it omitted, `.message` falls through to `Error.prototype.message`, which is `''`. So `"undefined"` was never reachable through this path — it would require explicitly passing the *string* `"undefined"`, or a `String(...)` coercion that isn't in the code.

The unchecked `as string` is a real hazard regardless — just not that symptom. I've fixed it with `message ?? ''`, which makes the assertion honest instead of load-bearing, and added `yields an empty message when the positional message is omitted` so the actual behavior is pinned.

Flagging it only because a future reader could otherwise go looking for a `"undefined"` string that never existed.

sdk/node/tests/integration/isolation-session-state-aware.test.ts
adpa-ms: Confirmed end to end, and this is now fixed — thank you, it's the finding with the most durable value.

I traced the whole chain to be sure the lift would actually work:

1. `Build.Windows.Job.yml` builds with `features: hyperlight isolation_session microvm wslc`, so the packaged `wxc-exec.exe` has the feature.
2. `Package.NpmSdk.Job.yml` pulls those artifacts into `sdk/node/bin` and packs them; the integration job consumes that artifact.
3. `mxc_engine::run_state_aware` dispatches with no host-availability probe beforehand.
4. `IsolationSessionRunner` is a stateless marker (`pub struct IsolationSessionRunner;`) — construction touches no WinRT.
5. `dispatch_state_aware` calls `validate_provision` before `provision`, which the existing `dispatch_provision_returns_validate_error_without_calling_provision` test already pins.
6. `validate_isolation_session_user` is a pure string shape check.
7. WinRT activation happens in `IsolationSessionManager::new`, reached only inside `provision`.

So these are genuinely host-independent, exactly as you said. `d453262` moves all four policy-validation cases — the three network-acknowledgment guards plus the structured-fields one — into a suite gated only on `os.platform() !== 'win32'`, leaving the lifecycle tests behind the runtime probe.

Reproducible locally without an isolation-capable host:

```
$env:MXC_SKIP_OS_BUILD_DEPENDENT_TESTS = "1"
npm run test:integration        # from sdk/node, with wxc-exec built --features isolation_session
```

Before this change the suite reported the whole file skipped. After it, the four cases execute — including `a policy rejection reaches the SDK with no structured failure fields`, which is the one that actually exercises Rust envelope serialization → dispatcher → SDK parse → typed `MxcError`.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
adpa-ms: Agreed — the "backend-neutral" language in that section is about which concepts deserve a top-level field, and it reads as availability. Nothing in any of the four changed docs said which backends actually populate them.

`d453262` adds an availability note where the fields are introduced, closely following your suggested wording: currently populated only by IsolationSession state-aware operations, absent for Windows Sandbox and the one-shot surface, other backends may adopt them, treat all three as optional and branch on `code` first. The same note is now in the SDK README beside the field docs, since that's where a TypeScript consumer is most likely to be reading.

Worth noting there was already an in-repo idiom for exactly this that the PR failed to apply — `sdk/node/README.md` describes `uiCapabilities` as *"currently populated only by the Windows native probe… Linux and macOS omit the field until their probes expose equivalent data."* The new notes are written to match.

PR #699, opened 2026-07-28
Add sealed ETL denial analysis

richiemsft:
## 📖 Description

Adds the denial-analysis layer on top of Integration PR #696.

- Introduces the cross-platform `learning_mode_core` denial model, summary, analysis, framing, and emission primitives.
- Adds the Windows sealed-ETL decoder using OpenTrace/ProcessTrace and TDH metadata.
- Decodes file, registry, UI, and capability-related Learning Mode events into normalized denial resources.
- Adds path normalization, typed property extraction, and composition tests using representative real event shapes.
- Adds `lm_analyze` for manually validating captured ETL files.

This PR is intentionally stacked on #696 and should merge after it.

## 🔗 References

- #696
- #661

## 🔍 Validation

- `cargo test -p learning_mode_core`
- `cargo test -p learning_mode_windows --all-targets`
- `cargo clippy -p learning_mode_core -p learning_mode_windows --all-targets -- -D warnings`
- `cargo check --workspace --all-targets`
- `cargo clippy --workspace --all-targets -- -D warnings`

## ✅ Checklist

- [x] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [x] Linked to an issue
- [ ] Updated documentation (if applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed)
- [ ] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes (see [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md))

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

GitHub Actions runs the PR validation build automatically. The ADO pipeline
(`MXC-PR-Build`) is the Azure version of the PR pipeline, kept in parity with the GitHub
Actions build; it runs on merge to `main`, and Microsoft reviewers with write access can trigger it
on a PR with `/azp run`. See [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md).

If the `dependency-feed-check` check fails on a new dependency, the crate must be added to
the feed before the PR can pass. See [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md)
for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/699)

PR #709, opened 2026-07-30
Add sealed ETL denial analysis (replacement for #699)

richiemsft:
## 📖 Description

Replacement for #699, which was mistakenly marked merged when the parent
branch temporarily received child-stack history during a stack-handling
mistake. This PR restores the intended denial-analysis review boundary after
merged Integration PR #696.

Adds the denial-analysis layer following #696:

- Introduces the cross-platform `learning_mode_core` denial model, summary, analysis, framing, and emission primitives.
- Adds the Windows sealed-ETL decoder using OpenTrace/ProcessTrace and TDH metadata.
- Decodes and normalizes file, registry, UI, and capability-related Learning Mode events.
- Caches manifest TDH schemas per trace while bypassing the cache for self-describing TraceLogging events.
- Contains all Rust panics inside the ETW callback boundary and resumes them only after the trace handle is closed.
- Bounds property counts, struct depth, total decode work, metadata references, and payload reads.
- Skips malformed individual payload events while preserving fatal schema/API errors.
- Handles fixed-width unsupported InTypes without losing payload alignment.
- Uses deterministic synthetic drive maps for path-normalization tests.
- Streams raw diagnostics and adds `lm_analyze` as a developer-only trace diagnostic.

PR #696 has merged. This PR now targets `main` directly.

## 🔗 References

- Replaces #699
- #696
- #661

## 🔍 Validation

- `cargo test -p learning_mode_core`
- `cargo test -p learning_mode_windows --all-targets` — 82 tests
- `cargo clippy -p learning_mode_core -p learning_mode_windows --all-targets -- -D warnings`
- Feature-enabled V1 Windows VM `172.17.7.153`:
  - captured a fresh 2,713-byte ETL
  - production analysis exited 0 and emitted 9 unique denials
  - raw diagnostic traversal exited 0

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---


src/backends/learning_mode/windows/Cargo.toml
```diff
@@ -9,6 +9,7 @@ thiserror = { workspace = true }
 
 [target.'cfg(target_os = "windows")'.dependencies]
 wxc_common = { workspace = true }
+learning_mode_core = { workspace = true }
```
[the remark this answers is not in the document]
richiemsft: Fixed in af3d5bcb: the architecture instructions now list `learning_mode_core` and document the `learning_mode_windows` dependency accurately.

src/backends/learning_mode/windows/examples/lm_analyze.rs
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Decode a sealed learning-mode `.etl` into the captureDenials RFC 7464
+//! JSON text sequence, or dump its raw ETW events for schema discovery.
+//!
+//! This is a developer diagnostic for inspecting captured traces manually.
+//! End users and SDK agents do not invoke it: the BaseContainer runner seals
+//! and consumes the trace automatically during normal `captureDenials` use.
```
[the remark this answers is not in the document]
richiemsft: Fixed in af3d5bcb: the diagnostic docs now state that BaseContainer only seals and reports the ETL path, and that analysis remains manual until runner integration.

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -0,0 +1,659 @@
...
+
+    if flags & PROPERTY_STRUCT != 0 {
+        let start_index = unsafe { prop_info.Anonymous1.structType.StructStartIndex } as usize;
+        let member_count = unsafe { prop_info.Anonymous1.structType.NumOfStructMembers } as usize;
+        for _ in 0..count {
+            for child_index in start_index..start_index + member_count {
+                let _ = decode_property(child_index, context, offset, numeric_values)?;
+            }
+        }
+        return Ok((prop_name, "<struct>".to_string()));
```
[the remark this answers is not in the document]
richiemsft: Fixed in af3d5bcb: struct decoding now retains the parent marker and flattens decoded child properties into the raw property list. Added a synthetic TDH metadata regression test.

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -0,0 +1,698 @@
...
+    pid: u32,
+    filetime: u64,
+) -> Option<RawDenial> {
+    // If the record explicitly reports the check as not-denied, skip it;
+    // absence of the field is treated as a denial (fail open on decode gap).
+    if let Some(denied) = find_prop(&parts.props, "Denied") {
+        if !denied.trim_matches('"').eq_ignore_ascii_case("true") {
+            return None;
+        }
+    }
```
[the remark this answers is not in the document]
richiemsft: Fixed in af3d5bcb: event 28 now requires `Denied` to be present and explicitly true, with a regression test for the missing-field case.

MGudgin (commented):
Adversarial review findings for the sealed ETL denial-analysis layer.

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -0,0 +1,624 @@
...
+        return;
+    }
+
+    let provider = header.ProviderId;
+    let event_id = header.EventDescriptor.Id;
+    if matches!(acc.mode, CollectionMode::Analyze) && !is_learning_mode_event(provider, event_id) {
+        return;
+    }
+
+    match unsafe { tdh_decode::decode_event_parts(event_record) } {
```
MGudgin: ### Prevent panics from escaping the ETW callback

`event_record_callback` is an `extern "system"` function invoked by `ProcessTrace`. The raw diagnostic visitor is protected with `catch_unwind`, but the production analyze path calls TDH decoding, denial extraction, path normalization, and accumulation without equivalent protection.

If any of that Rust code panics—for example, because a decoder invariant is violated or a future extractor introduces an unchecked assumption—the panic cannot unwind safely through the Win32 callback boundary. Rust will normally abort the process instead of returning an analysis error.

Could we wrap the complete callback processing body in `catch_unwind`? The callback could store the panic payload or a converted decode error in the accumulator and return normally to ETW. After `ProcessTrace` returns and `CloseTrace` has run, the Rust caller can safely resume the panic or return `AnalyzeError`.

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -0,0 +1,738 @@
...
+            std::ptr::null()
+        };
+        let (value, consumed) = format_property_value_with_pointer_size(
+            in_type,
+            declared_length,
+            data_ptr,
+            remaining,
+            context.pointer_size,
+        );
+        if consumed == 0 && remaining > 0 {
```
MGudgin, line 393: ### Do not fail the complete trace for an unsupported fixed-width property

The fallback formatter consumes `declared_length`, but TDH reports a zero declared length for several fixed-width InTypes whose size is implicit. In those cases the fallback returns `consumed == 0`, this branch converts that into a fatal decode error, and `EtlDenialAnalyzer` discards the entire capture after one unsupported field.

Examples include GUID, FLOAT, DOUBLE, FILETIME, and SYSTEMTIME properties. This also contradicts the module's stated behavior that unsupported values become placeholders while offset arithmetic continues.

Could we teach the formatter the fixed byte sizes of known-but-unformatted InTypes? If offset recovery is impossible, please reject only that event with precise provider/event/property context rather than poisoning every subsequent event in the trace.

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -0,0 +1,738 @@
...
+/// # Safety
+/// `event_record` must point to a valid `EVENT_RECORD` provided by the
+/// ETW callback; the caller must not retain references to its fields
+/// after the callback returns.
+pub unsafe fn decode_event_parts(
+    event_record: *mut EVENT_RECORD,
+) -> Result<DecodedEventParts, String> {
+    let mut buf_size: u32 = 0;
+    // First call: discover required buffer size. ERROR_INSUFFICIENT_BUFFER = 122.
+    let status = unsafe { TdhGetEventInformation(event_record, None, None, &mut buf_size) };
```
MGudgin, line 206: ### Cache TDH event schemas for the duration of the trace

This performs the two-call `TdhGetEventInformation` size/data sequence and allocates a new schema buffer for every matching event. The schema is invariant for a provider/event/version tuple, while Learning Mode traces can contain a large number of repeated access-check events.

Could we cache the validated `TRACE_EVENT_INFO` bytes by provider GUID, event ID, and version for the duration of the trace? That would avoid repeated manifest resolution, two Win32 calls, and an allocation on the per-event hot path.

src/backends/learning_mode/windows/src/path_norm.rs
```diff
@@ -0,0 +1,291 @@
...
+            return Some(format!(r"\\{rest}"));
+        }
+        return None;
+    }
+
+    if !kernel_path.starts_with(r"\Device\") {
+        return None;
+    }
+
+    let map = DRIVE_MAP.get_or_init(load_drive_map);
```
MGudgin: ### Add an injectable drive-map seam for deterministic tests

`to_user_visible` always initializes this process-global cache from the machine's live `QueryDosDeviceW` results. Tests that exercise the public `\Device\...` path therefore depend on the CI host's drive layout, reconstruct expectations from the same API, and may silently skip assertions when no suitable mapping is present.

Could we extract a `to_user_visible_with_map` helper, with `to_user_visible` passing the cached production map? Unit tests could then exercise the complete public normalization flow using fixed synthetic mappings, independently of the host machine and `OnceLock` state.

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -0,0 +1,738 @@
...
+
+struct PropertyDecodeContext<'a> {
+    info_buf: &'a [u8],
+    info: &'a TRACE_EVENT_INFO,
+    user_data: *const u8,
+    user_data_len: usize,
+    pointer_size: usize,
+}
+
+fn decode_property(
```
MGudgin, line 301: ### Add adversarial tests around the unsafe TDH decoder

The happy paths are well covered, but the branches immediately surrounding the unsafe payload reads need focused malformed-input tests. In particular, please cover out-of-range count/length metadata references, truncated ANSI and UTF-16 payloads, unsupported fixed-width InTypes with zero declared length, and invalid struct member ranges.

These tests would lock in the decoder's bounds and error behavior and would directly cover the all-or-nothing failure path discussed below, rather than relying only on valid synthetic events.

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -0,0 +1,738 @@
...
+            "property '{prop_name}' count {count} exceeds limit {MAX_PROPERTY_ELEMENTS}"
+        ));
+    }
+
+    if flags & PROPERTY_STRUCT != 0 {
+        let start_index = unsafe { prop_info.Anonymous1.structType.StructStartIndex } as usize;
+        let member_count = unsafe { prop_info.Anonymous1.structType.NumOfStructMembers } as usize;
+        let mut values = vec![(prop_name, "<struct>".to_string())];
+        for _ in 0..count {
+            for child_index in start_index..start_index + member_count {
```
MGudgin: ### Bound recursive TDH struct decoding explicitly

This recursion trusts the TDH struct graph and uses `start_index + member_count` without checked arithmetic or a depth budget. The production Learning Mode provider should have a valid schema, but this safety invariant is implicit, and the raw-event diagnostic path can encounter arbitrary registered providers.

Could we use `checked_add`, validate the complete member range against `PropertyCount`, and pass a small recursion-depth budget through `decode_property`? That would make the unsafe decoder's assumptions explicit and prevent cycles or malformed metadata from causing stack exhaustion.

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -0,0 +1,738 @@
...
+        values.push(value);
+    }
+    numeric_values[index] = numeric_value;
+
+    let rendered = if values.len() == 1 {
+        values.pop().unwrap_or_default()
+    } else {
+        format!("[{}]", values.join(", "))
+    };
+    Ok(vec![(prop_name, rendered)])
```
MGudgin: ### Avoid allocating a temporary one-element vector for scalar properties

Every scalar property allocates a new `Vec` containing one tuple, only for the caller to immediately `extend` another vector and discard this allocation. This occurs for every property of every decoded event.

Could `decode_property` push into a caller-owned result vector, or return a representation that does not require a heap allocation for the common scalar case?

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -0,0 +1,704 @@
...
+        27 => build_denial_from_learning_mode(parts, pid, filetime),
+        28 => build_denial_from_capability(parts, pid, filetime),
+        _ => None,
+    }
+}
+
+/// Whether a provider/event pair belongs to the Learning Mode capture schema.
+pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
+    if provider == KERNEL_GENERAL_PROVIDER {
+        matches!(event_id, 14 | 27 | 28)
```
MGudgin: ### Use shared constants for the accepted event IDs

The event vocabulary is repeated as numeric literals in both filtering and routing (`14`, `27`, `28`, and `4907`). If one match is updated without the other, an event can be accepted and then dropped or routed inconsistently.

Could we define named event-ID constants and use them in both `is_learning_mode_event` and `extract_denial`? This would make the supported provider/event vocabulary a single source of truth.

src/backends/learning_mode/windows/src/tdh_decode.rs
richiemsft: Fixed in f5c0f96a. Added adversarial coverage for out-of-range count/length references, truncated and odd-length strings, unsupported fixed-width zero-length values, invalid/empty/cyclic structs, TraceLogging schema detection, and total decode-work exhaustion.

src/backends/learning_mode/windows/src/tdh_decode.rs
richiemsft: Fixed in f5c0f96a. Added a per-trace TDH schema cache keyed by provider and the full event descriptor, capped at 4096 entries. Self-describing TraceLogging events are detected through `EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL` and always use their uncached embedded schema.

src/backends/learning_mode/windows/src/tdh_decode.rs
richiemsft: Fixed in f5c0f96a. Struct ranges use checked arithmetic and full PropertyCount validation, empty structs are rejected, recursion is capped at 32, and a shared 100,000-step work budget prevents multiplicative struct/count expansion.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: Fixed in f5c0f96a. Added named constants for access-check, UI violation, capability denial, and privacy access-check event IDs and reused them for both filtering and routing.

src/backends/learning_mode/windows/src/path_norm.rs
richiemsft: Fixed in f5c0f96a. Extracted `to_user_visible_with_map`; production passes the OnceLock drive map only for device paths, while canonicalization and component-boundary tests now use fixed synthetic mappings independent of the host.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Fixed in f5c0f96a. The complete production callback body now runs under `catch_unwind`; the panic payload is stored in the accumulator, `ProcessTrace` returns normally, `CloseTrace` runs, and only then is the panic resumed.

src/backends/learning_mode/windows/src/tdh_decode.rs
richiemsft: Fixed in f5c0f96a. `decode_property` now pushes directly into the caller-owned result vector; scalar properties no longer allocate a temporary one-element vector.

src/backends/learning_mode/windows/src/tdh_decode.rs
richiemsft: Fixed in f5c0f96a. Known fixed-width unsupported InTypes now use their implicit SDK sizes (including GUID/FLOAT/DOUBLE/FILETIME/SYSTEMTIME and the correct 306-308 character/size_t IDs). Unrecoverable payload corruption skips only that analyze event; raw diagnostics retain contextual errors.

MGudgin (approved):

MGudgin (dismissed):
You'll need to retarget at main (which may then require new signoff, not sure...)

richiemsft:
@mgudgin, yup it resetted the sign-off.

MGudgin (approved):

PR #710, opened 2026-07-30
Emit captureDenials JSON output (replacement for #701)

richiemsft:
## 📖 Description

Replacement for #701, restoring the intended output-layer review boundary
following replacement analysis PR #709.

Adds the caller-facing `captureDenials` output layer:

- Decodes the sealed ETL and writes one JSON document containing `denials` and
  `summary`.
- Generates unique per-run internal ETL and JSON paths.
- Returns the output path and summary as structured teardown metadata instead
  of writing from the backend to process-global stderr.
- Prints the structured pointer only at the `wxc-exec` CLI boundary, while the
  Rust SDK, C ABI, and C# SDK expose the metadata programmatically.
- Bounds ETL analysis to 10,000 unique denials and 1,000,000 processed events,
  stopping `ProcessTrace` and reporting truncation when either limit is reached.
- Uses the `DenialAnalyzer` abstraction at the output integration seam, with
  deterministic success, empty-result, decode-failure, and ETL-cleanup tests.
- Removes incomplete output files on write failure and propagates pointer JSON
  serialization errors instead of emitting an empty-path fallback.
- Resolves well-known capability SIDs to policy names and falls back to custom
  SID strings.

PR #709 has merged. This PR targets `main` directly and now contains only the
output layer.

## 🔗 References

- Replaces #701
- Related to #709
- Related to #696
- Related to #661
- Related to #663

## 🔍 Validation

- `cargo test -p learning_mode_core -p learning_mode_windows -p appcontainer_common -p mxc-sdk -p mxc_ffi --all-targets`
- `cargo test -p wxc --no-run`
- `cargo clippy -p learning_mode_windows -p appcontainer_common -p mxc-sdk -p mxc_ffi --all-targets -- -D warnings`
- `dotnet build sdk/dotnet/Microsoft.Mxc.Sdk/Microsoft.Mxc.Sdk.csproj --no-restore`
- Feature-enabled V1 Windows VM (`172.17.7.153`):
  - `wxc-exec` produced a valid JSON document and printed one matching pointer
    line at the CLI boundary.
  - An in-process `mxc_engine` run returned the same structured metadata while
    leaving host stderr empty.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

with write access can trigger it on a PR with `/azp run`. See
[docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md).

If the `dependency-feed-check` check fails on a new dependency, the crate must
be added to the feed before the PR can pass. See
[docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md)
for the steps.


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1646,6 +1675,163 @@ impl BaseContainerSandboxProcess {
...
+        );
+        let document = DenialsDocument::new(analysis.denials, summary);
+
+        write_denials_output_file(output_path, |writer| write_document(writer, &document))?;
+
+        // Surface a single structured pointer line on our own stderr so a
+        // caller can locate the deliverable (the authoritative summary lives in
+        // the file).
+        let pointer = DenialsOutputPointer::new(output_path.to_string_lossy(), &document.summary);
+        let stderr = std::io::stderr();
```
MGudgin: ### Do not publish the pointer through process-global stderr in the backend

`decode_and_emit_denials` writes the pointer directly to `std::io::stderr()`. This backend is used not only by `wxc-exec`, but also by the in-process Rust, C#, and streaming SDK paths through `mxc_engine` with `StdioMode::Pipes`.

For those callers, the pointer bypasses the sandbox process's returned stderr stream and is written into the embedding application's host stderr. The SDK therefore cannot capture or return it through its API, and a library call can unexpectedly write to global process output.

This also places the `wxc-exec` CLI presentation contract inside the backend execution layer.

Could we return the `DenialsOutputPointer` as structured teardown/output metadata instead? The CLI boundary could print it to stderr, while SDK and FFI callers could receive it programmatically through their defined result surface.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1593,23 +1613,32 @@ impl BaseContainerSandboxProcess {
...
             let output_path = self.capture_output_path.take();
-            match session.finish(output_path.as_deref()) {
-                Ok(()) => {
-                    if let Some(path) = &output_path {
-                        eprintln!("captureDenials: denials ETL written to {}", path.display());
+            let exit_code = self.last_exit_code.unwrap_or(-1);
+            let result = match session.finish(etl_path.as_deref()) {
+                Ok(()) => match (&etl_path, &output_path) {
+                    (Some(etl), Some(output)) => {
+                        Self::decode_and_emit_denials(etl, output, exit_code)
```
MGudgin: ### Bound the synchronous ETL processing performed before terminal completion

This adds complete ETL decoding, JSON serialization, file flushing, and pointer publication to `run_teardown()`. That path runs synchronously before `wait()` returns and also sits behind capture-aware `try_wait()` and `Drop`.

The 10,000-unique-denial limit bounds retained output, but `ProcessTrace` still scans every event in the ETL. A high-volume or long-running capture can therefore add unbounded post-exit latency to a polling, wait, or teardown call.

Could we add an explicit event/trace processing bound or stop `ProcessTrace` once the supported output cap is reached? If the output must be complete before returning the final result, the contract should remain synchronous but have a bounded worst-case cost.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1646,6 +1675,163 @@ impl BaseContainerSandboxProcess {
...
     }
+
+    /// Decodes the sealed ETL at `etl_path` into the JSON denials document at
+    /// `output_path`, then prints a one-line [`DenialsOutputPointer`] to stderr.
+    fn decode_and_emit_denials(
+        etl_path: &std::path::Path,
+        output_path: &std::path::Path,
+        exit_code: i32,
+    ) -> std::io::Result<()> {
+        let analysis = EtlDenialAnalyzer.analyze(etl_path).map_err(|error| {
```
MGudgin: ### Use the existing analyzer abstraction at this integration seam

`learning_mode_core::DenialAnalyzer` is object-safe, but this integration function hardcodes `EtlDenialAnalyzer`. That prevents ordinary unit tests from driving the complete decode → document → file → pointer pipeline without a real sealed ETL and feature-enabled Windows environment.

As a result, there is no deterministic test for analyzer failure propagation, ETL cleanup after decode failure, or a controlled successful analysis producing the expected document and pointer.

Could this function accept `&dyn DenialAnalyzer` (or a generic analyzer), with production passing `EtlDenialAnalyzer`? A fake analyzer could then cover success, empty results, and decode failure through the actual integration path.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1646,6 +1675,163 @@ impl BaseContainerSandboxProcess {
...
+    writer.flush()
+}
+
+fn publish_denials_output_pointer(
+    output_path: &Path,
+    writer: &mut impl std::io::Write,
+    pointer: &DenialsOutputPointer,
+) -> std::io::Result<()> {
+    if let Err(pointer_error) = write_denials_output_pointer(writer, pointer) {
+        return match std::fs::remove_file(output_path) {
```
MGudgin: ### Share the unpublished-output cleanup/error-combination logic

This block and the write-failure block below both remove the output file, ignore `NotFound`, and combine a cleanup failure with the primary error. Maintaining two copies makes it easy for their behavior or diagnostics to diverge.

Could we extract a helper that accepts the output path, primary error, and cleanup context, then performs the remove-and-combine operation once for both write and pointer-publication failures?

src/core/learning_mode_core/src/emit.rs
```diff
@@ -1,83 +1,166 @@
...
+            denied_resources_truncated: summary.denied_resources_truncated,
+        }
+    }
+
+    /// Serialises the pointer to a single-line JSON string (no trailing
+    /// newline). A fixed-shape struct of strings/ints/bools always
+    /// serialises, so this is effectively infallible.
+    #[must_use]
+    pub fn to_line(&self) -> String {
+        serde_json::to_string(self).unwrap_or_else(|_| {
```
MGudgin: ### Do not emit a success-shaped pointer with an empty `outputPath`

This structure contains only strings, integers, and a boolean, so normal JSON serialization is effectively infallible. If serialization nevertheless returns an error, the fallback emits a syntactically valid pointer while discarding the most important field: `outputPath`.

That converts an internal serialization failure into a misleading success-shaped message that callers cannot use.

Could `to_line` return `Result<String, serde_json::Error>` and let pointer publication propagate the failure? If this is intentionally treated as an invariant, an explicit invariant failure is still safer than publishing a pointer with an empty path.

MGudgin (commented):
Adversarial review findings for the captureDenials JSON output layer.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 956db58b. The backend now returns neutral `SandboxOutputMetadata` through `SandboxProcess`/`ScriptResponse` instead of writing to process-global stderr. `wxc-exec` prints the pointer at the CLI boundary, while the Rust SDK, C ABI, and C# SDK expose it programmatically. The V1 VM in-process probe returned metadata with empty host stderr.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 956db58b. ETL analysis now stops after 1,000,000 processed events or once an additional unique denial crosses the 10,000-result cap. The ETW buffer callback cancels `ProcessTrace`, and analysis reports `deniedResourcesTruncated: true`.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 956db58b. The integration seam accepts `&dyn DenialAnalyzer`; production passes `EtlDenialAnalyzer`. Fake-analyzer tests cover populated output, empty output, decode failure, and removal of the internal ETL after decode failure.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 956db58b. Pointer publication was removed from the backend entirely, so the duplicate unpublished-output cleanup path no longer exists. File-write failure retains the single incomplete-output cleanup path, and capture/ETL cleanup errors use the shared result combiner.

src/core/learning_mode_core/src/emit.rs
richiemsft: Fixed in 956db58b. `DenialsOutputPointer::to_line()` now returns `Result<String, serde_json::Error>` and no longer emits a success-shaped empty-path fallback. CLI-boundary serialization surfaces an explicit error.

MGudgin (approved):

bbonaby (approved):
merge conflict resolution fixes

MGudgin (approved):

PR #712, opened 2026-07-30
Add --wait-for-debugger flag for BaseContainer/AppContainer runners

asklar:
## 📖 Description

Adds a Windows-only `--wait-for-debugger` CLI flag to `wxc-exec.exe` for the BaseContainer/AppContainer (`processcontainer`) backends. When set, the sandboxed child is left suspended right after creation — before it has executed any code at all, not even its own loader init — instead of being resumed immediately. This lets an external, unsandboxed debugger attach to the *real* sandboxed process by PID at its own pace (no timeout, since nothing can run until the debugger releases it).

`wxc-exec` polls `CheckRemoteDebuggerPresent` (no timeout) and, the instant a debugger attaches, clears only its own `CREATE_SUSPENDED` increment — leaving the debugger's own attach-time freeze in place, so a plain `g` is enough afterward (no `~0 m` needed). Stderr/log output on suspend/resume is a single concise line (`Process suspended, PID: <pid>...` / `Process started, PID: <pid>.`) rather than a long banner.

Changes are scoped to just this feature: the new CLI flag, the `ExecutionRequest.wait_for_debugger` field threaded through the wire/domain config, and the two BaseContainer/AppContainer suspend points. `README.md` is updated with a new "Wait for Debugger" subsection under Debugging.

## 🔍 Validation

- `cargo build -p wxc -p appcontainer_common --target x86_64-pc-windows-msvc` (debug) — passes
- `cargo fmt --all -- --check` — passes
- `cargo clippy --target x86_64-pc-windows-msvc --locked --all-targets --all-features --release -- -D warnings` (full workspace, matching CI's Windows lint job exactly) — passes with zero warnings
- Manual: built `wxc-exec.exe` and confirmed it runs

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type
- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/712)

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,16 @@ wxc-exec.exe --audit policy.json
 
 > **Warning:** `--audit` injects `permissiveLearningMode` — AppContainer restrictions are **not** enforced for the duration of the run. Use only for policy authoring. It cannot be combined with `processContainer.captureDenials`; use `captureDenials.mode: "allow"` for permissive application-driven capture. `learningModeLogging` and `permissiveLearningMode` are reserved internal capability names and are rejected in `processContainer.capabilities`. See [docs/learning-mode/capabilities.md](docs/learning-mode/capabilities.md) for the three learning-mode flows.
 
+### Wait for Debugger
+
+`--wait-for-debugger` holds the sandboxed child suspended right after creation — before it has run any code, including its own loader initialization — instead of resuming it immediately. This lets an external, unsandboxed debugger attach to the *real* sandboxed process by PID at its own pace (there's no timeout, since nothing can run until a debugger attaches):
+
+```bash
+wxc-exec.exe --wait-for-debugger config.json
+```
+
+Attach a debugger to the printed PID and set breakpoints (necessarily deferred/pending ones, since no dependent DLL past the main image has loaded yet). The instant `wxc-exec` detects the attach it automatically clears its own suspend hold, so a plain `g` is all you need afterward — no `~0 m` ("Resume Thread") required. Windows-only; only the AppContainer tier of the `processcontainer` backend implements the suspend point (BaseContainer is skipped automatically, since it cannot guarantee `CREATE_SUSPENDED` is honored on every OS build).
+
 ## Telemetry (Experimental)
 
 MXC supports optional TraceLogging ETW telemetry for execution observability. When enabled, structured events (`MXC.Execution` and `MXC.Error`) are emitted to the local ETW subsystem via the Rust [`tracelogging`](https://crates.io/crates/tracelogging) crate. Every event includes common fields (Version, Channel, IsDebugging, `UTCReplace_AppSessionGuid`) as Part C custom event data.
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ wxc-exec.exe --audit policy.json
 wxc-exec.exe --wait-for-debugger config.json
 ```
 
-Attach a debugger to the printed PID and set breakpoints (necessarily deferred/pending ones, since no dependent DLL past the main image has loaded yet). The instant `wxc-exec` detects the attach it automatically clears its own suspend hold, so a plain `g` is all you need afterward — no `~0 m` ("Resume Thread") required. Windows-only; only the AppContainer tier of the `processcontainer` backend implements the suspend point (BaseContainer is skipped automatically, since it cannot guarantee `CREATE_SUSPENDED` is honored on every OS build).
+Attach a debugger to the printed PID and set breakpoints (necessarily deferred/pending ones, since no dependent DLL past the main image has loaded yet). The instant `wxc-exec` detects the attach it automatically clears its own suspend hold, so a plain `g` is all you need afterward — no `~0 m` ("Resume Thread") required. Windows-only; only the AppContainer tier of the `processcontainer` backend implements the suspend point (BaseContainer is skipped automatically, since it cannot guarantee `CREATE_SUSPENDED` is honored on every OS build). Combining `--wait-for-debugger` with any other `containment` backend (Windows Sandbox, WSLC, IsolationSession, Hyperlight, MicroVM, LXC, Bubblewrap, Seatbelt) is rejected up front, since none of them expose a pre-execution suspend point.
 
 ## Telemetry (Experimental)
 
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -1967,69 +1967,28 @@ impl BaseContainerRunner {
             }
         };
 
-        // `--wait-for-debugger`: block (no timeout) until an external
-        // debugger attaches to the real sandboxed PID, then clear only
-        // wxc-exec's own CREATE_SUSPENDED hold — see `debugger_wait` module
-        // docs for why this leaves the debugger's own attach-time freeze in
-        // place (so a plain `g` is enough, no `~0 m` needed). Unreachable in
-        // practice: `validate()` rejects `--wait-for-debugger` for this
-        // backend before `spawn_base` runs (BaseContainer cannot guarantee
-        // CREATE_SUSPENDED is honored on every OS build). Handled here
-        // anyway so a direct/test caller that bypasses `validate()` still
-        // fails closed instead of silently violating the suspend guarantee.
-        if request.wait_for_debugger {
-            if let Err(e) = crate::debugger_wait::wait_for_debugger_then_resume(
-                pi.hProcess,
-                pi.hThread,
-                pi.dwProcessId,
-                logger,
-            ) {
-                let _ = writeln!(
-                    logger,
-                    "Error: --wait-for-debugger failed ({e}); the sandboxed child has \
-                     been terminated."
-                );
-                // `wait_for_debugger_then_resume` already terminated the
-                // child on failure; reap it and tear down the same
-                // sandbox/proxy state the job-setup-failure path above does.
-                unsafe {
-                    let _ = WaitForSingleObject(pi.hProcess, u32::MAX);
-                    let _ = CloseHandle(pi.hProcess);
-                    let _ = CloseHandle(pi.hThread);
-                }
-                if request.lifecycle.destroy_on_exit {
-                    run_sandbox_cleanup(
-                        &identity,
-                        &sid_string,
-                        request.policy.network_proxy.is_enabled(),
-                        logger,
-                    );
-                    sandbox_tracking::unregister_ctrl_c_cleanup();
-                }
-                self.proxy_coordinator.stop(logger);
+        // `--wait-for-debugger` never reaches this point: `spawn()` (the only
+        // caller of `spawn_base`) runs `self.validate(request)?` first, and
+        // `validate()` unconditionally rejects the flag for this backend
+        // (BaseContainer cannot guarantee CREATE_SUSPENDED is honored on
+        // every OS build -- see `debugger_wait` module docs for what that
+        // guarantee protects). Asserted rather than handled with a second,
+        // untested parallel error path: if a future refactor of `spawn()`
+        // ever drops the `validate()` call, this fires instead of silently
+        // resuming a child the operator expected to stay suspended.
+        debug_assert!(
+            !request.wait_for_debugger,
+            "wait_for_debugger must be rejected by validate() before reaching spawn_base"
+        );
 
-                const WAIT_FOR_DEBUGGER_FAILED_MSG: &str =
-                    "--wait-for-debugger failed while waiting for a debugger to attach; \
-                     the sandboxed child was terminated.";
-                return Err(ScriptResponse {
-                    exit_code: -1,
-                    error_message: WAIT_FOR_DEBUGGER_FAILED_MSG.to_string(),
-                    standard_err: WAIT_FOR_DEBUGGER_FAILED_MSG.to_string(),
-                    extended_error: format!("wait_for_debugger_then_resume failed: {e}"),
-                    failure_phase: FailurePhase::LaunchFailed,
-                    ..Default::default()
-                });
-            }
-        } else {
-            // The child was created suspended; now that it is in the job object (so
-            // every descendant it spawns is captured), resume its main thread. If the
-            // create API ignored CREATE_SUSPENDED the thread is already running and
-            // this is a harmless no-op.
-            // SAFETY: `pi.hThread` is the just-created, still-owned main-thread
-            // handle; `ResumeThread` only adjusts its suspend count.
-            unsafe {
-                ResumeThread(pi.hThread);
-            }
+        // The child was created suspended; now that it is in the job object (so
+        // every descendant it spawns is captured), resume its main thread. If the
+        // create API ignored CREATE_SUSPENDED the thread is already running and
+        // this is a harmless no-op.
+        // SAFETY: `pi.hThread` is the just-created, still-owned main-thread
+        // handle; `ResumeThread` only adjusts its suspend count.
+        unsafe {
+            ResumeThread(pi.hThread);
         }
 
         // Hand ownership to the caller via `BaseChild`, which performs
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -1967,14 +1967,69 @@ impl BaseContainerRunner {
             }
         };
 
-        // The child was created suspended; now that it is in the job object (so
-        // every descendant it spawns is captured), resume its main thread. If the
-        // create API ignored CREATE_SUSPENDED the thread is already running and
-        // this is a harmless no-op.
-        // SAFETY: `pi.hThread` is the just-created, still-owned main-thread
-        // handle; `ResumeThread` only adjusts its suspend count.
-        unsafe {
-            ResumeThread(pi.hThread);
+        // `--wait-for-debugger`: block (no timeout) until an external
+        // debugger attaches to the real sandboxed PID, then clear only
+        // wxc-exec's own CREATE_SUSPENDED hold — see `debugger_wait` module
+        // docs for why this leaves the debugger's own attach-time freeze in
+        // place (so a plain `g` is enough, no `~0 m` needed). Unreachable in
+        // practice: `validate()` rejects `--wait-for-debugger` for this
+        // backend before `spawn_base` runs (BaseContainer cannot guarantee
+        // CREATE_SUSPENDED is honored on every OS build). Handled here
+        // anyway so a direct/test caller that bypasses `validate()` still
+        // fails closed instead of silently violating the suspend guarantee.
+        if request.wait_for_debugger {
+            if let Err(e) = crate::debugger_wait::wait_for_debugger_then_resume(
+                pi.hProcess,
+                pi.hThread,
+                pi.dwProcessId,
+                logger,
+            ) {
+                let _ = writeln!(
+                    logger,
+                    "Error: --wait-for-debugger failed ({e}); the sandboxed child has \
+                     been terminated."
+                );
+                // `wait_for_debugger_then_resume` already terminated the
+                // child on failure; reap it and tear down the same
+                // sandbox/proxy state the job-setup-failure path above does.
+                unsafe {
+                    let _ = WaitForSingleObject(pi.hProcess, u32::MAX);
+                    let _ = CloseHandle(pi.hProcess);
+                    let _ = CloseHandle(pi.hThread);
+                }
+                if request.lifecycle.destroy_on_exit {
+                    run_sandbox_cleanup(
+                        &identity,
+                        &sid_string,
+                        request.policy.network_proxy.is_enabled(),
+                        logger,
+                    );
+                    sandbox_tracking::unregister_ctrl_c_cleanup();
+                }
+                self.proxy_coordinator.stop(logger);
+
+                const WAIT_FOR_DEBUGGER_FAILED_MSG: &str =
+                    "--wait-for-debugger failed while waiting for a debugger to attach; \
+                     the sandboxed child was terminated.";
+                return Err(ScriptResponse {
+                    exit_code: -1,
+                    error_message: WAIT_FOR_DEBUGGER_FAILED_MSG.to_string(),
+                    standard_err: WAIT_FOR_DEBUGGER_FAILED_MSG.to_string(),
+                    extended_error: format!("wait_for_debugger_then_resume failed: {e}"),
+                    failure_phase: FailurePhase::LaunchFailed,
+                    ..Default::default()
+                });
+            }
+        } else {
+            // The child was created suspended; now that it is in the job object (so
+            // every descendant it spawns is captured), resume its main thread. If the
+            // create API ignored CREATE_SUSPENDED the thread is already running and
+            // this is a harmless no-op.
+            // SAFETY: `pi.hThread` is the just-created, still-owned main-thread
+            // handle; `ResumeThread` only adjusts its suspend count.
+            unsafe {
+                ResumeThread(pi.hThread);
+            }
         }
 
         // Hand ownership to the caller via `BaseChild`, which performs
@@ -2042,6 +2097,25 @@ struct BaseChild {
 
 impl SandboxBackend for BaseContainerRunner {
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
+        // `Experimental_CreateProcessInSandbox` is not guaranteed to honor
+        // CREATE_SUSPENDED on every OS build (see the creation-flags comment
+        // above), so this backend cannot guarantee the sandboxed process has
+        // run no code before a debugger attaches -- the core contract
+        // `--wait-for-debugger` promises. The dispatcher already routes
+        // `--wait-for-debugger` requests to an AppContainer tier
+        // automatically (its `CreateProcess*` contract always honors
+        // CREATE_SUSPENDED); this only fails closed for direct callers of
+        // this backend.
+        if request.wait_for_debugger {
+            return Err(ScriptResponse::error(
+                "--wait-for-debugger is not supported by the BaseContainer backend: \
+                 Experimental_CreateProcessInSandbox is not guaranteed to honor \
+                 CREATE_SUSPENDED on every OS build, so it cannot guarantee the \
+                 sandboxed process has run no code before a debugger attaches. Use \
+                 the AppContainer tier instead (the dispatcher does this \
+                 automatically for --wait-for-debugger).",
+            ));
+        }
         let capture_denials = request.policy.capture_denials.is_some();
         let use_process_security_environment = self.uses_process_security_environment(request);
         if capture_denials && !use_process_security_environment {
@@ -3588,6 +3662,27 @@ mod tests {
             "expected the capability-gate message, got: {}",
             err.error_message
         );
+    }
+
+    #[test]
+    fn validate_runner_rejects_wait_for_debugger() {
+        // BaseContainer's `Experimental_CreateProcessInSandbox` cannot
+        // guarantee CREATE_SUSPENDED is honored on every OS build, so this
+        // backend must fail closed for `--wait-for-debugger` rather than
+        // silently risk running code before a debugger attaches. The
+        // dispatcher already routes this flag to an AppContainer tier (see
+        // `dispatcher::tests::wait_for_debugger_skips_base_container_even_when_usable`);
+        // this covers a direct/test caller that bypasses the dispatcher.
+        let runner = BaseContainerRunner::new();
+        let request = ExecutionRequest {
+            wait_for_debugger: true,
+            ..ExecutionRequest::default()
+        };
+
+        let err = runner
+            .validate(&request)
+            .expect_err("--wait-for-debugger must be rejected by the BaseContainer backend");
+        assert!(err.error_message.contains("wait-for-debugger"));
     }
 
     #[test]
--- a/src/backends/appcontainer/common/src/debugger_wait.rs
+++ b/src/backends/appcontainer/common/src/debugger_wait.rs
@@ -34,26 +34,46 @@
 //!
 //! Once resumed this way, `wxc-exec` falls into its normal (already
 //! unbounded when `script_timeout` is 0) wait for the child to exit — no
-//! special-casing there. Ctrl-C during the attach-wait is handled by the
-//! existing `sandbox_tracking` console-control handler (registered well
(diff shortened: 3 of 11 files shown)


src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    loop {
+        let mut present = BOOL(0);
+        // SAFETY: `process` is a valid, open handle for the duration of this
+        // call (caller contract); `present` is a valid out-param on the stack.
+        let checked = unsafe { CheckRemoteDebuggerPresent(process, &mut present) };
+        if checked.is_ok() && present.as_bool() {
+            break;
+        }
+        std::thread::sleep(POLL_INTERVAL);
```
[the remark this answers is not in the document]
asklar: Fixed: CheckRemoteDebuggerPresent failures now break the loop and return Err(WxcError::Process(..)) (terminating the child first, fail-closed) instead of retrying forever. See wait_for_debugger_then_resume in 082e3e8.

src/core/wxc/src/main.rs
```diff
@@ -139,6 +139,29 @@ struct Cli {
...
+    /// only the debugger you attach runs unsandboxed.
+    ///
+    /// wxc-exec never calls ResumeThread in this mode. Attaching a debugger
+    /// adds its own +1 to the suspend count (symmetric with detach); after
+    /// setting breakpoints (necessarily deferred/pending ones, since no
+    /// dependent DLL past the main image has loaded yet), clear wxc-exec's
+    /// original CREATE_SUSPENDED increment yourself — e.g. WinDbg/cdb's
+    /// `~0 m` ("Resume Thread") — then `g`. Windows-only; only the
+    /// BaseContainer/AppContainer (`processcontainer`) backends implement
+    /// the suspend point.
```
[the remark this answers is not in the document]
asklar: Fixed: reworded the --wait-for-debugger CLI help. wxc-exec does call ResumeThread automatically the instant it detects the debugger attach (clearing its own CREATE_SUSPENDED increment), so operators only need a plain g -- no ~0 m. Also fixed the same stale wording in models.rs and README.md.

src/core/wxc_common/src/models.rs
```diff
@@ -740,6 +740,18 @@ pub struct ExecutionRequest {
...
+    /// thread after creation (it stays at its `CREATE_SUSPENDED` suspend
+    /// count, having executed no code at all — not even its own loader
+    /// init). An external, unsandboxed debugger attaches to the real PID at
+    /// its own pace, sets breakpoints (necessarily deferred/pending, since
+    /// no dependent DLL past the main image has loaded), then clears the
+    /// suspend count itself (e.g. WinDbg/cdb's `~0 m` "Resume Thread") and
+    /// `g`oes. `wxc-exec` never calls `ResumeThread` in this mode; it just
+    /// falls into its normal (already-unbounded-by-default) wait for the
+    /// child to exit. `false` (default) is the unchanged, immediate-resume
+    /// behavior. Consumed by the BaseContainer/AppContainer runners.
```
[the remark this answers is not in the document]
asklar: Fixed: reworded the wait_for_debugger field doc to match the implemented behavior (wxc-exec auto-clears its own suspend increment on attach detection via a single ResumeThread call; no ~0 m needed). Also clarified it's only consumed by the AppContainer runner now -- the dispatcher never routes wait_for_debugger to BaseContainer (see the dispatcher fix).

README.md
```diff
@@ -227,6 +227,16 @@ wxc-exec.exe --audit policy.json
...
 
+### Wait for Debugger
+
+`--wait-for-debugger` holds the sandboxed child suspended right after creation — before it has executed any code at all, not even its own loader init — instead of resuming it immediately. This lets an external, unsandboxed debugger attach to the *real* sandboxed process by PID at its own pace (there's no timeout, since nothing can run until the debugger releases it):
+
+```bash
+wxc-exec.exe --wait-for-debugger config.json
+```
+
+Attach a debugger to the printed PID and set breakpoints (necessarily deferred/pending ones, since no dependent DLL past the main image has loaded yet), then clear `wxc-exec`'s own suspend hold — e.g. WinDbg/cdb's `~0 m` ("Resume Thread") — and `g`. Windows-only; only the BaseContainer/AppContainer (`processcontainer`) backends implement the suspend point.
```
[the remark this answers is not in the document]
asklar: Fixed: updated the README wording to match the auto-resume behavior (plain g, no ~0 m) and to note BaseContainer is skipped automatically for --wait-for-debugger.

src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,95 @@
...
+    // SAFETY: `thread` is the caller's still-owned, still-suspended main
+    // thread handle; ResumeThread only adjusts its suspend count. This
+    // cancels exactly the one increment wxc-exec added via CREATE_SUSPENDED
+    // — the debugger's own attach-time freeze (a separate contribution to
+    // the same counter) is unaffected and keeps the thread from actually
+    // running until the operator issues `g`.
+    unsafe {
+        ResumeThread(thread);
+    }
+    let _ = writeln!(logger, "Process started, PID: {pid}.");
```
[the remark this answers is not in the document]
asklar: Fixed: wait_for_debugger_then_resume now checks the ResumeThread return value; on u32::MAX it fetches GetLastError, terminates the child (fail-closed, matching SpawnedChild::resume), and returns Err(WxcError::Process(..)). Both call sites (appcontainer_runner.rs, base_container_runner.rs) now propagate that Err and tear down instead of reporting 'Process started'.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1168,14 +1168,28 @@ impl BaseContainerRunner {
...
+        // wxc-exec's own CREATE_SUSPENDED hold — see `debugger_wait` module
+        // docs for why this leaves the debugger's own attach-time freeze in
+        // place (so a plain `g` is enough, no `~0 m` needed).
+        if request.wait_for_debugger {
+            crate::debugger_wait::wait_for_debugger_then_resume(
+                pi.hProcess,
+                pi.hThread,
+                pi.dwProcessId,
+                logger,
+            );
```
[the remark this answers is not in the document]
asklar: Good catch. Fixed by making --wait-for-debugger never reach BaseContainer at all: dispatcher::select_backend_with_fallback now passes prefer_base_container = !request.wait_for_debugger to fallback_detector::detect, so tier selection falls through to an AppContainer tier (2/3) whenever the flag is set -- AppContainer's CreateProcess* calls always honor CREATE_SUSPENDED per the documented Win32 contract, unlike Experimental_CreateProcessInSandbox. As defense-in-depth for direct/test callers that bypass the dispatcher, BaseContainerRunner::validate() now also rejects wait_for_debugger outright. Added tests for both (dispatcher::tests::wait_for_debugger_skips_base_container_even_when_usable, base_container_runner::tests::validate_runner_rejects_wait_for_debugger).

src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,134 @@
...
+    process: HANDLE,
+    thread: HANDLE,
+    pid: u32,
+    logger: &mut Logger,
+) -> Result<(), WxcError> {
+    let banner = format!("Process suspended, PID: {pid}. Attach a debugger to continue.");
+    let _ = writeln!(logger, "{banner}");
+    eprintln!("{banner}");
+
+    loop {
```
MGudgin: **Ctrl-C during the attach-wait orphans the suspended child permanently.**

The module doc (lines 41-44) says this case is covered:

> Ctrl-C during the attach-wait is handled by the existing `sandbox_tracking` console-control handler (registered well before this point), which runs on its own OS thread independent of this poll loop.

That handler runs, but it doesn't touch the child. Tracing what it actually does:

- `dacl_ctrl_handler` (`src/core/wxc/src/main.rs`) emits cancellation telemetry, drops the parked `DaclManager` to restore host ACEs, and cancels an in-flight audit trace. It never terminates the child.
- `sandbox_tracking::cleanup_sandbox` handles identity/SID/profile teardown, not process termination.
- The job object is never given `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` — `job_object.rs` only ever calls `SetInformationJobObject(JobObjectBasicUIRestrictions)` — so closing the job handle at exit doesn't reap it either.

Net effect: `wxc-exec` exits, the child's handles close, and a process that has **never executed a single instruction** (not even loader init) is left suspended at suspend-count 1 with no thread to resume it. It survives until someone finds it in Task Manager. This is squarely on the happy path for the feature's intended use — an operator starts the wait, changes their mind, and hits Ctrl-C. Each abandoned attempt leaks another one, and because they're pre-loader-init they're easy to miss.

Two options, both cheap:

1. **Kill from the signal path.** `JobObject::terminate()` already exists (`job_object.rs:282`). Park the job the same way the `DaclManager` is parked, and call it from `dacl_ctrl_handler` before yielding to the default handler. This mirrors the existing `register_ctrl_c_cleanup`/`unregister_ctrl_c_cleanup` shape in `sandbox_tracking.rs:250-264`, so it fits the established pattern.
2. **Let the OS do it.** Add `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` to the job's extended limits when `wait_for_debugger` is set. Fewer moving parts and it also covers the `TerminateProcess`-on-parent case that the comment block in `main.rs` notes bypasses every handler — but it needs a check that it doesn't disturb the normal (non-debugger) lifetime expectations.

I'd lean toward (2) gated on the flag, with (1) as the fallback if kill-on-close turns out to interact badly with the existing teardown ordering.

Worth pairing with a regression test: launch with `--wait-for-debugger`, send Ctrl-C during the wait, assert the PID printed in the `Process suspended, PID: <pid>` banner no longer exists.

Either way, please also fix the lines 41-44 comment — right now it documents this as solved, which is what would stop the next reader from noticing.

src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,134 @@
...
+        let mut present = BOOL(0);
+        // SAFETY: `process` is a valid, open handle for the duration of this
+        // call (caller contract); `present` is a valid out-param on the stack.
+        let checked = unsafe { CheckRemoteDebuggerPresent(process, &mut present) };
+        match checked {
+            Ok(()) if present.as_bool() => break,
+            Ok(()) => std::thread::sleep(POLL_INTERVAL),
```
MGudgin: **If the suspended child dies before a debugger attaches, this loop spins forever.**

The `Err` arm is doing the fail-closed work here, and its comment states the trigger:

> The handle went bad (e.g. the still-suspended child was terminated externally) -- this can never resolve on its own, so fail closed instead of polling forever.

But the handle can't go bad. `wait_for_debugger_then_resume`'s own `# Safety` contract requires the caller to keep `process` open for the entire call, and that's exactly what keeps the kernel object alive after the process dies. A `HANDLE` to a terminated process is still a perfectly valid handle — the object stays referenced until the last handle closes. `CheckRemoteDebuggerPresent` succeeds on it and reports `present = FALSE`.

So on external termination the loop takes the `Ok(())` arm, sleeps 100 ms, and repeats — forever. The `Err` arm never fires for the one scenario it was written for, and `wxc-exec` hangs instead of failing closed. Worth noting the module doc at lines 47-53 builds the whole failure-handling story on this premise, so the design intent is right and only the mechanism is wrong.

`WaitForSingleObject` fixes it and folds the sleep into the same call, so the loop gets a liveness check for free:

```rust
match checked {
    Ok(()) if present.as_bool() => break,
    Ok(()) => {
        // Doubles as the poll interval: returns early only if the child
        // died, which can never resolve into an attach.
        if unsafe { WaitForSingleObject(process, POLL_INTERVAL.as_millis() as u32) }
            == WAIT_OBJECT_0
        {
            return Err(WxcError::Process(format!(
                "PID {pid} exited before a debugger attached"
            )));
        }
    }
    Err(e) => { /* unchanged */ }
}
```

Keep the `Err` arm — it's still correct for genuine API failures, it just isn't the path that catches a dead child.

This is also the cheapest of the findings to lock in with a test: spawn any suspended process, `TerminateProcess` it, and assert the wait returns an error rather than hanging (with a test timeout, so a regression fails loudly instead of wedging CI).

src/core/wxc/src/main.rs
```diff
@@ -1019,6 +1044,10 @@ fn main() {
     request.experimental_enabled = cli.experimental;
     request.testing_features_enabled = cli.allow_testing_features;
     request.dry_run = cli.dry_run;
+    #[cfg(target_os = "windows")]
+    {
+        request.wait_for_debugger = cli.wait_for_debugger;
```
MGudgin, line 1071: **`--wait-for-debugger` is accepted for every backend but only honored by AppContainer.**

`wait_for_debugger` lives on the shared `ExecutionRequest` root (`models.rs:756`) and this line sets it unconditionally for every one-shot request, with no reference to `request.containment`. A full-repo grep shows only three files ever read it back:

- `appcontainer_runner.rs:1348` — honors it
- `base_container_runner.rs:1293` — rejects it
- `dispatcher.rs:343` — steers tier selection away from BaseContainer

Windows Sandbox, IsolationSession, Hyperlight, MicroVM, and WSLC never look at it. Those backends are dispatched directly rather than through `appcontainer::dispatcher`, so `wxc-exec --wait-for-debugger` against any of them prints nothing, suspends nothing, and runs the workload straight through. The operator is sitting there waiting to attach to a process that already came and went.

What makes this worth fixing rather than documenting is that the PR already establishes the right convention one file over. `BaseContainerRunner::validate` goes out of its way to fail closed with a genuinely excellent error — it explains the `CREATE_SUSPENDED` guarantee problem and names the workaround. Five sibling backends get silence on identical input. That asymmetry is the actual defect; either behavior would be defensible if it were uniform.

Suggest hoisting the check to `wxc_common::validator::validate_common`, rejecting when `wait_for_debugger && containment != ContainmentBackend::ProcessContainer`, in the spirit of the BaseContainer message:

> `--wait-for-debugger` is only supported by the `processcontainer` backend (AppContainer tier); `<backend>` provides no pre-execution suspend point.

That's one check covering every present and future backend, and it makes the `BaseContainerRunner::validate` rejection a narrower special case (BaseContainer *is* processcontainer, but its specific tier can't honor it) rather than the only guard rail in the system. It also closes the same gap on the state-aware path, which currently never copies the flag onto the request at all (`main.rs:1028-1029` propagates `--experimental` and `--dry-run` but not this one) — harmless today since state-aware never routes to processcontainer, but it's the same root cause and a central check makes it moot.

If you'd rather keep it permissive, then at minimum the README section needs to say the flag is inert outside processcontainer — right now it reads as a general `wxc-exec` capability.

MGudgin (commented):
Adversarial multi-axis review (9 axes). Three merge-blocking findings inline below; the first two are the same 20-line loop viewed from different angles. The suspend-count design itself is correct and no axis disputed it — these are all about the loop having no exits.

src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,134 @@
...
+///
+/// On failure the child is terminated (fail closed) before returning
+/// [`WxcError::Process`]; the caller only needs to tear down its own sandbox
+/// resources and surface the error.
+///
+/// # Safety
+/// `process` and `thread` must be valid, still-open handles to the just
+/// created, still-suspended child for the duration of this call; the caller
+/// retains ownership (this function does not close either handle).
+pub fn wait_for_debugger_then_resume(
```
MGudgin, line 140: **The wait loop's state machine can't be tested without a real debugger and a real process.**

`wait_for_debugger_then_resume` reaches straight through to `CheckRemoteDebuggerPresent`, `ResumeThread`, `TerminateProcess`, and `std::thread::sleep`, all inline. That bakes four decisions into a function whose actual logic — poll until attached, resume exactly once, terminate and fail closed on error — is pure state machine and would be trivially testable on its own.

The practical cost is visible in this PR: the branch coverage that exists is on the *guards* around the feature (`dispatcher::tests::wait_for_debugger_skips_base_container_even_when_usable`, `validate_runner_rejects_wait_for_debugger`), because those are the only parts that could be tested without an OS-level fixture. The behavior that actually implements the feature has none. The two hang findings on this PR are both single-branch defects that a mocked loop would have caught immediately.

A closure-based seam is enough here — no trait or new module needed:

```rust
fn wait_loop(
    mut is_attached: impl FnMut() -> Result<bool, WxcError>,
    mut is_alive:    impl FnMut() -> bool,
    mut tick:        impl FnMut(),
) -> Result<(), WxcError> { /* the loop, and only the loop */ }
```

The Win32 version passes real closures; tests pass canned sequences (`[false, false, true]`, `is_alive` flipping to `false` mid-wait, `is_attached` returning `Err`) and a no-op `tick`, so the whole matrix runs in microseconds with no child process at all. The public `wait_for_debugger_then_resume` signature doesn't have to change.

I'd treat this as a should-fix rather than a merge blocker — but it's the difference between the two hang fixes being verified and being merely believed, so it's worth doing in the same change rather than as follow-up.

src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,134 @@
...
+        // SAFETY: same fail-closed contract as above.
+        unsafe {
+            let _ = TerminateProcess(process, u32::MAX);
+        }
+        return Err(WxcError::Process(format!(
+            "ResumeThread failed after debugger attach for PID {pid}: {err:?}"
+        )));
+    }
+    let _ = writeln!(logger, "Process started, PID: {pid}.");
+    Ok(())
```
MGudgin, line 218: **The feature's entire runtime behavior ships with zero automated coverage.**

The PR's validation section lists build, `fmt`, `clippy`, and "built `wxc-exec.exe` and confirmed it runs". The tests that did land cover backend *selection* — that the dispatcher skips BaseContainer and that `validate` rejects it. Nothing exercises what happens after the process is created suspended.

Concretely, all of these would ship green today:

- `ResumeThread` called twice, or not at all (the suspend-count reasoning in the module doc is the subtlest part of this PR and is entirely unguarded)
- the terminate-on-error path silently not terminating, leaving the fail-closed contract unmet
- the flag not reaching `ExecutionRequest` from the CLI at all
- the parsed default flipping from `false` to `true`

That last pair is the part I'd push on regardless of how the seam question (see the testability comment above) is resolved, because it needs no Win32 fixture at all:

- a `main.rs` CLI test asserting `--wait-for-debugger` present/absent maps to `request.wait_for_debugger` true/false — `main.rs` already has CLI parsing tests to extend
- a `config_parser.rs` assertion that the default is `false` — the parser suite already constructs `ExecutionRequest` and now silently carries this field

Both are a few lines each and lock in the plumbing half of the feature.

For the Win32 half, if the loop gets a closure seam then the interesting cases (not-attached-then-attached, resume failure, check failure, child dies mid-wait) are all cheap unit tests. Without a seam, the fallback is a `wxc_e2e_tests` case that launches a trivial AppContainer process with the flag, asserts the thread's suspend count is non-zero, attaches with `DebugActiveProcess` from the test itself, and asserts the process then runs to completion — real coverage, but meaningfully more machinery than the seam route, which is part of why the seam is worth adding first.

MGudgin (commented):
Two follow-on findings from the same review, both about verification rather than defects: the wait loop has no test seam, and the feature's runtime behavior has no automated coverage at all.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1168,14 +1168,69 @@ impl BaseContainerRunner {
...
+                     the sandboxed child was terminated.";
+                return Err(ScriptResponse {
+                    exit_code: -1,
+                    error_message: WAIT_FOR_DEBUGGER_FAILED_MSG.to_string(),
+                    standard_err: WAIT_FOR_DEBUGGER_FAILED_MSG.to_string(),
+                    extended_error: format!("wait_for_debugger_then_resume failed: {e}"),
+                    failure_phase: FailurePhase::LaunchFailed,
+                    ..Default::default()
+                });
+            }
```
MGudgin: **The `prefer_base_container` fix made this branch dead code — worth deleting it now.**

Following on from the thread above: routing `--wait-for-debugger` away from BaseContainer via `prefer_base_container = !request.wait_for_debugger` is the right fix, and the `validate()` rejection is a sensible backstop for direct callers. Together they mean this branch in `spawn_base` can no longer execute. `spawn_base` is private, its only caller runs `self.validate(request)` first, and `validate` rejects the flag unconditionally at line 1293 — so `request.wait_for_debugger` is provably `false` by the time control reaches line 1181.

That leaves ~50 lines of unreachable code carrying its own error strings (`WAIT_FOR_DEBUGGER_FAILED_MSG`), its own teardown sequence, and its own `ScriptResponse` construction, all of which have to be kept in step with the job-setup cleanup path next to them and with the AppContainer copy of the same logic. Defense-in-depth is cheap when it's an assertion; it's less cheap when it's a full parallel error path that no test can reach and no reader can verify.

Suggest deleting the branch and keeping the unconditional resume. If you'd rather keep a tripwire for the "someone refactors `spawn_base`'s caller and drops the `validate` call" scenario, a one-liner does the same job without the maintenance surface:

```rust
debug_assert!(
    !request.wait_for_debugger,
    "wait_for_debugger must be rejected by validate() before reaching spawn_base"
);
```

Non-blocking — but the two guards you added are strictly better than this branch, and leaving all three in place obscures which one is actually load-bearing.

src/backends/appcontainer/common/src/debugger_wait.rs
```diff
@@ -0,0 +1,134 @@
...
+    thread: HANDLE,
+    pid: u32,
+    logger: &mut Logger,
+) -> Result<(), WxcError> {
+    let banner = format!("Process suspended, PID: {pid}. Attach a debugger to continue.");
+    let _ = writeln!(logger, "{banner}");
+    eprintln!("{banner}");
+
+    loop {
+        let mut present = BOOL(0);
```
MGudgin: **The loop has no cooperative exit — every way out is external.**

Flagging this mostly to note it's *not* fully addressed by the other two loop findings, in case it looks redundant. Adding a liveness check fixes the dead-child case, and killing the child from the Ctrl-C path fixes the orphan case — but both work by destroying the thing the loop is polling. The loop itself still has exactly one voluntary exit (`present == true`); nothing can ask it to stop.

Two places that matters:

- **Tests.** A test that wants to verify "loop is still waiting after N ticks, now cancel it" has no way to end the loop except by tearing down a real process. This compounds with the seam comment above — a closure seam plus a cancellation predicate is what makes the wait cheaply testable, and each is half a fix on its own.
- **Any future non-signal shutdown.** If the wait ever needs to end for a reason that isn't Ctrl-C and isn't child death — a timeout mode, a supervisor asking `wxc-exec` to stand down — there's no place to put it.

An `Option<&AtomicBool>` checked alongside the poll, or a `should_cancel: impl FnMut() -> bool` if you take the closure-seam route, covers both. Low priority relative to the two hangs — file it under "do this if you're already refactoring the loop for the seam", not as separate work.

MGudgin (commented):
Last two, both non-blocking: dead code left behind by the BaseContainer routing fix, and a note that the loop still has no cooperative exit even once the two hangs are fixed.

MGudgin (changes requested):

MGudgin (changes requested):
Various comments below? above?

asklar:
Rebased onto fresh `main` (resolved conflicts in `dispatcher.rs` tier-selection and `base_container_runner.rs` `validate()`) and addressed the adversarial review's three merge-blocking findings plus the should-fix and non-blocking ones, in `ac5db819`:

1. **Dead-child spin** (debugger_wait.rs) — the poll loop now uses `WaitForSingleObject` as both the poll-interval wait and a liveness check on the child process handle, failing closed the instant it observes the child has exited instead of retrying `CheckRemoteDebuggerPresent` forever.
2. **Ctrl-C orphans the suspended child** (debugger_wait.rs) — `UiJobObject` gained `set_kill_on_job_close()`, which sets `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. It's applied only when `--wait-for-debugger` is requested, so the OS itself kills the still-suspended child the instant the last handle to its job closes — covering Ctrl-C, console close, panics, and parent-kill, all of which bypass Rust destructors.
3. **Backend gating** (main.rs) — centralized in `wxc_common::validator::validate_common` (run by every backend before execution via `ScriptRunner::run`/`SandboxBackend::validate`), rejecting `--wait-for-debugger` for every non-`processcontainer` backend uniformly, present and future, rather than requiring each runner to remember it.

Should-fix:
- Extracted the loop into a pure, closure-seamed `wait_loop` state machine with unit tests covering immediate attach, delayed attach, child-exit-mid-wait, and API failure — no OS fixture needed.
- Added CLI (--wait-for-debugger flag parsing) and config-parser (default alse) plumbing tests.

Non-blocking:
- Deleted the now-dead `wait_for_debugger` branch in `BaseContainerRunner::spawn_base` (unreachable since `validate()` rejects the flag before `spawn_base` runs) in favor of a `debug_assert!` tripwire.

Verified: `cargo build`/`cargo test`/`cargo clippy --all-targets --all-features -D warnings`/`cargo fmt --check` all green on the full workspace (Windows target).

src/backends/appcontainer/common/src/debugger_wait.rs
asklar: Fixed: `UiJobObject` gained `set_kill_on_job_close()`, setting `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` (option 2, as you leaned toward). Applied only when `request.wait_for_debugger` is set, right where the job is created in `spawn_suspended`. The OS now kills the still-suspended child the instant the last handle to its job closes -- covering Ctrl-C, console close, panics, and parent-kill, all of which bypass the existing `sandbox_tracking`/`dacl_ctrl_handler` Rust-destructor paths. Also fixed the module doc's stale claim that Ctrl-C was already handled. See ac5db819400d6d0dbe76f8a6fe6b24312a9efc75.

src/backends/appcontainer/common/src/debugger_wait.rs
asklar: Fixed exactly as suggested: the loop now calls `WaitForSingleObject(process, POLL_INTERVAL_MS)` instead of `std::thread::sleep`, which doubles as the poll-interval wait and a liveness check on the child. It returns early (before the interval elapses) only when the child has become signaled, i.e. exited, and the loop then fails closed with `PID {pid} exited before a debugger attached` -- no `TerminateProcess` needed since the child is already dead. Kept the `Err` arm unchanged for genuine API failures. See ac5db819400d6d0dbe76f8a6fe6b24312a9efc75.

src/core/wxc/src/main.rs
asklar: Fixed by hoisting the check into `wxc_common::validator::validate_common`, exactly as suggested -- it now rejects any request where `wait_for_debugger` is set and `containment != ContainmentBackend::ProcessContainer`, with a message in the same style as the `BaseContainerRunner::validate` one. `validate_common` runs for every backend (every `ScriptRunner::run` and `SandboxBackend::validate` implementation calls it first), so this covers Windows Sandbox, WSLC, IsolationSession, Hyperlight, MicroVM, LXC, Bubblewrap, and Seatbelt uniformly, plus any future backend, with one check instead of five-plus scattered ones. Added unit tests for the reject/accept/false-on-any-backend cases. See ac5db819400d6d0dbe76f8a6fe6b24312a9efc75.

src/backends/appcontainer/common/src/debugger_wait.rs
asklar: Added the seam, close to what you sketched: `wait_for_debugger_then_resume`'s loop is now a pure `wait_loop(check_attached, wait_tick) -> WaitLoopOutcome` state machine (using an outcome enum instead of a plain bool/Result so the two failure modes -- API failure vs. child-exited -- stay distinguishable for the caller's terminate-or-not decision). The public function's signature is unchanged; it just wires the real `CheckRemoteDebuggerPresent`/`WaitForSingleObject` calls into the two closures. Unit tests cover immediate attach, delayed attach (asserting `wait_tick` call count), child-exit-mid-wait, and API failure -- all in microseconds, no process or debugger. See ac5db819400d6d0dbe76f8a6fe6b24312a9efc75.

src/backends/appcontainer/common/src/debugger_wait.rs
asklar: Covered the CLI/config-parser half exactly as you suggested (a `main.rs` CLI test asserting `--wait-for-debugger` present/absent maps to `request.wait_for_debugger`, and a `config_parser.rs` test asserting the default is `false`), plus the loop's state machine now has full unit coverage via the `wait_loop` seam (see the seam reply above) -- which incidentally covers exactly the two hang scenarios this PR fixes. Left the Win32-fixture-requiring end-to-end case (real suspend + real `DebugActiveProcess`) as follow-up, per your note that it's meaningfully more machinery than the seam route. See ac5db819400d6d0dbe76f8a6fe6b24312a9efc75.

src/backends/appcontainer/common/src/base_container_runner.rs
asklar: Fixed as suggested -- deleted the unreachable branch and kept a `debug_assert!(!request.wait_for_debugger, ...)` tripwire instead, so a future refactor that drops the `validate()` call in `spawn()` fails a debug build immediately rather than silently resuming a child the operator expected to stay suspended. See ac5db819400d6d0dbe76f8a6fe6b24312a9efc75.

src/backends/appcontainer/common/src/debugger_wait.rs
asklar: Left this one as-is for now -- per your note it's low priority and only worth doing if already refactoring the loop for the seam. The seam refactor (see the testability reply) makes wiring in a \should_cancel: impl FnMut() -> bool\ mechanical if we want it, but there's currently no cancellation signal plumbed from \main.rs\ down into a runner's \spawn()\ call to drive it, so adding the parameter alone wouldn't change any real behavior yet. Happy to pick this up as a follow-up once there's an actual caller for it.

asklar:
Superseded by IFEO-style approach instead; will add a doc for it.

PR #715, opened 2026-07-30
Publish MXC crates through ESRP

dhoehna:
## Summary

- makes the complete 18-crate `mxc-sdk` dependency closure publishable
- packages versioned `.crate` artifacts in the official-build pipeline
- publishes crates leaf-first through ESRP with checksum and propagation validation
- supports dry runs and safe retries without storing a crates.io token in the repository

## Current limitation

Multi-package `cargo package` with the private-feed `replace-with` configuration is blocked by rust-lang/cargo#17196. The crate-packaging stage is therefore explicitly opt-in and disabled by default, so ordinary official builds are unaffected while a workaround is selected.

## Validation

- packaged and validated all 18 crates locally
- verified the generated dependency order
- parsed the modified pipeline YAML
- verified crates.io sparse-index probing and checksum logic

No pipeline was queued and no crates were published.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/715)


dhoehna:
Superseded by #716, which is branched from current `main` and merges cleanly.

PR #717, opened 2026-07-30
Add backend support probe API design & discussion doc

huzaifa-d:

-----

## Summary

Adds a **design & discussion document** for a proposed read-only Rust `available_backends()` host-capability probe API.

**Changes:**
- Defines the proposed API and isolation-tier semantics.
- Documents existing detection methods and gaps.
- Outlines testing and follow-up work.

<details open>

| File | Description |
| ---- | ----------- |
| `docs/backend-support-probe-api-plan.md` | Adds the backend-support probe API design. |
</details>

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/717)

### DIFF
--- a/docs/backend-support-probe-api-plan.md
+++ b/docs/backend-support-probe-api-plan.md
@@ -105,7 +105,7 @@ have TypeScript read that result rather than compute its own.
 | Step | What | Why it gives parity |
 | --- | --- | --- |
 | 1. Consolidate detectors in Rust | Port the two TypeScript-only checks (`windows_sandbox` DISM/feature check, `lxc-ls`) into Rust<br>so `available_backends()` covers every backend. | Each backend has exactly one detector. |
-| 2. TypeScript stops probing itself | `getPlatformSupport()` reads the native probe instead of running its own `dism`/`lxc-ls`.<br>It already does this for the Windows tier via `populateIsolationFromProbe()` → `wxc-exec --probe`;<br>extend that JSON to carry the backend list. | TypeScript becomes a pure projection of the Rust result. |
+| 2. TypeScript stops probing itself | `getPlatformSupport()` reads the native backend-availability result instead of running its own `dism`/`lxc-ls`.<br>The transport must be **side-effect-free**: `wxc-exec --probe` runs *after* `recover_orphaned_state()` (which can restore/prune DACL state on the host), so it cannot be extended unchanged without violating the read-only contract.<br>Instead, expose backend availability via a dedicated mode handled **before** DACL recovery (e.g. `wxc-exec --available-backends`), or directly through `mxc_ffi`. | TypeScript becomes a pure projection of the Rust result, without triggering host mutation. |
 | 3. Names flow from serde | Backend names (`Containment`) and tier strings (`IsolationTier::as_str()`) are already Rust-serialized;<br>TypeScript consumes them as-is instead of hand-re-encoding. | Removes the wire-name drift class structurally. |
 
 See §7.9 for why the canonical probe stays in Rust rather than moving into the TypeScript layer.
@@ -138,8 +138,12 @@ Writing the missing detectors and wiring the TypeScript projection, one issue ea
 3. `microvm` / `hyperlight` - hypervisor-presence probe.
 4. `lxc` - port the `lxc-ls` presence check from TypeScript to Rust,
 so the probe (not just the SDK) can report it (§4.2, step 1).
-5. TypeScript projection - make `getPlatformSupport()` read the native probe (`wxc-exec --probe` / `mxc_ffi`)
-instead of running its own `dism`/`lxc-ls`, so the two layers can't drift (§4.2, step 2).
+5. TypeScript projection - make `getPlatformSupport()` read the native backend availability via a
+side-effect-free transport (e.g. a new `wxc-exec --available-backends` mode handled **before**
+`recover_orphaned_state()`, or `mxc_ffi`) instead of running its own `dism`/`lxc-ls`, so the two
+layers can't drift (§4.2, step 2). Do **not** extend the existing `--probe` flag: it runs after
+`recover_orphaned_state()`, which can restore/prune DACL state and would violate the read-only
+contract of this API.
 ---
 
 ## 7. Appendix - Decisions & Notes
--- a/docs/backend-support-probe-api-plan.md
+++ b/docs/backend-support-probe-api-plan.md
@@ -37,6 +37,12 @@ pub fn available_backends() -> Vec<AvailableBackend>;
 ```
 
 Example results:
+> **"Stock Windows"** here means a clean Windows install with only the default
+> optional features enabled (no BaseContainer, Windows Sandbox, etc.), so the
+> process-container backend falls to its `appcontainer-dacl` floor. See
+> [`docs/process-container/os-version-support.md`](process-container/os-version-support.md)
+> for the per-release policy-support matrix that determines the reachable tier.
+
 | Host | Result |
 | --- | --- |
 | Stock Windows | `[{ backend: "processcontainer", tier: Some("appcontainer-dacl") }]` |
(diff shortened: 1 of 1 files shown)


docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+/// An empty `Vec` means "no backend this API can currently affirm on this host"
+/// (e.g. an unsupported platform, Linux with neither `bwrap` nor `lxc`, or
+/// macOS without `sandbox-exec`) — it is a normal result, not an error.
+pub fn available_backends() -> Vec<AvailableBackend>;
+```
+
+Example results:
+| Host | Result |
+| --- | --- |
+| Stock Windows | `[{ backend: "processcontainer", tier: Some("appcontainer-dacl") }]` |
```
bbonaby, line 48: note: Might be good to have a comment above about what we consider Stock windows and link to https://github.com/microsoft/mxc/blob/main/docs/process-container/os-version-support.md


Also I think there are places in the appcontainer fallback code that use the fallback tier names as strings only. Might be useful to have a centralized place for these names so changes don't break you. One other thing to think about is that I believe Gudge might add a tier 4. We'd want to think about future additions too so we might need a PR pipeline check to make sure all tiers and all backends can be probed.

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+- For per-backend **diagnostics and reasons**, the tool is `wxc-exec --probe`
+- Each capability is **detected once, in Rust**, and the TypeScript SDK projects that result rather than re-checking.
+## 3. Detection & isolation tiers
+
+### 3.1 Detection methods used today (and their risks)
+
+Four backends already have *some* presence check, but they are spread across two
+layers and vary in how much they actually prove. Documenting them here so the
+probe API can reuse the Rust ones and knowingly accept the risk of the shallower
+ones.
```
bbonaby, line 66: question: is the rust ssdk covered in mxc-sdk as well?

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+| `isolation_session` | build ≥ 26300.8553 **and** `IsoSessionApp.dll` resolvable **and** feature compiled | build gate lives **only in the TypeScript SDK** | wrong OS builds falsely pass |
+| `microvm` | feature compiled, NanVix runtime files staged, and WHP usable on Windows or `/dev/kvm` readable/writable on Linux | nothing | checking only a hypervisor can report availability when required runtime files are missing |
+| `hyperlight` | hypervisor present + feature compiled | nothing | same VM-boot risk |
+
+### 4.2 The parity rule: detect once, project into TS
+
+Detection is split by layer today : `base-container`/`wslc` are Rust-only, while
+`windows_sandbox`/`lxc` are TypeScript-only, so the two layers can and already
+do disagree. The fix is: detect each capability in exactly one place (Rust), and
+have TypeScript read that result rather than compute its own.
```
bbonaby, line 109: note: rust first detection is the right direction. We need to make sure this probe is useable via the mxc-sdk rust crate as a p0. That's the sdk GitHub copilot is using now.

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+### 4.1 Backends still missing a Rust detector
+
+Four backends have no *probe-suitable* (cheap, no persistent host mutation, no
+process/VM launch) host detector in Rust today, so a truthful availability signal
+for them **cannot be built just yet**.
+
+| Backend | What a real probe needs | What exists today | Risk if faked |
+| --- | --- | --- | --- |
+| `windows_sandbox` | DISM/registry check of the *Containers-DisposableClientVM* optional feature | only a private "is the `.exe` on disk" check | reports available when the feature is off → launch fails |
+| `isolation_session` | build ≥ 26300.8553 **and** `IsoSessionApp.dll` resolvable **and** feature compiled | build gate lives **only in the TypeScript SDK** | wrong OS builds falsely pass |
```
bbonaby: @adpa-ms will sessions have an API we can call with our new probe? Just to see if it's available I mean. Possibly could be an Is-Apiset-Available call, assuming they apis are in an apiset that is.

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+
+| Step | What | Why it gives parity |
+| --- | --- | --- |
+| 1. Consolidate detectors in Rust | Port the two TypeScript-only checks (`windows_sandbox` DISM/feature check, `lxc-ls`) into Rust<br>so `available_backends()` covers every backend. | Each backend has exactly one detector. |
+| 2. TypeScript stops probing itself | `getPlatformSupport()` reads the native backend-availability result instead of running its own `dism`/`lxc-ls`.<br>The transport must be **side-effect-free**: `wxc-exec --probe` runs *after* `recover_orphaned_state()` (which can restore/prune DACL state on the host), so it cannot be extended unchanged without violating the read-only contract.<br>Instead, expose backend availability via a dedicated mode handled **before** DACL recovery (e.g. `wxc-exec --available-backends`), or directly through `mxc_ffi`. | TypeScript becomes a pure projection of the Rust result, without triggering host mutation. |
+| 3. Names flow from serde | Backend names (`Containment`) and tier strings (`IsolationTier::as_str()`) are already Rust-serialized;<br>TypeScript consumes them as-is instead of hand-re-encoding. | Removes the wire-name drift class structurally. |
```
bbonaby, line 115: question: so this is just a typescript change only?

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+
+See §7.9 for why the canonical probe stays in Rust rather than moving into the TypeScript layer.
+
+
+
+
+## 5. Testing
+- Every returned `backend` is a valid `wxc_common::wire::Containment` name.
+- On Windows the result contains `processcontainer` with a `tier` of one of the three known strings;
+`appcontainer-dacl` is the floor when nothing higher is reachable.
```
bbonaby, line 125: note: I think there are places in the appcontainer fallback code that use the fallback tier names as strings only. Might be useful to have a centralize place for these.

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,224 @@
...
+### 4.2 The parity rule: detect once, project into TS
+
+Detection is split by layer today : `base-container`/`wslc` are Rust-only, while
+`windows_sandbox`/`lxc` are TypeScript-only, so the two layers can and already
+do disagree. The fix is: detect each capability in exactly one place (Rust), and
+have TypeScript read that result rather than compute its own.
+
+| Step | What | Why it gives parity |
+| --- | --- | --- |
+| 1. Consolidate detectors in Rust | Port the two TypeScript-only checks (`windows_sandbox` DISM/feature check, `lxc-ls`) into Rust<br>so `available_backends()` covers every backend. | Each backend has exactly one detector. |
```
bbonaby, line 113: note: now that the rust sdk is an in process sdk (no executables) we'd want to make sure that callers don't have to elevate when using the probe.

docs/backend-support-probe-api-plan.md
adpa-ms: I improved the probe in #761 where it now looks for the API class being registered on the OS instead of checking a build number.

docs/backend-support-probe-api-plan.md
huzaifa-d: Thanks. Splitting into the three sub-points:

1. **"Stock Windows" definition + link**: added a note.
2. **Centralize tier-name strings**: Landed in #725.
3. **Tier 4 / future backends + pipeline check**: tracked in #769.

docs/backend-support-probe-api-plan.md
huzaifa-d: Yes. The probe is a export of the `mxc-sdk` Rust crate, not just internal to `mxc_engine`. Landed in #725.

docs/backend-support-probe-api-plan.md
huzaifa-d: Landed in #725.

docs/backend-support-probe-api-plan.md
huzaifa-d: @adpa-ms following up: could sessions expose an apiset-available style check (for example `IsApiSetImplemented` against the IsolationSession apiset) so the probe can confirm availability cheaply without attempting a real activation? Now that #761 detects the `Windows.AI.IsolationSession.Preview` `IsoSessionOps` class registration rather than a build number, a lightweight availability query would shape the `isolation_session` detector nicely.

docs/backend-support-probe-api-plan.md
huzaifa-d: Yes, it is a TypeScript-side change: TS stops hand-re-encoding the names and consumes the Rust-serialized values as-is.

docs/backend-support-probe-api-plan.md
huzaifa-d: Will be done in the implementation.

docs/backend-support-probe-api-plan.md
```diff
@@ -0,0 +1,230 @@
...
+### 7.6 `process` and `vm` are deliberately excluded
+
+They are *abstract intents*, not backends with their own runner.
+
+### 7.7 `base-container` detection caching
+
+`fallback_detector::is_base_container_usable()` caches its result in a
+`OnceLock`. So "fresh detection on every call" is not fully achievable for the
+`base-container` tier, and a cached `true` can go **stale** if BaseContainer
+enablement changes mid-process. Both are accepted and documented rather than
```
bbonaby, line 201: note: Richie added a check directly in base_container_runner now I believe that you can use.

bbonaby (approved):
FYI: @mgudgin would probably be interested in seeing how we implement this

docs/backend-support-probe-api-plan.md
adpa-ms: This API isn't in an apiset unfortunately. I do not think I can do better than I have put in right now. However, I'm noting this down as a follow up. The isolation session API is also attempting to lift out of the OS, so that may open up some better options down line.

PR #724, opened 2026-07-31
[LXC] Filter IPv6 destinations and CIDR ranges in firewall mode (AB#62830559)

dhoehna:
Linked work item: AB#62830559 — LXC IPv6 + CIDR destination filtering in firewall mode

## Summary

In firewall mode (`enforcementMode: "firewall"`), the LXC backend resolved `allowedHosts` / `blockedHosts` to IPv4 only, so IPv6 destinations were silently unfiltered and any CIDR entry (a string containing `/`) was discarded without becoming a rule.  This PR resolves and programs both address families and forwards validated CIDRs verbatim to `iptables` / `ip6tables`, closing roadmap item 19 (IPv6 + CIDR parsing).  Bubblewrap picks this up too, because it delegates to the same manager.

- Hostnames now resolve to both A and AAAA records; IPv4 rules go to `iptables`, IPv6 rules to `ip6tables`, each with its own per-container chain and `FORWARD` hook.
- Validated CIDRs (`140.82.112.0/20`, `2606:50c0::/32`) are forwarded verbatim; the prefix must be ASCII digits within family range (`<=32` v4, `<=128` v6).  Invalid CIDR is skipped with a warning rather than treated as fatal.  On `main` a `/`-containing entry failed IP parse, then failed DNS, and produced no rule at all, so these entries take effect for the first time here.
- Empty or whitespace-only entries resolve to nothing, avoiding a Winsock `":0"` lookup that would otherwise program the host's own interface addresses.
- `ip6tables` is probed once.  If the binary is missing or IPv6 is disabled *and the host has no active IPv6*, the IPv4 chain still applies and the count of unapplied IPv6 rules is logged.  If the host has live IPv6 but `ip6tables` is unusable, setup now fails closed; on `main` that combination silently succeeded with IPv4-only rules while IPv6 egress went unfiltered.

## Ownership of installed rules

Adding a second address family doubles the objects a partly-failed setup can leave behind, so the manager now records exactly which chains and hooks it created and removes only those.

- Each chain and hook is recorded the moment its command succeeds, not after the whole apply does, so a failure or a signal mid-apply rolls back precisely what exists.
- A teardown returns the objects whose removal command itself failed.  Ownership is retained exactly when something survived, so `remove_firewall_rules` and `Drop` retry it instead of reporting success over a leaked chain.
- **A chain is never flushed while its `FORWARD` hook is still installed.**  `-F` succeeds regardless of who references the chain, and an emptied user chain returns to its caller instead of reaching its own closing DROP, so flushing a still-hooked chain unfilters a container that may still be running.  The `-X` would have failed anyway, since iptables refuses to delete a referenced chain, so the flush bought nothing and cost the container its filtering.  The whole step is gated per address family on that family's hook delete having succeeded, and the chain stays published so a later pass retries.  Leaving it costs a leaked chain that still filters correctly.
- A second apply is refused while the manager still owns resources.  Both arms of `apply_firewall_rules` replace the record with the current attempt's set, so re-applying would drop the earlier one and strand what it named.  Every caller builds a manager immediately before its single apply, so refusing costs nothing and makes the hazard unreachable rather than leaving it to callers to avoid.
- Signal-time cleanup acts on the same record.  A signal arriving before anything was created runs no `iptables` command at all, which matters because chain names truncate to 20 characters and can collide with a different live container.

## Merge-order constraints

- **#632** also edits `network_iptables.rs` and conflicts.  Whoever merges second must resolve keeping #632's deny-before-allow ordering; resolving toward this PR regresses deny-wins precedence.
- **#632 and #633** independently fix the same flush-before-hooks-gone fail-open in this file.  `teardown_chain` here is deliberately byte-identical to the one #632 landed, so that conflict resolves to a no-op.  #633 reaches the same outcome by re-reading `FORWARD` instead of trusting the delete's exit code; either resolution is correct, and neither may be resolved by taking `main`'s unconditional flush.

## Interim rule ordering

Allow-list rules are emitted before block-list rules, and iptables/ip6tables are first-match-wins, so a destination present in both lists is ACCEPTed.  This allow-wins behavior already exists on `main` and is preserved, not introduced, here.  GA deny-precedence is owned by AB#62830341 (net-model-2, PR #632) and is documented in a `NOTE` on `build_policy_rules_logged`.  Callers must not assume deny-precedence.

## Scope

IPv6 and CIDR filtering, plus the rule-ownership work that partial failure in two address families makes necessary.  Port and protocol filtering are out of scope: the shipping `allowedHosts` / `blockedHosts` lists are flat host strings with nowhere to attach a port, and the GA `egress.allow[]/deny[]` schema is not in `main`.  DNS hostnames are still accepted, and the GA schema migration is untouched, so this PR does not close roadmap item 15.  It does not modify the roadmap document.

Two known limits are documented rather than fixed: a mapped-range CIDR with a prefix below 96 stays classified IPv6, and shared CIDR/address-family handling is still duplicated across backends (#766).

## Validation

Re-verified on this branch (`f8c1cd2`) on the Windows dev box:

- `cargo test -p lxc_common --lib` — 113 passed, 0 failed, 0 ignored.
- `cargo clippy -p lxc_common --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.
- The three ownership repairs are mutation-tested: reintroducing any of them fails the corresponding test rather than passing quietly.  Each carries a negative control, so an implementation that simply never flushes, or retains ownership unconditionally, fails too.
- Four integration configs and scripts (`lxc_network_ipv6_cidr`, `lxc_network_invalid_cidr`, `lxc_network_dualstack_hostname`, `lxc_network_cidr_boundary`) are wired into `run_lxc_all_tests.sh`.  They need a root LXC host and were not executed in this pass.

One limit of testing on Windows is worth stating precisely, because it bounds what the flush gate's tests prove.  Every firewall command fails at *spawn* here, which is before `run_firewall_command` logs anything — so the logger observes nothing and the returned residual is identical whether the gate fires or not.  The gate is therefore asserted through a pure seam (`teardown_chain`), which pins the decision exhaustively: hooked, unhooked, and never-created.  What no test on this host can reach is the wiring one line above it, or a real `-X` succeeding.  Nothing here was exercised against live `iptables`, `ip6tables`, or real signal delivery.

AB#62830559


tests/configs/lxc_network_cidr_boundary.json
```diff
@@ -0,0 +1,33 @@
+{
+  "version": "0.6.0-alpha",
```
SohamDas2021: shall we use the latest schema version? 0.8?
Applies to all the configs you created

src/backends/lxc/common/src/network_iptables_ip6status_spec_tests.rs
```diff
@@ -0,0 +1,158 @@
+//! Spec-derived tests for the `ip6tables` usability classification and the
```
SohamDas2021: We use inline mod tests throughout the repo and in src/backends/lxc/common/src/network_iptables.rs. Why create new files for these tests?

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -81,162 +182,632 @@ impl NetworkIptablesManager {
...
+                )?;
+                created.v6_hook = true;
+                logger.log_line(&format!(
+                    "FORWARD hook installed on {} for chain {} (ip6tables).",
+                    iface, self.chain_name
+                ));
+            }
         } else {
             // Without a veth interface, we cannot safely scope rules to the container.
             // Refuse to apply host-wide rules to avoid affecting all host traffic.
```
SohamDas2021, line 1053: Fyi, this seems to be pre-existing TODO, not related to this change. 
Filed https://github.com/microsoft/mxc/issues/755 to track and assigned to you.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -81,162 +182,632 @@ impl NetworkIptablesManager {
...
-                .collect(),
-            Err(_) => Vec::new(),
+        // Try as IP address first.
+        if let Ok(addr) = host.parse::<IpAddr>() {
+            return match addr {
+                IpAddr::V4(_) => ResolvedDestinations {
+                    ipv4: vec![host.to_string()],
+                    ipv6: Vec::new(),
+                },
+                IpAddr::V6(_) => ResolvedDestinations {
```
SohamDas2021, line 288: ::ffff:a.b.c.d  is bucketed as V6 and gets an  ip6tables -d ::ffff:…  rule, but Linux emits mapped destinations as IPv4 on the wire, so that rule never matches — a mapped  blockedHosts  entry fails open under default-allow. Map these to their embedded IPv4 ( to_ipv4_mapped ) and emit an iptables rule instead?
Related test needs to be fixed as well

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -246,49 +817,93 @@ impl NetworkIptablesManager {
...
         if let Some(v) = veth_interface {
             mgr.set_veth_interface(v);
         }
-        // Bypass the rules_applied gate; if there's nothing to remove the
-        // iptables `-D`/`-F`/`-X` calls just no-op.
+        // Bypass the rules_applied gate and assume all resources may exist; if
+        // there's nothing to remove the iptables `-D`/`-F`/`-X` calls just
+        // no-op.
         mgr.rules_applied = true;
+        mgr.created = CreatedResources {
```
SohamDas2021: Hardcoding v4_chain: true, v6_chain: true here (instead of the real CreatedResources) makes the signal-path teardown -F / -X the chain unconditionally. Chain names sanitize+truncate to 20 chars, so on a name collision a signal to container A flushes container B's live chain (-X fails but -F empties it) → B fails open. The apply/rollback path is correctly guarded by CreatedResources ; the signal path should be too.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -81,162 +182,632 @@ impl NetworkIptablesManager {
...
+                        None => false,
+                    }
+                });
+                if has_egress_capable_interface {
+                    HostIpv6State::Active
+                } else {
+                    HostIpv6State::Inactive
+                }
+            }
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => HostIpv6State::Inactive,
```
SohamDas2021: Err(e) if …NotFound => Inactive treats an unmounted/unreadable /proc the same as IPv6-being-disabled (both ENOENT). The comment above over-claims certainty; a host without /proc would be misclassified as IPv6-inactive. Narrow, but worth distinguishing Unknown (fail-closed) from a genuine "IPv6 off" signal.

SohamDas2021 (changes requested):

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -4,14 +4,111 @@
...
+    Allow,
+    Deny,
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+struct ResolvedDestinations {
+    ipv4: Vec<String>,
+    ipv6: Vec<String>,
+}
+
```
bbonaby, line 36: issue: For modularity purposes I think the handling of ipv4 and ipv6  addresses and parsing of CIDRs for ipv6 should be placed in a common module that all the other backends can use. E.g windows process containers would benefit from this kind of thing as well. I'll have to see if we already do any of this CIDR parsing in other places, maybe we can reuse it.

SohamDas2021: We can take this as a follow-up and file an issue to track.

SohamDas2021 (dismissed):

SohamDas2021 (dismissed):

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -145,98 +405,106 @@ impl NetworkIptablesManager {
...
+                // Roll back whatever was created before the failure. Without
+                // this, `remove_firewall_rules` short-circuits on
+                // `rules_applied == false` and the orphan chain(s) survive, so
+                // the next attempt fails permanently on `-N` ("chain already
+                // exists") until someone cleans up by hand.
+                logger.log_line(&format!(
+                    "Firewall setup failed: {}. Cleaning up partial iptables state.",
+                    e
+                ));
+                self.teardown_chains(logger);
```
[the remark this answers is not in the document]
dhoehna: This is now guarded by the CreatedResources record.  install_firewall_rules sets created.v4_chain/v6_chain only after the corresponding -N returns success (network_iptables.rs:849-855), and teardown_created runs -F/-X only for a family whose flag is set (network_iptables.rs:988-1001).  A failed -N therefore leaves the flag false, so rollback never deletes a chain this attempt did not create, and a 20-character name collision with another container's chain is left untouched.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -145,98 +405,106 @@ impl NetworkIptablesManager {
...
-        for host in &policy.allowed_hosts {
-            let ips = Self::resolve_host(host);
-            if ips.is_empty() {
+        for host in policy
+            .allowed_hosts
+            .iter()
+            .chain(policy.blocked_hosts.iter())
+        {
+            if Self::resolve_host(host).is_empty() {
                 logger.log_line(&format!("Warning: could not resolve host '{}'", host));
```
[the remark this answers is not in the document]
dhoehna: The production apply path now resolves each entry exactly once.  build_policy_rules_logged (network_iptables.rs:493-529) calls resolve_host a single time per host and reuses that result for both the unresolved-host warning and the rule args, so the rule that gets installed is the one that was logged.  The old two-pass build_policy_rule_args is now #[cfg(test)]-only and is never on the apply path.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -145,98 +405,106 @@ impl NetworkIptablesManager {
...
-                Self::run_iptables(&["-A", &self.chain_name, "-d", ip, "-j", "DROP"], logger)?;
-            }
+        let policy_rules = Self::build_policy_rule_args(&self.chain_name, policy);
+        Self::run_iptables_rule_args(&policy_rules.ipv4, logger)?;
+        if ipv6_enabled {
+            Self::run_ip6tables_rule_args(&policy_rules.ipv6, logger)?;
+        } else if !policy_rules.ipv6.is_empty() {
+            logger.log_line(&format!(
+                "Warning: {} IPv6 firewall rule(s) not applied because ip6tables \
+                 is unavailable; IPv6 egress is unfiltered on this host.",
```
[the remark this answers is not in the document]
dhoehna: Fixed by the three-way Ip6tablesStatus classification.  When the ip6tables probe fails but the host has active IPv6, classify_ip6tables_status returns UnusableButIpv6Active (network_iptables.rs:551-557) and install_firewall_rules returns an error rather than marking the policy applied (network_iptables.rs:838-844).  Only a kernel with IPv6 inactive (KernelIpv6Disabled) is treated as safe to skip.  This path is unit-tested but has not been run against a live host without ip6tables, since I am on a Windows box.

tests/scripts/run_lxc_network_ipv6_cidr_test.sh
```diff
@@ -0,0 +1,104 @@
...
+
+if ! echo "$OUTPUT" | grep -q "Default network policy: DROP"; then
+    fail "default-deny policy was not applied."
+fi
+
+# The v6 half is the point of the test: if ip6tables is unusable the v6 rules
+# are skipped with a warning, which would make this a v4-only run.
+if echo "$OUTPUT" | grep -q "IPv6 firewall rule(s) not applied"; then
+    fail "IPv6 rules were skipped; ip6tables is unusable on this host."
+fi
```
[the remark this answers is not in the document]
dhoehna: The script now fails on a skipped hook.  run_lxc_network_ipv6_cidr_test.sh:149-153 fails if the output contains "Skipping FORWARD hook" and also requires the positive "FORWARD hook installed" confirmation before PASS, so a run where veth discovery failed and the chain was never hooked no longer passes.  I have not run this against a live LXC host; this is a Windows workstation.

tests/scripts/run_lxc_network_invalid_cidr_test.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# Invalid CIDRs are warned about and omitted from rule generation; applying the
+# remaining firewall policy should still succeed.
+if echo "$OUTPUT" | grep -qE "^(ip6?tables) .* failed:|Firewall setup failed:"; then
+    fail "invalid CIDR entry caused firewall setup to fail."
+fi
+
+if ! echo "$OUTPUT" | grep -q "Default network policy: DROP"; then
+    fail "default-deny policy was not applied."
+fi
```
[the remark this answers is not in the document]
dhoehna: Same fix as the IPv6/CIDR script.  run_lxc_network_invalid_cidr_test.sh:78-82 now fails on the "Skipping FORWARD hook" warning and requires "FORWARD hook installed", after the default-policy check, so the invalid-entry run cannot report success with an unhooked chain.  Not yet run against live iptables/LXC on this Windows box.

tests/scripts/run_lxc_all_tests.sh
```diff
@@ -46,6 +46,10 @@ run_test "LXC Object Validation" "$SCRIPT_DIR/run_lxc_object_test.sh"
 run_test "LXC Most-Specific Path" "$SCRIPT_DIR/run_lxc_most_specific_test.sh"
 run_test "LXC Denied Masking" "$SCRIPT_DIR/run_lxc_denied_masking_test.sh"
 run_test "LXC Network" "$SCRIPT_DIR/run_lxc_network_test.sh"
+run_test "LXC Network IPv6+CIDR" "$SCRIPT_DIR/run_lxc_network_ipv6_cidr_test.sh"
+run_test "LXC Network Invalid CIDR" "$SCRIPT_DIR/run_lxc_network_invalid_cidr_test.sh"
+run_test "LXC Network Dual-Stack Hostname" "$SCRIPT_DIR/run_lxc_network_dualstack_test.sh"
+run_test "LXC Network CIDR Boundary" "$SCRIPT_DIR/run_lxc_network_cidr_boundary_test.sh"
```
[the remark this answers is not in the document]
dhoehna: All four network integration scripts now assert the FORWARD hook before PASS: run_lxc_network_ipv6_cidr_test.sh:149-153, run_lxc_network_invalid_cidr_test.sh:78-82, run_lxc_network_dualstack_test.sh:220-224, and run_lxc_network_cidr_boundary_test.sh:199-203.  Each fails on "Skipping FORWARD hook" and requires the "FORWARD hook installed" line, so an unhooked chain fails the test.  These assertions have not been executed on a live LXC host.

src/backends/lxc/common/src/network_iptables_lifecycle_spec_tests.rs
```diff
@@ -0,0 +1,85 @@
...
+fn every_enforcement_mode_takes_the_contractual_firewall_gate() {
+    const SKIP_MESSAGE: &str = "Network enforcement mode does not use firewall, skipping iptables.";
+
+    for (mode, uses_firewall) in enforcement_modes_with_firewall_contract() {
+        let mut manager = NetworkIptablesManager::new(&format!("gate-{mode:?}"));
+        manager.set_veth_interface("veth-gate");
+        let policy = policy_with_enforcement_mode(mode.clone());
+        let mut logger = Logger::new(Mode::Buffer);
+
+        let _ = manager.apply_firewall_rules(&policy, &mut logger);
```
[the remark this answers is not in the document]
dhoehna: The lifecycle spec file was removed and no unit test invokes the host firewall anymore.  The enforcement-mode check is now the pure predicate enforcement_mode_uses_firewall, tested by every_enforcement_mode_takes_the_contractual_firewall_gate (network_iptables.rs:2072-2080), and the only apply_firewall_rules test uses the non-firewall Capabilities mode, which returns before touching iptables (network_iptables.rs:2051-2067).  Nothing under #[cfg(test)] shells out to iptables/ip6tables for the Firewall or Both cases.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -81,162 +159,539 @@ impl NetworkIptablesManager {
...
+    /// `/proc/net/if_inet6` is populated by the kernel only when the IPv6
+    /// module is loaded, and lists every interface IPv6 address (including the
+    /// link-local `fe80::` address present on any interface with IPv6 up). A
+    /// host booted with `ipv6.disable=1` never creates the file, and a host
+    /// with IPv6 fully disabled via sysctl has no addresses to list; either
+    /// way there is no IPv6 egress to filter. A non-empty file means IPv6 is
+    /// live, so a broken `ip6tables` is a real gap rather than a no-op.
+    fn host_has_active_ipv6() -> bool {
+        match std::fs::read_to_string("/proc/net/if_inet6") {
+            Ok(contents) => contents.lines().any(|line| !line.trim().is_empty()),
```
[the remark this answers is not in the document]
dhoehna: Loopback is now excluded.  classify_host_ipv6_state counts a line in /proc/net/if_inet6 as active only when its final field (the device) is not lo (network_iptables.rs:604-618), so a host that lists only ::1 on lo classifies as Inactive, which maps to KernelIpv6Disabled and skips the v6 chain — IPv4-only setup then succeeds even when ip6tables is absent.  This is covered by unit tests over injected file contents but has not been run on a live host.

tests/configs/lxc_network_cidr_boundary.json
dhoehna: Done — all four configs I added now use "version": "0.8.0-alpha": lxc_network_cidr_boundary.json, lxc_network_dualstack_hostname.json, lxc_network_invalid_cidr.json, and lxc_network_ipv6_cidr.json.  That matches maxSupported in schemas/schema-version.json, which is the latest 0.8 dev schema line, so this is just a version-string bump and not a schema change.

src/backends/lxc/common/src/network_iptables_ip6status_spec_tests.rs
dhoehna: Moved inline.  I deleted the five separate network_iptables_*_spec_tests.rs files and put all of the tests in a #[cfg(test)] mod tests at the bottom of network_iptables.rs (network_iptables.rs:1073), matching the repo convention.  There are 71 #[test] functions there now.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Done exactly as suggested.  ipv4_mapped_destination (network_iptables.rs:315-335) rewrites ::ffff:a.b.c.d, and mapped CIDRs inside ::ffff:0:0/96, to their embedded IPv4 form before the family split so they are filed under iptables, and bucket_resolved_addrs routes mapped AAAA DNS results to the IPv4 bucket via to_ipv4_mapped (network_iptables.rs:290-292).  The tests were updated too: ipv4_mapped_ipv6_literal_is_filed_as_ipv4 and ipv4_mapped_cidr_is_translated_to_its_ipv4_prefix assert ::ffff:127.0.0.1 to 127.0.0.1 and ::ffff:192.0.2.0/120 to 192.0.2.0/24 (network_iptables.rs:1402-1421).

src/backends/lxc/common/src/network_iptables.rs
dhoehna: The signal path now carries the real CreatedResources.  publish_created forwards the actual accumulated record to set_active_created after each resource is installed (network_iptables.rs:939-940), the watchdog snapshots it and hands it to force_cleanup (signal_cleanup.rs:180-185), and force_cleanup returns immediately when the record is empty (network_iptables.rs:1047-1051).  Nothing hardcodes v4_chain/v6_chain true outside #[cfg(test)], so a signal to container A no longer flushes a colliding chain it never created.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Fixed with an explicit Unknown state.  classify_host_ipv6_state returns Inactive for a NotFound only when /proc/net itself exists; if /proc/net is absent it returns Unknown, and any other read error is Unknown as well (network_iptables.rs:619-633).  Unknown is then treated as potentially active so setup fails closed rather than silently proceeding IPv4-only (network_iptables.rs:660-667).  So an unmounted /proc is no longer read as a confirmed "IPv6 off".

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -246,49 +817,93 @@ impl NetworkIptablesManager {
             );
         }
 
-        self.rules_applied = true;
-        Ok(true)
+        Ok(())
```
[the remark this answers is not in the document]
dhoehna: You are right that the production path still returns success with an unhooked chain when no veth is set, and that Bubblewrap (bwrap_runner.rs:205-206) never sets one.  Soham confirmed this is pre-existing on main and filed #755 to track the fail-closed fix across both LXC and Bubblewrap, so that change is out of scope for this PR.  What this PR does change is the LXC FORWARD direction (-o to -i, network_iptables.rs:894-897) and the four integration scripts, which now fail when the hook is skipped.  I have not run any of this against live iptables or LXC.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Agreed — this is the pre-existing no-veth fail-open, not something this PR introduced.  Thanks for filing #755; I will take it as the follow-up that closes the gap for both LXC and Bubblewrap.  I have left the corrected FORWARD direction and the test-side hook assertions in this PR.

src/backends/lxc/common/src/network_iptables.rs
dhoehna: Agreed, and I filed #766 for it rather than doing it here.  The CIDR parsing, IPv4-versus-IPv6 classification, and per-family bucketing are genuinely duplicated across backends, and duplicated parsing of security-relevant input is how two backends end up disagreeing about what a policy means.

I'm keeping it out of this PR because #724 is a scoped behavior fix for IPv6 destinations and CIDR ranges; extracting a shared module touches every backend that does address handling and turns a reviewable fix into a cross-backend refactor that is also worse to revert.  #766 notes that the first step should be checking whether the implementations currently disagree on any input, since any disagreement found is a live bug rather than just duplication.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -81,162 +212,711 @@ impl NetworkIptablesManager {
...
+                // The inner call has already rolled back exactly what it
+                // created, so nothing is torn down that this attempt did not
+                // install. Report and propagate.
+                logger.log_line(&format!(
+                    "Firewall setup failed: {}. Partial iptables state rolled back.",
+                    e
+                ));
+                Err(e)
             }
         }
```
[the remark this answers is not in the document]
dhoehna: You're right, and it's fixed in 3969778.  `apply_firewall_rules_inner` rolled back what it created but discarded the result with `let _ =`.  A removal command can itself fail, so the outer arm then left `self.created` empty and `rules_applied` false while a chain or FORWARD hook was still installed.  Since `rules_applied` gates both `remove_firewall_rules` and `Drop`, nothing afterward knew the survivors were ours and the leak was permanent for the life of the process.

The inner call now returns the residual alongside the error and the outer arm adopts it, so ownership is retained exactly when something survived.  The signal path was already covered -- `teardown_created` publishes the residual before returning -- so this closes the ordinary Drop and remove path specifically.

Mutation-tested: discarding the residual again makes the new test fail on the retention assertion.  The test drives `remove_firewall_rules` and observes the log rather than asserting `rules_applied`, and carries a negative control so it cannot pass by retaining unconditionally.  Not exercised against live iptables.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -246,49 +926,137 @@ impl NetworkIptablesManager {
...
+            &self.chain_name,
+            self.veth_interface.as_deref(),
+            &self.created,
+            logger,
+        );
 
         self.rules_applied = false;
+        self.created = residual;
         Ok(())
     }
```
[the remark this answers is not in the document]
dhoehna: You're right, and it's fixed in b974fb1.  This is the same defect I had just fixed on the failed-apply rollback, still present on the removal path -- `remove_firewall_rules` cleared `rules_applied` even when `teardown_created` reported a non-empty residual, and since `Drop` is gated on that same flag, a teardown whose commands failed reported itself done and threw away the last retry.

Both paths now share one `retain_residual_ownership` helper that keeps the gate open exactly when something survived.  Having one path get this right and the other not was the underlying problem, so I merged them rather than patching the second copy.

I kept the return type as `Ok(())` rather than erroring on a non-empty residual: callers treat cleanup failure as non-fatal today, and making it fatal would change behavior beyond this fix.  Retaining ownership is what actually recovers the resource.

Mutation-tested: clearing the flag unconditionally again fails the new test, which drives the second removal -- the call `Drop` makes -- and observes the log.  Not exercised against live iptables.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -81,162 +212,772 @@ impl NetworkIptablesManager {
...
+                        .to_string(),
+                );
+            }
         };
+
+        // Create custom chains, recording each family as created so rollback
+        // removes only the chains this attempt installed.
+        Self::run_iptables(&["-N", &self.chain_name], logger)?;
+        created.v4_chain = true;
+        Self::publish_created(created);
```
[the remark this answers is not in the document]
dhoehna: This is the same window raised on #632 (`:809`), and my answer is the same: real, understood, and deliberately left in place because every fix I can reach from here is worse than the gap.

`iptables -N` can succeed and the process can be signalled before ownership is published, stranding a chain nobody records as theirs.

Publishing *before* the `-N` does not fix it, it inverts it into something worse. Two starts race for the same chain name; the loser publishes ownership of a chain it did not create, and its cleanup then deletes the winner's chain and leaves the winner's container running unfiltered. That trades a leak for a fail-open.

The leak is the benign side of that trade. A stranded chain is not hooked into FORWARD by anything -- `-N` only creates it -- so it filters nothing and blocks nothing. It costs a chain name until the next teardown for that container reclaims it. Nothing escapes containment.

Closing it properly needs the create and the publish to be one atomic step against host state shared across processes, which iptables does not offer. The available answer is a lock file or a reconciler that sweeps unowned `MXC_`-prefixed chains at startup, and both are more machinery than the leak justifies right now.

I have not been able to exercise it either: it needs Linux plus real signal delivery, and this branch's suite runs on Windows where every iptables command fails at spawn. Noting it as a known gap in the PR description rather than claiming a fix I cannot demonstrate.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -305,13 +1172,313 @@ impl Drop for NetworkIptablesManager {
...
+
+        // Positive control: the same call with one resource published does
+        // reach the teardown, so the assertion above discriminates between the
+        // two cases rather than observing a permanently silent function.
+        let mut noisy = Logger::new(Mode::Buffer);
+        NetworkIptablesManager::force_cleanup(
+            "racer-that-won",
+            Some("mxcv-winner"),
+            CreatedResources::for_test(true, false, false, false),
+            &mut noisy,
```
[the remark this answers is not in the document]
SohamDas2021: Agree — this is a real test-isolation gap, and it isn't confined to this one positive control. The new ownership tests (`a_rollback_that_could_not_finish_keeps_ownership_of_what_survived` / `survivor`, `a_failed_apply_adopts_the_residual_its_rollback_left_behind` / `adopted`, `a_removal_whose_commands_failed_stays_owned_for_the_drop_retry` / `stubborn`, `a_manager_that_owns_nothing_still_reaches_the_apply_path` / `fresh`) all drive `force_cleanup` / `remove_firewall_rules` / `apply_firewall_rules`, which reach `run_firewall_command` → `Command::new("iptables"/"ip6tables")` (network_iptables.rs:743). There is no mock/injection, so they only stay side-effect-free because a non-root process makes every `iptables` command fail — their own comments say "on this host every iptables command fails." Run as **root** (a privileged CI runner, a dev running `cargo test` as root, or inside a `--privileged` container) they execute real `-F`/`-X` against predictable `MXC-<name>` chains and can flush/delete a colliding live chain, making the suite both host-mutating and privilege-dependent (non-deterministic pass/fail).

Suggest injecting the command runner (a `FnMut(&[&str]) -> Result<..>` seam, or a trait the manager holds) so these tests assert the *planned* command sequence against a fake, and keep the real-binary path for the integration scripts. That also lets you test the success arms (`-X` succeeds → ownership released) which the fail-closed-because-non-root approach can't reach today. Not a production-correctness blocker, but worth closing before merge so CI can't mutate host firewall state.

SohamDas2021 (approved):

src/backends/lxc/common/src/network_iptables.rs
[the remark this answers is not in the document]
dhoehna: Fixed in 5a6d25e.  You were right that it was not confined to the one positive control -- all four of the tests you named drove the real binary, and two of them were inheriting their pass/fail from the host rather than arranging it.

I used a `#[cfg(test)]` interception point inside `run_firewall_command` rather than a runner the manager holds, for two reasons:

- `NetworkIptablesManager` has to stay `Send` (`SandboxProcess: Send` -> `BubblewrapSandboxProcess` -> `BwrapChild.fw_manager`), so a `FnMut` field would need a `Send` bound threaded through it.  The seam adds no field at all.
- `force_cleanup` is an associated function and `Drop::drop` takes no arguments, so neither can reach an injected runner -- and those are exactly the two paths under test.

`run_firewall_command` is the narrowest waist that still works.  `run_iptables` and `run_ip6tables` only prepend the binary name, and below it there is nothing but `std`, so it is the single place holding both the complete argv to record and the `Result<bool, String>` to script.  Production signatures are unchanged, and interception is opt-in: a test that installs no fake behaves exactly as it does today, so the roughly seventy tests that never touch the firewall needed no churn.  The `ip6tables -S` probe in `ip6tables_status` got its own gate for the same reason.

The success arms are covered now.  `a_removal_whose_commands_all_succeeded_releases_ownership` asserts that `-F` and `-X` both succeed and ownership is released -- the arm you noted was unreachable while the tests depended on being non-root.  The "on this host every iptables command fails" comment is gone; that test scripts its own failure instead.

Verified as root against a real iptables (WSL2 Ubuntu 24.04, iptables 1.8.10 nf_tables), A/B against the parent commit, with a logging shim on `PATH` counting spawns and a live `MXC-racer-that-won` chain planted before each run:

|  | parent f8c1cd2 | 5a6d25e |
| --- | --- | --- |
| firewall binaries spawned | 35 | 0 |
| planted live chain | deleted | intact |
| suite result | 117 passed | 118 passed |

The old run created `MXC-fresh` in both families, added five real rules to it, and deleted the planted chain -- while reporting all green.  That is the non-determinism you described, and it is worth noting that a before/after dump of the table does not catch it: the suite deletes the chains it creates, so only a planted collision or a spawn trace shows the mutation.

PR #730, opened 2026-08-01
Add fail-closed base resolution and SemVer libraries for the versioning gates

MGudgin:
This PR adds the two shared libraries every history-aware versioning gate needs
-- resolving what a pull request is being compared against, and ordering schema
versions -- together with a guard that the versioning job cannot report success
without running its tests.

Both libraries fail closed. A gate that cannot determine its base, or cannot
order two versions, reports failure rather than skipping, because a check that
goes quiet when confused is a check that passes on exactly the inputs it exists
to catch.

## Details
* `lib/git-base.js` resolves the pull-request base commit and reads files as
  they were at that commit. It returns `null` only when a file genuinely does
  not exist there, and throws when git itself fails, so "I could not look" is
  never mistaken for "it was deleted".
* Paths are normalised to git's forward-slash syntax, and `ls-tree` is read
  NUL-delimited. Git C-quotes any path containing non-ASCII bytes, quotes,
  backslashes or control characters, and a Windows caller naturally produces
  backslashes; either would miss the literal comparison and make an existing
  file read as absent.
* The same normalisation collapses `.` and `..` segments and doubled slashes,
  and an absolute path is rebased onto the repository root -- `path.join(repoRoot,
  ...)` is what a caller naturally writes, and it matches no `ls-tree` entry. A
  path that cannot name a file in the repository at all is refused outright
  rather than reported as absent, so `null` keeps meaning exactly "not present
  at that commit" and a gate cannot read a malformed path as "newly added,
  nothing to compare".
* `lib/version.js` parses strict SemVer. Numeric identifiers reject leading
  zeros, and the core components reject anything not exactly representable,
  since an unbounded digit run converts to `Infinity` and `Infinity - Infinity`
  is `NaN`, for which every ordinary ordering check is false. Build metadata is
  accepted, kept for round-tripping, and excluded from precedence as the
  specification requires.
* Numeric *prerelease* identifiers are instead ordered exactly, without ever
  converting them to `Number`: SemVer leaves them unbounded, and the grammar has
  already rejected leading zeros, so the longer digit string is the larger value
  and equal lengths compare lexically. Converting would collapse distinct values
  above 2^53 onto one float and long ones onto `Infinity`, whose difference is
  `NaN` -- falsy, so the comparison would report equality.
* The parsers take strings only. `RegExp.exec` coerces its argument, so a
  single-element array, a boxed `String` or any object with a `toString` would
  otherwise parse, and `raw` would carry the non-string.
* The comparison helpers reject anything that is not a parsed version. Reading
  `.major` off a raw string or a failed parse yields `undefined`, falls through
  every branch, and reports the two versions as equal. `compareVersions` applies
  the stricter check, since a `parseMajorMinor` result carries no `patch` or
  `prerelease` and would otherwise subtract `undefined` and yield `NaN`;
  comparing version lines is what `compareMajorMinor` is for. Both guards check
  the domain the parsers actually produce -- non-negative components, and a
  prerelease matching the grammar `parseVersion` applies -- since
  `Number.isSafeInteger` alone admits negatives and a bare `typeof` check admits
  a prerelease the parser would have refused, either of which would be ordered
  rather than rejected.
* `check-tests-present.js` runs as `pretest` and fails when the tests directory
  is missing or holds no test files. It reads entry types rather than names, so a
  *directory* named `foo.test.js` does not satisfy it. `node --test tests/*.test.js` exits 0 when
  the pattern matches nothing, so moving or renaming the directory would leave
  the job green while executing nothing.
* The versioning workflow checks out full history so a base ref can be
  resolved.

## Tests
* 39 unit and integration tests pass. The git helpers are exercised against
  real scratch repositories, since fail-closed base resolution and path
  handling cannot be tested any other way.
* Path coverage includes a backslash-separated lookup and a non-ASCII filename,
  each of which read as absent before normalisation and NUL-delimited output.
* A shallow clone with no reachable base is asserted to raise, not to skip.
* SemVer coverage includes leading zeros, unrepresentable components, build
  metadata parsing and its exclusion from ordering, malformed prerelease and
  build identifiers, prerelease precedence, and the comparison type guards.
* Numeric prerelease ordering is pinned across the 2^53 boundary and at 400
  digits, in both directions and for equal values, together with the full
  precedence chain from the specification's own example, which fixes
  numeric-before-alphanumeric and numeric-not-lexical ordering at once.
* `compareVersions` is asserted to throw for a version line, in either argument
  position, and for hand-rolled objects missing `patch` or `prerelease`, while
  `compareMajorMinor` still accepts both shapes.
* The presence check now has its own tests, run as real processes against
  scratch layouts so the exit code -- the only thing CI reads -- is what is
  asserted: a missing directory, an empty one, a directory named `foo.test.js`
  with nothing else, and that same directory alongside a real test file.
* Path coverage adds the absolute `path.join(repoRoot, ...)` form, doubled
  slashes and a `..` round trip, each of which read as absent before; a file
  that is genuinely missing still reads as absent, and a path outside the
  repository raises instead.
* The parsers are asserted to reject arrays, boxed strings, `toString` objects
  and numbers, and the comparison guards to reject negative components and every
  prerelease spelling `parseVersion` refuses, while still accepting the shapes
  it produces.
* `check-schema-versions.js` and `validate-configs.js` both still pass, so the
  parser change does not disturb existing consumers.
* The resolver takes its options as a single `{ argv, env }` object, and a test
  pins that: passing them positionally silently falls back to `process.argv`
  and `process.env`, which would make these tests read whatever the ambient job
  happens to set. The suite is run both with and without the workflow's own
  `MXC_VERSIONING_BASE_REF` and `GITHUB_ACTIONS` values.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/730)

### DIFF
--- a/scripts/versioning/check-tests-present.js
+++ b/scripts/versioning/check-tests-present.js
@@ -1,0 +1,42 @@
+#!/usr/bin/env node
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Guards the guard: fail when there are no test files to run.
+//
+// `node --test tests/*.test.js` exits 0 when the pattern matches nothing, so
+// renaming or moving the tests directory would leave the versioning job green
+// while executing nothing at all. Every other gate in this directory is only as
+// trustworthy as its tests actually running, so this runs ahead of them.
+//
+//   node scripts/versioning/check-tests-present.js
+
+const { readdirSync, existsSync } = require("fs");
+const { join, resolve } = require("path");
+
+const testsDir = resolve(__dirname, "tests");
+
+if (!existsSync(testsDir)) {
+  console.error(
+    `Test presence check FAILED: ${join("scripts", "versioning", "tests")} does not exist.`
+  );
+  process.exit(1);
+}
+
+// Read the entry types, not just the names. A *directory* named `foo.test.js`
+// matches the suffix but runs nothing, so a name-only check would report the
+// tests are present while `node --test` still executes nothing -- exactly the
+// silent pass this script exists to prevent.
+const files = readdirSync(testsDir, { withFileTypes: true })
+  .filter((entry) => entry.isFile() && entry.name.endsWith(".test.js"))
+  .map((entry) => entry.name);
+
+if (files.length === 0) {
+  console.error(
+    "Test presence check FAILED: no *.test.js files found; the test step " +
+      "would report success without executing anything."
+  );
+  process.exit(1);
+}
+
+console.log(`Test presence OK: ${files.length} test file(s).`);
--- a/scripts/versioning/lib/git-base.js
+++ b/scripts/versioning/lib/git-base.js
@@ -140,9 +140,18 @@ function listFilesAtCommit(repoRoot, commit, path) {
   // containing non-ASCII bytes, quotes, backslashes or control characters, and
   // the literal comparison below would silently miss it.
   const gitPath = repoRelativeGitPath(repoRoot, path);
+  const args = [
+    "--literal-pathspecs",
+    "ls-tree",
+    "-r",
+    "-z",
+    "--name-only",
+    commit,
+  ];
+  if (gitPath) args.push("--", gitPath);
   const output = git(
     repoRoot,
-    ["ls-tree", "-r", "-z", "--name-only", commit, "--", gitPath],
+    args,
     { trim: false }
   );
   return output ? output.split("\0").filter(Boolean) : [];
--- a/scripts/versioning/lib/git-base.js
+++ b/scripts/versioning/lib/git-base.js
@@ -27,6 +27,7 @@ function git(repoRoot, args, { trim = true } = {}) {
   const output = execFileSync("git", args, {
     cwd: repoRoot,
     encoding: "utf8",
+    maxBuffer: Infinity,
     stdio: ["ignore", "pipe", "pipe"],
   });
   return trim ? output.trimEnd() : output;
@@ -81,7 +82,10 @@ function resolveBaseCommit(
 // slashes are all meaningful to the filesystem but never appear in git's own
 // output, so they are collapsed here rather than compared literally.
 function toGitPath(path) {
-  const text = String(path).split("\\").join("/");
+  if (typeof path !== "string") {
+    throw new TypeError(`path must be a string, got ${typeof path}`);
+  }
+  const text = path.split("\\").join("/");
   const segments = [];
   for (const segment of text.split("/")) {
     if (segment === "" || segment === ".") continue;
@@ -111,17 +115,20 @@ function toGitPath(path) {
 function repoRelativeGitPath(repoRoot, path) {
   const normalized = toGitPath(path);
   const root = toGitPath(resolve(repoRoot));
-  if (!isAbsolute(String(path).split("\\").join("/")) && !hasDriveLetter(normalized)) {
+  if (!isAbsolute(path.split("\\").join("/")) && !hasDriveLetter(normalized)) {
     return normalized;
   }
-  if (normalized === root) return "";
-  const prefix = `${root}/`;
-  if (!normalized.startsWith(prefix)) {
+  const caseInsensitive = process.platform === "win32" || hasDriveLetter(root);
+  const comparablePath = caseInsensitive ? normalized.toLowerCase() : normalized;
+  const comparableRoot = caseInsensitive ? root.toLowerCase() : root;
+  if (comparablePath === comparableRoot) return "";
+  const prefix = `${comparableRoot}/`;
+  if (!comparablePath.startsWith(prefix)) {
     throw new Error(
       `path "${path}" is outside the repository root "${repoRoot}"`
     );
   }
-  return normalized.slice(prefix.length);
+  return normalized.slice(root.length + 1);
 }
 
 function hasDriveLetter(path) {
--- a/scripts/versioning/lib/git-base.js
+++ b/scripts/versioning/lib/git-base.js
@@ -1,0 +1,163 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const { execFileSync, spawnSync } = require("child_process");
+const { isAbsolute, resolve } = require("path");
+
+function argumentValue(argv, name) {
+  const index = argv.indexOf(name);
+  if (index < 0) return null;
+  if (!argv[index + 1]) throw new Error(`${name} requires a ref`);
+  return argv[index + 1];
+}
+
+function requestedBaseRef(argv = process.argv.slice(2), env = process.env) {
+  const fromArgs = argumentValue(argv, "--base-ref");
+  if (fromArgs) return fromArgs;
+  if (env.MXC_VERSIONING_BASE_REF) return env.MXC_VERSIONING_BASE_REF;
+  if (env.GITHUB_ACTIONS) {
+    throw new Error(
+      "MXC_VERSIONING_BASE_REF is required in GitHub Actions; refusing to skip history checks"
+    );
+  }
+  return null;
+}
+
+function git(repoRoot, args, { trim = true } = {}) {
+  const output = execFileSync("git", args, {
+    cwd: repoRoot,
+    encoding: "utf8",
+    stdio: ["ignore", "pipe", "pipe"],
+  });
+  return trim ? output.trimEnd() : output;
+}
+
+function refExists(repoRoot, ref) {
+  return (
+    spawnSync("git", ["rev-parse", "--verify", "--quiet", `${ref}^{commit}`], {
+      cwd: repoRoot,
+      stdio: "ignore",
+    }).status === 0
+  );
+}
+
+function resolveBaseCommit(
+  repoRoot,
+  { argv = process.argv.slice(2), env = process.env } = {}
+) {
+  let ref = requestedBaseRef(argv, env);
+  if (ref) {
+    if (!refExists(repoRoot, ref)) {
+      throw new Error(`versioning base ref "${ref}" is unavailable`);
+    }
+  } else {
+    ref = ["origin/main", "HEAD^"].find((candidate) =>
+      refExists(repoRoot, candidate)
+    );
+    if (!ref) {
+      throw new Error(
+        "could not resolve a versioning base; pass --base-ref <ref> or set MXC_VERSIONING_BASE_REF"
+      );
+    }
+  }
+
+  let commit;
+  try {
+    commit = git(repoRoot, ["merge-base", "HEAD", ref]);
+  } catch (error) {
+    throw new Error(
+      `could not compute merge-base between HEAD and "${ref}": ${error.message}`
+    );
+  }
+  if (!commit) throw new Error(`empty merge-base for HEAD and "${ref}"`);
+  return { ref, commit };
+}
+
+// Git speaks repository-relative paths with forward slashes. Callers on Windows
+// naturally produce backslashes (path.join), which would never match `ls-tree`
+// output and would make an existing file look absent.
+//
+// Non-canonical spellings fail the same way: `.` and `..` segments and doubled
+// slashes are all meaningful to the filesystem but never appear in git's own
+// output, so they are collapsed here rather than compared literally.
+function toGitPath(path) {
+  const text = String(path).split("\\").join("/");
+  const segments = [];
+  for (const segment of text.split("/")) {
+    if (segment === "" || segment === ".") continue;
+    if (segment === "..") {
+      // A `..` with nothing to pop escapes the repository root, so it can never
+      // name a tracked file. Throw rather than silently normalising it away,
+      // which would make an out-of-tree path read as an in-tree one.
+      if (segments.length === 0) {
+        throw new Error(`path "${path}" escapes the repository root`);
+      }
+      segments.pop();
+      continue;
+    }
+    segments.push(segment);
+  }
+  return segments.join("/");
+}
+
+// Reduces a caller's path to the repository-relative form git expects. The
+// natural call shape is `path.join(repoRoot, ...)`, which yields an absolute
+// path that matches no `ls-tree` entry, so an absolute path inside the
+// repository is rebased onto its root and one outside it is refused.
+//
+// Refusing is the point: returning a value that simply fails to match would be
+// reported as "the file is not present at that commit", and a gate reading that
+// as "newly added, nothing to compare" would pass without checking anything.
+function repoRelativeGitPath(repoRoot, path) {
+  const normalized = toGitPath(path);
+  const root = toGitPath(resolve(repoRoot));
+  if (!isAbsolute(String(path).split("\\").join("/")) && !hasDriveLetter(normalized)) {
+    return normalized;
+  }
+  if (normalized === root) return "";
+  const prefix = `${root}/`;
+  if (!normalized.startsWith(prefix)) {
+    throw new Error(
+      `path "${path}" is outside the repository root "${repoRoot}"`
+    );
+  }
+  return normalized.slice(prefix.length);
+}
+
+function hasDriveLetter(path) {
+  return /^[A-Za-z]:\//.test(path);
+}
+
+function listFilesAtCommit(repoRoot, commit, path) {
+  // `-z` gives NUL-delimited, unquoted names. Without it git C-quotes any path
+  // containing non-ASCII bytes, quotes, backslashes or control characters, and
+  // the literal comparison below would silently miss it.
+  const gitPath = repoRelativeGitPath(repoRoot, path);
+  const output = git(
+    repoRoot,
+    ["ls-tree", "-r", "-z", "--name-only", commit, "--", gitPath],
+    { trim: false }
+  );
+  return output ? output.split("\0").filter(Boolean) : [];
+}
+
+/// Returns the file's content at `commit`, or `null` when the file does not
+/// exist there. Throws when git itself fails, or when the path cannot name a
+/// file in this repository at all, so neither a lookup error nor a malformed
+/// path is ever mistaken for an absent file.
+function readFileAtCommit(repoRoot, commit, path) {
+  const gitPath = repoRelativeGitPath(repoRoot, path);
+  if (!listFilesAtCommit(repoRoot, commit, gitPath).includes(gitPath)) {
+    return null;
+  }
+  return git(repoRoot, ["show", `${commit}:${gitPath}`], { trim: false });
+}
+
+module.exports = {
+  listFilesAtCommit,
+  readFileAtCommit,
+  repoRelativeGitPath,
+  requestedBaseRef,
+  resolveBaseCommit,
+  toGitPath,
+};
--- a/scripts/versioning/lib/git-base.js
+++ b/scripts/versioning/lib/git-base.js
@@ -4,6 +4,8 @@
 const { execFileSync, spawnSync } = require("child_process");
 const { isAbsolute, resolve } = require("path");
 
+const GIT_MAX_BUFFER = 100 * 1024 * 1024;
+
 function argumentValue(argv, name) {
   const index = argv.indexOf(name);
   if (index < 0) return null;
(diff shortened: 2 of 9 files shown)


scripts/versioning/lib/version.js
```diff
@@ -0,0 +1,139 @@
...
+  const right = b.split(".");
+  for (let i = 0; i < Math.max(left.length, right.length); i++) {
+    if (left[i] === undefined) return -1;
+    if (right[i] === undefined) return 1;
+
+    const leftNumeric = /^\d+$/.test(left[i]);
+    const rightNumeric = /^\d+$/.test(right[i]);
+    if (leftNumeric && rightNumeric) {
+      const difference = Number(left[i]) - Number(right[i]);
+      if (difference) return difference < 0 ? -1 : 1;
```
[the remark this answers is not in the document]
MGudgin: Fixed, and the gap was self-inconsistent: `parseNumericIdentifier` bounds `major`, `minor` and `patch` for exactly this reason, and `version.test.js` has a test literally named *"parseVersion rejects components too large to order"* -- but numeric *prerelease* identifiers never went through that check. Confirmed both cases you named:

```
1.0.0-9007199254740992 vs ...93   => 0   (equal)
1.0.0-<400 ones>       vs <400 twos> => 0   (equal)
```

Took your suggested fix rather than rejecting. Leading zeros are already rejected by the grammar, so both operands are canonical: more digits means larger, and at equal length lexical order *is* numeric order. That is exact at any length, so unlike the core components these need no bound at all -- and it avoids turning away versions that SemVer considers valid, since the spec leaves numeric identifiers explicitly unbounded.

Kept the `Number` bound on major/minor/patch: those are stored as numbers and compared by subtraction, so the bound is intrinsic there, and `version.test.js:45` pins it deliberately. That test now also covers `patch`, which was enforced but untested.

New tests pin ordering across the 2^53 boundary and at 400 digits, in both directions and for equal values, plus the full precedence chain from SemVer 11.4.12 -- which fixes numeric-before-alphanumeric and numeric-not-lexical ordering at the same time.

scripts/versioning/lib/version.js
```diff
@@ -0,0 +1,139 @@
...
+function compareVersions(a, b) {
+  assertParsed(a, "a");
+  assertParsed(b, "b");
+  if (a.major !== b.major) return a.major - b.major;
+  if (a.minor !== b.minor) return a.minor - b.minor;
+  if (a.patch !== b.patch) return a.patch - b.patch;
+  // Build metadata is intentionally not consulted: SemVer gives it no
+  // precedence, so 1.2.3+a and 1.2.3+b are the same version.
+  return comparePrerelease(a.prerelease, b.prerelease);
+}
```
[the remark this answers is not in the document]
MGudgin: Fixed. Confirmed the fail-open exactly as described:

```
compareVersions(parseMajorMinor("1.2"), parseVersion("1.2.3"))  =>  NaN
  ... < 0  =>  false
  ... > 0  =>  false
```

The irony is that `assertParsed` exists specifically to stop this shape, but checked only `major` and `minor` -- which is precisely the shape `parseMajorMinor` returns, so the guard admitted the one object that then produced `NaN`.

Split the guard as you suggest: `assertFullVersion` additionally requires a safe-integer `patch` and a string `prerelease`, and `compareVersions` uses it, while `compareMajorMinor` keeps the loose check since comparing a version line is its legitimate purpose. A hand-rolled `{major, minor}` or `{major, minor, patch}` is now rejected too.

Tests assert the throw in both argument positions and for `line`-vs-`line`, and separately that `compareMajorMinor` still accepts both shapes so the intended use is not collateral damage.

scripts/versioning/tests/git-base-integration.test.js
```diff
@@ -0,0 +1,155 @@
...
+test("resolveBaseCommit honours an explicit base ref", () => {
+  const dir = scratchRepo((d) => {
+    writeFileSync(join(d, "a.txt"), "one\n");
+    commitAll(d, "first");
+    writeFileSync(join(d, "a.txt"), "two\n");
+    commitAll(d, "second");
+  });
+  try {
+    const first = git(dir, ["rev-parse", "HEAD^"]);
+    const base = resolveBaseCommit(dir, [], { MXC_VERSIONING_BASE_REF: "HEAD^" });
```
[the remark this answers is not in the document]
MGudgin: This one is stale rather than wrong -- it was raised against `beab2f51`, which was force-pushed away before the comment landed.

That commit did have three positional calls; the current head has none:

| | `beab2f51` (reviewed) | `1feaca3b` (current) |
| --- | --- | --- |
| `resolveBaseCommit(dir, [], ...)` | 3 | 0 |

All call sites use `{ argv, env }`, and there is a dedicated regression test -- *"resolveBaseCommit takes its options as a single object"* -- that sets a conflicting ambient `MXC_VERSIONING_BASE_REF` and asserts the injected value still wins, which is exactly the coincidental-pass failure mode you describe. The two suppressed comments on lines 134 and 149 of the same file are the same stale diff.

Your second review, against the current commit, generated 0 new comments and re-raised only the two `version.js` items -- both of which were real and are now fixed.

MGudgin:
Both `version.js` findings were real fail-opens and are fixed; the test-signature finding was raised against a commit that had already been force-pushed away. Details are in the per-comment replies.

**Stack rebase.** This PR is the base of #731 (7c) and #732 (7d), plus the unopened 8a branch, so all three were rebased onto the amended base and force-pushed with lease:

| Branch | Was | Now |
| --- | --- | --- |
| `versioning_phase6a_gate_libraries` (this PR) | `1feaca3b` | `18daed71` |
| `versioning_phase7c_detector_lib` (#731) | `d087d3c8` | `41ddb792` |
| `versioning_phase7d_dev_schema_gate` (#732) | `8977c188` | `04ca0fea` |
| `versioning_phase8a_version_windows` | `3ff44853` | `a9dd485e` |

7d and 8a were additionally stale against #731's own fix rounds from earlier today -- both still carried the original `14fced3e` -- so this brings the whole stack back onto one line.

**Verification at each rewritten tip:**

* `npm test` in `scripts/versioning`: 31 here, 82 on 7c, 90 on 7d, 112 on 8a, all passing.
* `check-schema-versions.js`, `validate-configs.js`, `check-rust-toolchain-sync.js` and `check-version-sync.js` all pass on the amended base.
* The 7d gate (`check-dev-schema-compat.js`) was run end to end against `origin/main` and reports no breaking change -- that is the real consumer of both libraries, so it exercises the tightened `compareVersions` in situ.
* 8a's `src/`, `sdk/`, `schemas/` and `docs/` trees are byte-identical to their pre-rebase state, so its Rust and SDK code needed no re-verification; the only delta reaching it is this JS library, which its 112-test suite covers.

Worth flagging that `compareVersions` is now stricter and will *throw* rather than return `NaN` when handed a `parseMajorMinor` result. The current consumers in 8a (`check-version-windows.js`, `lib/version-windows.js`) only ever pass version lines to `compareMajorMinor`, which is unaffected, but any future caller comparing a line against a full version needs `compareMajorMinor`.

scripts/versioning/lib/git-base.js
```diff
@@ -0,0 +1,111 @@
...
+    );
+  }
+  if (!commit) throw new Error(`empty merge-base for HEAD and "${ref}"`);
+  return { ref, commit };
+}
+
+// Git speaks repository-relative paths with forward slashes. Callers on Windows
+// naturally produce backslashes (path.join), which would never match `ls-tree`
+// output and would make an existing file look absent.
+function toGitPath(path) {
```
SohamDas2021, line 86: Fail-open on absolute / non-canonical paths.  toGitPath  only flips separators and strips a leading  ./, so absolute paths ( C:\mxc\README.md ),  ..  segments ( schemas/../README.md ), and doubled/ ././  slashes don't match ls-tree  output-  readFileAtCommit  returns null, which consumers read as "file absent at base / newly added, skip the check." Silent on Windows; POSIX absolute paths throw instead. The natural consumer call join(repoRoot, …)  produces exactly this absolute form.

scripts/versioning/lib/version.js
```diff
@@ -0,0 +1,173 @@
...
+
+function validDotSeparated(text, identifier) {
+  const parts = text.split(".");
+  return parts.length > 0 && parts.every((part) => identifier.test(part));
+}
+
+function parseVersion(value) {
+  const match =
+    /^(\d+)\.(\d+)\.(\d+)(?:-([^+]*))?(?:\+(.*))?$/.exec(value || "");
```
SohamDas2021, line 50: SemVer parsers coerce non-string values. RegExp.exec(value || "") accepts single-element arrays / boxed strings -  parseVersion(["1.2.3"]) and parseMajorMinor(["1.2"]) both succeed, violating strict fail-closed parsing.

SohamDas2021 (changes requested):

scripts/versioning/check-tests-present.js
```diff
@@ -0,0 +1,36 @@
...
+const testsDir = resolve(__dirname, "tests");
+
+if (!existsSync(testsDir)) {
+  console.error(
+    `Test presence check FAILED: ${join("scripts", "versioning", "tests")} does not exist.`
+  );
+  process.exit(1);
+}
+
+const files = readdirSync(testsDir).filter((name) => name.endsWith(".test.js"));
```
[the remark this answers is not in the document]
MGudgin: Fixed. Confirmed the exact case -- a scratch `tests/` containing nothing but a *directory* named `placeholder.test.js`:

```
before: Test presence OK: 1 test file(s).   exit 0
after:  Test presence check FAILED: no *.test.js files found ...   exit 1
```

Now reads entry types via `readdirSync(testsDir, { withFileTypes: true })` and filters `isFile()`.

Worth saying this one stung: the script exists precisely because `node --test tests/*.test.js` exits 0 on an empty glob, so a name-only check reproduced the same silent pass one level up. It also had no tests of its own -- the PR body claimed only that it "was run against an empty tests directory and a missing one". It now has four, spawned as real processes against scratch layouts so the exit code is what gets asserted: missing directory, empty directory, a `foo.test.js` *directory* alone, and that same directory alongside a real test file.

scripts/versioning/lib/version.js
MGudgin: Confirmed and fixed. Reproduced exactly as you describe, and it is a little broader than arrays and boxed strings -- any object with a `toString` gets coerced, and `raw` then retains the non-string, so round-tripping breaks too:

```
parseVersion(["1.2.3"])              => {major:1,...,raw:["1.2.3"]}
parseVersion(new String("1.2.3"))    => {major:1,...}
parseVersion({toString:()=>"1.2.3"}) => {major:1,...,raw:{}}
```

Both parsers now require `typeof value === "string"` up front, so all of the above return `null` while `"1.2.3"` is unaffected. Tests pin arrays, boxed strings, `toString` objects, numbers, booleans, `null` and `undefined` for `parseVersion` and `parseMajorMinor` alike.

Thanks -- this is the kind of thing the `|| ""` was quietly inviting.

scripts/versioning/lib/git-base.js
MGudgin: Confirmed and fixed, though I want to be precise about the impact since it changes how urgent this is.

**The bug is real.** Every non-canonical spelling you name returned `null` from `readFileAtCommit`:

| path | before | after |
| --- | --- | --- |
| `README.md` | content | content |
| `path.join(repoRoot, "README.md")` | **`null`** | content |
| `schemas/../README.md` | **`null`** | content |
| `scripts//versioning/lib/version.js` | **`null`** | content |
| `././README.md` | **`null`** | content |
| `no/such/file.txt` | `null` | `null` (correctly absent) |
| `C:/elsewhere/README.md` | **`null`** | **throws** |
| `../README.md` | **`null`** | **throws** |

And you are right that `path.join(repoRoot, ...)` is the natural consumer call shape, so this was a trap laid for the next caller rather than a theoretical one.

**On the fail-open framing** -- I checked the only consumer that exists today, `check-dev-schema-compat.js` on the 7d branch. It passes literal relative forward-slash constants (`"schemas/schema-version.json"`, and a template literal for the dev schema), and it treats *every* `null` as `fail(...)`, never as a skip:

```js
if (!baseSchema) fail([`${basePath} is missing at ${base.ref}`]);
```

So no gate is currently fail-open on this. What was actually wrong is that `readFileAtCommit`'s own contract was false: its doc comment promises `null` means "the file does not exist there", and for an absolute path that was simply untrue. A future consumer writing the obvious `path.join` call, or treating a missing file at the base as "newly added, nothing to compare", would have inherited exactly the failure you describe.

**The fix.** `toGitPath` still normalises separators but now also collapses `.` segments, doubled slashes and `..`, and throws when a `..` would escape above the root. A new `repoRelativeGitPath(repoRoot, path)` rebases an absolute path onto the repository root and **throws** for anything outside it, rather than returning a value that merely fails to match. That keeps `null` meaning exactly one thing.

Because the failure mode you found was "silently returns something that reads as absent", I deliberately made every unrepresentable path raise instead. I also ran an adversarial pass over the fix specifically hunting for new fail-opens in it -- drive-relative `C:foo.txt`, UNC `\\server\share\x`, mismatched drive letters, `repoRoot` passed as the path itself, and `repoRoot` given with a trailing slash or relatively. All either resolve correctly or throw; none return a silently-wrong path. Windows case-insensitivity versus git's case-sensitive index throws with "outside the repository root", which is noisy but fail-closed.

New tests cover the absolute form, doubled slashes and a `..` round trip, that a genuinely missing file still returns `null`, and that an out-of-repo path raises.

MGudgin:
Both **suppressed** findings from the latest Copilot review were real and are fixed. Recording them here since they have no reply thread.

**`version.js:113` -- `Number.isSafeInteger` admits negatives.** Confirmed: `compareVersions({major:-1,minor:0,patch:0,prerelease:""}, parseVersion("1.0.0"))` returned `-2`, ordering a value no parser can produce instead of rejecting it.

**`version.js:128` -- the full-version guard was too weak.** This one is my own code from earlier today, added to fix the previous round. Confirmed all four cases:

| object | before | after |
| --- | --- | --- |
| `patch: -5` | compared (`-5`) | throws |
| `prerelease: "01"` | compared (`-1`) | throws |
| `prerelease: "alpha..1"` | compared (`-1`) | throws |
| `prerelease: "has space"` | compared (`-1`) | throws |

`parseVersion` rejects every one of those inputs, so the guard had drifted from the parser it stands in for.

Both are fixed by checking the domain the parsers actually produce: a shared `isVersionComponent` requires a non-negative safe integer for `major`, `minor` and `patch`, and `assertFullVersion` validates `prerelease` against the same `PRERELEASE_IDENTIFIER` grammar via the existing `validDotSeparated`, rather than merely checking `typeof`. Reusing the parser's own predicate is deliberate -- it is what stops the two drifting again. The legitimate shapes still pass: an empty prerelease compares equal, and `"alpha.1"` still sorts below the release.

---

**Stack.** Rebased and force-pushed with lease:

| Branch | Was | Now |
| --- | --- | --- |
| `versioning_phase6a_gate_libraries` (this PR) | `18daed71` | `3cb26ce8` |
| `versioning_phase7c_detector_lib` (#731) | `41ddb792` | `523b4dba` |
| `versioning_phase7d_dev_schema_gate` (#732) | `04ca0fea` | `2b42c509` |

The unopened `versioning_phase8a_version_windows` branch is deliberately **not** rebased this round -- it has uncommitted work in its worktree, so I left it alone. It is one commit behind and will need a rebase onto `2b42c509` before it opens.

**Verification at each rewritten tip:** `npm test` in `scripts/versioning` gives 39 here, 90 on 7c and 98 on 7d, all passing; `check-schema-versions.js`, `validate-configs.js`, `check-rust-toolchain-sync.js` and `check-version-sync.js` pass on this branch; and `check-dev-schema-compat.js` -- the real consumer of both libraries -- was run end to end on 7d against `origin/main` and reports no breaking change.

Given that this is the third consecutive round where findings landed in freshly written fix code, I also ran a dedicated adversarial pass over this diff before pushing, targeting new fail-opens in the fixes themselves. It came back clean, having probed drive-relative and UNC paths, mismatched drive letters, `repoRoot` as the path, prerelease grammar drift between guard and parser, and whether each new test actually fails against the pre-fix code.

SohamDas2021 (approved):

bbonaby:
Note:  I clicked the  "Rebase stack" button so this PR gets rebased on top of the latest main. No other changes added.

image

SohamDas2021 (approved):

bbonaby:
did another rebase of the stack like above. No direct code changes from me. Some changes on main fixed a pipeline issue, so rebase of the stack on the latest main was necessary for the CI to be green again.

SohamDas2021 (dismissed):

scripts/versioning/lib/version.js
```diff
@@ -0,0 +1,198 @@
...
+  if (a.major !== b.major) return a.major - b.major;
+  if (a.minor !== b.minor) return a.minor - b.minor;
+  if (a.patch !== b.patch) return a.patch - b.patch;
+  // Build metadata is intentionally not consulted: SemVer gives it no
+  // precedence, so 1.2.3+a and 1.2.3+b are the same version.
+  return comparePrerelease(a.prerelease, b.prerelease);
+}
+
+function majorMinor(value) {
+  const parsed = typeof value === "string" ? parseVersion(value) : value;
```
shschaefer: This allows malformed objects to be reported as undefined.undefined which will let upstream callers equate two unrelated calls.

scripts/versioning/lib/git-base.js
```diff
@@ -0,0 +1,163 @@
...
+
+// Git speaks repository-relative paths with forward slashes. Callers on Windows
+// naturally produce backslashes (path.join), which would never match `ls-tree`
+// output and would make an existing file look absent.
+//
+// Non-canonical spellings fail the same way: `.` and `..` segments and doubled
+// slashes are all meaningful to the filesystem but never appear in git's own
+// output, so they are collapsed here rather than compared literally.
+function toGitPath(path) {
+  const text = String(path).split("\\").join("/");
```
shschaefer: This coercion forces downstream incorrect errors to result instead of throwing/returning an error if the path is not a string/path.

scripts/versioning/lib/git-base.js
```diff
@@ -0,0 +1,163 @@
...
+// Refusing is the point: returning a value that simply fails to match would be
+// reported as "the file is not present at that commit", and a gate reading that
+// as "newly added, nothing to compare" would pass without checking anything.
+function repoRelativeGitPath(repoRoot, path) {
+  const normalized = toGitPath(path);
+  const root = toGitPath(resolve(repoRoot));
+  if (!isAbsolute(String(path).split("\\").join("/")) && !hasDriveLetter(normalized)) {
+    return normalized;
+  }
+  if (normalized === root) return "";
```
shschaefer: Absolute path containment is case-sensitive on Windows.  D:\mxc\folder is the same as d:\mxc\folder.

scripts/versioning/lib/git-base.js
```diff
@@ -0,0 +1,163 @@
...
+  if (env.GITHUB_ACTIONS) {
+    throw new Error(
+      "MXC_VERSIONING_BASE_REF is required in GitHub Actions; refusing to skip history checks"
+    );
+  }
+  return null;
+}
+
+function git(repoRoot, args, { trim = true } = {}) {
+  const output = execFileSync("git", args, {
```
shschaefer, line 29: execFileSync executes a command within a shell with a limited buffer ~ 1MB for execution output.   If git succeeds and returns to large of an output, this will get ENOBUFS and fail.

scripts/versioning/lib/version.js
bbonaby: should be resolved now

scripts/versioning/lib/git-base.js
bbonaby: I used infinitiy before thinking about it more I just changed it to 100MB instead that should be enough for our repository.

scripts/versioning/lib/git-base.js
bbonaby: we check for a string now and the place where we do call this function we do end up doing a path exists check now

scripts/versioning/lib/git-base.js
bbonaby: should be fixed now as well

shschaefer (approved):

PR #739, opened 2026-08-03
Update Learning Mode to V2 API contracts

richiemsft:
## 📖 Description

Updates `captureDenials` to the official V2 Windows Learning Mode and process
security-environment contracts.

- Uses the V2 Learning Mode ABI:
  - `HRESULT StartLearningModeTrace`
  - retryable, non-consuming `HRESULT StopLearningModeTrace`
  - infallible `CloseLearningModeTrace`
- Applies three bounded delivery attempts for transient Stop failures
  (sharing/lock contention, busy, or retry requested), while permanent errors
  remain single-attempt and the final HRESULT is preserved.
- Uses the V2 process security-environment ABI:
  - `HRESULT CreateProcessSecurityEnvironment`
  - `QueryProcessSecurityEnvironmentSupport`
  - infallible, by-value `CloseProcessSecurityEnvironment`
- Adds generated bindings for the official PSEC 1.0 FlatBuffer contract.
- Preserves the supported `CreateProcessW` +
  `PROC_THREAD_ATTRIBUTE_SECURITY_ENVIRONMENT` launch path; no legacy
  `CreateProcessAsUserInsideSecurityEnvironment` dependency remains.
- Routes BaseContainer by request and host capability: schema 0.8+ prefers PSEC,
  while ordinary requests retain SBOX/AppContainer fallback when PSEC is
  unavailable or cannot represent the requested policy. `captureDenials`
  remains PSEC-only and fails closed.
- Uses `IsApiSetImplemented("api-win-appmodel-processmodel~securityenvironment")`
  before resolving the official security-environment exports.
- Encodes explicit PSEC egress defaults and routes unsupported ordinary
  `leastPrivilege`, proxy, and deny-policy combinations through compatible
  lower tiers; capture and host-list combinations still fail closed.
- Requires the complete official V2 export set, so older V1 hosts return
  `backend_unavailable` instead of invoking an incompatible ABI.
- Memoizes successful and failed export resolution plus the process-wide PSEC
  deny-support query.
- Adds host-independent lifecycle, export-completeness, V1-degradation, and
  PSEC fail-closed tests.
- Records PSEC schema provenance (validated Windows build, schema SHA-256, and
  exact `flatc` release assets/checksums) and adds a CI gate that regenerates
  and byte-compares the bindings with the pinned compiler.
- Retains the existing JSON denial document and structured output-metadata
  contract introduced by #710.

## Proxy compatibility fallback

Windows OS PR 16153076 changed the capability-aware SBOX proxy contract. Proxy
requests now require an AppContainer-hosted proxy named through
`allowed_appcontainer_peer`, require `privateNetworkClientServer`, and reject
`internetClient`. MXC does not yet author that model-2 peer identity.

Until that integration is implemented, hosts exposing
`Experimental_QuerySandboxSupport` route proxy requests to the AppContainer
compatibility path instead of launching an invalid SBOX request. Query-less
hosts retain the legacy SBOX proxy behavior. This is intentionally a temporary
compatibility fallback; model-2 SBOX support should be implemented separately.

VM validation on build `26663.1000` confirmed minimal built-in WinHTTP proxy
E2Es pass for schema `0.7.0-alpha` and `0.8.0-alpha`. The full Node integration
policy additionally grants SDK tool paths such as `C:\Debuggers`; when the
fallback reaches AppContainer+DACL, the test account must have `WRITE_DAC` on
those paths. That host permission is not required once MXC supports model-2 T1.
The separate OS-side ETW provider-readiness improvement is intentionally out of
scope for this ABI update.

## 🔗 References

- Learning Mode OS contract update: [OS PR 16307987](https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/16307987)
- SBOX model-2 proxy contract: [OS PR 16153076](https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/16153076)
- SBOX model-2 proxy tests: [OS PR 16153125](https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/16153125)
- Proxy Velocity-key move: [OS PR 16213867](https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/16213867)
- Related Issues:
  - #661
  - #663
- Related PRs:
  - #696
  - #709
  - #710

## 🔍 Validation

- `cargo test -p learning_mode_core -p learning_mode_windows -p appcontainer_common --all-targets`
- `cargo test -p wxc --no-run`
- `cargo clippy -p learning_mode_core -p learning_mode_windows -p appcontainer_common --all-targets -- -D warnings`
- `cargo check --workspace --all-targets`
- `node scripts/versioning/check-schema-codegen.js`
- `node scripts/versioning/check-sdk-types-codegen.js`
- `node scripts/versioning/check-psec-codegen.js`
- `node scripts/versioning/validate-configs.js`
- Independent adversarial, performance, reliability, leak, API-contract,
  lifecycle/concurrency, data-validation, and error-handling reviews found no
  blocking MXC issues.
- V1 host (`26657.1002`):
  - V2 probes reject the incomplete export set.
  - `captureDenials` returns `FailurePhase::BackendUnavailable`.
  - no output document is created; schema 0.7 ProcessContainer execution remains
    functional through the legacy SBOX path.
- V2 host (`26663.1000`):
  - all official V2 exports resolve.
  - standalone Start → launch → Stop → Close produced a 3,302-byte ETL.
  - block and allow captures produced JSON documents and structured pointer
    metadata.
  - schema 0.8 ordinary and capture ProcessContainer requests use PSEC.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

with write access can trigger it on a PR with `/azp run`. See
[docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md).

If the `dependency-feed-check` check fails on a new dependency, the crate must
be added to the feed before the PR can pass. See
[docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md)
for the steps.


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1600,21 +1670,71 @@ struct BaseChild {
...
+                    ))
+                })?
+            } else {
+                crate::fallback_detector::base_container_supports_deny_paths()
+            };
+            if !deny_supported {
+                return Err(ScriptResponse::error(
+                    wxc_common::error::DENIED_PATHS_FEATURE_DISABLED_MSG,
+                ));
+            }
```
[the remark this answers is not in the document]
richiemsft: Fixed in b21163a8. Capture denied-path failures now name `QueryProcessSecurityEnvironmentSupport` and `PSE_SUPPORT_FS_DENY`, explicitly state that capture cannot use AppContainer/host-DACL fallback, and surface `FailurePhase::BackendUnavailable`. The legacy SBOX message remains only for ordinary BaseContainer.

src/backends/learning_mode/windows/examples/lm_probe.rs
```diff
@@ -4,12 +4,11 @@
...
-//! exports (`StartLearningModeTrace` / `StopLearningModeTrace`) and the 2-phase
-//! security-environment exports (`CreateProcessSecurityEnvironment` /
-//! `CloseProcessSecurityEnvironment`),
-//! reporting the exact resolved name for each (plain vs `Experimental_`). Intended to
-//! be run on a feature-enabled Windows build to confirm the runtime FFI resolves
-//! against the real API.
+//! exports (`StartLearningModeTrace` / `StopLearningModeTrace` /
+//! `CloseLearningModeTrace`) and the 2-phase security-environment exports
+//! (`CreateProcessSecurityEnvironment` / `CloseProcessSecurityEnvironment`),
+//! reporting each official export that resolves. Intended to be run on a
```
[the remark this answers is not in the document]
richiemsft: Fixed in b21163a8. The `lm_probe` documentation now lists the complete required security-environment surface: Create, QueryProcessSecurityEnvironmentSupport, and Close.

src/backends/learning_mode/windows/src/lifecycle.rs
```diff
@@ -110,97 +101,33 @@ impl CaptureSession {
...
             None => Ok(()),
         };
-        let close_result = match self.environment.as_mut() {
-            Some(environment) => self.secenv_api.close(environment),
-            None => Ok(()),
-        };
-        if close_result.is_ok() {
-            self.environment.take();
+        if let Some(trace) = self.trace.take() {
+            trace.close();
```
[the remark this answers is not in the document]
richiemsft: Fixed in dfca3cc4. `CaptureSession::finish` now uses the non-consuming V2 Stop contract for three bounded attempts on transient delivery failures (`ERROR_SHARING_VIOLATION`, `ERROR_LOCK_VIOLATION`, `ERROR_BUSY`, and `ERROR_RETRY`) with 25/75 ms backoff. Permanent failures are not retried, exhausted retries preserve the final HRESULT, and trace/environment close still occur exactly once after the retry policy completes. Added tests for transient success, permanent single-attempt failure, and exhausted retries.

src/backends/learning_mode/windows/src/lifecycle.rs
```diff
@@ -110,97 +102,35 @@ impl CaptureSession {
...
+    /// Stop the trace and deliver it to `output_path` (or skip delivery when
+    /// `None`), retry transient delivery failures, close the trace, then close
+    /// the security environment. Call **after** the child has exited.
     ///
     /// # Errors
-    /// - [`LearningModeError::ApiCall`] from `StopLearningModeTrace` or
-    ///   `CloseProcessSecurityEnvironment`.
-    /// - [`LearningModeError::CleanupFailed`] if both teardown calls fail.
+    /// - [`LearningModeError::HResultCall`] from `StopLearningModeTrace`.
     pub fn finish(mut self, output_path: Option<&Path>) -> Result<(), LearningModeError> {
```
MGudgin, line 111: **High (reliability + correctness) — this discards the retryable, non-consuming Stop that `ffi.rs` implements correctly.**

`finish(mut self, ...)` calls `stop_trace` exactly once, then unconditionally closes the trace (line 115-117) and the environment (118-120) before returning `stop_result`. Because it takes `self` by value, **no caller can retry** — every HRESULT is effectively terminal.

That's a shame, because the layer below gets this exactly right: `StopLearningModeTrace` is declared non-consuming (`ffi.rs:42-43`, handle by value, `HRESULT` return), and `ffi.rs` has two tests proving the handle survives a failure — `failed_hresult_keeps_trace_live_until_close` (`ffi.rs:400`) and `failed_start_preserves_hresult_and_does_not_close` (`ffi.rs:421`). The capability is implemented and tested one layer down, then thrown away here.

Concretely: a transient ETL-delivery failure fails an **already-completed** sandbox execution and permanently loses the denial document, with no second chance.

**Fix:** keep the session live across a bounded retry loop with backoff, retrying only documented transient HRESULTs, and close only after success or exhaustion. A `finish(&mut self)` (or an explicit `retry_stop`) would make the non-consuming contract expressible. Test with a fake that fails Stop once then succeeds.

src/backends/learning_mode/windows/src/lifecycle.rs
```diff
@@ -110,97 +102,35 @@ impl CaptureSession {
...
-            cleanup: Box::new(cleanup),
-        }),
+        if let Some(environment) = self.environment.take() {
+            environment.close();
+        }
+        stop_result
     }
 }
 
 impl Drop for CaptureSession {
```
MGudgin, line 128: **High (test coverage + testability) — this file went from 2 unit tests to 0 while gaining the V2 state machine.**

I verified this against base rather than inferring it:

- `git show 80cd261c:...lifecycle.rs` → **2** `#[test]` functions, 188 lines
- HEAD → **0** `#[test]` functions, 122 lines

So this isn't "new code without new tests" — existing coverage was removed from the module that now owns Create→Start→Stop→Close ordering, teardown-on-failure, and the Drop-without-Stop discard path. Across all the changed Learning Mode + AppContainer files the count moves 66 → 70 for ~1,579 added hand-written lines.

The root cause is structural: `LearningModeApi` and `SecurityEnvironmentApi` hold private function pointers with no injection seam, so `ffi.rs`/`secenv.rs` can fake them for their *own* tests but `lifecycle.rs` cannot fake anything. That is also why validation had to fall back to manual runs on two OS builds — nothing here is assertable in CI.

**Fix:** add a `pub(crate)` constructor taking the function pointers (or extract a small trait, as `CaptureSessionOps` already does in the runner), then restore and extend the tests: create→start ordering, start-failure cleanup, launch-failure discard, `finish` stop→close ordering, and `Drop` without `Stop`. That one seam also unlocks the export-set and fail-closed tests noted on `ffi.rs` and `base_container_runner.rs`.

src/backends/learning_mode/windows/src/ffi.rs
```diff
@@ -89,11 +140,13 @@ impl LearningModeApi {
```
MGudgin, line 154: **High (test coverage) — the V1-host degradation path is only exercised against whatever DLL the build machine happens to have.**

The loader tests call the real `processmodel.dll` and accept **either** outcome (`available == LearningModeApi::load().is_ok()`, `ffi.rs:331`), which by construction cannot fail. Nothing simulates a missing individual export or a partial V1 export set, so the `BackendUnavailable` contract at `base_container_runner.rs:1729-1745` — the centrepiece of the "older V1 hosts degrade safely" claim — is verified only by the manual run on `26657.1002` recorded in the PR description.

Manual host validation confirms today's behavior but locks in nothing: a later refactor could drop an export from the required set and re-enable an incompatible ABI call on old hosts with CI fully green.

**Fix:** inject the resolver (see the seam suggested on `lifecycle.rs:128`) and table-test the export sets — all present, each one individually missing, and a representative V1 subset — asserting `BackendUnavailable` and that no V2 entry point is invoked.

---

Separately, credit where due: I checked all six `extern` declarations against the V2 contract and they're correct, including the two easy ones to get wrong — `StopLearningModeTrace` taking the handle by value with an `HRESULT` return (`:42-43`), and `CloseLearningModeTrace` returning nothing (`:46`).

src/backends/learning_mode/windows/src/ffi.rs
```diff
@@ -3,77 +3,128 @@
...
         f.debug_struct("LearningModeApi")
             .field("start", &(self.start as *const ()))
             .field("stop", &(self.stop as *const ()))
+            .field("close", &(self.close as *const ()))
             .finish()
     }
 }
 
 impl LearningModeApi {
+    const STOP_DELIVERY_ATTEMPTS: usize = 3;
```
MGudgin, line 119: **Medium (performance) — export resolution is repeated per call, and failures are never cached.**

`LearningModeApi::load()` (and `SecurityEnvironmentApi::load()` in `secenv.rs:334`) does a fresh `LoadLibraryExW` + six `GetProcAddress` lookups every time it is called, and it is called more than once per launch — at least in `validate()` (`base_container_runner.rs:1698`) and again on the launch path (`:1730`).

The worse half is the negative case: on a V1 host the resolution **fails** every time and the failure isn't remembered, so each launch repeats the full load-and-fail sequence to reach the same `backend_unavailable`.

Worth noting this design is *safe* — a sibling concurrency pass specifically confirmed that building an immutable API value locally publishes no partially-initialised shared state. A `OnceLock<Result<Api, _>>` preserves that property exactly while paying the cost once per process.

**Fix:** memoise both loaders (and the `supports_deny_paths()` result) in `OnceLock`s, caching the failure as well as the success.

src/backends/learning_mode/windows/src/secenv.rs
```diff
@@ -306,32 +299,27 @@ impl SecurityEnvironmentExportReport {
...
-/// `Experimental_`-prefixed name kept as a fallback for older feature builds.
-const CREATE_NAMES: &[&core::ffi::CStr] = &[
-    c"CreateProcessSecurityEnvironment",
-    c"Experimental_CreateProcessSecurityEnvironment",
-];
-const CLOSE_NAMES: &[&core::ffi::CStr] = &[
-    c"CloseProcessSecurityEnvironment",
-    c"Experimental_CloseProcessSecurityEnvironment",
-];
+const CREATE_NAMES: &[&core::ffi::CStr] = &[c"CreateProcessSecurityEnvironment"];
```
MGudgin, line 311: **Medium (testability) — `is_complete()` is a pure function and still isn't unit-tested.**

The export-completeness decision is exactly the kind of logic that needs no OS at all: given a `SecurityEnvironmentExportReport`, is the V2 set complete? But the only test covering it (`probe_does_not_panic_and_agrees_with_load`, `:530`) runs against the live host and asserts agreement with `load()` — which means on any given machine it exercises one arm and tautologically agrees.

Nothing asserts the interesting combinations: `Create` present but `Close` missing, `Query` present but `Create` missing, and so on. These are the exact shapes a real V1 host presents.

**Fix:** construct `SecurityEnvironmentExportReport` directly with each partial combination and assert `is_complete()` is `false`, plus the all-present case being `true`. No V2 host, no DLL, runs on every platform — this is a handful of lines for a meaningful guarantee.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1600,21 +1675,76 @@ struct BaseChild {
...
 impl SandboxBackend for BaseContainerRunner {
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
-        // deniedPaths reaches the OS via the SandboxSpec `fs_deny` field, honored
-        // only when the OS advertises SANDBOX_CAP_FS_DENY. The dispatcher only
-        // routes deny here when supported; fail closed for direct callers.
-        if !request.policy.denied_paths.is_empty()
-            && !crate::fallback_detector::base_container_supports_deny_paths()
-        {
+        let capture_denials = request.policy.capture_denials.is_some();
+        if capture_denials && request.policy.least_privilege_mode {
```
MGudgin: **Medium (documentation-drift / backward-compatibility) — these new rejections never reached the generated schema.**

I confirmed against base that both of these are genuinely new: there is no `capture_denials && least_privilege_mode` or `capture_denials && network_proxy` rejection at `80cd261c` — base only threads `least_privilege` through to the spec (`:655`). So this is a real, newly-observable config-validation break for anyone who previously combined these fields.

`docs/schema.md` documents it (thank you), but `src/core/wxc_common/src/wire.rs` is **unchanged** — and `wire.rs` is the generator input for `schemas/dev/`. The result is that `schemas/dev/mxc-config.schema.0.8.0-dev.json` still describes the old contract, so editor validation and IDE hovers won't warn anyone authoring a config, and `--experimental` users discover the constraint by hitting it.

**Fix:** add the incompatibility notes to the `capture_denials` doc comment in `wire.rs` and regenerate the dev schema (`cargo run -p mxc_schema_gen -- schemas/dev/mxc-config.schema.<dev>.json`). Worth a release-note line too, since it's an observable break rather than an addition.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1600,21 +1675,76 @@ struct BaseChild {
...
+                "processContainer.captureDenials cannot be combined with network.proxy \
+                 until the V2 process-security-environment path can supply the required \
+                 proxy AppContainer peer identity",
             ));
         }
+        // deniedPaths reaches ordinary BaseContainer through SBOX and capture
+        // through PSEC. Each path has a distinct support query; fail closed
+        // rather than silently dropping the deny policy.
+        if !request.policy.denied_paths.is_empty() {
+            let deny_supported = if capture_denials {
```
MGudgin: **High (test coverage) — the PSEC fail-closed branch has no host-independent test, and the existing test drives the *other* probe.**

This block now forks: capture mode loads `SecurityEnvironmentApi` and queries PSEC deny support (`:1698-1709`), while ordinary mode uses `fallback_detector::base_container_supports_deny_paths()` (`:1711`). The existing unsupported-deny test uses `DenyPathsGuard`, which controls the **legacy** probe — so it exercises the `else` arm and tells you nothing about the capture arm.

That leaves three new failure paths untested: API unavailable (`:1698`), query failure (`:1704`), and support-bit clear (`:1713`). All three must produce `BackendUnavailable` and must **not** launch. Given this is the fail-closed guarantee the PR description leads with, it deserves a test that runs everywhere rather than only on a V2 host.

**Fix:** inject the PSEC capability provider (same seam as `lifecycle.rs:128`) and assert each of the three outcomes returns `BackendUnavailable` with no launch attempted.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1600,21 +1675,76 @@ struct BaseChild {
...
+                let api = SecurityEnvironmentApi::load().map_err(|error| ScriptResponse {
+                    failure_phase: FailurePhase::BackendUnavailable,
+                    ..ScriptResponse::error(&format!(
+                        "process security-environment API unavailable: {error}"
+                    ))
+                })?;
+                api.supports_deny_paths().map_err(|error| ScriptResponse {
```
MGudgin: **Medium (performance) — a blocking ALPC round-trip on the cold-start path, repeated every launch.**

`api.supports_deny_paths()` is a synchronous `QueryProcessSecurityEnvironmentSupport` call into the AppInfo broker, executed during `validate()` whenever `captureDenials` is set and `denied_paths` is non-empty. Combined with the unmemoised `SecurityEnvironmentApi::load()` immediately above (`:1698`), each launch pays a DLL load, six `GetProcAddress` lookups, and a cross-process query before anything runs.

Sandbox launch is a cold path, but MXC is used for interactive agent workloads where start latency is user-visible, and the support bit cannot change within a process lifetime.

**Fix:** cache the boolean in a `OnceLock` alongside the loader memoisation. Scoped to the feature — I confirmed the feature-off path pays none of this, which is exactly right.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -666,6 +625,84 @@ impl BaseContainerRunner {
         builder.finished_data().to_vec()
     }
 
+    fn effective_capabilities(request: &ExecutionRequest) -> Vec<String> {
```
MGudgin: **Low (performance) — avoidable clone of the capability vector.**

`effective_capabilities` clones the whole `Vec<String>` (`:632`) purely to conditionally append `"internetClient"`, and the only consumer immediately `.join(",")`s it into a FlatBuffer string (`:656`).

Genuinely minor on a once-per-launch path — flagging it only because it's a two-line change in new code.

**Fix:** build the joined string directly from an iterator chain, e.g. `request.policy.capabilities.iter().map(String::as_str).chain(needs_internet.then_some("internetClient")).collect::<Vec<_>>().join(",")`, dropping the intermediate `Vec<String>` entirely.

docs/learning-mode/capabilities.md
```diff
@@ -121,6 +121,16 @@ ungranted access is handled while it is recorded:
 > build exposing the BaseContainer security-environment and Learning Mode APIs.
 > It is not supported by the AppContainer fallback tiers; unsupported hosts
 > return `backend_unavailable`.
+>
```
MGudgin, line 132: **Medium (documentation-drift) — the host requirement is stated but never made actionable, and it's now V2-specific.**

To be fair on attribution: base already carried a host-requirement note (lines 120-123: "requires a feature-enabled Windows build exposing the BaseContainer security-environment and Learning Mode APIs… return `backend_unavailable`"), so the *absence* of a floor isn't new. What's new is that the requirement changed — it is now the **complete official V2 export set**, and the PR body records concrete validation points (V1 `26657.1002` fails, V2 `26663.1000` works) that appear nowhere in the docs.

A user on `26657` gets `backend_unavailable` and has no documented way to learn whether their build is too old or their config is wrong.

Two more gaps in this section:
- Only `leastPrivilege` and `network.proxy` are listed as incompatible. The new path also rejects `deniedPaths` when the PSEC deny capability is absent (`base_container_runner.rs:1697-1722`).
- `docs/process-container/os-version-support.md` is a per-Windows-release support matrix and is unchanged — it has no Learning Mode row at all, which is where someone checking OS support would actually look.

**Fix:** record the validated build numbers and the complete-V2-export requirement here and in `os-version-support.md`, and complete the incompatibility list.

src/Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
     "core/mxc_build_common",
     "host/plm",
     "core/generated/base_container_specification",
+    "core/generated/process_security_environment_specification",
```
MGudgin, line 14: **Medium (cross-platform-parity) — a Windows-only generated crate is compiled on every target.**

`process_security_environment_spec` is added as an unconditional workspace member, so its ~2,256 lines of generated FlatBuffers bindings compile on Linux and macOS for a feature that only exists on Windows. A parity pass verified this directly with `cargo check -p process_security_environment_spec --target x86_64-unknown-linux-gnu`, and the whole-workspace Linux check compiles it too.

To be clear about the blast radius: this adds **no** new transitive dependency — `flatbuffers` already backs `sandbox_spec` — so it is pure build cost, not a supply-chain change. It's a judgement call whether that's worth acting on; I'd lean yes because the crate is placed under `core/` (the documented "cross-platform foundation" tier) while encoding a Windows OS contract, which makes the placement itself a bit of a signpost problem.

Also worth noting the stated validation was scoped to `-p learning_mode_core -p learning_mode_windows -p appcontainer_common`; no whole-workspace or cross-target build is recorded, despite this PR adding a workspace member.

**Fix:** either move the generated bindings into the Windows-only Learning Mode crate, or target-gate the member and its `flatbuffers` dependency to `cfg(windows)`.

external/windows-sdk/ProcessSecurityEnvironment.fbs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
MGudgin: **Medium (supply-chain) — a vendored two-sided OS contract with no recorded provenance.**

This schema is a copy of an OS-side contract (the PR links ADO OS PR 16307987), and MXC's encoder must stay byte-compatible with the OS decoder. But the file records no source revision, SDK build number, or hash, and the crate README only names the file and a minimum `flatc` version.

That means nobody can later answer "is our copy still current?" — and a silent divergence between this schema and the OS's isn't a build break, it's a **wrong sandbox network-egress policy** encoded at runtime.

The repo already has the pattern worth copying here: `docs/wsl/wslc-sdk-bindings.md` documents the WSLC SDK bindings with a `WSLC_SDK_VERSION` + hash that must be bumped together, plus a regeneration runbook and an explicit do-not-hand-edit rule.

**Fix:** record the upstream OS revision/build and a SHA-256 of the `.fbs` next to it, and validate the hash in CI or in `regenerate.ps1`.

src/core/generated/process_security_environment_specification/regenerate.ps1
```diff
@@ -0,0 +1,65 @@
...
+$fbs = "external\windows-sdk\ProcessSecurityEnvironment.fbs"
+
+if (-not (Test-Path $fbs)) {
+    throw "FlatBuffers schema not found: $fbs"
+}
+if (-not (Test-Path $Flatc) -and -not (Get-Command $Flatc -ErrorAction SilentlyContinue)) {
+    throw "flatc not found: $Flatc"
+}
+
+$minFlatcVersion = [version]'25.12.19'
```
MGudgin: **Medium (supply-chain / maintainability) — regeneration isn't reproducible and isn't drift-gated.**

The script accepts **any** `flatc` newer than `25.12.19`, while the Rust runtime is pinned separately as `flatbuffers = "25"` (`src/Cargo.toml:131`). Two engineers on different `flatc` builds can therefore produce different committed bindings, or bindings that don't match the runtime — and the next person to run this gets a large unrelated generated diff mixed into their change.

There's also no CI check that regenerates and compares. The repo has this exact pattern twice already — `scripts/versioning/check-schema-codegen.js` and `scripts/check-dotnet-bindings-codegen.js` both regenerate and fail on drift — so the precedent and the tooling shape already exist.

**Fix:** pin (and ideally checksum) the exact `flatc` version used to produce the committed output, record the schema hash, and add a regeneration-diff gate to the versioning checks job. A do-not-hand-edit header on the generated files would round it out.

MGudgin (dismissed):
## Adversarial review — 13 axes

Ran a multi-axis adversarial review (security, reliability, performance, maintainability, correctness, testability, test coverage, documentation-drift, cross-platform-parity, concurrency, backward-compatibility, supply-chain, proportionality), each axis in an isolated context on a non-Claude model, then consolidated.

**I then re-verified every finding that cited code outside the diff, to make sure this PR is not being blamed for pre-existing issues.** That pass moved four findings out of scope and made one worse. Details below — I'd rather hand you a shorter, accurate list than a longer one.

### What the review found clean, with receipts

The hardest parts of an ABI migration are right, and that's worth saying plainly:

- **ABI fidelity** — all six V2 exports match the documented contract, including the by-value `CloseProcessSecurityEnvironment` (`secenv.rs:78-79`) and the `void`-returning infallible `CloseLearningModeTrace` (`ffi.rs:46`). The pointer-vs-by-value trap is handled correctly.
- **HRESULT handling** — everything goes through `HRESULT::is_err()` (`self.0 < 0`), so `S_FALSE` is correctly treated as success — and there's an explicit regression test for it (`successful_hresult_starts_retryable_trace`, `ffi.rs:381`).
- **Security** — System32-only DLL loading, exact+complete export matching, unpredictable 128-bit denial filenames with create-new semantics, and job-object/suspended-launch enforcement retained through the legacy removal.
- **Concurrency** — exact-once close, correct trace-before-environment drop order on success/error/unwind, handles nulled before `Drop`, and the Ctrl-C handler shares no trace/PSEC state.
- **Proportionality** — the hand-written churn maps to an externally forced ABI change; no scope creep found.
- **Performance of the feature-off path** — executions without `captureDenials` pay **zero**; all new DLL loading, OS queries and PSEC construction are cleanly gated.

### The one thing I'd fix first

`CaptureSession::finish` (`lifecycle.rs:111`) calls `StopLearningModeTrace` once, treats every HRESULT as terminal, and then closes the trace and environment unconditionally. Your `ffi.rs` layer implements the non-consuming/retryable contract **correctly** — `failed_hresult_keeps_trace_live_until_close` proves it — and then `finish` takes `self` by value and throws that capability away. A transient delivery failure fails an already-completed sandbox execution and permanently loses the denial document.

### The verification pass — findings I am NOT attributing to this PR

These were raised by agents but I confirmed against `git show 80cd261c:<path>` that they are pre-existing. Flagging them only as context / possible follow-ups, **not** as things to fix here:

1. **Unchecked `ResumeThread`** (`base_container_runner.rs:1614`). Reported as a new hang risk. It is **byte-identical** between base and HEAD, and there is exactly one `CreateProcessW` call site in both revisions — so there is no new capture-specific launch. Pre-existing; the contrast with the AppContainer path (which does terminate and report) is a pre-existing inconsistency worth a separate issue.
2. **`LearningModeError` flattened into `io::Error::other`** (`combine_capture_and_cleanup_results`, HEAD `:2014`). Reported as a High testability blocker. The function is **unchanged** by this PR (base `:1896`); it arrived with #710. Your new `LearningModeError` variants do flow through it and get stringified, so the diagnosability ceiling is real — but it isn't yours to fix in this PR.
3. **The capture E2E is `#[ignore]` and asserts vacuously** (`e2e_windows.rs:284`). `e2e_windows.rs` is **not modified** by this PR and the test is identical at base — same `cmd.exe /c echo` workload producing an empty `denials` array. Pre-existing. The fair criticism is only that it wasn't *extended* to cover V2.
4. **`--probe` disagrees with the dispatcher.** Reported as a High. The dispatcher already had a `capture_denials` override at base (`dispatcher.rs:342`), and `probe.rs` is unchanged, so the tier-misprediction divergence pre-dates this PR. **However** — see below, there is a genuinely new part.

### Out-of-diff findings that ARE attributable

**Medium — `--probe` doesn't reflect the *new* V2 export requirement.** `probe.rs` is unchanged and still calls `fallback_detector::detect` directly (`probe.rs:131`), with no reference to `capture_denials` or Learning Mode anywhere in the file. The tier divergence is pre-existing, but this PR **changes the availability threshold** to the complete V2 export set, so `--probe` now reports success on a V1 host for a request that will fail `backend_unavailable` at `validate()`. Worth teaching `run_probe` about the V2 check and exposing it in `ProbeFacts`.

**Medium — `wire.rs` / generated dev schema are now stale.** `src/core/wxc_common/src/wire.rs` is unchanged, and I confirmed the `captureDenials + leastPrivilege` and `captureDenials + network.proxy` rejections (`base_container_runner.rs:1679`, `:1686`) have **no equivalent at base** — they are new here. Since `wire.rs` drives `schemas/dev/`, the generated schema and IDE hovers still describe the old contract. `docs/schema.md` got the note; its generator input didn't.

**Medium — no Learning Mode row in `docs/process-container/os-version-support.md`.** That file is unchanged. Given this PR establishes a hard OS build floor (validated V1 `26657.1002` fails, V2 `26663.1000` works), the per-release support matrix is where a user would look.

---

### What I'd want before approving

1. **Retry Stop** rather than discarding it (`lifecycle.rs:111`) — bounded, transient-HRESULT-only.
2. **Restore lifecycle test coverage** — see the inline comment; `lifecycle.rs` went from 2 tests to 0 while gaining the V2 state machine.
3. **Add an FFI seam** so export-set gating and the PSEC fail-closed paths can be tested without a V2 host.
4. **Update `wire.rs`** and regenerate `schemas/dev/`, and record the V2 build floor in `os-version-support.md`.
5. **Record `.fbs` provenance and pin `flatc`** — the repo already has the pattern in the WSLC bindings runbook.

The `OnceLock` memoisation, the workspace-member gating, and the `spawn_base` extraction all read as follow-ups. Nothing here suggests the migration is wrong — the implementation is in good shape; it's the verification that's thin relative to the blast radius. Happy to be pushed back on any of it, particularly the parity/build-cost point which is a judgement call.

src/backends/learning_mode/windows/src/lifecycle.rs
richiemsft: Fixed across dfca3cc4 and 5f4237be. `finish` uses three bounded attempts for transient Stop delivery HRESULTs before exactly-once trace/environment close. Lifecycle tests now assert retry ordering, permanent failure teardown, and final error propagation.

src/backends/learning_mode/windows/src/lifecycle.rs
richiemsft: Fixed in 5f4237be. Added test-only raw-function constructors and host-independent lifecycle tests for create→start ordering, start-failure environment cleanup, transient Stop retry→trace close→environment close, permanent Stop failure teardown, and Drop-without-Stop discard.

src/backends/learning_mode/windows/src/ffi.rs
richiemsft: Fixed in 5f4237be. Added a pure LearningMode export report and tests for all-present, each individually missing, and the representative V1 Start+Stop subset. Runner validation also has an injectable platform-support seam proving missing V2 APIs return BackendUnavailable before capture begin/launch.

src/backends/learning_mode/windows/src/ffi.rs
richiemsft: Fixed in 5f4237be. Both API loaders now memoize success and typed failure through `OnceLock<Result<...>>`; `LearningModeError` is clonable so diagnostics are preserved. The PSEC deny-support result is memoized separately.

src/backends/learning_mode/windows/src/secenv.rs
richiemsft: Fixed in 5f4237be. Added pure `SecurityEnvironmentExportReport::is_complete` tests for the complete set and every missing-export combination, independent of the host DLL.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 5f4237be. Updated the `wire.rs` captureDenials documentation and regenerated both `schemas/dev/mxc-config.schema.0.8.0-dev.json` and `sdk/node/src/generated/wire.ts`. The incompatibilities are also documented in the user-facing Learning Mode and OS-support docs.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 5f4237be. Added injectable V2 platform-support checks and host-independent tests for API unavailable, support-query failure, and deny bit clear. All assert BackendUnavailable and verify the capture factory was never invoked.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 5f4237be. `QueryProcessSecurityEnvironmentSupport` is now memoized once per process (including typed errors); non-cacheable test fakes bypass the global cache to avoid test poisoning.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 5f4237be for the new PSEC path. Capability serialization now joins the borrowed capability slice directly and appends `internetClient` only when needed, avoiding the intermediate `Vec<String>` clone. The legacy SBOX builder is unchanged.

docs/learning-mode/capabilities.md
richiemsft: Fixed in 5f4237be. The Learning Mode guide and OS-version support matrix now list the complete V2 export set, validated build points, runtime-probe guidance, leastPrivilege/proxy incompatibilities, and the PSE_SUPPORT_FS_DENY requirement/no-fallback behavior.

src/Cargo.toml
richiemsft: Acknowledged. This is a pure build-cost optimization, not a correctness or dependency issue. We intentionally deferred workspace exclusion/moving the generated crate to a separate change; it remains a member matching the existing generated SBOX crate. `cargo check --workspace --all-targets` passes and is now recorded in validation.

external/windows-sdk/ProcessSecurityEnvironment.fbs
richiemsft: Fixed in 5f4237be. Added `ProcessSecurityEnvironment.provenance.toml` recording the validated Windows build, LF-normalized schema SHA-256, related trace ABI PR (explicitly not claimed as schema origin), and exact checksummed flatc release assets. Regeneration and CI verify the schema hash.

src/core/generated/process_security_environment_specification/regenerate.ps1
richiemsft: Fixed in 5f4237be. Regeneration now requires exact flatc 25.12.19 and validates the schema hash. The versioning workflow runs a new gate that downloads the pinned platform asset, verifies its SHA-256, performs a real temp regenerate + rustfmt + byte diff across all 12 generated files, checks generated headers, and compiles the crate.

scripts/versioning/check-psec-codegen.js
```diff
@@ -0,0 +1,367 @@
...
+
+const sha256Hex = (buf) => crypto.createHash("sha256").update(buf).digest("hex");
+const sha256File = (p) => sha256Hex(fs.readFileSync(p));
+// LF-normalize so autocrlf checkouts don't produce false differences.
+const lfNormalize = (buf) =>
+  Buffer.from(buf.toString("binary").replace(/\r\n/g, "\n"), "binary");
+
+// --- Minimal TOML reader (targeted, no dependency) --------------------------
+function tomlSection(text, header) {
+  const escaped = header.replace(/[.[\]]/g, "\\$&");
```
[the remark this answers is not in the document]
richiemsft: Fixed in c9934c8a. Removed dynamic regular-expression construction entirely; the minimal provenance parser now identifies TOML sections and keys with line-based comparisons and uses only a fixed regex for a quoted value.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -666,6 +672,98 @@ impl BaseContainerRunner {
...
+                ..Default::default()
+            },
+        );
+        let network_policy = Some(PsecNetworkPolicy::create(
+            &mut builder,
+            &PsecNetworkPolicyArgs {
+                egress: Some(egress),
+                ..Default::default()
+            },
+        ));
```
bbonaby: issue: you'll need to sync with main. I have a function now that should be used for network policy [here](https://github.com/microsoft/mxc/blob/c47ab4f1db7215ca7305ba32a0996ea243806b76/src/backends/appcontainer/common/src/base_container_runner.rs#L539).  When you sync , be sure to run `https://github.com/microsoft/mxc/blob/main/tests/scripts/run_base_container_network_tests.ps1` on a build of directwinpd, one since the 31st. SHould be able to run it locally on your dev machine. Just want to make sure we don't break GHCP again.

external/windows-sdk/ProcessSecurityEnvironment.fbs
bbonaby: tbh this one might be overkill. The schema will be copied to msdn for folks to use and one day maybe added to windows SDK. Another thing is the OS will reject a flatbuffer blob that does not adhere to the schema, so if this file is changed without an OS change then our tests would fail. That said this file should include a similar comment that the OS contains:

```rust
//
// FlatBuffers schema for the BaseContainer sandbox specification used in the CreateProcessSecurityEnvironment Windows API.
//
// This is the CPSE wire format. The legacy CPIS path continues to use
// SandboxSpec.fbs and its "SBOX" file identifier.
//
// ---- FlatBuffers Schema Evolution Rules ----
//
// Compatible (non-breaking) changes:
//   - Add new fields to the END of a table (they get the next vtable slot).
//   - Add new tables, structs, enums, or union members.
//   - Deprecate a field with (deprecated) - the slot is preserved, readers skip it.
//   - Rename a field or table (wire format uses slot indices, not names).
//
// BREAKING changes (will corrupt existing buffers):
//   - Remove or reorder fields in a table or struct.
//   - Change a field's type (e.g. uint32 -> int32, or scalar -> string).
//   - Change a field's default value.
//   - Add/remove/reorder values in an enum that is already serialized.
//   - Change the file_identifier or root_type.
//   - Change a field from scalar to non-scalar (or vice versa).
//
// To machine-check for breaking changes, keep a copy of the last shipped schema
// (e.g. ProcessSecurityEnvironment.previous.fbs) and run:
//
//   flatc --conform ProcessSecurityEnvironment.previous.fbs ProcessSecurityEnvironment.fbs
//
// --conform verifies that every field/enum/table in the old schema still exists
// at the same vtable slot, type, and default in the new schema.
```

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -916,15 +1027,21 @@ impl BaseContainerRunner {
...
+                Err(e) => return Err(ScriptResponse::error(&e)),
+            };
+            let _ = writeln!(
+                logger,
+                "loaded Experimental_CreateProcessInSandbox from processmodel.dll"
+            );
+            Some(api)
+        } else {
+            None
         };
```
bbonaby: issue: seems with this we base whether or not to use the new API based on if capture_denials was used in the  policy. I believe we should use schema version here. If schema version is 0.7.0 or below we  should be going down the old path. If it's 0.8.0 pr above  we should be going down the new path with the new APIs. This will allow us to not have to make a code change here for 0.8.0 in the future when we make it the stable schema.

I know gudge has plans for how he wants code to diverge based on versioning, however that isn't out yet so I think we should be able to do this versioning check here without issue. At least until Gudge's PR for that comes out, in which case He'll probably update it at that time.


EDIT: I looked at the current code and looks like we only do the second phase set up for the new API if there is a capture session available here: 

https://github.com/microsoft/mxc/blob/c9934c8acd5f884025bc67c2f327ce4351b3eca8/src/backends/appcontainer/common/src/base_container_runner.rs#L1392

Since this affect base container runner  and not just learning mode APIs I believe we should be going into this block when schema version is also 0.8.0 or higher. Not just if there is a capture session available.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -940,10 +1057,15 @@ impl BaseContainerRunner {
...
-        // Identity: when destroy_on_exit is true we generate a random ephemeral
+        let legacy_destroy_on_exit = capture_denials.is_none() && request.lifecycle.destroy_on_exit;
+
+        // Identity applies only to the SBOX one-shot API. PSEC creates and owns
+        // its own AppContainer identity and profile.
         // identity so each sandbox gets a unique, cleanable AppContainer profile.
         // Otherwise we honour whatever the caller passed in (or the default).
-        let (identity, sid_string) = if request.lifecycle.destroy_on_exit {
+        let (identity, sid_string) = if capture_denials.is_some() {
+            ("<process-security-environment>".to_string(), String::new())
```
bbonaby: issue: see comment above about using schema version number

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 580c19ef. I merged current `main` and now route both SBOX and PSEC serialization through the same resolved network-policy semantics. On DirectWinPD build 10.0.26663.1000, the provided script selected BaseContainer and logged the expected `allow`/`deny` egress actions, but its PowerShell workload exited with `STATUS_DLL_INIT_FAILED` even after all UI limits were enabled. I therefore repeated the enforcement check with direct `curl.exe`: allow exited 0, while deny blocked DNS with curl exit 6.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in 580c19ef. Routing now uses semantic schema-version comparison: empty/0.6/0.7 stay on SBOX, while 0.8 prereleases, 0.8 stable, and later versions use PSEC regardless of whether `captureDenials` is present. Non-capture PSEC ownership is retained until the child process tree has terminated and been reaped.

external/windows-sdk/ProcessSecurityEnvironment.fbs
richiemsft: Added the requested OS-style schema purpose and FlatBuffers evolution comment in 580c19ef. The provenance hash was updated and the pinned `check-psec-codegen.js` regeneration/diff gate passes with no generated drift.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed with the same schema-version routing change in 580c19ef. The launch/setup block is now entered for every schema 0.8+ request; Learning Mode trace setup remains conditional only on `captureDenials`.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -535,45 +558,156 @@ impl BaseContainerRunner {
...
+                )
+            }
+        }
+    }
+
+    fn uses_process_security_environment(request: &ExecutionRequest) -> bool {
+        Version::parse(&request.schema_version).is_ok_and(|version| {
+            let comparable = Version::new(version.major, version.minor, version.patch);
+            comparable >= Version::new(0, 8, 0)
+        })
```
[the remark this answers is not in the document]
richiemsft: The schema-based predicate is intentional and follows the explicit review direction in #discussion_r3718467312 and #discussion_r3718490422: schema versions through 0.7 use SBOX, while every schema 0.8+ BaseContainer request uses PSEC, even without `captureDenials`. Gating PSEC on capture would restore the behavior those comments asked us to remove.

The real issue was the stale description: the PR body and two code comments still said ordinary execution remained on SBOX. Fixed in 873ddb10 by documenting schema-based routing in the runner comments, OS support guide, Copilot architecture instructions, and PR description. The V1 behavior is now stated precisely: schema 0.7 remains functional through SBOX; schema 0.8+ requires the official PSEC exports and fails `backend_unavailable` when they are absent.

src/backends/appcontainer/common/src/dispatcher.rs
```diff
@@ -338,13 +338,18 @@ fn select_backend_with_fallback(
...
     DispatchError,
 > {
-    let decision = fallback_detector::detect(&request.policy, /*prefer_bc=*/ true)?;
-    if request.policy.capture_denials.is_some() && decision.tier != IsolationTier::BaseContainer {
-        return Err(DispatchError::CaptureDenialsUnsupported {
-            tier: decision.tier,
-        });
+    // captureDenials uses the official V2 process-security-environment API,
+    // independently of the legacy SBOX tier probe/fallback chain.
+    if request.policy.capture_denials.is_some() {
```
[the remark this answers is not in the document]
richiemsft: Fixed in 5222015e. This was valid: schema 0.8+ routing inside `BaseContainerRunner` was not sufficient because the dispatcher could select T2/T3 before constructing that runner. `select_backend_with_fallback` now checks the same schema predicate and bypasses legacy tier detection for every schema 0.8+ request, not only `captureDenials`. Added `schema_0_8_without_capture_selects_psec_without_legacy_fallback`, which forces the T3 fallback and verifies BaseContainer/PSEC is still selected without a DACL guard.

external/windows-sdk/ProcessSecurityEnvironment.fbs
```diff
@@ -0,0 +1,93 @@
...
+//   - Change the file_identifier or root_type.
+//   - Change a field from scalar to non-scalar (or vice versa).
+//
+// To machine-check for breaking changes, keep a copy of the last shipped schema
+// (e.g. ProcessSecurityEnvironment.previous.fbs) and run:
+//
+//   flatc --conform ProcessSecurityEnvironment.previous.fbs ProcessSecurityEnvironment.fbs
+//
+// --conform verifies that every field/enum/table in the old schema still exists
+// at the same vtable slot, type, and default in the new schema.
```
jsidewhite, line 32: remove

src/backends/learning_mode/windows/src/secenv.rs
```diff
@@ -306,45 +300,61 @@ impl SecurityEnvironmentExportReport {
...
         f.debug_struct("SecurityEnvironmentApi")
             .field("create", &(self.create as *const ()))
+            .field("query_support", &(self.query_support as *const ()))
             .field("close", &(self.close as *const ()))
+            .field("cacheable", &self.cacheable)
             .finish()
     }
 }
 
 impl SecurityEnvironmentApi {
```
jsidewhite, line 346: >SecurityEnvironmentApi [](http://example.com/codeflow?start=5&length=22)

why is the secenv code in `learning_mode` folder?

richiemsft: it's the only place that uses it at the moment. I introduced it since it was needed to call Brian's API.

src/backends/learning_mode/windows/src/secenv.rs
```diff
@@ -357,75 +367,142 @@ impl SecurityEnvironmentApi {
                 .map_err(|e| LearningModeError::DllLoad(e.to_string()))?;
 
             let create_proc = resolve_any(hmodule, CREATE_NAMES)?;
+            let query_support_proc = resolve_any(hmodule, QUERY_SUPPORT_NAMES)?;
             let close_proc = resolve_any(hmodule, CLOSE_NAMES)?;
```
jsidewhite, line 390: These functions are in a *APISet Named Group* and should be queried for with something like,

```
if (!IsApiSetImplemented("api-win-appmodel-processmodel~securityenvironment"))
```

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -535,45 +558,156 @@ impl BaseContainerRunner {
...
+                    &PsecNetworkPolicyArgs {
+                        egress: Some(egress),
+                        ..Default::default()
+                    },
+                )
+            }
+        }
+    }
+
+    pub(crate) fn uses_process_security_environment(request: &ExecutionRequest) -> bool {
```
jsidewhite: >uses_process_security_environment [](http://example.com/codeflow?start=18&length=33)

should this also do the `IsApiSetImplemented` check or how does the fallback logic somehow handle that?

jsidewhite:
>             job.terminate(u32::MAX);

@sergeyshilov does this work?

---
Refers to: src/backends/appcontainer/common/src/base_container_runner.rs:2125 in b9368ff. [](commit_id = b9368ff0e2605a6dcafc9053e779f9342aebb4cf, deletion_comment = False)

richiemsft:
@bbonaby FYI: PR #739 now includes a temporary MXC compatibility fallback for the SBOX model-2 proxy contract introduced by OS PR 16153076.

The updated OS contract requires `allowed_appcontainer_peer`, an AppContainer-hosted proxy, and `privateNetworkClientServer`, while rejecting the current `internetClient` shape. MXC does not yet author that peer identity, so capability-aware hosts now route proxy requests to the AppContainer compatibility path instead of failing with `ERROR_INVALID_PARAMETER (87)`. Query-less hosts retain legacy SBOX proxy behavior.

Validated on build `26663.1000`: minimal built-in WinHTTP proxy E2Es pass for schema 0.7 and 0.8. A separate MXC change is still needed to implement the model-2 T1 proxy architecture and remove this fallback.

src/backends/learning_mode/windows/src/secenv.rs
richiemsft: Addressed in 6961b4af. `SecurityEnvironmentApi::load()` now calls `IsApiSetImplemented("api-win-appmodel-processmodel~securityenvironment")` before loading/resolving the named-group exports, and reports the unavailable API set as backend unavailability.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed in 6961b4af. The request-aware PSEC usability probe calls `SecurityEnvironmentApi::load()`, which now performs the API-set named-group check. A missing group makes PSEC unusable, so ordinary requests continue through SBOX/AppContainer fallback; `captureDenials` reports `backend_unavailable`.

external/windows-sdk/ProcessSecurityEnvironment.fbs
richiemsft: This schema is copied from the official OS PSEC schema, including the FlatBuffer evolution guidance, so I retained it to keep the checked-in contract aligned with its source. If you intended a different specific section on line 32 to be removed, could you clarify which portion?

src/backends/learning_mode/windows/src/secenv.rs
richiemsft: Good question. `secenv` is now used by both ordinary PSEC execution and Learning Mode capture, so it is broader than Learning Mode. Moving it directly into `appcontainer_common` would create a dependency cycle because the Learning Mode lifecycle also consumes it; the clean separation would be a dedicated Windows process-security-environment support crate. I kept that structural extraction out of this ABI/fallback fix, but agree it should be tracked separately.

richiemsft:
@jsidewhite On the `job.terminate(u32::MAX)` question: `UiJobObject::terminate` is a best-effort wrapper around `TerminateJobObject`, so it terminates the sandbox child and all descendants assigned to that job. In the BaseContainer teardown path it is followed by an infinite wait on the root process to reap it before capture/proxy teardown continues. The AppContainer streaming path uses the same tree-kill behavior.

shschaefer (approved):

bbonaby (approved):
We can follow up with Sergey about jeffs comment, and the refactoring.

Issue #686, opened 2026-07-27
[WSLC] Consolidate unsafe FFI into a single auditable module with safe wrappers


SohamDas2021:
### Description of the task

Today the WSLC backend's  unsafe extern  calls into  wslcsdk.dll  (via  libloading ) are interleaved with backend logic in  src/backends/wslc/common/src/wslc_bindings.rs  and  wsl_container_runner.rs . This makes the unsafe surface hard to audit — unsafe blocks are scattered across the runner rather than confined to one boundary.

Move the WSLC unsafe FFI behind a single, auditable module that exposes safe Rust wrappers, so that all  unsafe  for this backend is contained in one place and everything outside it can be treated as safe Rust. Follow-on from PR review feedback on  user/sodas/wslc-bindgen  (reviewer: this "doesn't have to be" that PR).

Scope:

• Keep the bindgen-generated  wslcsdk_sys.rs  as-is (raw FFI, do not hand-edit).
• Wrap each raw SDK entry point (session/container lifecycle, exec, I/O and exit callbacks,  WslcGetMissingComponents , etc.) in a safe function that owns the  unsafe  block, null/HRESULT checking, and pointer/lifetime invariants.
• Refactor  wsl_container_runner.rs  call sites to use the safe wrappers, removing  unsafe  from backend logic.
• Preserve current behavior — no functional change; validate with existing WSLC E2E on a live WSL host.
• Address https://github.com/microsoft/mxc/pull/745#discussion_r3739560871

SohamDas2021:
Tracked in ADO.

PR #745, opened 2026-08-04
[WSLC] State-aware lifecycle (PR 1/3): persistent daemon + named-pipe IPC control plane

SohamDas2021:
## 📖 Description
First of three PRs delivering the **state-aware WSLc lifecycle**. The WSLc SDK (2.9.3) has **no cross-process re-attach** — `WslcSession`/`WslcContainer`/`WslcProcess` handles are in-process only — so a persistent, per-Windows-user **daemon**
must hold the handles across the separate `provision`/`start`/`exec`/`stop`/`deprovision` phase processes. This PR lands that daemon and its IPC control plane. It has **no product-visible surface** (no `wire.rs`/schema changes; nothing invokes the daemon yet), so it merges independently.

### What's included
- **New `wxc-wslc-daemon` crate** (`src/backends/wslc/daemon/`):
  - `main.rs` — bootstrap, bind-before-publish daemon record, dedicated WSLc worker thread/COM apartment, idle watchdog (teardown on no live containers + no in-flight clients).
  - `control_server.rs` — owner-only-SDDL named pipe, length-prefixed JSON frame dispatch
    (`PROVISION`/`START`/`EXEC`/`STOP`/`DEPROVISION`/`PING`).
  - `session_manager.rs` — real `WslcSession`/container/process lifecycle over `container_steps`, holding the refcounted `sandbox_id -> WslcContainer` map on a thread-affine worker.
- **`wslc_common` additions**: `daemon_protocol.rs` (control frames + internal per-phase config structs, deliberately separate from the public wire schema), `daemon_record.rs` (discovery record, pid+creation-time liveness, owner-only DACL hardening, transition lock), `daemon_client.rs` (spawn/poll, connect, exec result accumulation — live bidirectional stdio bridge deferred to PR 2).
- **New `container_steps.rs`**: daemon-side WSLc step primitives (create session/container, exec, stop, delete) adapted for a warm keepalive-init session. The one-shot `wsl_container_runner.rs` is unchanged except two visibility relaxations so the daemon steps can reuse its `import_image_from_tar` + `wslc_prerequisite_error`.
- **Build wiring**: `build.bat` stages the daemon exe next to `wxc-exec` + SDK bin.
- **Tests**: Rust integration test drives the daemon directly over the pipe (real container + 2× exec on a warm container); the full-lifecycle test is `#[ignore]`d (requires a WSL2 host with `alpine:latest` pre-pulled). Unit tests for framing, refcount, idle-teardown, transition lock, and record trust run by default.

### Coming in the pipeline
- **PR 2/3 — state-aware backend + wire/schema + E2E**: `wslc/common/state_aware.rs` (`StatefulSandboxBackend`, prefix `wslc`) translating the public `experimental.wslc.*` wire schema into daemon protocol frames; `mxc_engine` state-aware arm + config-parser wiring; regenerated dev schema + generated TS wire types; multi-invocation E2E script with warm-reuse + idle-teardown assertions;
- **PR 3/3 — TypeScript SDK**: `Wslc*Config`/`*Result` types + branded `SandboxId<'wslc'>` and helper prefix wiring, mirroring the LXC state-aware SDK surface.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/745)

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -140,6 +140,10 @@ for %%T in (x86_64-pc-windows-msvc aarch64-pc-windows-msvc) do (
             )
         )
         if "%WITH_WSLC%"=="1" (
+            if exist "!BIN_DIR!\wxc-wslc-daemon.exe" (
+                copy /Y "!BIN_DIR!\wxc-wslc-daemon.exe" "sdk\node\bin\!SDK_ARCH!\" >nul
+                echo   Copied !SDK_ARCH!\wxc-wslc-daemon.exe
+            )
             if exist "!BIN_DIR!\wslcsdk.dll" (
                 copy /Y "!BIN_DIR!\wslcsdk.dll" "sdk\node\bin\!SDK_ARCH!\" >nul
                 echo   Copied !SDK_ARCH!\wslcsdk.dll
--- a/src/Cargo.toml
+++ b/src/Cargo.toml
@@ -27,6 +27,7 @@ members = [
     "backends/lxc/common",
     "backends/bubblewrap/common",
     "backends/wslc/common",
+    "backends/wslc/daemon",
     "backends/seatbelt/common",
     "host/wxc_host_prep",
     "host/wxc_winhttp_proxy_shim",
--- a/src/backends/wslc/common/src/container_steps.rs
+++ b/src/backends/wslc/common/src/container_steps.rs
@@ -62,6 +62,23 @@ pub(crate) fn sdk_error(context: &str, hr: HRESULT, sdk_msg: &str) -> ScriptResp
     ScriptResponse::error(&msg)
 }
 
+/// NUL-terminate `value` for a C string the WSLc SDK will read, rejecting an
+/// interior NUL instead of letting the SDK silently observe a truncated value.
+///
+/// The daemon protocol accepts arbitrary JSON strings, so a command, env entry,
+/// image, or container path can contain `\u{0}`. Rust keeps such a string whole,
+/// but the SDK stops at the first NUL — so validation/logging and execution
+/// would disagree about what actually ran. Reject at the marshalling boundary.
+fn cstr_bytes(field: &str, value: &str) -> Result<Vec<u8>, ScriptResponse> {
+    std::ffi::CString::new(value)
+        .map(|c| c.into_bytes_with_nul())
+        .map_err(|_| {
+            ScriptResponse::error(&format!(
+                "{field} contains an interior NUL byte, which is not a valid C string"
+            ))
+        })
+}
+
 // ---------------------------------------------------------------------------
 // Process I/O capture plumbing
 // ---------------------------------------------------------------------------
@@ -323,7 +340,7 @@ impl ProcessSettings {
         // script heaps; all are owned by the returned struct.
         let sh = b"/bin/sh\0".to_vec();
         let dash_c = b"-c\0".to_vec();
-        let script_cstr = format!("{}\0", script_code).into_bytes();
+        let script_cstr = cstr_bytes("command", script_code)?;
         let argv: Vec<PCSTR> = vec![
             sh.as_ptr() as PCSTR,
             dash_c.as_ptr() as PCSTR,
@@ -341,8 +358,8 @@ impl ProcessSettings {
         if !env.is_empty() {
             env_cstrings = env
                 .iter()
-                .map(|e| format!("{}\0", e).into_bytes())
-                .collect();
+                .map(|e| cstr_bytes("environment variable", e))
+                .collect::<Result<Vec<_>, _>>()?;
             env_ptrs = env_cstrings.iter().map(|e| e.as_ptr() as PCSTR).collect();
             let hr =
                 sdk.WslcSetProcessSettingsEnvVariables(&mut raw, env_ptrs.as_ptr(), env_ptrs.len());
@@ -365,7 +382,7 @@ impl ProcessSettings {
                     "working_directory must be an absolute in-container path (got {working_directory:?})"
                 )));
             }
-            let c = format!("{}\0", working_directory).into_bytes();
+            let c = cstr_bytes("working_directory", working_directory)?;
             let hr = sdk.WslcSetProcessSettingsWorkingDirectory(&mut raw, c.as_ptr() as PCSTR);
             if hr != S_OK {
                 return Err(sdk_error(
@@ -452,7 +469,7 @@ impl ContainerSettings {
         flags: WslcContainerFlags,
         logger: &mut Logger,
     ) -> Result<Self, ScriptResponse> {
-        let image_cstr = format!("{}\0", image).into_bytes();
+        let image_cstr = cstr_bytes("image", image)?;
         let mut raw = std::mem::zeroed::<WslcContainerSettings>();
         let hr = sdk.WslcInitContainerSettings(image_cstr.as_ptr() as PCSTR, &mut raw);
         if hr != S_OK {
@@ -503,10 +520,10 @@ impl ContainerSettings {
             .iter()
             .map(|m| {
                 let win: Vec<u16> = to_wide(&m.windows_path);
-                let ctr: Vec<u8> = format!("{}\0", m.container_path).into_bytes();
-                (win, ctr)
+                let ctr: Vec<u8> = cstr_bytes("mount container_path", &m.container_path)?;
+                Ok((win, ctr))
             })
-            .collect();
+            .collect::<Result<Vec<_>, ScriptResponse>>()?;
 
         let mut volumes: Vec<WslcContainerVolume> = Vec::new();
         if !mounts.is_empty() {
@@ -1063,4 +1080,33 @@ pub unsafe fn delete_daemon_container(
     }
     let _ = writeln!(logger, "[WSLC][daemon] Container deleted");
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn cstr_bytes_nul_terminates_a_plain_value() {
+        let bytes = cstr_bytes("command", "echo hi").expect("plain value is a valid C string");
+        assert_eq!(bytes, b"echo hi\0");
+    }
+
+    #[test]
+    fn cstr_bytes_accepts_empty() {
+        let bytes = cstr_bytes("working_directory", "").expect("empty value is a valid C string");
+        assert_eq!(bytes, b"\0");
+    }
+
+    #[test]
+    fn cstr_bytes_rejects_interior_nul() {
+        let err = cstr_bytes("command", "echo\0rm -rf /")
+            .expect_err("an interior NUL must be rejected, not silently truncated");
+        let msg = format!("{:?}", err);
+        assert!(msg.contains("command"), "error names the field: {msg}");
+        assert!(
+            msg.contains("interior NUL"),
+            "error explains the cause: {msg}"
+        );
+    }
 }
--- a/src/backends/wslc/common/src/container_steps.rs
+++ b/src/backends/wslc/common/src/container_steps.rs
@@ -140,6 +140,11 @@ unsafe extern "C" fn exit_callback(_exit_code: i32, context: *mut c_void) {
 pub struct ProcessSettings {
     raw: WslcProcessSettings,
     io_ctx: Arc<IoContext>,
+    // Reclaims the SDK's `IoContext` reference if settings build or process
+    // creation fails before a live process adopts it. Disarmed by
+    // `mark_process_created` once a process is created, after which
+    // `exit_callback` (or a deliberate leak on kill) owns reclamation.
+    sdk_io_ref: SdkIoRef,
     _sh: Vec<u8>,
     _dash_c: Vec<u8>,
     _script_cstr: Vec<u8>,
@@ -147,6 +152,38 @@ pub struct ProcessSettings {
     _env_cstrings: Vec<Vec<u8>>,
     _env_ptrs: Vec<PCSTR>,
     _cwd_cstr: Option<Vec<u8>>,
+}
+
+/// Owns an `Arc<IoContext>` reference handed to the SDK and reclaims it on drop
+/// unless disarmed. Used both as a local guard while building settings and as
+/// the stored owner inside [`ProcessSettings`], so any error path before a live
+/// process adopts the reference frees it instead of leaking.
+struct SdkIoRef(Option<*const IoContext>);
+
+impl SdkIoRef {
+    fn none() -> Self {
+        Self(None)
+    }
+
+    /// Disarm reclamation, returning the raw pointer (dropped on the floor by the
+    /// caller so ownership passes to `exit_callback`).
+    fn disarm(&mut self) -> Option<*const IoContext> {
+        self.0.take()
+    }
+}
+
+impl Drop for SdkIoRef {
+    fn drop(&mut self) {
+        if let Some(p) = self.0.take() {
+            // SAFETY: `p` came from `Arc::into_raw`. This runs only when no live
+            // process ever adopted the SDK reference (settings build or process /
+            // container creation failed), so the SDK will never invoke
+            // `exit_callback` for it and reclaiming here is sound.
+            unsafe {
+                drop(Arc::from_raw(p));
+            }
+        }
+    }
 }
 
 impl ProcessSettings {
@@ -206,6 +243,9 @@ impl ProcessSettings {
         // SDK's reference is reclaimed inside `exit_callback`, so a process that
         // is killed without its exit callback firing deliberately leaks the
         // reference rather than freeing it while the SDK may still call back.
+        // Until a process actually adopts the reference, `sdk_io_ref` guards it
+        // so any early return below frees it instead of leaking.
+        let mut sdk_io_ref = SdkIoRef::none();
         if register_callbacks {
             let io_ctx_raw = Arc::into_raw(Arc::clone(&io_ctx)) as *mut c_void;
             let callbacks = WslcProcessCallbacks {
@@ -219,6 +259,7 @@ impl ProcessSettings {
                 drop(Arc::from_raw(io_ctx_raw as *const IoContext));
                 return Err(sdk_error("WslcSetProcessSettingsCallbacks failed", hr, ""));
             }
+            sdk_io_ref = SdkIoRef(Some(io_ctx_raw as *const IoContext));
         }
 
         // Command line: /bin/sh -c <script>. argv points into the sh/dash_c/
@@ -279,6 +320,7 @@ impl ProcessSettings {
         Ok(ProcessSettings {
             raw,
             io_ctx,
+            sdk_io_ref,
             _sh: sh,
             _dash_c: dash_c,
             _script_cstr: script_cstr,
@@ -287,6 +329,21 @@ impl ProcessSettings {
             _env_ptrs: env_ptrs,
             _cwd_cstr: cwd_cstr,
         })
+    }
+
+    /// Transfer ownership of the SDK's `IoContext` reference to the newly-created
+    /// process. After this, `Drop` will not reclaim it — `exit_callback` (or a
+    /// deliberate leak on kill) owns reclamation, preserving the invariant that
+    /// the reference is never freed while the SDK may still call back.
+    ///
+    /// # Safety
+    /// Call exactly once, immediately after a process has been successfully
+    /// created from these settings (`WslcCreateContainerProcess`, or
+    /// `WslcCreateContainer` + `WslcStartContainer` when used as a container
+    /// init). Skipping this after a successful create would let `Drop` free a
+    /// reference the SDK still owns.
+    pub unsafe fn mark_process_created(&mut self) {
+        let _ = self.sdk_io_ref.disarm();
     }
 
     /// The I/O-capture context shared with the SDK callbacks.
@@ -746,6 +803,9 @@ pub unsafe fn exec_in_container(
         return Err(sdk_error("WslcCreateContainerProcess failed", hr, &msg));
     }
     let process_guard = WslcProcessGuard::from_raw(process, sdk.release_process_fn());
+    // The process now owns the SDK's IoContext reference; `exit_callback` will
+    // reclaim it. Disarm the settings guard so it is not freed underneath the SDK.
+    process_settings.mark_process_created();
 
     let mut exit_event: HANDLE = ptr::null_mut();
     let hr = sdk.WslcGetProcessExitEvent(process_guard.as_raw(), &mut exit_event);
@@ -768,7 +828,14 @@ pub unsafe fn exec_in_container(
                 wait_ms
             );
             // Kill only this process; the keepalive init keeps the container up.
-            let _ = sdk.WslcSignalProcess(process_guard.as_raw(), WslcSignal::WSLC_SIGNAL_SIGKILL);
+            let kill_hr =
+                sdk.WslcSignalProcess(process_guard.as_raw(), WslcSignal::WSLC_SIGNAL_SIGKILL);
+            if kill_hr != S_OK {
+                let _ = writeln!(
+                    logger,
+                    "[WSLC][daemon] Warning: WslcSignalProcess(SIGKILL) failed (hr={kill_hr:?}); process may still be running"
+                );
+            }
         }
     }
 
--- a/src/backends/wslc/common/src/container_steps.rs
+++ b/src/backends/wslc/common/src/container_steps.rs
@@ -75,36 +75,14 @@ pub struct IoContext {
     pub(crate) exited: Arc<(Mutex<bool>, Condvar)>,
 }
 
-/// RAII guard that reclaims an `Arc<IoContext>` from a raw pointer on drop.
-/// Prevents leaking the `Arc` reference count on early returns.
-pub(crate) struct IoCtxRawGuard {
-    ptr: *mut c_void,
-}
-
-impl IoCtxRawGuard {
-    fn new(ptr: *mut c_void) -> Self {
-        Self { ptr }
-    }
-}
-
-impl Drop for IoCtxRawGuard {
-    fn drop(&mut self) {
-        if !self.ptr.is_null() {
-            unsafe {
-                eprintln!("[WSLC][debug] IoCtxRawGuard dropped -- reclaiming Arc<IoContext>");
-                let _ = Arc::from_raw(self.ptr as *const IoContext);
-            }
-        }
-    }
-}
-
 /// Callback invoked by the WSLc SDK for stdout/stderr data.
 ///
 /// # Safety
 /// `context` must be a valid pointer obtained from `Arc::into_raw(Arc<IoContext>)`.
-/// The `Arc` is kept alive by the owning [`ProcessSettings`] (via [`IoCtxRawGuard`],
-/// which reclaims it on drop), so the pointer remains valid for the duration of
-/// all callbacks. The SDK guarantees `data` is valid for `data_size` bytes.
+/// The `Arc` reference handed to the SDK is released inside [`exit_callback`]; the
+/// SDK guarantees no I/O callback fires after exit, and the owning
+/// [`ProcessSettings`] holds an independent reference, so the pointer stays valid
+/// for every callback. The SDK guarantees `data` is valid for `data_size` bytes.
 unsafe extern "C" fn io_callback(
     io_handle: WslcProcessIOHandle,
     data: *const BYTE,
@@ -131,15 +109,21 @@ unsafe extern "C" fn io_callback(
 
 /// Callback invoked when the process exits and all I/O has been flushed.
 /// Per SDK docs: "Once this callback is invoked, any registered IO callbacks
-/// will no longer be called." This guarantees buffers are complete.
+/// will no longer be called." This guarantees buffers are complete and that the
+/// SDK will not touch `context` again, so this reclaims the `Arc` reference the
+/// SDK was handed.
 ///
 /// # Safety
-/// Same lifetime requirements as [`io_callback`].
+/// `context` must be the pointer handed to `WslcSetProcessSettingsCallbacks`,
+/// obtained from `Arc::into_raw(Arc<IoContext>)`, and must be invoked at most
+/// once.
 unsafe extern "C" fn exit_callback(_exit_code: i32, context: *mut c_void) {
     if context.is_null() {
         return;
     }
-    let ctx = &*(context as *const IoContext);
+    // Reclaim the SDK's reference. The owning `ProcessSettings` holds its own
+    // reference, so the `IoContext` stays alive for the waiting thread.
+    let ctx = Arc::from_raw(context as *const IoContext);
     let mut exited = ctx.exited.0.lock().unwrap_or_else(|e| e.into_inner());
     *exited = true;
     ctx.exited.1.notify_all();
@@ -156,7 +140,6 @@ unsafe extern "C" fn exit_callback(_exit_code: i32, context: *mut c_void) {
 pub struct ProcessSettings {
     raw: WslcProcessSettings,
     io_ctx: Arc<IoContext>,
-    _io_guard: IoCtxRawGuard,
     _sh: Vec<u8>,
     _dash_c: Vec<u8>,
     _script_cstr: Vec<u8>,
@@ -212,21 +195,19 @@ impl ProcessSettings {
             return Err(sdk_error("WslcInitProcessSettings failed", hr, ""));
         }
 
-        // Register I/O callbacks to capture stdout/stderr. We hand the SDK an
-        // Arc reference via raw pointer; IoCtxRawGuard reconstructs it on drop
-        // so the reference count is not leaked, and the memory stays alive while
-        // the SDK may still invoke callbacks on its internal threads.
         let io_ctx = Arc::new(IoContext {
             stdout: Arc::new(Mutex::new(Vec::new())),
             stderr: Arc::new(Mutex::new(Vec::new())),
             exited: Arc::new((Mutex::new(false), Condvar::new())),
         });
-        let io_ctx_raw = Arc::into_raw(Arc::clone(&io_ctx)) as *mut c_void;
-        let io_guard = IoCtxRawGuard::new(io_ctx_raw);
 
         // Callbacks are registered only for the streamed path; a detached init
-        // shares no IoContext with the SDK.
+        // shares no IoContext with the SDK and mints no extra reference. The
+        // SDK's reference is reclaimed inside `exit_callback`, so a process that
+        // is killed without its exit callback firing deliberately leaks the
+        // reference rather than freeing it while the SDK may still call back.
         if register_callbacks {
+            let io_ctx_raw = Arc::into_raw(Arc::clone(&io_ctx)) as *mut c_void;
             let callbacks = WslcProcessCallbacks {
                 onStdOut: Some(io_callback),
                 onStdErr: Some(io_callback),
@@ -234,6 +215,8 @@ impl ProcessSettings {
             };
             let hr = sdk.WslcSetProcessSettingsCallbacks(&mut raw, &callbacks, io_ctx_raw);
             if hr != S_OK {
+                // The SDK never took ownership; reclaim the reference now.
+                drop(Arc::from_raw(io_ctx_raw as *const IoContext));
                 return Err(sdk_error("WslcSetProcessSettingsCallbacks failed", hr, ""));
             }
         }
@@ -296,7 +279,6 @@ impl ProcessSettings {
         Ok(ProcessSettings {
             raw,
             io_ctx,
-            _io_guard: io_guard,
             _sh: sh,
             _dash_c: dash_c,
             _script_cstr: script_cstr,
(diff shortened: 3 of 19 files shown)


src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -0,0 +1,625 @@
...
+            while let Some(cmd) = rx.blocking_recv() {
+                match cmd {
+                    WorkerCommand::Provision { config, reply } => {
+                        let _ = reply.send(worker.provision(config));
+                    }
+                    WorkerCommand::Start { config, reply } => {
+                        let _ = reply.send(worker.start(config));
+                    }
+                    WorkerCommand::Exec { config, reply } => {
+                        let _ = reply.send(worker.exec(config));
```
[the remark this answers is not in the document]
SohamDas2021: Acknowledged — this is real. PR2 makes admission atomic (validate + start on the worker without yielding), but the full fix — keeping SDK calls thread-affine while moving only process-event waiting off-thread so Stop / Deprovision /idle/shutdown aren't queued behind a long exec — is scoped to the 2c increment

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+    };
+    write_frame(&mut pipe, &terminal).await?;
+    Ok(())
+}
+
+/// Map a `Result<()>` to `Ok` / `Err` response.
+fn ok_or_err(result: Result<()>) -> DaemonResponse {
+    match result {
+        Ok(()) => DaemonResponse::Ok,
+        Err(e) => err_response(ErrKind::Backend, e),
```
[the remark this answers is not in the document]
SohamDas2021: Will be addressed in PR2

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+/// TODO(fill-in): bidirectional live stdio (client `Stdin` frames -> process,
+/// process stdout/stderr -> `Stdout`/`Stderr` frames). The skeleton runs the
+/// command to completion and emits only the terminal frame.
+async fn handle_exec(
+    mut pipe: NamedPipeServer,
+    session: SessionHandle,
+    config: wslc_common::daemon_protocol::ExecConfig,
+) -> Result<()> {
+    write_frame(&mut pipe, &DaemonResponse::Ok).await?;
+    let terminal = match session.exec(config).await {
```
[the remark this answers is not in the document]
SohamDas2021: Will be addressed in PR2

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+    Ok(())
+}
+
+/// Exec: admit with `Ok`, then stream the run's outcome as [`StreamFrame`]s.
+///
+/// TODO(fill-in): bidirectional live stdio (client `Stdin` frames -> process,
+/// process stdout/stderr -> `Stdout`/`Stderr` frames). The skeleton runs the
+/// command to completion and emits only the terminal frame.
```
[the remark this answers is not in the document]
SohamDas2021: PR description updated. Doc change will be addressed in PR2

build.bat
```diff
@@ -140,6 +140,10 @@ for %%T in (x86_64-pc-windows-msvc aarch64-pc-windows-msvc) do (
             )
         )
         if "%WITH_WSLC%"=="1" (
+            if exist "!BIN_DIR!\wxc-wslc-daemon.exe" (
+                copy /Y "!BIN_DIR!\wxc-wslc-daemon.exe" "sdk\node\bin\!SDK_ARCH!\" >nul
+                echo   Copied !SDK_ARCH!\wxc-wslc-daemon.exe
```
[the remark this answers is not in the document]
SohamDas2021: Partially done — PR2 added wxc-wslc-daemon.exe to the ADO copy list ( Rust.Build.Job.yml ). Completing it: I'll also add it to the GitHub artifact list ( Build.Windows.Job.yml ) and the ADO signing list so CI/release artifacts contain the co-located executable  DaemonClient  requires.

src/Cargo.toml
```diff
@@ -27,6 +27,7 @@ members = [
     "backends/lxc/common",
     "backends/bubblewrap/common",
     "backends/wslc/common",
+    "backends/wslc/daemon",
```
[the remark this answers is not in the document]
SohamDas2021: Will be addressed in PR2

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+use wxc_common::models::{PortMapping, ScriptResponse};
+use wxc_common::string_util::{to_wide, CoTaskMemPWSTR};
+
+use crate::policy_mapping::{self, VolumeMount};
+use crate::wsl_container_runner::{wslc_prerequisite_error, WSLContainerRunner};
+use crate::wslc_bindings::*;
+
+// ---------------------------------------------------------------------------
+// Shared error helper
+// ---------------------------------------------------------------------------
```
bbonaby, line 52: thought: probably would do away with these kinds of comments since they take up space, no clue why the agent always does them. They do it to me too.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+
+/// Build a `ScriptResponse` error from an HRESULT failure with an optional
+/// SDK-provided message.
+pub(crate) fn sdk_error(context: &str, hr: HRESULT, sdk_msg: &str) -> ScriptResponse {
+    let msg = if sdk_msg.is_empty() {
+        format!("{}: HRESULT 0x{:08X}", context, hr as u32)
+    } else {
+        format!("{}: {} (HRESULT 0x{:08X})", context, sdk_msg, hr as u32)
+    };
+    ScriptResponse::error(&msg)
```
bbonaby, line 62: question: I thought for ScriptResponse errors now we try to encapsulate the error in an enum error type since this would help the rust sdk. Could be wrong though but that seemed like a good error pattern.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+
+/// Callback invoked by the WSLc SDK for stdout/stderr data.
+///
+/// # Safety
+/// `context` must be a valid pointer obtained from `Arc::into_raw(Arc<IoContext>)`.
+/// The `Arc` reference handed to the SDK is released inside [`exit_callback`]; the
+/// SDK guarantees no I/O callback fires after exit, and the owning
+/// [`ProcessSettings`] holds an independent reference, so the pointer stays valid
+/// for every callback. The SDK guarantees `data` is valid for `data_size` bytes.
+unsafe extern "C" fn io_callback(
```
bbonaby, line 156: issue: can we move all the unsafe functions to a   `wslc_sys.rs` file under `src/backends/wslc/common/src/`  ? That would be helpful so we have all the unsafe functions in one place that we can review.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+    // `mark_process_created` once a process is created, after which
+    // `exit_callback` (or a deliberate leak on kill) owns reclamation.
+    sdk_io_ref: SdkIoRef,
+    _sh: Vec<u8>,
+    _dash_c: Vec<u8>,
+    _script_cstr: Vec<u8>,
+    _argv: Vec<PCSTR>,
+    _env_cstrings: Vec<Vec<u8>>,
+    _env_ptrs: Vec<PCSTR>,
+    _cwd_cstr: Option<Vec<u8>>,
```
bbonaby, line 228: question: why the need for underscores here? Is it because these fields won't be used in the PR but another one?

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+/// context. Safe to move (all referenced data is heap/`'static`); do not move
+/// after taking `&raw`.
+pub struct ProcessSettings {
+    raw: WslcProcessSettings,
+    io_ctx: Arc<IoContext>,
+    // Reclaims the SDK's `IoContext` reference if settings build or process
+    // creation fails before a live process adopts it. Disarmed by
+    // `mark_process_created` once a process is created, after which
+    // `exit_callback` (or a deliberate leak on kill) owns reclamation.
+    sdk_io_ref: SdkIoRef,
```
bbonaby, line 217: thought: haven't read through the rest of this yet but seeing `deliberate leak on kill` makes me think we need a simple design doc (mostly diagrams, headings and brief text, nothing long form). That said maybe it'll make sense further down.

src/backends/wslc/common/src/daemon_record.rs
```diff
@@ -0,0 +1,683 @@
...
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()),
+        Err(e) => Err(e),
+    }
+}
+
+/// True iff the daemon described by `record` is still the live process it claims
+/// to be: a process with its PID is currently running AND its creation time
+/// matches the recorded one.
+pub fn daemon_alive(record: &DaemonRecord) -> bool {
+    running_process_creation_time(record.pid) == Some(record.pid_creation_time)
```
bbonaby, line 235: **praise:** Pairing PID liveness with creation time and validating the record owner and pipe prefix is a strong fail-closed discovery design.

src/backends/wslc/daemon/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "wxc_wslc_daemon"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[[bin]]
+name = "wxc-wslc-daemon"
+path = "src/main.rs"
+
+[dependencies]
```
bbonaby, line 11: **issue:** This workspace member imports Windows-only APIs unconditionally, so documented non-Windows workspace commands fail. We should target-gate the dependencies/code and provide a non-Windows stub `main`.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+) {
+    if context.is_null() || data.is_null() || data_size == 0 {
+        return;
+    }
+    let ctx = &*(context as *const IoContext);
+    let bytes = std::slice::from_raw_parts(data, data_size as usize);
+    match io_handle {
+        WslcProcessIOHandle::WSLC_PROCESS_IO_HANDLE_STDOUT => {
+            let mut buf = ctx.stdout.lock().unwrap_or_else(|e| e.into_inner());
+            buf.extend_from_slice(bytes);
```
bbonaby: **issue (blocking):** stdout/stderr comes from untrusted code and grows these buffers without limit, so one exec can OOM the daemon and terminate every owned sandbox. Let's cap each stream or use bounded streaming/backpressure.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+    }
+
+    let wait_ms = if timeout_ms > 0 { timeout_ms } else { u32::MAX };
+    let mut timed_out = false;
+    if !exit_event.is_null() {
+        let wait_result = windows::Win32::System::Threading::WaitForSingleObject(
+            windows::Win32::Foundation::HANDLE(exit_event),
+            wait_ms,
+        );
+        if wait_result == windows::Win32::Foundation::WAIT_TIMEOUT {
```
bbonaby: **issue (blocking):** A timeout is reported even if SIGKILL fails, and unexpected waits or a missing exit callback still fall through to `WslcGetProcessExitCode` (possibly `STILL_ACTIVE`). Could we require positive exit confirmation and retire the warm container when termination cannot be confirmed?

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+                    hr,
+                    "",
+                ));
+            }
+        }
+
+        // Working directory (mapped Windows -> container path; skip if unmapped).
+        let mut cwd_cstr: Option<Vec<u8>> = None;
+        if !working_directory.is_empty() {
+            if let Some(container_cwd) =
```
bbonaby: **issue:** `working_directory` is an in-container path, but `/work` is treated as a Windows path and silently ignored. We can avoid that fallback by validating an absolute container path and passing it directly to the SDK.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -0,0 +1,632 @@
...
+                config.timeout_ms,
+                &mut self.logger,
+            )
+        }
+        .map_err(sr_err)?;
+
+        if outcome.timed_out {
+            anyhow::bail!("exec timed out after {}ms", config.timeout_ms);
+        }
+        // NOTE: outcome.stdout/stderr are captured but not yet forwarded — live
```
bbonaby, line 360: **issue (blocking):** This drops captured stdout/stderr although `DaemonClient::exec` exposes them, so output-dependent commands always return empty buffers. We should return the outcome and emit bounded stdout/stderr frames before `Exit`.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -0,0 +1,632 @@
...
+            while let Some(cmd) = rx.blocking_recv() {
+                match cmd {
+                    WorkerCommand::Provision { config, reply } => {
+                        let _ = reply.send(worker.provision(config));
+                    }
+                    WorkerCommand::Start { config, reply } => {
+                        let _ = reply.send(worker.start(config));
+                    }
+                    WorkerCommand::Exec { config, reply } => {
+                        let _ = reply.send(worker.exec(config));
```
bbonaby, line 471: **issue (blocking):** `worker.exec` occupies the only SDK worker for the full command lifetime; one no-timeout exec blocks unrelated sandboxes, the idle watchdog, and shutdown. Could we keep the SDK calls affine here while moving only process-event waiting off-thread?

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+async fn handle_exec(
+    mut pipe: NamedPipeServer,
+    session: SessionHandle,
+    config: wslc_common::daemon_protocol::ExecConfig,
+) -> Result<()> {
+    write_frame(&mut pipe, &DaemonResponse::Ok).await?;
```
bbonaby, line 397: **issue:** `Ok` is sent before validating sandbox/state, so routine admission failures become untyped stream errors while `NotProvisioned`, `NotStarted`, and `Busy` already exist. Let's validate first and return the stable error kind.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+                    server = create_secured_instance(&pipe_name, &security, false)?;
+                    continue;
+                }
+                let connected = server;
+                // Create the next instance before servicing this one.
+                server = create_secured_instance(&pipe_name, &security, false)?;
+                let session = session.clone();
+                let active = active_clients.clone();
+                active.fetch_add(1, Ordering::SeqCst);
+                tokio::spawn(async move {
```
bbonaby: **issue:** These handlers are detached, so shutdown can tear down the worker while an accepted request is still running. We should retain the tasks, stop admission atomically, and drain them before worker shutdown.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+            .to_string()
+            .context("SID string was not valid UTF-16")?;
+        let _ = LocalFree(Some(HLOCAL(sid_wstr.0 as *mut c_void)));
+        Ok(sid)
+    }
+}
+
+/// Service exactly one request on a freshly-connected pipe instance.
+async fn handle_client(mut pipe: NamedPipeServer, session: SessionHandle) -> Result<()> {
+    let request: DaemonRequest = read_frame(&mut pipe).await?;
```
bbonaby: **issue:** The first frame has no deadline and each connection creates an unbounded task, so an abandoned same-user client can retain daemon resources indefinitely. Could we add a short request timeout and a bounded client semaphore?

src/backends/wslc/daemon/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+        shutdown.clone(),
+    ));
+
+    // Wait for the control server to finish (it stops when `shutdown` fires).
+    let server_result = server.await;
+
+    // Ensure the watchdog and worker are torn down.
+    shutdown.notify_waiters();
+    watchdog.abort();
+    let _ = session.shutdown().await;
```
bbonaby: **issue:** The record remains ready while the server stops and is removed outside the transition lock, so a client can select a daemon that is already disappearing. Let's mark it not-ready and serialize admission shutdown, draining, worker teardown, and record removal with the transition lock.

src/backends/wslc/daemon/tests/daemon_ipc.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+impl Drop for DaemonProcess {
+    fn drop(&mut self) {
+        let _ = self.child.kill();
+        let _ = self.child.wait();
+        // The daemon was killed before it could clean up; drop the stale record
+        // so it does not confuse a later run.
+        let _ = remove_daemon_record();
```
bbonaby: **issue:** This test uses and unconditionally deletes the real per-user daemon record, which can orphan a live daemon or race another test run. Could we use an isolated test state root and remove only the record owned by this child?

src/backends/wslc/common/src/daemon_client.rs
```diff
@@ -0,0 +1,366 @@
...
+use serde::Serialize;
+
+use crate::daemon_protocol::{
+    encode_frame, DaemonRequest, DaemonResponse, DeprovisionConfig, ErrKind, ExecConfig,
+    ProvisionConfig, StartConfig, StopConfig, StreamFrame, MAX_FRAME_SIZE, PROTOCOL_VERSION,
+};
+use crate::daemon_record::{live_daemon, DaemonRecord, TransitionLock};
+
+/// Max wait to acquire the daemon spawn/teardown transition lock.
+const SPAWN_LOCK_TIMEOUT: Duration = Duration::from_secs(30);
```
bbonaby: **issue:** This 30-second lock can expire before the 60-second readiness wait it protects, so a second client can fail while the first daemon is still within its valid startup window. We should make the lock timeout longer than `READY_TIMEOUT`.

bbonaby (commented):
**issue (blocking):** `wxc-wslc-daemon.exe` is required beside `wxc-exec.exe`, but it is absent from the ADO signing/copy pattern, GitHub artifact upload, and playground resources. Could we add it to all three manifests so release builds sign and ship the daemon?

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Will be fixed in PR2.

src/backends/wslc/daemon/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+    active_clients: Arc<AtomicUsize>,
+    shutdown: Arc<Notify>,
+) {
+    let mut idle_for = Duration::ZERO;
+    loop {
+        tokio::time::sleep(IDLE_POLL).await;
+        match session.container_count().await {
+            // A provision holds the count at zero while it runs, so also require
+            // no in-flight client requests before counting as idle.
+            Ok(0) if active_clients.load(Ordering::SeqCst) == 0 => {
```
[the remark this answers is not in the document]
SohamDas2021: Addressed in PR2 (the daemon-hardening commit). 

Exec is now validate-then-admit — the worker validates existence/started-state and admits atomically with the start of the run, so unknown/not-started sandboxes return a pre-admission typed  DaemonResponse::Err { NotProvisioned/NotStarted }  instead of an untyped stream  Error . All phase failures now carry a typed  WorkerError  whose  kind()  maps to the stable  ErrKind . The idle watchdog also gained a monotonic  activity  generation counter so a connect+finish between two polls can't be mistaken for idle. These land on the PR2 branch; I'll ensure the PR1 branch carries them before merge.

src/backends/wslc/daemon/src/session_manager.rs
SohamDas2021: Will be addressed in PR2c

src/backends/wslc/daemon/src/session_manager.rs
SohamDas2021: Will be addressed in PR2c

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Good catch- proper bounded streaming/backpressure is coupled to the 2c stdio work, but I'll add an interim per-stream byte cap (truncate with a marker) in this PR so a single exec can't OOM the daemon and take down every owned sandbox.

src/backends/wslc/common/src/daemon_client.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    /// Open a fresh synchronous connection to the daemon pipe, retrying past the
+    /// brief windows where every instance is momentarily busy or being recreated.
+    fn open_pipe(&self) -> Result<File> {
+        let deadline = Instant::now() + OPEN_TIMEOUT;
+        loop {
+            match OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(&self.pipe_name)
```
MGudgin: `src/backends/wslc/common/src/daemon_client.rs:220`

**High (security) - authenticate the pipe server and restrict impersonation.**

This opens the endpoint named by the same-user-owned record with default named-pipe QoS. Windows defaults that to `SecurityImpersonation`, so a hostile same-user pipe server can impersonate an elevated phase process after planting a valid-looking record containing its own PID, creation time, and prefixed pipe name. The client also never verifies that the connected server PID matches the record. The existing diagnostic-pipe client already demonstrates the repository's `GetNamedPipeServerProcessId` plus integrity-token verification pattern.

**Fix:** open with `SECURITY_SQOS_PRESENT | SECURITY_IDENTIFICATION` (or a stricter level), then verify the connected server PID/creation time against the trusted record or use a per-spawn nonce authenticated on the wire.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// SAFETY: `psd` is a self-contained LocalAlloc'd security descriptor with no
+// thread affinity; ownership can move across threads freely.
+unsafe impl Send for OwnerOnlySecurity {}
+
+impl OwnerOnlySecurity {
+    fn new() -> Result<Self> {
+        let sid = current_user_sid_string()?;
+        let sddl = format!("D:P(A;;FA;;;{sid})(A;;FA;;;SY)");
```
MGudgin, line 268: `src/backends/wslc/daemon/src/control_server.rs:132`

**Medium (security) - the pipe ACL does not preserve the caller's integrity boundary.**

The ACL grants full access to the user SID and SYSTEM but does not add a mandatory label or authenticate the connecting token. Elevated and non-elevated processes for a split-token user have the same user SID, so a medium-integrity process can drive a daemon spawned elevated. `Provision` then passes caller-supplied host mounts to the SDK without validating that the caller could access those paths itself, turning the elevated daemon into a confused deputy.

**Fix:** reject clients whose integrity/elevation does not match the daemon (or stamp the appropriate mandatory label), and validate host mount paths under the authenticated caller's security context.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -0,0 +1,632 @@
...
+    }
+}
+
+/// Spawn the WSLc worker thread and return a handle to it.
+///
+/// The thread joins the MTA for its entire lifetime (WSLc SDK apartment
+/// affinity) and processes [`WorkerCommand`]s until a [`WorkerCommand::Shutdown`]
+/// is received or the command channel closes.
+pub fn spawn() -> Result<SessionHandle> {
+    let (tx, mut rx) = mpsc::unbounded_channel::<WorkerCommand>();
```
MGudgin, line 447: `src/backends/wslc/daemon/src/session_manager.rs:428`

**High (reliability/performance) - a long exec freezes the entire daemon while requests accumulate without bound.**

All lifecycle operations share this one worker and unbounded channel. `WorkerCommand::Exec` calls `exec_in_container`, which synchronously waits on the process exit event; `timeout_ms == 0` becomes an effectively unbounded wait. During that time, unrelated provision/start/stop/deprovision requests and the watchdog's `ContainerCount` request queue indefinitely, consuming memory and preventing recovery.

**Fix:** use a bounded channel with explicit backpressure/`Busy`, and move process-exit waiting off the SDK command worker (or introduce per-container execution workers) so unrelated sandboxes remain serviceable. Add a regression test with one blocked exec and a second sandbox operation.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+        tokio::select! {
+            connect = server.connect() => {
+                if let Err(e) = connect {
+                    eprintln!("[wslc-daemon] pipe connect error: {e}");
+                    server = create_secured_instance(&pipe_name, &security, false)?;
+                    continue;
+                }
+                let connected = server;
+                // Create the next instance before servicing this one.
+                server = create_secured_instance(&pipe_name, &security, false)?;
```
MGudgin: `src/backends/wslc/daemon/src/control_server.rs:82`

**Medium (reliability) - failure to create the next pipe instance terminates every live sandbox.**

This `?` runs after a client has connected but before that connection is handed to `handle_client`. A transient handle/resource-exhaustion failure therefore abandons the accepted client, exits the control server, and causes `main` to shut down the shared session and all containers. The connect-error path above has the same propagation when re-creating its instance.

**Fix:** retry instance creation with bounded backoff while retaining or explicitly failing the accepted client; do not turn one transient accept-capacity failure into daemon-wide teardown.

src/backends/wslc/common/src/container_steps.rs
MGudgin: `src/backends/wslc/common/src/container_steps.rs:100`

**High (reliability/performance) - daemon exec output can exhaust process memory.**

The new daemon callback appends all stdout/stderr into unbounded vectors until the process exits. A container command that continuously writes output grows the long-lived daemon without limit; because exec also occupies the sole worker, other lifecycle requests cannot make progress while memory is consumed.

**Fix:** enforce a documented per-stream or aggregate capture limit, stop extending after the limit, and report truncation in the terminal result. Cover a command that exceeds the limit.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+
+/// Read one length-prefixed frame and deserialise it.
+async fn read_frame<T: DeserializeOwned>(pipe: &mut NamedPipeServer) -> Result<T> {
+    let mut len_buf = [0u8; 4];
+    pipe.read_exact(&mut len_buf).await?;
+    let len = u32::from_le_bytes(len_buf) as usize;
+    if len > MAX_FRAME_SIZE {
+        bail!("incoming frame length {len} exceeds maximum {MAX_FRAME_SIZE}");
+    }
+    let mut body = vec![0u8; len];
```
MGudgin, line 431: `src/backends/wslc/daemon/src/control_server.rs:291`

**Medium (security/performance) - each unauthenticated connection can reserve 16 MiB and stall forever.**

After sending only a four-byte prefix, a client can make the daemon allocate up to `MAX_FRAME_SIZE` and then stop sending. Pipe instances and handler tasks are not globally bounded, and there is no read timeout, so repeated same-user connections can exhaust memory and also keep `active_clients` nonzero indefinitely.

**Fix:** bound concurrent clients and aggregate in-flight frame bytes, apply a read/idle timeout, and read large bodies incrementally instead of reserving the full declared size immediately.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+                policy_mapping::windows_path_to_container_path(working_directory)
```
MGudgin: `src/backends/wslc/common/src/container_steps.rs:305`

**Medium (correctness) - the documented in-container working directory is treated as a Windows host path.**

`ExecConfig::working_directory` is documented as a path inside the container, and the protocol round-trip test even uses `/work`. This code instead calls `windows_path_to_container_path`; a normal Linux path returns `None`, after which the setting is silently omitted and the command runs in the container default directory.

**Fix:** pass an already-container-native path directly, or explicitly define and validate a host-path contract. Do not silently discard a non-empty value; add an exec test using `/tmp` or `/work`.

src/backends/wslc/daemon/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+                idle_for += IDLE_POLL;
+                if idle_for >= IDLE_TIMEOUT {
+                    shutdown.notify_waiters();
```
MGudgin: `src/backends/wslc/daemon/src/main.rs:132`

**High (correctness/reliability) - the idle shutdown notification can be lost permanently.**

`Notify::notify_waiters()` does not retain a permit when no task is currently waiting. The control server awaits `shutdown.notified()` only inside its `select`; if the watchdog reaches this line while the server is executing the connected-client branch between select iterations, the notification is lost. The watchdog then returns, leaving the daemon alive indefinitely with no future idle checks. The separate `container_count` and `active_clients` observations also make the final-idle boundary race-prone.

**Fix:** use a cancellation primitive or watch channel with persistent state, or at minimum a retained `notify_one` permit, and add a deterministic test that triggers idle expiry while the accept branch is active.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -0,0 +1,302 @@
...
+async fn handle_client(mut pipe: NamedPipeServer, session: SessionHandle) -> Result<()> {
+    let request: DaemonRequest = read_frame(&mut pipe).await?;
+    match request {
+        DaemonRequest::Ping => {
+            write_frame(&mut pipe, &DaemonResponse::Pong).await?;
+        }
+        DaemonRequest::Provision(config) => {
+            let resp = match session.provision(config).await {
+                Ok(sandbox_id) => DaemonResponse::Provisioned { sandbox_id },
+                Err(e) => err_response(ErrKind::Backend, e),
```
MGudgin, line 364: `src/backends/wslc/daemon/src/control_server.rs:224`

**Medium (correctness/API contract) - every production failure is classified as `Backend`.**

The protocol defines `NotProvisioned`, `NotStarted`, `Busy`, `NotReady`, and `Protocol` specifically so the later state-aware client can map failures without parsing strings. Production dispatch constructs only `Backend`, even for explicit `unknown sandbox` and `not started` errors, and no single-flight path emits `Busy`. The advertised stable classification is therefore unusable before PR 2 consumes it.

**Fix:** introduce a typed worker error and map each state/protocol failure to its declared `ErrKind`; add dispatch tests that assert the wire classification, not just the message.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -0,0 +1,944 @@
...
+                return Err(sdk_error("WslcSetProcessSettingsCallbacks failed", hr, ""));
+            }
+            sdk_io_ref = SdkIoRef(Some(io_ctx_raw as *const IoContext));
+        }
+
+        // Command line: /bin/sh -c <script>. argv points into the sh/dash_c/
+        // script heaps; all are owned by the returned struct.
+        let sh = b"/bin/sh\0".to_vec();
+        let dash_c = b"-c\0".to_vec();
+        let script_cstr = format!("{}\0", script_code).into_bytes();
```
MGudgin: `src/backends/wslc/common/src/container_steps.rs:269`

**Medium (correctness/security) - interior NULs silently truncate values passed to the SDK.**

The protocol accepts arbitrary JSON strings, but command, environment, image, and container-path C strings are created by appending `\0`. An input containing `\u0000` is accepted in full by Rust and then observed only up to the first NUL by the SDK, so validation/logging and execution can disagree about the command or path.

**Fix:** construct every SDK C string with `CString::new` and reject interior NULs at the protocol boundary with a typed protocol error. Add round-trip rejection tests for command, environment, image, and mount fields.

MGudgin (changes requested):

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Will address in https://github.com/microsoft/mxc/issues/686

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Not addressing in this PR, will do a follow up, tracked in #794

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Addressed.

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Addressed.

src/backends/wslc/daemon/src/session_manager.rs
SohamDas2021: Planned in PR 2c/3.

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Planned in PR 2a/3.

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Filed as #795 to track this. 

Deferring the fix because its premise doesn't hold today: the daemon is spawned via a plain Command::spawn() that inherits the caller's token (spawn_daemon, DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP), so daemon and caller always run at the same integrity; **it's never elevated**, and there's no privilege gradient for the confused-deputy path. The mandatory-label/token-match + per-path  AccessCheck -under-impersonation work is a larger design change worth doing only once an elevated daemon variant exists (e.g. Entra/cloud-agent host).

The higher-value same-user vector — a hostile pipe server impersonating the phase process — is fixed in this PR (see G1): open with  SECURITY_SQOS_PRESENT | SECURITY_IDENTIFICATION  + authenticate the server's PID/creation-time against the discovery record.

src/backends/wslc/daemon/src/main.rs
SohamDas2021: Fixed.

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Fixed.

src/backends/wslc/common/src/daemon_client.rs
SohamDas2021: Fixed.

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Fixed.

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: The stall-forever half is closed: concurrent clients are bounded MAX_CONCURRENT_CLIENTS = 128 semaphore) and the full read_frame is under a 30s FIRST_FRAME_TIMEOUT, so a connect-then-stall drops in ≤30s and can't pin active_clients.

src/backends/wslc/common/src/container_steps.rs
MGudgin: Thanks — the 8 MiB per-stream cap addresses the original daemon-memory-exhaustion finding, so I no longer consider this blocking. One small correctness follow-up remains: if a callback fills the buffer to exactly the cap, a later callback returns before appending the truncation marker, making truncated output look complete. That residual marker bug can be handled separately.

src/backends/wslc/daemon/src/control_server.rs
MGudgin: Thanks — the 128-client semaphore and 30-second first-frame timeout address the unbounded/stall-forever portion of this finding. The eager 16 MiB allocation remains, so the theoretical aggregate is still about 2 GiB across 128 handlers, but given the same-user control plane and the other hardening in this PR, I no longer consider that residual risk blocking. Incremental reads or an aggregate in-flight-byte budget would be a worthwhile follow-up.

MGudgin (approved):
Re-verified at `0a43d4f6`. The blocking findings from my review are fixed here or have concrete follow-up coverage: the integrity-boundary design is tracked in #795 and typed error classification is implemented in #767. PR #801 documents the remaining serialized-exec limitation and intended redesign.

Local verification passed: 132 `wslc_common` tests, 5 daemon tests with 1 ignored, and the live named-pipe ping integration test. The two remaining resource-bound observations are non-blocking and noted in their threads. Approving.

PR #592, opened 2026-06-30
Migrate isolation_session backend + SDK to the IsolationSession Preview API

adpa-ms:
## 📖 Description

Migrates the experimental `isolation_session` backend (Rust backend + TypeScript SDK) from the in-development `Windows.AI.IsolationSession` WinRT surface to the **frozen `Windows.AI.IsolationSession.Preview`** namespace, reshaping every consumer to the reduced, stable API. This is a breaking change to the backend's experimental surface (it is gated behind `--experimental`).

Applied symmetrically across bindings → backend → domain → wire → schema → SDK → tests → docs:

- **Host-folder sharing** (`ShareFolderBatchAsync` + the protected-paths filter) is removed. Filesystem policy (`readwritePaths`/`readonlyPaths`/`deniedPaths`) is now **rejected at every phase** with `policy_validation` on both the one-shot and state-aware paths — the folder-sharing sink is gone entirely, eliminating the subtree-inheritance hazard.
- **Sizing profile** (`configurationId`) is removed from the config, wire types, and dev schema; an unknown `configurationId` is now silently ignored rather than rejected.
- **Entra cloud-agent support is retained** — the optional `user` bundle (`upn` + `wamToken`) is still accepted at provision and start. Only the call shape changes: the previous split between the v1 and v2 provisioning interfaces collapses into a single token-carrying `AddUserAsync`/`StartSessionAsync` pair, and the separate app-registration step is dropped. The service now assigns an opaque agent-user name at provision and validates the token itself, so the `sandboxId` tail is that opaque handle (no client-side UPN cross-check).
- **Availability gating** moves from a hardcoded OS-build pin in `platform.ts` to a runtime probe fact (`probes.isolationSessionAvailable` from `wxc-exec --probe`).

The Rust bindings are regenerated against the Preview metadata (provenance in `GENERATION_INFO.toml`), and a `build.rs` path bug is fixed so the "bindings must be regenerated" version-gate is actually live.

## 🔗 References

- IsolationSession backend specs updated in this PR: `docs/isolation-session/oneshot.md`, `docs/isolation-session/state-aware-rust.md`, `docs/isolation-session/state-aware-typescript.md`.

## 🔍 Validation

- **Local CI gates:** `cargo fmt --check`; clippy `--all-targets --all-features -D warnings`; build + unit tests with the `isolation_session` feature **on** (359 pass) and **off** (397 pass); `wxc_host_prep` (17, elevated); SDK unit tests; schema/SDK-type codegen; config validation — all green.
[...]

PR #761, opened 2026-08-06
Migrate the isolation_session backend and SDK to the IsolationSession Preview API

adpa-ms:
## 📖 Description

Migrates the `isolation_session` backend and its SDK surface to the `Windows.AI.IsolationSession.Preview` in-proc runtime API, and brings the backend's policy behaviour in line with what that API can actually deliver.

**Lifecycle shape.** The agent user account name is now OS-assigned rather than client-minted: provision returns it, along with the agent SID and an OS-provided `ephemeralWorkspacePath` — a directory shared between the calling user and the isolated agent user. Sizing profiles (`configurationId`) have no equivalent in the Preview API and are removed, leaving `provision` as the only phase that carries a per-phase config.

**Host folder sharing is gone.** The Preview API exposes no folder-sharing primitive, so `readwritePaths` / `readonlyPaths` / `deniedPaths` are refused at every phase rather than partially honoured; staging files in and out of a session goes through the ephemeral workspace instead. This also retires the emergency path filter added for the subtree-ACE regression, because the call it guarded no longer exists.

**Three policy sections move from "silently dropped" to "explicitly refused"**, so the backend never appears to offer a guarantee it cannot enforce.

- **`network`** — the container's network is unrestricted on both axes (outbound is open, and a process inside can listen on a localhost-reachable port), and MXC has no primitive to filter or deny it. Provision accepts **only** the canonical unrestricted-network acknowledgment (`defaultPolicy: allow` + `allowLocalNetwork: true`, no host rules, no proxy, default enforcement) and refuses everything else — including an **absent** policy, which would otherwise default to an unenforceable `block`. Post-provision phases reject any supplied network policy and inherit the posture fixed at provision.
- **`ui`** — an isolation session isolates the *host's* UI from contained code, but does not deny that code UI capabilities: window creation, GDI and the session's own clipboard all work inside it. No `ui` posture would be truthful, so a supplied `ui` is refused at every phase on both surfaces. Unlike `network` there is no acknowledgment form, because there is no value a caller could supply that would be honest — so there is nothing to accept. The refusal is **presence-based rather than value-based**: `UiPolicy`'s defaults are full lockdown, which makes an explicitly-supplied lockdown `ui` indistinguishable by value from an absent one. An omitted `ui` is accepted and applies no restriction — note this means the schema's default-deny reading does not hold on this backend.
- **`lifecycle`** — the in-proc API exposes no session-lifetime knob, so one-shot refuses `destroyOnExit: false` and `preservePolicy: true` (the default `destroyOnExit: true` matches actual behaviour and is accepted), and the state-aware parser refuses the section outright.

**`sandboxId` becomes opaque, structured and versioned** — `iso:<base64url-nopad(JSON)>` — carrying the OS-assigned agent user name plus an optional caller-supplied **`appId`** (the Package Family Name for a packaged application). `appId` is validated structurally and is carried, not consumed: it exists so that a future OS contract keyed on calling-application identity does not require a breaking id change. Ids minted by a newer MXC are reported as such rather than as corrupt. **Pre-migration ids are no longer addressable** — those were `iso:<token>`, carrying an opaque MXC-minted provisioning token rather than the agent user name, so no manual recovery path is being removed. An accepted consequence of the format change, since no sandbox survives an executor upgrade.

**State-aware failures carry structured fields** on the wire envelope — `operation`, `nativeCode`, `remediation` — as siblings of `code` and `message` rather than prose folded into a single string, so callers can branch without parsing text. The invariant is that `nativeCode` and `remediation` never appear without `operation`; a failure MXC raises before any API call carries neither. `0x80070490` maps to `stale_id` rather than a generic backend error because, on the lifecycle operations, it is raised only for an unknown agent user — every other failure mode on those paths, including a missing executable, surfaces a different HRESULT.

**Host support detection** moves from a hardcoded Windows build-number gate in the TypeScript SDK to a runtime probe (`wxc-exec --probe` → `probes.isolationSessionAvailable`). The old gate pinned an exact Insider build and would have reported "unsupported" on any newer one.

### Known gaps and deliberate exclusions

These were considered and consciously left out; they are not oversights.

- **The Entra enterprise `user` bundle (`{ upn, wamToken }`) is not carried here.**
- **The C# SDK (`sdk/dotnet/`) is not updated for the new wire contract.** It still emits `configurationId` and a `user` bundle for this backend, and its comments describe the pre-migration shape. No files under `sdk/dotnet/` are touched by this PR. Nothing breaks at runtime, because the C# path cannot reach an experimental backend today. Tracked separately.
- **Two pre-existing behaviours in the session manager were examined and deliberately left.** The `STILL_ACTIVE` (259) exit-code ambiguity in the graceful-shutdown wait, and the fact that a bounded join on the stdio relay threads is best-effort by construction — a relay blocked in a read cannot be interrupted, so no join can be guaranteed to complete. Relatedly, the relay handle lifetime is sound for a one-shot process but deserves revisiting for a long-lived state-aware process performing many `exec` calls, where handle values are recycled aggressively. That is a follow-up, not a regression introduced here.

## 🔗 References

No tracking issue. These commits were previously reviewed as PRs into the feature branch: #592, #683, #682, #708, #718, #746.

## 🔍 Validation

`fmt`, `clippy --all-targets --all-features -D warnings`, and build + test for x64 with `isolation_session` **ON and OFF**, plus an aarch64 build. The OFF configuration is covered deliberately: a change that only compiles with the feature enabled would otherwise reach `main` unnoticed.

Versioning and codegen gates all pass: schema versions, schema codegen, SDK wire-types codegen, config validation, version sync, and toolchain sync. `cargo fetch --locked` resolves the whole graph through the public feed across all five CI target triples.

SDK unit tests and the SDK integration suite pass. On a host with isolation-session runtime support: one-shot suite 16, state-aware lifecycle suite 62, SDK integration 12 — **90 passed, 0 failed, 0 skipped** — followed by a manual interactive pass covering TTY resize, streaming, and interactive PowerShell. A name-agnostic diff of local accounts taken before and after the run showed no leaked agent accounts.

## ✅ Checklist

## 📋 Issue Type

- [x] Feature

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/761)


jeffstall (dismissed):
:shipit:

docs/schema.md
```diff
@@ -144,6 +150,35 @@ containment tier selected at runtime:
...
+or supplied, so an omitted `ui` there is not lockdown. Check the backend's own
+documentation before relying on the default.
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `disable` | boolean | `true` | Disable all UI access. On Windows ProcessContainer this maps to the Win32k system-call disable mitigation, so the process cannot create windows, use GDI, or make `NtUser*` / `NtGdi*` calls. |
+| `clipboard` | enum | `"none"` | Clipboard access level: `"none"`, `"read"`, `"write"`, or `"all"`. |
+| `injection` | boolean | `false` | Whether the process may inject synthetic keyboard/mouse input (`SendInput` and friends). |
+
+**Per-backend support.** `ui` is enforced by the Windows ProcessContainer
```
shschaefer, line 168: We need to remove this from here and the process-container document.  We don't claim that it will ALWAYS be enforced by Job Object, etc...  The whole point of MXC is to separate policy from the underlying mechanism.  Not just ship more mechanisms.

shschaefer (approved):

PR #767, opened 2026-08-07
[WSLC] Harden state-aware daemon (PR 2a/3): typed errors, validate-then-admit exec, idle-watchdog

SohamDas2021:
## Summary
PR 1/3 of the WSLc state-aware split landed the per-user daemon + owner-only named-pipe IPC. This PR (call it 2a/3) hardens the daemon internals with **no public wire or schema surface**, so it lands independently. 

Four changes:

- **Typed worker errors** — `WorkerError{NotProvisioned, NotStarted, Backend}` with a `kind() -> ErrKind` mapping instead of collapsing every worker failure to `ErrKind::Backend`. The control server returns the real classification, so clients can distinguish an unknown/not-started sandbox from a backend fault without string-matching. `Display` preserves the existing `"unknown sandbox"` message.
- **Validate-then-admit exec** — the `Exec` worker command now carries two reply channels (`admit` + `done`). The worker calls `validate_exec()` (sandbox exists + started) and writes the `admit` decision *before* running, so an unknown/not-started sandbox comes back as a pre-admission typed `Err` rather than a post-admission stream `Error` frame. Validation and run share one worker handler with no yield between them, so admission is atomic against concurrent `Stop`/`Deprovision`.
- **Idle-watchdog hardening** — a monotonic `activity` counter bumped on each accepted connection; the watchdog declares idle only when the container count is zero, no request is in flight, and the activity generation is unchanged since the last poll (generation read last, so a connect that starts and finishes between two polls is not missed).
- **Dedup** — the one-shot runner now uses the shared `container_steps::sdk_error` instead of a byte-identical private copy.

## Testing

- Daemon unit tests (incl. new worker-error `kind` + `validate_exec` assertions) and the `#[ignore]`d WSL2-host lifecycle tests (in-proc **and** over the pipe): provision → start → exec → exec → stop → deprovision.
- `wslc_common` suite; full `run_wslc_all_tests.ps1` corpus **24/24**.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` clean.
- Reviewed with a 3-model pass; the one flagged finding (idle-watchdog atomic read ordering) is fixed.

## Coming in the pipeline

- **PR 2b/3 — state-aware backend + wire/schema + E2E:** `wslc/common/state_aware.rs` (`StatefulSandboxBackend`, prefix `wslc`) translating the public `experimental.wslc.*` wire schema into daemon protocol frames; `mxc_engine` state-aware arm + config-parser wiring; regenerated dev schema + generated TS wire types; multi-invocation E2E script with warm-reuse + idle-teardown assertions.
- **PR 3/3 — TypeScript SDK:** `Wslc*Config`/`*Result` types + branded `SandboxId<'wslc'>` and helper prefix wiring, mirroring the LXC state-aware SDK surface.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/767)

### DIFF
--- a/src/backends/wslc/daemon/src/control_server.rs
+++ b/src/backends/wslc/daemon/src/control_server.rs
@@ -17,7 +17,7 @@
 //! connect to the control plane.
 
 use std::ffi::c_void;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -41,10 +41,10 @@ use windows::Win32::Security::{
 use windows::Win32::System::Threading::{GetCurrentProcess, OpenProcessToken};
 
 use wslc_common::daemon_protocol::{
-    encode_frame, DaemonRequest, DaemonResponse, ErrKind, StreamFrame, MAX_FRAME_SIZE,
+    encode_frame, DaemonRequest, DaemonResponse, StreamFrame, MAX_FRAME_SIZE,
 };
 
-use crate::session_manager::SessionHandle;
+use crate::session_manager::{SessionHandle, WorkerError};
 
 /// Upper bound on concurrently-serviced client connections. At capacity the
 /// accept loop applies backpressure (a new connection waits for a slot) instead
@@ -85,13 +85,16 @@ pub fn bind(pipe_name: &str) -> Result<(OwnerOnlySecurity, NamedPipeServer)> {
 /// `active_clients` tracks in-flight requests so the idle watchdog does not tear
 /// the daemon down mid-request. Concurrency is bounded by a semaphore, and on
 /// shutdown all in-flight handlers are drained before returning so the caller
-/// can release the WSLc session without racing a live handler.
+/// can release the WSLc session without racing a live handler. `activity` is a
+/// monotonic connection counter the watchdog compares across polls to catch
+/// bursts that start and finish between two of its samples.
 pub async fn run(
     session: SessionHandle,
     pipe_name: String,
     security: OwnerOnlySecurity,
     first_instance: NamedPipeServer,
     active_clients: Arc<AtomicUsize>,
+    activity: Arc<AtomicU64>,
     shutdown: Arc<Notify>,
 ) -> Result<()> {
     let mut server = first_instance;
@@ -119,6 +122,10 @@ pub async fn run(
                     continue;
                 }
                 let connected = server;
+                // Record the connection so an idle streak that spans this
+                // request is invalidated even if it completes between polls.
+                activity.fetch_add(1, Ordering::SeqCst);
+
                 // Recreate the next listening instance before servicing this one
                 // so the next client is not refused. Transient failures are
                 // retried with bounded backoff rather than tearing the whole
@@ -361,7 +368,7 @@ async fn handle_client(mut pipe: NamedPipeServer, session: SessionHandle) -> Res
         DaemonRequest::Provision(config) => {
             let resp = match session.provision(config).await {
                 Ok(sandbox_id) => DaemonResponse::Provisioned { sandbox_id },
-                Err(e) => err_response(ErrKind::Backend, e),
+                Err(e) => worker_err_response(e),
             };
             write_frame(&mut pipe, &resp).await?;
         }
@@ -384,7 +391,15 @@ async fn handle_client(mut pipe: NamedPipeServer, session: SessionHandle) -> Res
     Ok(())
 }
 
-/// Exec: admit with `Ok`, then stream the run's outcome as [`StreamFrame`]s.
+/// Exec: validate-then-admit, then stream the run's outcome as [`StreamFrame`]s.
+///
+/// The sandbox is validated (exists + started) *before* the `Ok` admission is
+/// written, and — critically — admission is **atomic** with the start of the
+/// run on the worker thread (see [`SessionHandle::exec`]): the worker validates
+/// and begins running within one command handler, so no `Stop`/`Deprovision`
+/// can invalidate the checked state between the admission and the run. An
+/// unknown/not-started sandbox therefore comes back as a pre-admission typed
+/// [`DaemonResponse::Err`] rather than a post-admission stream `Error` frame.
 ///
 /// TODO(fill-in): bidirectional live stdio (client `Stdin` frames -> process,
 /// process stdout/stderr -> `Stdout`/`Stderr` frames). The skeleton runs the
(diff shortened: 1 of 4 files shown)


src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -394,29 +409,42 @@ async fn handle_exec(
     session: SessionHandle,
     config: wslc_common::daemon_protocol::ExecConfig,
 ) -> Result<()> {
+    // Await the worker's admission decision before writing anything: a rejected
+    // exec is a pre-admission typed error, never a post-admission stream frame.
+    let done = match session.exec(config).await {
```
MGudgin: **Medium (reliability) - an exec can start before the client receives the admission response.**

This ordering is introduced here: `session.exec(config).await` returns once the worker sends `admit`, and that worker immediately enters the blocking execution. If the subsequent `DaemonResponse::Ok` write fails because the client disconnected, the completion receiver is dropped but the command continues with real side effects. A retry can therefore execute a non-idempotent script twice, with no record of the orphaned result.

**Fix:** Add an acknowledged start phase so execution begins only after the admission frame is written, or provide cancellation plus explicit logging/result capture when that write fails.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -467,8 +541,23 @@ pub fn spawn() -> Result<SessionHandle> {
...
+                    } => {
+                        // Validate and run in one handler so admission is atomic
+                        // with the start of the run: the worker never yields
+                        // between the two, so no Stop/Deprovision can interleave.
+                        match worker.validate_exec(&config.sandbox_id) {
+                            Err(e) => {
+                                let _ = admit.send(Err(e));
+                            }
+                            Ok(()) => {
+                                let _ = admit.send(Ok(()));
```
MGudgin: **Medium (reliability/maintainability) - skip execution when the admission receiver is already gone.**

`admit.send(Ok(()))` can fail when the handler was cancelled, but the result is ignored and the worker always enters `worker.exec(config)`. Because exec blocks the single worker, an abandoned request can starve every lifecycle operation for its full timeout. The blocking worker itself predates this PR; the new unchecked admission channel is what makes dead work reachable.

**Fix:** Gate execution on a successful send, for example `if admit.send(Ok(())).is_ok() { let _ = done.send(worker.exec(config)); }`.

src/backends/wslc/daemon/src/main.rs
```diff
@@ -154,40 +161,62 @@ async fn run() -> Result<()> {
...
             }
+        };
+        // Read the monotonic generation last — after count and active_clients —
+        // so a client that connects and completes entirely within this poll
+        // (bumping `activity`, then dropping `active_clients` back to zero) is
+        // still observed here as `generation != last_activity`, instead of
+        // slipping through with both zero-reads while the generation bump goes
+        // unsampled.
+        let active = active_clients.load(Ordering::SeqCst);
+        let generation = activity.load(Ordering::SeqCst);
```
MGudgin, line 224: **Medium (correctness) - the generation check still leaves a post-sample shutdown race.**

The counter closes the between-polls window, but a client can connect after this final generation load and before the accept loop consumes the shutdown permit. With both branches ready, the unbiased `select!` may accept and fully service a `Provision`; shutdown then removes the discovery record and deletes the newly returned container. The client receives a sandbox ID that immediately becomes invalid.

**Fix:** Enter an authoritative draining state before accepting more work, prioritize shutdown and reject new requests, or re-check activity before teardown. Add a paused-time regression that connects after the final sample.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -301,13 +363,22 @@ impl Worker {
         Ok(())
     }
 
-    fn exec(&mut self, config: ExecConfig) -> Result<i32> {
+    /// Validate that a sandbox exists and is started, without touching the SDK.
+    fn validate_exec(&self, sandbox_id: &str) -> Result<(), WorkerError> {
+        match self.containers.get(sandbox_id) {
+            None => Err(WorkerError::NotProvisioned(sandbox_id.to_string())),
+            Some(entry) if !entry.started => Err(WorkerError::NotStarted(sandbox_id.to_string())),
```
MGudgin, line 373: **Medium (testability) - the new `NotStarted` admission contract has no SDK-free regression path.**

The PR adds two tests, both for unknown/`NotProvisioned` sandboxes. Reaching this branch requires a real `WslcContainerGuard` inserted by provisioning, so unit tests cannot construct a provisioned-but-not-started entry; even the ignored host lifecycle starts before exec. `worker.exec` then repeats the same existence/start checks immediately after this non-yielding validation, leaving two owners of the invariant.

**Fix:** Add a test fixture or container-handle abstraction that can create a provisioned/not-started entry, test `ErrKind::NotStarted`, and make either validation or execution the single owner of these checks.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -394,29 +409,42 @@ async fn handle_exec(
...
+            write_frame(&mut pipe, &worker_err_response(e)).await?;
+            return Ok(());
+        }
+    };
     write_frame(&mut pipe, &DaemonResponse::Ok).await?;
-    let terminal = match session.exec(config).await {
-        Ok(code) => StreamFrame::Exit { code },
-        Err(e) => StreamFrame::Error {
-            message: format!("{e:#}"),
+    let terminal = match done.await {
```
MGudgin, line 451: **Medium (testability) - the new completion-channel failure path has no transport-level test.**

This file remains at zero tests, while the existing two daemon IPC tests remain happy-path only. Nothing verifies that dropping `done` after admission produces exactly one terminal `StreamFrame::Error` rather than a hang or malformed stream. Draft #801's lifecycle E2E also covers only normal completion.

**Fix:** Extract the handler over `AsyncRead + AsyncWrite` for an in-memory duplex test, or add a named-pipe integration test that admits exec, drops the completion sender, and asserts the terminal error frame.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -394,29 +409,42 @@ async fn handle_exec(
...
     match result {
         Ok(()) => DaemonResponse::Ok,
-        Err(e) => err_response(ErrKind::Backend, e),
+        Err(e) => worker_err_response(e),
     }
 }
 
-fn err_response(kind: ErrKind, e: anyhow::Error) -> DaemonResponse {
+/// Build a [`DaemonResponse::Err`] carrying the worker error's protocol `kind`.
+fn worker_err_response(e: WorkerError) -> DaemonResponse {
```
MGudgin, line 473: **Medium (test coverage) - round-trip the new typed error response through the actual pipe.**

The unit tests validate `WorkerError::kind()`, but no IPC test proves that `DaemonResponse::Err { kind, message }` survives frame encoding, named-pipe transport, decoding, and client handling. Draft #801 tests client-side mappings by constructing error values directly, so it does not cover this boundary.

**Fix:** Send an exec request for an unknown sandbox through the daemon IPC test and assert the client receives `ErrKind::NotProvisioned` with the expected message.

MGudgin (changes requested):

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Fixed here- if the DaemonResponse::Ok / completion-channel write fails after admission, the worker now logs the orphaned result instead of silently discarding it.

Deferred and planned in PR 2c: the full acknowledged-start phase- begin execution only after the admission frame is written. In 2a's design admission and run-start are deliberately atomic on the single worker thread, which is what closes the Stop/Deprovision interleave window. Decoupling them to gate on the write reintroduces that window, so a clean fix belongs with the streaming/cancellation rework in 2c where the frame protocol already needs revisiting.

MGudgin (approved):

MGudgin (approved):

PR #774, opened 2026-08-07
Add capture ETL retention option

### LINKED ISSUE #772: Add captureDenials option to retain analyzed ETL traces
### Description of the task

Add an optional boolean to `processContainer.captureDenials` that controls whether MXC deletes or preserves the sealed ETL trace after analysis.

Proposed API:

```json
{
  "processContainer": {
    "captureDenials": {
      "retainEtl": true
    }
  }
}
```

Expected behavior:

- `retainEtl` defaults to `false`, preserving the current behavior of deleting the internal ETL after analysis.
- When `true`, keep the sealed ETL for diagnostics and report its actual path through capture output metadata and the CLI output pointer.
- Apply the option consistently to all capture strategies, including native V2 capture and legacy WPR fallback.
- Preserve the ETL when analysis fails if retention was requested, so the trace remains available for troubleshooting.
- Continue treating cleanup failures as explicit errors when retention is disabled.
- Add wire/domain parsing, generated schema and SDK wire-type updates, documentation, and tests for default deletion, requested retention, analysis failure, and unique concurrent trace paths.

### Additional context

The current BaseContainer capture flow always calls `remove_internal_capture_file` after decoding the trace. Retained ETLs may contain sensitive resource paths or identifiers, so documentation should note that callers are responsible for securing and deleting retained traces.

richiemsft:
## 📖 Description

Adds `processContainer.captureDenials.retainEtl`, an opt-in diagnostic setting that preserves a sealed denial-capture ETL after a terminal wait.

- `retainEtl` defaults to `false`; the runner uses a protected per-run system-temp directory and removes the ETL and directory after analysis.
- Retention-enabled captures start in `%LOCALAPPDATA%\Microsoft\MXC\capture-denials\working\<run-id>` and move to a protected `retained\<run-id>` directory only after the trace seals successfully.
- Per-run directories use protected owner+SYSTEM DACLs and unpredictable 128-bit suffixes.
- Abandoned sandbox handles stop capture without sealing, analysis, retained ETL, or managed JSON output because no caller can observe output metadata.
- Successful capture metadata is exposed as `captureDenials`; post-seal retention/finalization failures are exposed as `captureDenialsError { message, etlPath }`. Both may be present when denial analysis succeeds but retained-storage promotion fails.
- Seal failures clean internal storage and do not advertise a retained path. Analysis failures after sealing report the known protected ETL path without probing the filesystem.
- Rust SDK/FFI/C# consumers receive the structured metadata programmatically; the CLI emits the capture pointer on stderr.
- Callers own retained ETL cleanup, including removal of the empty per-run parent directory.

## 🔗 References

Closes #772

## 🔍 Validation

- Rust unit tests for `wxc_common` and `appcontainer_common`.
- Rust format and Clippy with warnings denied for affected crates and E2E tests.
- Host-gated timeout-plus-retention E2E test compilation.
- Node wire-conformance tests and focused C# metadata tests.
- Schema codegen, SDK wire codegen, schema-version synchronization, and 198 config validations.
- Windows build 26666 VM lifecycle matrix:
  - retained capture creates JSON and a protected ETL under `retained`;
  - default capture omits `etlPath` and leaves no temporary directory;
  - an actual 1-second timeout retains JSON/ETL metadata while reporting timeout;
  - FFI handle abandonment creates no JSON and does not increase working or retained directory counts.
- Ten-axis adversarial review covering security, reliability, performance, maintainability, correctness, testability, coverage, platform parity, documentation, and simplicity.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task


richiemsft:

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2380,22 +2389,58 @@ impl BaseContainerSandboxProcess {
...
+    retain_etl: bool,
+) -> std::io::Result<CaptureDenialsOutput> {
+    if retain_etl {
+        let Some(etl_path) = etl_path else {
+            return capture_result;
+        };
+        let retained_path = etl_path.to_string_lossy().into_owned();
+        return capture_result
+            .map(|mut output| {
+                output.etl_path = Some(retained_path);
```
MGudgin, line 2532: **Blocking (#1, security) - Do not orphan a retained ETL when the process is dropped.**

This stores the retained path only in `self.output_metadata`. `Drop::drop` calls `run_teardown()` and then destroys the process object, so a successful drop-time teardown can preserve the sensitive ETL while discarding the only structured path to it. This affects early returns, panic unwinds, `kill()` followed by drop, and `mxc_sandbox_free`; before this PR the same path deleted the ETL.

**Fix:** Define an observable drop-time contract: either delete the ETL when no terminal `wait()` can return its metadata, or emit/persist the retained path through a channel the owner can consume and clean up.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2380,22 +2389,58 @@ impl BaseContainerSandboxProcess {
...
         )
     }
 }
 
+fn finalize_capture_result(
+    capture_result: std::io::Result<CaptureDenialsOutput>,
+    etl_path: Option<&Path>,
+    retain_etl: bool,
+) -> std::io::Result<CaptureDenialsOutput> {
+    if retain_etl {
```
MGudgin, line 2525: **Blocking (#2, security) - Protect retained traces as long-lived sensitive artifacts.**

This branch converts the runner-managed ETL from a short-lived internal file into a persistent artifact. `managed_capture_output_path()` still places it directly under `std::env::temp_dir()` with no explicit owner-only ACL, protected per-run directory, expiry, or cleanup policy. In service contexts that can be a shared temp root, while the trace contains sensitive resource paths and identifiers.

**Fix:** Put retained traces in a per-run directory with an explicit owner-only DACL, or require/use a caller-selected protected destination and document the lifetime contract.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2380,22 +2389,58 @@ impl BaseContainerSandboxProcess {
...
+            return capture_result;
+        };
+        let retained_path = etl_path.to_string_lossy().into_owned();
+        return capture_result
+            .map(|mut output| {
+                output.etl_path = Some(retained_path);
+                output
+            })
+            .map_err(|error| {
+                if etl_path.exists() {
```
MGudgin: **Blocking (#3; also #8/#13, reliability/security/testability) - Do not probe the filesystem to decide whether to report the recovery path.**

`Path::exists()` adds another unbounded filesystem operation precisely after capture has already failed. It can delay teardown, and it returns `false` for metadata errors such as access denied, causing MXC to preserve the ETL but suppress its location. It also forces tests to create a real file merely to select an error-formatting branch.

**Fix:** Report the known path unconditionally, or use `try_exists()` only to distinguish `NotFound` while still reporting the path on metadata errors. Prefer carrying the retained path in structured error metadata.

src/core/wxc_common/src/sandbox_process.rs
```diff
@@ -432,13 +432,23 @@ impl<B: SandboxBackend> ScriptRunner for Runner<B> {
...
-                error_message: format!("script timed out after {}ms", request.script_timeout),
-                output_metadata: child.output_metadata().cloned().map(Box::new),
-                failure_phase: FailurePhase::Timeout,
-                ..Default::default()
-            },
+            Err(e) if e.kind() == std::io::ErrorKind::TimedOut => {
+                let timeout_message =
+                    format!("script timed out after {}ms", request.script_timeout);
+                let detail = e.to_string();
+                let error_message = if detail.starts_with(&timeout_message) {
```
MGudgin: **Medium (#4; also #12, maintainability/testability) - Do not parse a cross-module error contract from display text.**

The backend formats `"script timed out after {}ms"` and appends teardown detail; this layer independently reconstructs the same text and uses `starts_with` to detect it. A wording or unit change in either module silently produces doubled/garbled messages, and tests can only assert magic substrings.

**Fix:** Preserve timeout and teardown failures as structured variants/fields until the final user-facing rendering.

src/core/wxc_common/src/sandbox_process.rs
```diff
@@ -432,13 +432,23 @@ impl<B: SandboxBackend> ScriptRunner for Runner<B> {
...
+                    detail
+                } else {
+                    format!("{timeout_message}; {detail}")
+                };
+                ScriptResponse {
+                    exit_code: -1,
+                    error_message,
+                    output_metadata: child.output_metadata().cloned().map(Box::new),
```
MGudgin: **Medium (#6, test coverage) - Add an end-to-end timeout-plus-retention test.**

The helper tests cover retention and combined errors separately, but no test exercises timeout -> teardown -> `output_metadata` -> SDK consumer. A regression here could return `TimedOut` while silently dropping `etlPath`.

**Fix:** Add a host-gated test using a short timeout and `retainEtl=true`; assert the timeout response contains `CaptureDenials.EtlPath` and that the reported file exists.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2380,22 +2389,58 @@ impl BaseContainerSandboxProcess {
...
+    etl_path: Option<&Path>,
+    retain_etl: bool,
+) -> std::io::Result<CaptureDenialsOutput> {
+    if retain_etl {
+        let Some(etl_path) = etl_path else {
+            return capture_result;
+        };
+        let retained_path = etl_path.to_string_lossy().into_owned();
+        return capture_result
+            .map(|mut output| {
```
MGudgin, line 2531: **Medium (#7, test coverage) - Lock in the drop-time retention contract.**

The new tests call finalization helpers directly, but none lets a `BaseContainerSandboxProcess` fall out of scope with retention enabled. That leaves the path-loss behavior from #1 untested and makes either deletion or reporting easy to regress.

**Fix:** Add a lifecycle test for abandoned/drop-owned teardown and assert the chosen contract on both successful analysis and analysis failure.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2380,22 +2389,58 @@ impl BaseContainerSandboxProcess {
...
+fn finalize_capture_result(
+    capture_result: std::io::Result<CaptureDenialsOutput>,
+    etl_path: Option<&Path>,
+    retain_etl: bool,
+) -> std::io::Result<CaptureDenialsOutput> {
+    if retain_etl {
+        let Some(etl_path) = etl_path else {
+            return capture_result;
+        };
+        let retained_path = etl_path.to_string_lossy().into_owned();
```
MGudgin, line 2529: **Low (#9, performance) - Avoid allocating the retained path before the result branch is known.**

`to_string_lossy().into_owned()` runs before `capture_result` is mapped, so analysis failures allocate a string that the success closure never consumes.

**Fix:** Move the owned conversion into the success closure, or reuse one structured path value for both success and error reporting.

src/core/wxc_common/src/sandbox_process.rs
```diff
@@ -432,13 +432,23 @@ impl<B: SandboxBackend> ScriptRunner for Runner<B> {
...
             }
-            Err(e) if e.kind() == std::io::ErrorKind::TimedOut => ScriptResponse {
-                exit_code: -1,
-                error_message: format!("script timed out after {}ms", request.script_timeout),
-                output_metadata: child.output_metadata().cloned().map(Box::new),
-                failure_phase: FailurePhase::Timeout,
-                ..Default::default()
-            },
+            Err(e) if e.kind() == std::io::ErrorKind::TimedOut => {
+                let timeout_message =
```
MGudgin: **Low (#10, performance) - Avoid constructing a timeout string solely to reparse it.**

This allocation is immediately discarded when `detail` already begins with the same text. More importantly, it is part of the stringly typed coupling described in #4.

**Fix:** Prefer the structured-error fix from #4; as a local improvement, format only when constructing the final fallback message.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2271,39 +2279,40 @@ impl BaseContainerSandboxProcess {
...
-                    _ => combine_capture_and_cleanup_results(
+                    (Some(etl), Some(output)) => Self::decode_write_and_finalize(
+                        &EtlDenialAnalyzer,
+                        etl,
+                        output,
+                        exit_code,
+                        self.retain_capture_etl,
+                    )
+                    .map(Some),
+                    _ => finalize_capture_result(
```
MGudgin: **Low (#11, maintainability) - Centralize capture finalization after `session.finish`.**

The success-with-paths, missing-paths, and finish-error branches now each invoke `finalize_capture_result(...).map(Some)`. That spreads the retention/cleanup transition across multiple branches and makes future failure-path changes easy to apply inconsistently.

**Fix:** Normalize `session.finish()` into one `Result<CaptureDenialsOutput>` and call `finalize_capture_result` once.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcSandboxTests.cs
```diff
@@ -66,4 +66,24 @@ public void SandboxPolicy_SerializesToCamelCaseJson()
         Assert.Equal("read", root.GetProperty("ui").GetProperty("clipboard").GetString());
         Assert.True(root.GetProperty("ui").GetProperty("allowWindows").GetBoolean());
     }
+
+    [Fact]
+    public void CaptureDenialsOutput_DeserializesRetainedEtlPath()
```
MGudgin, line 71: **Low (#14, test coverage) - Exercise the actual Rust/FFI/C# metadata round trip.**

This verifies deserialization from a hardcoded JSON string, but not Rust serialization, FFI transport, and managed deserialization together.

**Fix:** Extend the host-gated retention test through the C# SDK and assert `EtlPath` is populated from a real retained capture.

docs/schema.md
```diff
@@ -77,10 +77,12 @@ production configs and the dev schema when working on experimental features:
...
-            "outputPath": "C:\\logs\\denials.json" // JSON denials file the app reads. The parent
-        }                                  // dir must already exist; a unique per-run id is stamped
-                                           // into the stem (denials.<run-id>.json) and the actual
-                                           // path printed on stderr. Omit outputPath for a managed temp file.
+            "outputPath": "C:\\logs\\denials.json", // JSON denials file the app reads. Parent dir
+                                           // must exist; a unique per-run id is stamped into the
+                                           // stem and the actual path is printed on stderr.
+            "retainEtl": false             // Keep the sealed ETL after analysis and report its
+                                           // path in output metadata. Defaults to false.
+        }                                  // Omit outputPath for a managed temp file.
```
MGudgin: **Low (#15, documentation) - Keep the managed-temp note attached to `outputPath`.**

After adding `retainEtl`, `// Omit outputPath for a managed temp file.` is now attached to the closing brace, which makes it read like a note about the whole block rather than the field.

**Fix:** Move the note into the `outputPath` comment block before `retainEtl`.

MGudgin (changes requested):
## Summary

Requesting changes for findings #1-#3. The remaining comments are non-blocking but should be considered before merge. Finding #5 was intentionally omitted as requested.

Attribution and anchor verification:

- Re-fetched PR #774 at `a389495` and verified its merge-base is exactly the PR base, `331fbee`.
- The local three-dot diff and `gh pr diff` differ only by line-ending bytes.
- Every inline comment below is anchored to an added line in the live PR diff.
- Every cited file is modified by this PR; no surviving finding is being charged to unchanged base code.
- The PR increases the BaseContainer Rust tests from 54 to 58 and the C# file from 5 to 6 tests. The new tests cover parsing, helper-level retention, and C# JSON deserialization, but searches found no timeout-plus-retention or drop-plus-retention lifecycle test.

To avoid duplicate threads, one inline comment covers #3/#8/#13, and another covers #4/#12.

### Finding without a safe inline anchor

**#16 Low (documentation) - Keep the domain and wire documentation consistent.** `src/core/wxc_common/src/models.rs:570` removes the parenthetical explaining that `output_path` is separate from the runner-managed intermediate ETL, while `wire.rs` retains that explanation. The defect is in removed text rather than an added line, so it is recorded here instead of forcing an invalid inline anchor.

**Fix:** Restore the intermediate-ETL explanation on `CaptureDenialsConfig::output_path`.

None. The pre-existing Drop implementation and managed temp-path helper become problematic only because this PR newly allows the ETL to survive teardown.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed in 7f9df574. Drop now uses the abandonment path (`run_teardown(false)` -> `session.finish(None)`), so it does not seal, analyze, retain an ETL, or create managed JSON. The FFI abandonment VM case confirms no working/retained directory count increase and no output file.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed in 7f9df574. Internal traces now use unique per-run directories with protected owner+SYSTEM DACLs. Retention-enabled traces use `%LOCALAPPDATA%\\Microsoft\\MXC\\capture-denials\\working` and are promoted to the protected `retained` root after sealing; default traces use protected per-run system-temp storage. The lifetime and caller cleanup contract is documented.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed in 7f9df574. Failure reporting no longer probes `Path::exists()`. After a successful seal the known protected path is carried directly in structured `captureDenialsError` metadata; seal failure cleans storage and intentionally emits no retained path.

src/core/wxc_common/src/sandbox_process.rs
richiemsft: Addressed in 7f9df574. Shared timeout handling no longer constructs and reparses display text; it emits the canonical timeout message from the structured `TimedOut` kind. Capture finalization details and recovery paths are carried through `SandboxOutputMetadata`.

src/core/wxc_common/src/sandbox_process.rs
richiemsft: Addressed in 7f9df574. Added the host-gated `test_processcontainer_capture_denials_timeout_retention` E2E test, which forces an actual timeout and asserts both the denials JSON and retained ETL exist. The same flow passed on Windows build 26666 with a 1-second timeout.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed in 7f9df574. `abandoned_capture_discards_without_sealing_output` locks in the `finish(None)` contract, and the native FFI VM harness frees a live sandbox without waiting and verifies no JSON or working/retained artifact is added.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed in 7f9df574. The managed ETL path remains a single structured `ManagedCapturePath` lifecycle value and string conversion occurs only when constructing success or failure metadata.

src/core/wxc_common/src/sandbox_process.rs
richiemsft: Addressed in 7f9df574. The timeout arm no longer creates a string solely to compare or reparse it; it renders the canonical timeout message directly from `request.script_timeout`.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Addressed with an explicit lifecycle split in 7f9df574. Successful sealing flows through the common decode/finalize helper, while seal failure intentionally remains separate because no sealed ETL is guaranteed and storage must be cleaned without advertising a recovery path. This preserves truthful failure semantics rather than normalizing unlike states.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcSandboxTests.cs
richiemsft: Covered at the available API boundaries in 7f9df574. The C# SDK does not expose `captureDenials` policy authoring, so a real C#-initiated retained capture would require a separate public API expansion. Rust/CLI E2E verifies real native metadata production, FFI VM validation verifies lifecycle transport, and focused C# tests verify managed deserialization of both success and failure metadata.

docs/schema.md
richiemsft: Addressed in 7f9df574. The managed-output note is attached directly to `outputPath`, separately from the `retainEtl` documentation.

MGudgin (approved):
Verified the update against the prior review findings. All blocking findings are addressed.

Validation reviewed:

* Rust format passed.
* 814 targeted Rust tests passed.
* 7 focused C# tests passed.
* All 40 PR checks are green.
* Timeout-retention and abandonment lifecycle behavior was validated in the PR's Windows VM matrix; local E2E execution was limited by this host selecting the unsupported AppContainer fallback tier.

The remaining observations are non-blocking.

PR #775, opened 2026-08-07

bbonaby:
> [!IMPORTANT]
> This PR replaces #738 using the same phase8a implementation, with the original Gudge and Copilot attribution preserved in the feature commit. #738 was mistakenly rebased and pushed without its original phase8a changes while attempting to fix merge conflicts, so GitHub immediately marked the empty PR as merged in the original stack; this restores the full change and adds it back to the stack. PR description also taken from that PR.

## 📖 Description

> **Stacked on #732.** Base is `user/gudge/versioning_phase7d_dev_schema_gate`; review the top commit only.

## Summary

This PR adds per-field schema version availability: a wire field can declare the range of config schema versions it is valid in, and the parser rejects any use outside that range.

The governing requirement is that a breaking config-schema change must be possible *without* dropping support for an earlier version, where "support" means shape only — an older config keeps parsing and is enforced with today's semantics. #732 blocks the dev schema from accepting less than it did, which forces breaking changes to be additive; this PR supplies the other half. Without parse-time enforcement, an availability annotation is documentation that nothing honours.

## Details

* New `mxc_version_derive` proc-macro crate. `#[derive(VersionAvailability)]` lifts `#[mxc_version(since = "0.8")]` / `until` off `wire.rs` into metadata normal builds carry, so one declaration feeds both the parser and schema generation (published as `x-mxc-since` / `x-mxc-until`). It has to be a derive: `#[schemars(extend(...))]` sits behind the `schema-gen` feature, which only `mxc_schema_gen` enables, so it can annotate the schema but can never be consulted by the parser.
* The mechanism fails **open** if a derived JSON name ever disagrees with what serde accepts — the range simply never fires — so that case is guarded twice. The macro compile-errors on every serde construct it cannot model exactly (`flatten`, split `rename` / `rename_all`, unrecognised `rename_all` rules, data-carrying variants, malformed literals), and a conformance test cross-checks all 32 wire types against the property names `schemars` derives independently from the same attributes.
* Enforcement runs immediately after deserialisation in every entry point, because `convert_wire_config` moves fields out of the config and the document can no longer be checked as a whole. State-aware requests are gated on the original document, not the experimental-masked copy.
* `version` is now **required**. It selects the legal field surface, so an absent one would silently opt out of every range rather than defaulting to something safe.
* New `version_incompatible` error code across all five coupled surfaces (Rust `MxcErrorCode`, engine `ErrorCode`, TS, C#, `MXC_STATUS_VERSION_INCOMPATIBLE = 13`) carrying structured `details: { field, declaredVersion, since, until }`. The supported-range error migrates onto it, so one code covers both classes.
* Three annotations, each checked against real corpus usage before landing: `seatbelt` since 0.7, `processContainer.captureDenials` and `processContainer.learningMode` since 0.8.
* New `check-version-availability.js` oracle gate derives each field's true first appearance from the frozen 0.6/0.7 and dev schemas and fails when a declared `since` disagrees.
* Corpus and callers migrated: 61 previously-unversioned configs, ~200 Rust test literals, the PowerShell lifecycle helpers (stamped centrally in each `Invoke-StateAware*` helper rather than at ~20 call sites), and the SDK config builders.

## Tests

* `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, and the per-package test suites all clean on the rebased tip.
* New corpus test asserts all 195 configs declare a version and still parse, with the out-of-range fixture pinned as a negative case by exact code and bounds.
* Feature-gated builds covering every flag this diff can reach: `wxc_common` {schema-gen, microvm}, `mxc_ffi` {dotnetsdk}, `mxc_engine` {isolation_session}, `wxc` {isolation_session, microvm, tier2_bfs, wslc, hyperlight}.
* Versioning gate suite 94 → 120 tests (26 of those arrive with #730's hardening). Node SDK build + 223 tests; C# SDK 35 tests; ErrorCode parity 17 codes; bindings codegen OK. All 11 CI gates pass.
* Non-regression: the #676 replay still yields exactly 6 findings; detector baselines hold (dev vs dev = 0, 0.6→0.7 = 6, 0.7→dev = 12); #732's gate passes; `SUPPORTED_VERSION` unchanged at `>=0.6, <=0.8`.
* Converged through a 2-round adversarial review (14 findings: 12 fixed, 1 pushback accepted, 1 pre-existing and filed separately).

## Notes for reviewers

**Annotation is opt-in, and "unannotated" means *no claim*, not "since 0.6".** An unannotated field is unbounded, which today is indistinguishable from `since: 0.6, until: 0.8` because the supported-range check already rejects everything outside that window — but it becomes observable when the range moves, and unbounded is what lets fields keep working as a new dev line opens.

**The most important part of this PR is what is deliberately *not* annotated.** A field's first appearance in the JSON Schema is only a *lower bound* on how long it has been accepted:

* `experimental` declared no properties before 0.8, so anything under it validated vacuously and has always been accepted.
* State-aware requests declare `0.6.0-alpha` while carrying `phase` / `sandboxId` / `correlationVector`, which the schema only described from 0.8.

Deriving bounds from schema data alone would have approved `since: 0.8` on `phase` and rejected every state-aware request ever sent. Measured: 66 properties are unannotated yet absent from the 0.6 schema — 8 are covered transitively by an annotated ancestor (the walker checks a field's range before descending, so a rejected parent is never traversed into), and the rest legitimately carry none. The oracle gate is therefore **fail-closed** on those surfaces: it refuses a declaration it cannot justify rather than checking it against a bound that would be wrong. `corpus_parses.rs` is the behavioural counter-check the oracle structurally cannot provide.

**Put the range on the containing field, never inside a shared struct.** `Seatbelt` is reachable from both the top-level `seatbelt` section and `experimental.seatbelt` — it is one node, so a range on its inner fields would leak onto the unconstrained experimental surface. The oracle gate catches this class automatically.

**One deliberate observable break:** migrating the supported-range error onto `version_incompatible` changes an existing error's shape, so a consumer string-matching the old "older/newer than supported" message is affected. This was flagged and accepted in review.

**Not executed on this host** (Windows): the macOS Seatbelt paths, the Windows Sandbox and IsolationSession PowerShell lifecycle suites, and the host-gated MicroVM / Hyperlight E2E configs. The macOS code does cross-compile — `cargo check --target aarch64-apple-darwin --all-targets` is clean for `mxc_engine`, `wxc_common` and `mxc-sdk`, including the new `cfg(target_os = "macos")` regression tests — but it has not been run. `mxc_darwin` cannot be cross-checked at all, for the pre-existing reason in #735.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/738)
## 🔗 References

- Replaces #738
- Stacked on #732

## 🔍 Validation

- `cargo fmt --manifest-path src/Cargo.toml --all -- --check`
- `cargo test --manifest-path src/Cargo.toml -p mxc_version_derive -p wxc_common -p mxc_engine`
- `npm test --prefix scripts/versioning`
- Schema, SDK wire-type, and version-availability codegen gates

## ✅ Checklist

- [ ] Linked to an issue
- [ ] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes (pending CI)

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/775)


scripts/versioning/lib/version-availability.js
```diff
@@ -0,0 +1,254 @@
...
+ * `timeline` is ordered oldest-first: `[{ label, version, paths }]`.
+ */
+function checkAvailability({ declared, timeline, compareMajorMinor }) {
+  const errors = [];
+  let checked = 0;
+
+  const firstAppearance = (path) => timeline.find((t) => t.paths.has(path));
+  const atLabel = (label) => timeline.find((t) => t.label === label);
+
+  for (const record of declared) {
```
bbonaby, line 148: Originally posted by @SohamDas2021 on #738 ([original comment](https://github.com/microsoft/mxc/pull/738#discussion_r3732041224)):

> checkAvailability only iterates declared (records that already carry an x-mxc-since / x-mxc-until). A brand-new field added after the floor with no annotation produces no record, so it is never checked?

scripts/versioning/lib/version-availability.js
```diff
@@ -0,0 +1,254 @@
...
+    if (record.until !== undefined) {
+      const at = atLabel(record.until);
+      if (!at) {
+        errors.push(
+          `${name}: declares ${UNTIL_KEY} '${record.until}', which is not a version this ` +
+            `gate has a schema for (${timeline.map((t) => t.label).join(", ")})`
+        );
+      } else {
+        for (const path of record.paths) {
+          if (!at.paths.has(path)) {
```
bbonaby, line 219: Originally posted by @SohamDas2021 on #738 ([original comment](https://github.com/microsoft/mxc/pull/738#discussion_r3732051146)):

> This only confirms the field exists at the until version; it never checks that the field is absent from every later frozen schema. A field still present in a frozen 0.7 schema can declare x-mxc-until:"0.6" and pass here after which the runtime wrongly rejects valid 0.7 configs that use it?

scripts/versioning/check-version-availability.js
```diff
@@ -0,0 +1,120 @@
...
+// Derived from the canonical version file plus the stable schemas on disk: a
+// hard-coded timeline would attach a stale label to a bumped dev schema.
+function discoverTimeline() {
+  const stableDir = join(repoRoot, "schemas", "stable");
+  const floor = parseMajorMinor(majorMinor(schemaVer.min));
+  if (!floor) {
+    throw new Error(`schema-version.json: 'min' (${schemaVer.min}) is not a version`);
+  }
+
+  const stable = readdirSync(stableDir)
```
bbonaby, line 52: Originally posted by @SohamDas2021 on #738 ([original comment](https://github.com/microsoft/mxc/pull/738#discussion_r3732057644)):

> discoverTimeline reads schemas/stable/* from the HEAD worktree (readdirSync + readJson off disk), so a PR that edits or adds a stable schema shifts the very timeline the since / until claims are validated against. This is asymmetric with #732/#730, which read the comparison baseline from the base commit via git-base?

src/core/mxc_engine/src/policy.rs
```diff
@@ -831,19 +832,29 @@ pub fn build_request_with_containment(
...
+    // Reject here too, so the SDK reports the same typed failure without
+    // building a config first.
     if policy.version.is_empty() {
-        return Err(MxcError::malformed_request("Policy version is required").into());
+        return Err(MxcError::version_incompatible("Policy version is required")
+            .with_details(serde_json::json!({
+                "field": "version",
+                "declaredVersion": "",
+                "since": null,
+                "until": null,
```
bbonaby, line 843: Originally posted by @SohamDas2021 on #738 ([original comment](https://github.com/microsoft/mxc/pull/738#discussion_r3732090865)):

> nit: for a missing/empty version, the SDK path emits "since": null, "until": null here, while the parser path emits since: MIN_SUPPORTED / until: MAX_SUPPORTED for the equivalent error in config_parser.rs?

bbonaby (commented):
Reposting @SohamDas2021's review feedback inline for continuity from accidentally merged #738.

scripts/versioning/check-dev-schema-compat.js
bbonaby, line 89: Originally posted by @SohamDas2021 on #738 ([original comment](https://github.com/microsoft/mxc/pull/738#discussion_r3737807652)):

> This gate advertises a floor-bump escape hatch it doesn't implement. detectBreaking(baseSchema, headSchema) compares the complete schemas; neither baseVersions.min nor headVersions.min scopes it, so a removal is blocked unconditionally- even in a PR that raises the floor. 
> Fixes:
> • A (now): make the message honest- drop "or move the supported-availability range in the same change" from L98 and correct the L17–20 comment to say removals are always blocked and window-based retirement isn't implemented yet.
> • B (real fix): when headVersions.min > baseVersions.min , scope the comparison to surface still supported after the new floor — prune base nodes whose x-mxc-until < headMin before  detectBreaking (watch $ref -shared definitions so a retired field sharing a type with a live one doesn't over-prune), plus the integration test that raises min and removes retired surface. Note there's no x-mxc-until surface today, so B needs a synthetic fixture to exercise it.

jsidewhite:
I still think we can just nuke all parser code and use a standard json schema validator.

MGudgin:
> I still think we can just nuke all parser code and use a standard json schema validator.

No. JSON schema is NOT the right source of truth here. It's a projection.

MGudgin:
Closing in favour of a fresh PR against a rebuilt stack.

The stack has been rebased onto the current `main` tip (`ad5a5f8c`), one commit newer than the base this PR was built on (`0be1cccd`).

While rebasing, three `SandboxPolicy` literal sites were found that `main`'s #748 (`capture_denials` on the typed `SandboxPolicy`) breaks. Git auto-merges them cleanly, so they surface only at compile time:

- `src/core/mxc_engine/src/policy.rs` (4 test sites) - already fixed here by `309bc30c`
- `src/core/mxc-sdk/tests/sandbox.rs:149` - **not** fixed; this head does not compile
- two `#[cfg(target_os = "macos")]` tests in `policy.rs` - **not** fixed; invisible to a Windows `cargo check`, so a green Windows run does not catch them

All three are fixed in the replacement, and the fixes are folded into the phase-8a commit rather than carried as a follow-up.

PR #778, opened 2026-08-08
feat(sdk): probe denial capture support

caarlos0:
## 📖 Description

Expose `captureDenials` through the Rust SDK's existing `available_backends()` host-capability probe.

ProcessContainer reports `BackendCapability::CaptureDenials` only after the host successfully creates a minimal PSEC environment and starts a Learning Mode trace. The probe session is immediately discarded. This remains advisory; launch still returns `ErrorCode::BackendUnavailable` if support changes before execution.

## 🔗 References

No linked issue.

## 🔍 Validation

- `cargo test -p mxc_engine -p mxc-sdk`
- `cargo clippy -p mxc_engine -p mxc-sdk --all-targets -- -D warnings`
- `cargo test -p mxc_engine probe::tests`
- `cargo clippy -p mxc_engine --all-targets -- -D warnings`

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/778)


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -500,6 +500,34 @@ impl BaseContainerRunner {
         })
     }
 
+    /// Whether this host can create a PSEC environment and start a Learning Mode trace.
+    ///
+    /// The successful probe session is dropped immediately, which closes and discards the trace.
+    pub fn is_capture_denials_usable() -> bool {
+        static USABLE: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+        *USABLE.get_or_init(|| {
```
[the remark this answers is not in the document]
caarlos0: Fixed in 82f121a. The capture-denials probe no longer caches its result, so a transient StartLearningModeTrace failure is retried by the next available_backends() call while successful probe sessions are still discarded immediately.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -500,6 +500,31 @@ impl BaseContainerRunner {
         })
     }
 
+    /// Whether this host can create a PSEC environment and start a Learning Mode trace.
+    ///
+    /// The successful probe session is dropped immediately, which closes and discards the trace.
+    pub fn is_capture_denials_usable() -> bool {
```
SohamDas2021, line 506: Both siblings cache their host probe behind OnceLock, but this one doesn't. So, every available_backends() call re-creates a PSEC env and starts/closes a real Learning Mode ETW trace. Please either memoize behind a static OnceLock<bool> like the siblings, or add a one-line comment explaining the deliberate no-cache choice (if it's intentional to allow retry after a transient StartLearningModeTrace  failure), since diverging silently from the cached siblings is surprising.

src/core/mxc_engine/src/probe.rs
```diff
@@ -297,6 +324,24 @@ mod tests {
...
+        let backends = available_backends();
+        let process_container = backends
+            .iter()
+            .find(|backend| backend.backend == "processcontainer")
+            .expect("processcontainer must always be reported on Windows");
+        assert_eq!(
+            process_container
+                .capabilities
+                .contains(&BackendCapability::CaptureDenials),
+            BaseContainerRunner::is_capture_denials_usable()
```
SohamDas2021: This runs a second live is_capture_denials_usable() probe and asserts it equals the value available_backends() already produced. Because the probe is uncached and each call does a real  StartLearningModeTrace, a transient trace-start failure on either probe makes the two disagree and fails the assert on a healthy host — and it's near-tautological (same fn both sides). Capture the probe once and assert  capabilities.contains(&CaptureDenials)  against that single bool, or inject a forced result via the existing  MXC_FORCE_BC_USABLE  pattern.

src/backends/appcontainer/common/src/base_container_runner.rs
caarlos0: Addressed in 8acb98f. Added an explicit comment explaining that the live trace probe is deliberately uncached so transient StartLearningModeTrace failures can be retried.

SohamDas2021 (approved):

PR #783, opened 2026-08-08

MGudgin:
This PR adds per-field schema version availability: a wire field can declare the range of config schema versions it is valid in, and the parser rejects any use outside that range. It is what makes shape-only support for older schema versions real — until now such an annotation would have been documentation that nothing honoured.

Replaces #775 (closed), rebased onto the current `main` tip with three compile breaks fixed.

Details

* New `mxc_version_derive` proc-macro crate. `#[derive(VersionAvailability)]` lifts `#[mxc_version(since = "0.8")]` / `until` off `wire.rs` into metadata normal builds carry, so one declaration feeds both the parser and schema generation (published as `x-mxc-since` / `x-mxc-until`). A derive rather than `#[schemars(extend(...))]`, which sits behind the `schema-gen` feature and so can never be consulted by the parser.
* The mechanism would fail **open** if a derived JSON name ever disagreed with what serde accepts — the range would simply never fire — so that case is guarded twice: the macro compile-errors on every serde construct it cannot model exactly (`flatten`, split `rename`/`rename_all`, unknown `rename_all` rules, data-carrying variants), and a conformance test cross-checks all 32 wire types against the names `schemars` derives independently.
* The gate runs immediately after deserialisation in every entry point, because `convert_wire_config` moves fields out of the config. State-aware requests are gated on the original document, not the experimental-masked copy.
* `version` is now **required** — it selects the legal field surface, so an absent one would silently opt out of every range.
* New `version_incompatible` code across all five surfaces (Rust `MxcErrorCode`, engine `ErrorCode`, TS, C#, `MXC_STATUS_VERSION_INCOMPATIBLE = 13`) carrying `details: { field, declaredVersion, since, until }`. The supported-range error migrates onto it. **This changes an existing error's observable shape** — a consumer string-matching the old range message is affected.
* Only three fields are annotated (`seatbelt` since 0.7, `processContainer.captureDenials` / `learningMode` since 0.8). What is deliberately *not* annotated matters more: schema-first-appearance is only a lower bound on accepted surface, so `experimental` (an open block before 0.8) and the state-aware discriminators carry no range — annotating them from schema data would reject configs that have always worked. The new `check-version-availability.js` oracle gate is fail-closed on exactly those surfaces.
* Rebase fixes: `main`'s #748 added `capture_denials` to the typed `SandboxPolicy`, which breaks `SandboxPolicy` literals that git auto-merges cleanly, so they surface only at compile time. `src/core/mxc-sdk/tests/sandbox.rs` and two `#[cfg(target_os = "macos")]` tests in `mxc_engine/src/policy.rs` did not compile on #775's head; the macOS pair is invisible to a Windows `cargo check`.

Tests

* `cargo test --workspace` — **2031 passed, 0 failed**. `cargo fmt --all -- --check`, `cargo check --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` all clean.
* `cargo check --target aarch64-apple-darwin -p mxc_engine -p wxc_common -p mxc-sdk --all-targets` clean — this is what caught the two macOS test breaks that a green Windows run had hidden.
* Feature-gated builds clean: `wxc_common` {schema-gen, microvm}, `mxc_ffi` {dotnetsdk}, `mxc_engine` {isolation_session}, `wxc` {isolation_session, microvm, tier2_bfs, wslc, hyperlight}.
* Versioning gates: **125 pass / 0 fail**. `check-version-availability` (3 declared ranges agree with the 0.6 / 0.7 / 0.8 schemas), `validate-configs` (201 configs), `check-schema-versions`, `check-schema-codegen`, `check-sdk-types-codegen`, and `check-dev-schema-compat` against phase 7d all green.
* Node SDK build + **223 pass / 6 skipped**; C# SDK **35 pass**; ErrorCode parity 17 codes; C# bindings codegen OK; version sync OK.
* Non-regression: the #676 replay still yields exactly **6** findings naming `allowLocalNetwork`, `allowedHosts`, `blockedHosts`, `defaultPolicy`, `enforcementMode`, `proxy`. `SUPPORTED_VERSION` unchanged at `>=0.6, <=0.8`.
* **Not executed on this host** (Windows): the macOS Seatbelt runtime paths, the Windows Sandbox and IsolationSession PowerShell lifecycle suites, and the host-gated MicroVM / Hyperlight E2E configs. The macOS code cross-compiles clean but has not been run.

---

Stacked on #732 (phase 7d). Replaces #775.

Co-authored-by: Branden Bonaby <bbonaby@microsoft.com>


MGudgin:
Revisiting the versioning approach for policy/config.

PR #790, opened 2026-08-09
[LXC] Fail closed when firewall rules cannot be scoped to the container

dhoehna:
## The problem

`install_firewall_rules` built the container's full deny-all chain and filled it with the network policy, then — when it did not know the container's veth interface — logged a warning and returned `Ok(())`.

The chain is only ever reached from `FORWARD` via `-i <veth>`.  With no hook, no packet traverses it.  So the caller was told the network policy had been applied while **zero packets were being filtered**.

That is worse than having no firewall, because it looks enforced.  Nothing downstream had any way to tell the difference.

## Why erroring is the right answer, and not the other two

There are exactly three things this code can do when it cannot scope the rules:

| Option | Result |
|---|---|
| Return `Ok` (before) | Chain exists, filters nothing, caller believes it is confined |
| Install the rules unscoped | They *would* filter — and would hit every other container and the host's own traffic |
| Return `Err` (this PR) | Container does not start; nothing is silently unprotected |

Erroring loses no enforcement, because there was none to lose.

## Blast radius — verified, not assumed

- **Only reachable when the caller explicitly asked for a firewall.**  `apply_firewall_rules` returns early unless the enforcement mode is `Firewall` or `Both`, and `NetworkEnforcementMode` defaults to `Capabilities`.  A container that never wanted firewalling is untouched — and there is a test pinning exactly that.
- **Rollback and teardown already handle `Err` correctly.**  `apply_firewall_rules_inner` converts it into a precise teardown of what was created plus residual adoption, and `lxc_runner.rs:221` destroys the container rather than starting a workload that believes it is confined.  No new branch was needed.
- **CI already proves veth discovery works on Linux.**  All four E2E scripts that exercise firewall enforcement (`run_lxc_network_cidr_boundary_test.sh`, `..._dualstack_test.sh`, `..._invalid_cidr_test.sh`, `..._ipv6_cidr_test.sh`) already **fail** if `"FORWARD hook installed"` is absent from the output.  So in every environment where those pass today, a veth exists and this change is a no-op.  It only alters the case those scripts already treat as a failure.
- **No existing test pinned the old `Ok(())`** — all 115 still pass unchanged.  That the fail-open was untested is itself the finding.

## Tests

Six black-box tests, written from the documented contract by an author who did not read the implementation, so they describe the intended behavior rather than mirroring whatever the code happens to do.

They pin the refusal under **both** firewall-requesting modes separately, the error naming the chain, the negative control (same policy succeeds once an interface is set — without which an apply that always returned `Err` would pass everything else), teardown ordering relative to chain creation, and the `Capabilities` no-op that bounds the blast radius.

**Mutation tested: 7 seeded defects, 7 caught by a failing test, 0 survivors.**  The seeds include restoring the old `Ok(())`, dropping the chain name from the message, applying the check to `Firewall` but not `Both`, inverting the interface check, skipping rollback, and swallowing the error one layer up in `record_apply_outcome`.  Each mutant is compiled with lints silenced, so a defect caught only by the compiler counts as a harness failure rather than a pass — the tests have to answer for themselves.

The spec is attached as a `#[path]` child module because the fake-firewall seam is `#[cfg(test)]` and private, which an integration test — a separate crate — can never reach.

## Scope

Slice 3 of the PR #632 re-cut (#632 was closed; its merge resolution had grown past the point of being reviewable).  Follows #788 and #789.

**Deliberately not fixed here:** the second fail-open in the same file — `build_policy_rules_logged` warns `"could not resolve host"` and continues, so under `defaultPolicy: allow` a `blockedHosts` entry that fails DNS silently becomes reachable.  Four Linux E2E scripts grep that exact string and one *requires* it, so changing it needs those scripts updated in the same PR.  It moves to the enforcement slice.

Also noted and not fixed: host lists are only enforced under `Firewall`/`Both`, while the runner treats a non-empty host list as "needs network" — so under the default `Capabilities` mode a container with `blockedHosts` gets no firewall at all.  That belongs with the enforcement work.

Refs AB#62830341.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/790)

### DIFF
--- a/src/backends/lxc/common/src/network_iptables.rs
+++ b/src/backends/lxc/common/src/network_iptables.rs
@@ -1049,12 +1049,26 @@ impl NetworkIptablesManager {
                 ));
             }
         } else {
-            // Without a veth interface, we cannot safely scope rules to the container.
-            // Refuse to apply host-wide rules to avoid affecting all host traffic.
-            logger.log_line(
-                "Warning: No veth interface set for container. \
-                 Cannot scope iptables rules. Skipping FORWARD hook.",
-            );
+            // Without a veth interface there is nothing to hook the chain to,
+            // and an unhooked chain is never traversed: FORWARD only reaches it
+            // via `-i <veth>`. Reporting success here would hand the caller a
+            // fully populated deny-all chain that filters nothing, which is
+            // strictly worse than no firewall at all because it looks enforced.
+            //
+            // The alternative -- installing the rules host-wide so they do take
+            // effect -- is not acceptable either: unscoped they would apply to
+            // every container and to the host's own traffic.
+            //
+            // So the only honest outcome is to fail. `apply_firewall_rules_inner`
+            // rolls back the chains recorded in `created`, and `lxc_runner`
+            // destroys the container rather than starting a workload that
+            // believes it is confined.
+            return Err(format!(
+                "No veth interface for container; cannot scope iptables rules to chain {}. \
+                 The chain would never be reached from FORWARD, so the network policy would \
+                 not be enforced. Refusing to report success for an unenforceable policy.",
+                self.chain_name
+            ));
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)


dhoehna:
Superseded by #798.

PR #6, opened 2026-02-16
Build fixes, update documentation.

shschaefer:
Cleanup the build scripts.  Update the documentation, creating a doc folder and first files.

PR #791, opened 2026-08-09
Fix scoped ETW telemetry validation

RamonArjona4:
## 📖 Description
This PR completes the Windows-local Microsoft.MXC ETW logging remediation. It separates ETW emission from local diagnostic sinks, corrects ETW payload semantics, strengthens payload and sink-gating tests, makes the ETW smoke test provider-aware, and documents the contract.

The scope is limited to logging/ETW emission, tests, performance, and documentation. It does not change network/DNS enforcement, fallback/DACL behavior, cleanup/kill behavior, execution policy, or add collectors/uploaders/services/daemons.

## 🔗 References
No linked issue.

## 🔍 Validation
- Targeted telemetry, policy identity, IsolationSession, and AppContainer tests
- `cargo build --workspace`
- PowerShell smoke-test script parsing

## ✅ Checklist
- [ ] Linked to an issue

## 📋 Issue Type
- [x] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/791)


RamonArjona4:
## Addressing In-Scope Review Comments

All 6 in-scope comments have been addressed and CI is now passing ✅

### Fix #1: M-ETW-5 Gate Reordering ✅
**File:** \src/backends/appcontainer/common/src/appcontainer_runner.rs\
Reordered the ETW emission gate so \log_sandbox_torn_down()\ is called before the \is_active()\ check. This ensures M-ETW-5 fires on all successful T2/T3 containment paths independent of diagnostic sink state.

### Fix #2: Policy Hash Logging (M-ETW-3) ✅
**File:** \src/core/wxc/src/main.rs\ (~line 565)
Moved \log_policy_hash()\ outside the telemetry-only gate so it emits when either telemetry OR diagnostics is active. This ensures M-ETW-3 is captured for all IsolationSession state-aware phases.

### Fix #3: Console Message Escaping ✅
**File:** \src/tools/mxc_diagnostic_console/src/main.rs\
Implemented \scape_display_message()\ function that escapes control chars (0x00–0x1F except \\t/\\n) and non-ASCII via \char::escape_default()\. Applied to all DisplayEvent::Message rendering paths to prevent ANSI escape sequence injection.

### Fix #4: M-ETW-7 Config Rejection ✅
**File:** \src/core/wxc/src/main.rs\ (~line 619–625)
Verified: M-ETW-7 (ConfigRejected) is already implemented at dispatch-time validation. Config validation failures correctly emit ConfigRejected before provider shutdown.

### Fix #5: Provider Initialization Order ✅  
**File:** \src/core/wxc/src/main.rs\ (~line 1206–1210)
Verified: Telemetry provider is initialized before command-override checks in the state-aware path. Early-exit validation failures can now call \log_config_rejected\ with ETW active.

### Fix #6: Case-Insensitive Secret Detection ✅
**File:** \src/core/wxc_common/src/policy_identity.rs\
Verified: Case-insensitive secret-name matching using substring predicates is already implemented to prevent secret-bearing fields from entering the policy hash.

---

### CI Status: ALL PASSING ✅
- ✅ Windows x64 (9m19s)
- ✅ Windows arm64 (5m9s)  
- ✅ Linux x64/arm64
- ✅ macOS arm64
- ✅ All linting & SDK tests

### Out-of-Scope Comments
Comments #7–9 (named-pipe token enumeration, policy hash identity oracle, malformed env logging) remain out-of-scope as discussed.

schemas/dev/mxc-config.schema.0.dev.json
```diff
@@ -0,0 +1,1065 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://github.com/microsoft/mxc/schemas/dev/mxc-config.schema.0.8.0-dev.json",
+  "title": "MXC Configuration",
+  "description": "MXC container execution configuration. Defines the recommended config format for both one-shot and state-aware sandbox lifecycle requests. A few deprecated field spellings not listed here are also accepted via serde aliases.",
+  "additionalProperties": false,
```
jsidewhite, line 6: what is this file for?

PR_791_FAILURE_ANALYSIS.md
```diff
@@ -0,0 +1,218 @@
+# PR #791 - GitHub Actions Build Failure Analysis
+
+## Summary
```
jsidewhite: remove?

run_31447349167_full.txt
```diff
@@ -0,0 +1,283 @@
+WXC-Exec Hyperlight	Set up job	﻿2026-08-11T00:49:51.4764641Z Current runner version: '2.336.0'
+WXC-Exec Hyperlight	Set up job	2026-08-11T00:49:51.4831513Z ##[group]Runner Image Provisioner
```
jsidewhite: remove these logs?

docs/diagnostics.md
```diff
@@ -14,34 +14,106 @@ All layers stream into a single `mxc-diagnostic-console.exe` window in real time
 ## Quick Start
 
 ```powershell
-# Terminal 1: start the diagnostic console (run as admin for ETW)
+# Terminal 1: choose one token and start the diagnostic console
+$env:MXC_DIAG_PIPE_TOKEN = [guid]::NewGuid().ToString("N")
+$env:MXC_DIAG_PIPE_TOKEN
 mxc-diagnostic-console.exe
```
jsidewhite, line 20: should diag console just emit one of these if not specified by the user?

`[guid]::NewGuid().ToString("N")`

jsidewhite (dismissed):
:shipit:

docs/diagnostics.md
RamonArjona4: Updated docs/diagnostics.md to clarify that the caller generates one token and supplies the same value to both processes before startup. The console does not silently generate its own token because the executor would not receive it, so the processes would not select the same authenticated pipe.

schemas/dev/mxc-config.schema.0.dev.json
RamonArjona4: This file is the generated development JSON schema for the current Rust wire model. It is intentionally committed for editor validation and CI schema-codegen checks, and should be regenerated rather than removed.

run_31447349167_full.txt
RamonArjona4: Removed the captured workflow log from the PR. No replacement log was added.

PR_791_FAILURE_ANALYSIS.md
RamonArjona4: Removed this temporary failure-analysis artifact from the PR.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1502,7 +1774,24 @@ impl SandboxProcess for AppContainerSandboxProcess {
...
         // Terminate the whole job: the child and every descendant assigned to
         // it die together (tree-kill).
-        self.job.terminate(u32::MAX);
+        //
+        self.kill_requested = true;
+        if let Err(error) = self.job.terminate(u32::MAX) {
+            wxc_common::telemetry::log_process_event(
+                sanitize_identity(&self.identity),
+                self.pid,
+                wxc_common::telemetry::ProcessEvent::KillFailed(
```
MGudgin: **High (correctness) - `ProcessKillFailed` ETW drops the actual OS error.**

This new ETW call passes only the kill method (`terminate_job_object`). The adjacent local audit record includes `error.code().0`, so ETW consumers learn what was attempted but not why it failed, despite M-ETW-1 requiring the process outcome error.

**Fix:** Carry both the bounded kill method and numeric OS error through `ProcessEvent::KillFailed`, emit both to ETW, and extend the payload test.

src/core/wxc/src/main.rs
```diff
@@ -338,6 +340,148 @@ fn log_state_aware_dispatch_error(logger: &mut Logger, error: &MxcError) {
...
+/// reach it.
+fn log_config_rejected(
+    logger: &mut Logger,
+    reason: RejectionReason,
+    backend: &str,
+    offending_field: &str,
+    phase: &str,
+) {
+    let correlation_id = wxc_common::audit::process_correlation_id();
+    wxc_common::telemetry::log_config_rejected(
```
MGudgin, line 356: **High (correctness) - `ConfigRejected` ETW omits the state-aware phase.**

`phase` is available to this function and is included in the local audit record, but this telemetry call does not pass it. The documented M-ETW-7 matrix says state-aware rejections identify the rejecting phase.

**Fix:** Thread `phase` through the `wxc_common` and `mxc_telemetry` wrappers, emit `mxc.phase`, and assert it in the payload test.

src/core/wxc_common/src/telemetry/events.rs
```diff
@@ -1,249 +1,662 @@
...
+        assert_eq!(FailureReason::Timeout.as_str(), "timeout");
+        assert_eq!(FailureReason::InitError.as_str(), "init_error");
+        assert_eq!(FailureReason::InternalError.as_str(), "internal_error");
+        assert_eq!(FailureReason::Cancelled.as_str(), "cancelled");
+        assert_eq!(FailureReason::Unknown.as_str(), "unknown");
+    }
+
+    #[test]
+    fn requirement_events_use_bounded_event_names() {
+        test_sink::install();
```
MGudgin, line 513: **High (testability) - This test clears process-global capture buffers without the required lock.**

`test_sink::install()` documents that the telemetry `TEST_LOCK` must be held, but this test cannot acquire the lock because it is private to a different test module. Parallel tests can clear records while another test is asserting them.

**Fix:** Move/expose one shared capture lock and make `install()` require or return a guard so lock ownership is enforced.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1439,24 +1596,139 @@ impl AppContainerSandboxProcess {
...
+                record = record.str("skip_reason", reason.as_str());
+            }
+            self.audit_logger.log_audit_event(&record);
         }
+        let released_resources = if wxc_common::telemetry::is_active() {
+            format_released_resources(network.rules_removed, bfs_removed, network.proxy_stopped)
+        } else {
+            String::new()
+        };
+        wxc_common::telemetry::log_sandbox_torn_down(
```
MGudgin: **High (test coverage) - The runner-to-telemetry wiring is not exercised end-to-end.**

The event helpers have direct payload tests, but the runner tests never install the telemetry capture sink or assert that real prepare/timeout/kill/teardown paths emit the expected events. A missing call or incorrect gate can therefore leave all helper tests green.

**Fix:** Add representative runner-level tests that capture and assert the emitted requirement events and payloads.

src/core/wxc_common/src/logger.rs
```diff
@@ -104,6 +116,11 @@ impl Logger {
 
         match std::fs::OpenOptions::new()
             .write(true)
+            .access_mode(
```
MGudgin, line 153: **Medium (security) - The pipe client can be impersonated before the new server-integrity check runs.**

The new access/verification path opens the named pipe without `SECURITY_SQOS_PRESENT | SECURITY_IDENTIFICATION`. Windows therefore permits the server to impersonate the client as soon as the handle opens, before `verify_server_integrity` runs. That is dangerous when `wxc-exec` is elevated. The failure branch also writes into the pipe it just rejected.

**Fix:** Add identification-level SQOS flags and drop a failed-verification handle without writing to it.

src/core/wxc/src/main.rs
```diff
@@ -338,6 +340,148 @@ fn log_state_aware_dispatch_error(logger: &mut Logger, error: &MxcError) {
...
+    }
+}
+
+fn rejection_reason_for(error: &MxcError) -> RejectionReason {
+    config_rejection_reason_for(error).unwrap_or(RejectionReason::RunnerUnavailable)
+}
+
+fn state_aware_policy_identity(sandbox_id: Option<&str>) -> String {
+    let identity = sandbox_id.filter(|id| !id.is_empty()).unwrap_or("CLI");
+    wxc_common::policy_identity::redact_identity(identity)
```
MGudgin: **Medium (security) - Caller-shaped `iso:`/`wsb:` IDs can bypass the intended redaction invariant.**

`redact_identity` ultimately accepts any short `iso:`/`wsb:` value with the allowed token characters, and this path emits the result before backend-specific ID validation. A caller can therefore control a privacy-tagged identity field despite the stated rule that caller-supplied IDs are redacted.

**Fix:** Emit a fixed minted-identity marker or a hash of a successfully decoded/round-tripped MXC ID; redact everything else.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1502,7 +1774,24 @@ impl SandboxProcess for AppContainerSandboxProcess {
     fn kill(&mut self) -> std::io::Result<()> {
         // Terminate the whole job: the child and every descendant assigned to
         // it die together (tree-kill).
-        self.job.terminate(u32::MAX);
+        //
+        self.kill_requested = true;
```
MGudgin: **Medium (reliability) - Explicit cancellation can suppress every terminal outcome record.**

`kill_requested` is set before termination succeeds. `wait()` then suppresses `ProcessExited` whenever this flag is true. A successful SDK/FFI cancellation emits neither exit, timeout, nor kill-failure; a failed kill followed by natural exit is also hidden.

**Fix:** Record a dedicated successful-kill outcome, or suppress normal exit only after a successful kill while preserving exactly one terminal record.

src/core/wxc/src/main.rs
```diff
@@ -1179,24 +1492,35 @@ fn main() {
...
-            let _ = writeln!(logger, "{}", json.trim());
+            // Redact secret-bearing fields (e.g. `experimental.isolationSession.user.{upn,wamToken}`)
+            // before writing: a one-shot IsolationSession request's credential
+            // bundle is only rejected by the runner after this point, so the
+            // raw config as received from the caller may still contain it.
+            let _ = writeln!(logger, "SECTION: JSON Config (redacted)");
+            let _ = writeln!(
+                logger,
+                "{}",
+                wxc_common::diagnostic::redact_raw_config_json(json.trim())
```
MGudgin: **Medium (performance) - Newly added config redaction runs without a diagnostic sink.**

The second config read predates this PR, but the new JSON parse/redaction and full `ExecutionRequest` rendering are performed even when no file/pipe diagnostic sink is attached and the buffer is discarded on success.

**Fix:** Gate the newly added redaction/rendering work on `logger.has_diagnostic_sink()` or the relevant console enablement.

src/core/wxc_common/src/diagnostic.rs
```diff
@@ -158,6 +183,84 @@ pub fn redacted_request_json(request: &ExecutionRequest) -> String {
...
+                if crate::config_deserialize::is_secret_path_field(&key_lower) {
+                    *entry = serde_json::Value::String("<redacted>".to_string());
+                } else if key_lower == "env" {
+                    redact_environment_values(entry);
+                } else if key_lower == "url" && path.iter().any(|parent| parent == "proxy") {
+                    if let serde_json::Value::String(url) = entry {
+                        *url = crate::proxy_env::redact_proxy_url(url);
+                    }
+                } else {
+                    let mut child_path = path.to_vec();
```
MGudgin: **Medium (performance) - Recursive redaction clones the full path for every key.**

`path.to_vec()` clones the complete path and its strings at each object key, causing allocation growth with both JSON size and depth.

**Fix:** Reuse one mutable path stack: push before recursion and pop afterward.

src/core/wxc_common/src/policy_identity.rs
```diff
@@ -0,0 +1,736 @@
...
+    match value {
+        Value::Object(map) => {
+            let mut keys: Vec<&String> = map.keys().collect();
+            keys.sort_unstable();
+            out.push('{');
+            for (i, key) in keys.iter().enumerate() {
+                if i > 0 {
+                    out.push(',');
+                }
+                out.push_str(&scalar_json(&Value::String((*key).clone())));
```
MGudgin: **Medium (performance) - Canonical hashing allocates intermediate strings for every key and scalar.**

Each key is cloned into a new `Value::String`, then serialized into another `String`; scalar values allocate similarly. This work runs whenever policy identity is computed.

**Fix:** Serialize keys/scalars directly into a safe `Vec<u8>` writer and convert once at the end.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1439,24 +1596,139 @@ impl AppContainerSandboxProcess {
...
+                .bool("proxy_stopped", network.proxy_stopped)
+                .bool("preserve_policy", self.preserve_policy)
+                // The AppContainer tiers delete no profile here (the runner's own
+                // `Drop` owns profile deletion), so this handle releases no
+                // container.
+                .bool("container_released", false);
+            if let Some(reason) = skip_reason {
+                record = record.str("skip_reason", reason.as_str());
+            }
+            self.audit_logger.log_audit_event(&record);
```
MGudgin, line 1975: **Medium (maintainability) - The new teardown sink preservation leaves cleanup warnings in a throwaway buffer.**

This structured teardown record correctly uses `self.audit_logger`, but the immediately preceding firewall/BFS cleanup still receives a fresh buffer-only logger. Partial cleanup warnings are silently dropped. The old local logger is pre-existing; the gap is that the new sink-preservation fix does not cover it.

**Fix:** Pass `self.audit_logger` or a sink-preserving clone into cleanup.

src/core/wxc_common/src/logger.rs
```diff
@@ -242,28 +298,195 @@ impl Logger {
...
+    /// handle is `try_clone`-duplicated (append mode, so a shared file pointer
+    /// is safe) and, on Windows, the diagnostic pipe is duplicated the same
+    /// way (`PIPE_TYPE_MESSAGE`, so each `write_all` remains a discrete
+    /// message that cannot interleave). Publication is per-thread, so
+    /// concurrent runs on different threads never share sinks by accident.
+    ///
+    /// The installed sink is reclaimed with
+    /// [`Logger::clear_thread_diagnostic_sink`]; leaving one installed at
+    /// process exit is harmless (the sinks close with the process).
+    pub fn install_thread_diagnostic_sink(&self) {
```
MGudgin: **Medium (testability, maintainability, simplicity) - Thread-local sink installation is not scoped by RAII.**

The caller must manually pair install and clear. A panic or future early return leaves stale ambient state for later work on the same thread.

**Fix:** Return a drop guard from installation, or pass a diagnostic-sink handle explicitly through the state-aware path.

schemas/dev/mxc-config.schema.0.dev.json
```diff
@@ -0,0 +1,1065 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://github.com/microsoft/mxc/schemas/dev/mxc-config.schema.0.8.0-dev.json",
```
MGudgin: **Medium (documentation-drift, proportionality) - The dev schema is committed under the wrong filename.**

The file is named `mxc-config.schema.0.dev.json`, but its own `$id` and `schemas/schema-version.json` identify the canonical suffix as `0.8.0-dev`. This near-duplicate is outside the normal schema drift gate.

**Fix:** Delete this file and regenerate/update only `schemas/dev/mxc-config.schema.0.8.0-dev.json`.

src/core/wxc_common/src/telemetry/events.rs
```diff
@@ -1,249 +1,662 @@
...
+/// contract. The event kind and payload are coupled in a single enum
+/// to make mismatches unrepresentable at compile time.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProcessEvent<'a> {
+    Exited(i32),
+    TimedOut(u64),
+    KillFailed(&'a str),
+}
+
+/// Backwards compatibility: deprecated wrappers around ProcessEvent variants.
```
MGudgin: **Medium (simplicity, proportionality) - Compatibility types are added and deprecated in the same PR.**

`ProcessEventKind` and `ProcessEventData` did not exist on the base and have no call sites at the head, so there are no existing callers to preserve.

**Fix:** Remove both deprecated enums and keep only `ProcessEvent`.

src/backends/appcontainer/common/src/network_manager.rs
```diff
@@ -98,12 +166,61 @@ struct RuleContext<'a> {
...
+        if firewall_mode_selected {
+            let default_policy = if policy.default_network_policy == NetworkPolicy::Block {
+                DefaultPolicy::Block
+            } else {
+                DefaultPolicy::Allow
+            };
+            return NetworkPolicyPlan {
+                default_policy,
+                firewall_mode_selected,
+                rules_will_be_installed: true,
```
MGudgin: **Medium (simplicity) - The new policy-description decision duplicates and disagrees with apply-time behavior.**

This sets `rules_will_be_installed=true` for every firewall mode. The pre-existing `initialize_policy` only installs rules when hosts are present or the default policy blocks. Audit status can therefore diverge from what the manager actually does.

**Fix:** Make one `NetworkPolicyPlan` function drive both auditing and side effects.

CODE_REVIEW_TRACELOGGING_COMPLIANCE.md
```diff
@@ -0,0 +1,420 @@
+# Code Review: TraceLogging Best Practices Compliance
```
MGudgin: **Medium (proportionality; also documentation-drift/maintainability) - This is a transient review artifact, not durable repository documentation.**

The 420-line file evaluates an intermediate commit and duplicates the durable telemetry guide. The final cleanup commit removed similar PR analysis/log artifacts but missed this one.

**Fix:** Remove it; keep any useful conclusions in the PR discussion or durable telemetry documentation.

src/core/wxc_common/src/diagnostic.rs
```diff
@@ -15,20 +15,45 @@ use windows::Win32::Security::{GetTokenInformation, TokenUser, TOKEN_QUERY, TOKE
...
+/// entropy for a pipe name.
+pub fn diagnostic_pipe_token() -> Option<String> {
+    let token = env::var(ENV_PIPE_TOKEN).ok()?;
+    if !is_valid_pipe_token(&token) {
+        return None;
+    }
+    Some(token)
+}
+
+fn is_valid_pipe_token(token: &str) -> bool {
```
MGudgin, line 42: **Low (security) - The validator accepts low-entropy repeated token patterns.**

Length, character class, and four distinct bytes allow a repeated four-character token while diagnostics call this a high-entropy token.

**Fix:** Require a canonical random-token shape with a meaningful entropy floor, or describe it only as a naming discriminator.

src/core/wxc_common/src/policy_identity.rs
```diff
@@ -0,0 +1,736 @@
...
+/// explicitly.
+///
+/// Used to excise credential-bearing sub-objects from an otherwise
+/// blanket-serialized section, so the section's enforcement-relevant fields can
+/// still be hashed.
+fn strip_keys(value: &mut Value, keys: &[&str]) {
+    match value {
+        Value::Object(map) => {
+            map.retain(|key, _| {
+                let normalized = key.to_ascii_lowercase();
```
MGudgin: **Low (performance) - Case-insensitive key matching allocates for every key.**

`to_ascii_lowercase()` creates a new string per key during recursive policy projection.

**Fix:** Use `eq_ignore_ascii_case` for exact names and an allocation-free marker check.

src/core/wxc_common/src/logger.rs
```diff
@@ -104,6 +116,11 @@ impl Logger {
 
         match std::fs::OpenOptions::new()
             .write(true)
+            .access_mode(
+                windows::Win32::Storage::FileSystem::FILE_GENERIC_WRITE.0
+                    | 0x0002_0000
```
MGudgin: **Low (maintainability) - The security-relevant access mask contains an unexplained raw constant.**

`0x0002_0000` is `READ_CONTROL`, but the intent is not visible here and the repository uses a named constant for the same value elsewhere.

**Fix:** Use/name `READ_CONTROL` and briefly explain why the pipe handle needs it.

src/Cargo.toml
```diff
@@ -100,11 +100,12 @@ windows-core = "0.62"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 serde_path_to_error = "0.1"
+# SHA-256 for the canonical policy hash (`wxc_common::policy_identity`).
+sha2 = "0.10"
```
MGudgin: **Low (supply-chain) - Core use of `sha2 0.10` further entrenches duplicate versions.**

This new direct workspace use expands `sha2 0.10` into `wxc_common` while `sha2 0.11` is already present transitively. No new crate or MSRV regression is introduced, but the binary now carries the older implementation more broadly.

**Fix:** Track unification on one supported `sha2` version when dependencies retaining 0.10 can move.

MGudgin (changes requested):
## Summary

I verified these findings against PR #791 at head `424a486` before filing. The review covers 4 High, 13 Medium, and 4 Low findings; 20 are anchored to added lines and one PR-description finding is recorded below.

The branch is behind its current `main` base: `baseRefOid=aa05d77`, while the PR merge-base is `3a5d0f1`. This review uses GitHub's actual three-dot PR range. The local and GitHub diffs both contain 10,436 lines (their normalized byte counts differ by 76 bytes from headers/encoding).

- All 16 unique cited paths have non-empty `git diff --numstat` output in this PR. No finding relies on a byte-identical out-of-diff file.
- Every inline comment below was checked against the captured diff and lands on an added (`+`) line.
- Runner test counts increased (`appcontainer_runner.rs`: 29 to 34; IsolationSession `manager.rs`: 1 to 4), but neither runner test module calls the telemetry capture sink; the integration-coverage finding is therefore about the new wiring, not an absence of all tests.
- Cross-platform parity was reviewed separately and no accidental shared-build/API asymmetry was found.

### Attribution adjustments

- The config file reread predates this PR; the performance comment is limited to the newly added ungated redaction and full-request rendering.
- The teardown buffer logger predates this PR; the maintainability comment is framed as a gap in the new diagnostic-sink preservation fix.
- `initialize_policy` predates this PR; the simplicity comment targets the newly added `describe_policy` copy and its divergent decision.

### Finding in the PR description

**Medium (proportionality) - The stated scope understates changes to enforcement paths.** The description says network/DNS enforcement and fallback/DACL behavior are unchanged, but this PR rewrites control flow in `fallback_detector.rs` and `network_manager.rs` to collect telemetry fields. Final behavior appears restored, but commit `bf8f8eb` was required to repair a BaseContainer denied-path DACL regression introduced during that refactor. Please update the PR description to name these enforcement-path changes so the appropriate reviewers are engaged.

No byte-identical out-of-diff finding is being charged to this PR. The three pre-existing structures noted above are included only where the new code adds an ungated cost, leaves a gap in the new fix, or duplicates an existing decision.

src/Cargo.toml
RamonArjona4: Thanks for flagging this. Confirmed via `cargo tree -i sha2`: `sha2 0.10.9` is pulled in transitively by several existing backends (lxc_common, bwrap_common, etc.) independent of this PR, and `sha2 0.11.0` is present as a separate transitive dependency of an unrelated crate. This PR's new direct `wxc_common` -> `sha2 0.10` dependency does not introduce a new duplicate pair -- it adds one more consumer of a version that was already duplicated on `main` before this change.

Unifying on a single `sha2` version workspace-wide is a legitimate follow-up, but it requires coordinating a version bump across multiple unrelated backend crates that do not otherwise touch this PR's ETW/logging scope, so I'm deferring it rather than bundling it into this change. Tracking it as a separate housekeeping item is the right call here; no code change is being made in this PR for this comment.

src/core/wxc/src/main.rs
RamonArjona4: Fixed: the raw-config redaction/logging work in `main.rs` is now gated behind `logger.has_diagnostic_sink()` up front, instead of always re-reading, decoding, and redacting the config JSON only to have `log_diagnostic_line` discard the result when no sink is attached.

src/core/wxc_common/src/diagnostic.rs
RamonArjona4: Fixed: `redact_secret_fields_at_path` in `diagnostic.rs` now takes `&mut Vec<String>` and pushes/pops the current key around each recursive call, instead of cloning the whole path vector at every object key -- allocation no longer grows with both JSON size and nesting depth.

src/backends/appcontainer/common/src/appcontainer_runner.rs
RamonArjona4: Fixed: `ProcessEvent::KillFailed` now carries the OS error code alongside the bounded kill-method string. `mxc_telemetry::log_process_kill_failed` gained an `error_code: i32` parameter emitted as `mxc.error_code`, the `appcontainer_runner.rs`/`base_container_runner.rs`/`isolation_session::manager.rs` call sites now pass the in-scope error's code, and the requirement test in `events.rs` asserts the field's presence and value.

src/core/wxc/src/main.rs
RamonArjona4: Fixed: the `log_config_rejected` wrapper in `main.rs` already accepted a `phase: &str` parameter but never forwarded it. It now threads `phase` through `wxc_common::telemetry::log_config_rejected` -> `mxc_telemetry::log_config_rejected` (both the Windows `write_event!` call, which now includes `mxc.phase`, and the non-Windows no-op stub).

src/core/wxc_common/src/telemetry/events.rs
RamonArjona4: Fixed: `telemetry::mod.rs`'s test module now reuses `events::test_sink::TEST_LOCK` instead of declaring its own private `TEST_LOCK`, since both modules touch the same process-global capture state and a lock private to one module wouldn't serialize against the other.

src/backends/appcontainer/common/src/appcontainer_runner.rs
RamonArjona4: Added: three runner-level e2e tests in `appcontainer_runner.rs` (`runner_e2e` submodule) that spawn a real Win32 process, assign it to a real `UiJobObject`, and drive it through the actual `wait`/`kill`/`Drop` code paths with a real file-backed `Logger`, then assert the emitted audit records in the log file. Covers: natural exit (ProcessExited + SandboxTornDown), proactive kill before natural exit (still emits ProcessExited -- this is the regression test for the kill-telemetry-suppression fix below), and timeout (ProcessTimedOut).

src/core/wxc_common/src/logger.rs
RamonArjona4: Fixed: the diagnostic pipe's `CreateFile` call now includes `SECURITY_SQOS_PRESENT | SECURITY_IDENTIFICATION` so the client can't be impersonated at a level higher than intended, plus an explicit named `READ_CONTROL` constant replacing the previous raw hex literal.

src/core/wxc/src/main.rs
RamonArjona4: Fixed: `main.rs`'s state-aware policy-hash identity now goes through `redact_identity`, which recognizes UPN-shaped `iso:`/caller-supplied identities via the `@` discriminator and substitutes the bounded `entra-upn` marker instead of ever hashing/logging the raw value.

src/backends/appcontainer/common/src/appcontainer_runner.rs
RamonArjona4: Fixed: verified `AppContainerSandboxProcess::kill()` unconditionally records `ProcessExited`/`ProcessKillFailed` telemetry regardless of whether the cancellation was proactive (caller-initiated) or reactive (timeout/natural exit) -- added the `proactive_kill_before_natural_exit_still_emits_process_exited` e2e test above as the regression lock-in for this exact scenario.

src/core/wxc_common/src/policy_identity.rs
RamonArjona4: Fixed: `canonical_json`/`write_canonical` in `policy_identity.rs` now write directly into a single growing `Vec<u8>` buffer (converted to `String` once at the end) instead of allocating a separate `String` per object key (via an intermediate `Value::String` clone) and per scalar value.

src/core/wxc_common/src/policy_identity.rs
RamonArjona4: Fixed: added an ASCII-case-insensitive `is_secret_path_field_ci` in `config_deserialize.rs` (whole-field `eq_ignore_ascii_case` + allocation-free substring marker match) and switched `policy_identity.rs`'s `strip_keys` to use it, removing the per-key `to_ascii_lowercase()` allocation that existed solely to feed the existing lower-cased-only `is_secret_path_field`.

src/backends/appcontainer/common/src/appcontainer_runner.rs
RamonArjona4: Verified: `run_teardown()`'s cleanup path already uses `self.audit_logger` (the real duplicated diagnostic sink captured at construction), not a throwaway buffer -- confirmed via the new `natural_exit_emits_process_exited_and_torn_down_records` e2e test, which asserts a real `mxc.SandboxTornDown` record lands in a real log file after teardown runs.

src/core/wxc_common/src/logger.rs
RamonArjona4: Fixed: `Logger::install_thread_diagnostic_sink` now returns a `ThreadDiagnosticSinkGuard` whose `Drop` clears the thread-local slot, instead of relying on the caller to remember to call `clear_thread_diagnostic_sink()` on every exit path (including an unwind through a panic in `run_state_aware`). The `wxc-exec` call site in `main.rs` now binds the guard instead of pairing an explicit install/clear.

src/core/wxc_common/src/logger.rs
RamonArjona4: Fixed: added an explicit named `READ_CONTROL` constant (with a doc comment distinguishing it from the SQOS flags) replacing the previous unexplained raw hex literal in the diagnostic pipe's access-mask construction.

schemas/dev/mxc-config.schema.0.dev.json
RamonArjona4: Fixed: removed the stray `schemas/dev/mxc-config.schema.0.dev.json`. Confirmed the canonical, correctly-named `mxc-config.schema.0.8.0-dev.json` (per `schemas/schema-version.json`'s `devSchemaFile: "0.8.0-dev"`) already exists on `main`, untouched by this PR -- this file was a wrongly-named duplicate produced by a stale schema-gen invocation during development, not a second required artifact. This also corrects an earlier reply I gave suggesting it should be regenerated rather than removed.

src/core/wxc_common/src/telemetry/events.rs
RamonArjona4: Fixed: removed `ProcessEventKind`/`ProcessEventData` from `telemetry/events.rs` -- confirmed via grep across `wxc_common`, `appcontainer_common`, and `isolation_session_common` that they had zero call sites anywhere in the codebase.

src/backends/appcontainer/common/src/network_manager.rs
RamonArjona4: Fixed: `network_manager.rs`'s `initialize_policy` now calls `Self::describe_policy(policy)` and reuses its `NetworkPolicyPlan` instead of recomputing the same default-policy/firewall-mode decision inline with separate (and previously slightly disagreeing) logic. No enforcement behavior changed -- only the audit-description/log-gating layer.

CODE_REVIEW_TRACELOGGING_COMPLIANCE.md
RamonArjona4: Fixed: removed `CODE_REVIEW_TRACELOGGING_COMPLIANCE.md`, a transient review artifact that should not have been committed to the PR.

src/core/wxc_common/src/diagnostic.rs
RamonArjona4: Fixed: `is_valid_pipe_token` in `diagnostic.rs` now computes Shannon entropy (bits/char) over the token's non-separator bytes and requires at least 3 bits/char, replacing the previous "at least 4 distinct bytes" check that a short pattern repeated to reach the length floor (e.g. "abcd" x8) could satisfy with only 2 bits/char of real entropy. Added a regression test for exactly that repeated-pattern case.

MGudgin (dismissed):
Re-reviewed current head `b60e655`.

Confirmed the ETW payload, shared telemetry test lock, pipe SQOS, identity redaction, no-sink performance gates, allocation fixes, teardown diagnostics, RAII cleanup, artifact removal, policy-plan reuse, entropy validation, and named access-mask fixes. The targeted verification suite passed 38 tests with 0 failures across `wxc_common`, `appcontainer_common`, and `mxc_telemetry`.

The `sha2` version-unification follow-up is tracked in #844.

The remaining BaseContainer cancellation-outcome parity, broader runner-to-ETW integration coverage, and PR-description scope wording are non-blocking observations for this approval.

bbonaby (approved):

PR #792, opened 2026-08-09
[LXC] Hook the firewall chain onto the bridge port so it actually filters

dhoehna:
## The problem

The per-container firewall chain was hooked into `FORWARD` with `-i <veth>`
only.  That matches nothing whenever the veth is enslaved to a bridge, which
is the default LXC topology: the packet is bridged onto `lxcbr0` and then
routed off it, so by the time `FORWARD` sees it the input interface is the
bridge, never the veth.

The chain was created correctly, populated correctly, hooked without error,
and traversed by zero packets.

Measured on a live container before this change, with `defaultPolicy: block`
and no allowed hosts:

```
Chain MXC-CLI-LXC-Net-Probe (1 references)
 pkts bytes target     prot opt in     out
    0     0 ACCEPT     0    --  lo     *
    0     0 ACCEPT     0    --  *      *      state RELATED,ESTABLISHED
    0     0 ACCEPT     17   --  *      *      udp dpt:53
    0     0 ACCEPT     6    --  *      *      tcp dpt:53
    0     0 DROP       0    --  *      *
```

Every counter zero, the closing DROP included, and a fetch from inside the
container succeeded anyway.  Adding a counting rule to the same `FORWARD`
chain on the same traffic separated the two cases cleanly:

| rule in FORWARD | packets |
|---|---|
| `-i lxcbr0 -j <counter>` | 11 |
| `-i <veth> -j MXC-…` (what the code installed) | 0 |

## The fix

Install a second hook per family matching `-m physdev --physdev-in <veth>`.
That match names the bridge port the packet entered on, so the chain stays
scoped to a single container -- matching the bridge itself would apply one
container's policy to every container sharing it.  The two rules are mutually
exclusive for any given packet, so a directly routed veth is still carried by
the `-i` rule and nothing is counted twice.

Fail closed on the two conditions that would leave the chain unreachable
again, in the same voice as the missing-veth refusal from #790:

- a bridged veth whose `bridge-nf-call-{ip,ip6}tables` toggle is absent or `0`,
  because bridged packets are then not delivered to iptables at all;
- a bridged veth whose physdev hook will not install, because that rule is the
  only one its packets can match.

On a directly routed veth the physdev rule is redundant, so a kernel without
the match warns rather than refusing a container that is in fact filtered
correctly.

Teardown removes both forms, built from the same builders used at insertion so
a delete cannot drift from the insert it has to match, and the chain delete now
waits on both hooks because either surviving one still references the chain.

## Why the diff is larger than the earlier slices in this series

A hook that installs cleanly and matches nothing is indistinguishable from a
working firewall in every test this repository had.  Shrinking the change to
match the shape of #788/#789/#790 would have meant shipping one half of it.

## Verification

Live containers, same binary, same configs, before and after:

| scenario | before | after |
|---|---|---|
| `block`, no allowed hosts | fetch succeeded | **blocked** |
| `block`, `api.github.com` allowed | — | **reachable** |
| all five existing network E2E scripts | pass | **pass** |
| FORWARD references / chains after teardown | 0 / 0 | **0 / 0** |

The positive control matters as much as the negative one: blocking everything
would also have produced a blocked result.

- `cargo test -p lxc_common --lib` — **144 passed, 0 failed** (121 before).
- `cargo clippy -p lxc_common --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- Mutation testing over nine seeded defects, including the exact bug fixed
  here — **9 caught, 0 survivors**.

## Tests

The 23 unit specs were written against the documented contract by an author
who did not read the implementation, and live in their own file.

The new E2E script exists because unit tests cannot see this failure at all.
Every existing network script asserts that the FORWARD hook was *installed* --
a log line that printed correctly throughout.  The new one asserts the
guarantee: a destination the policy does not allow must be unreachable from
inside the container, and an explicitly allowed one must still be reachable.

It was checked in both directions.  Against this branch it passes.  Against
the implementation from the parent commit it fails on the deny case with
`egress succeeded under a default-block policy with no allowed hosts`, which
is the regression it exists to catch.

Note for reviewers: no CI workflow currently runs the LXC E2E scripts, which
is a large part of how this shipped unnoticed.

## Base

Targets `main` so CI runs: the `Build` workflow only triggers for pull requests
whose base is `main`, `feature/*`, or `user/*`, and `*` does not match `/`, so
a PR based on `user/dahoehna/...` gets no build at all.

It therefore carries #790's two commits as well as its own, because it rewrites
the hook block immediately adjacent to that PR's `else` branch in the same
function.  Once #790 merges this diff shrinks to its own two commits with no
action needed.  **Review commits `f51dc1b` and `b9946e3`;** the two below them
are #790, already reviewed there.

## Not in this PR

- **Deny-wins precedence** and **unresolvable-Deny fail-closed** move to the
  next slice.  Both rearrange or add rules *inside* the chain, which is
  cosmetic until the chain is reachable -- which is what this PR fixes.
- **The blanket port-53 hole** stays open for now.  Narrowing it requires
  knowing which resolver addresses are legitimate, and no schema field carries
  that today; it is also load-bearing, since containers resolve names through
  it.  It waits on the GA network schema alongside #634 and #627.

Refs AB#62830341.

### DIFF
--- a/src/backends/lxc/common/src/network_iptables.rs
+++ b/src/backends/lxc/common/src/network_iptables.rs
@@ -8,6 +8,7 @@
 //! interface.
 
 use std::net::{IpAddr, Ipv6Addr, ToSocketAddrs};
+use std::path::Path;
 use std::process::Command;
 
 use wxc_common::logger::Logger;
@@ -18,6 +19,16 @@ enum IpFamily {
     V4,
     V6,
 }
+
+/// Where the kernel reports per-interface attributes. Injectable in tests via
+/// the `_in` form of the lookup below.
+const SYSFS_NET_ROOT: &str = "/sys/class/net";
+
+/// Toggles that decide whether bridged packets are handed to iptables and
+/// ip6tables at all. A bridged container's chain is unreachable unless the
+/// matching one reads `1`.
+const BRIDGE_NF_CALL_IPTABLES: &str = "/proc/sys/net/bridge/bridge-nf-call-iptables";
+const BRIDGE_NF_CALL_IP6TABLES: &str = "/proc/sys/net/bridge/bridge-nf-call-ip6tables";
 
 /// Whether a host-list entry produces an ACCEPT or a DROP rule. Local to this
 /// backend: it distinguishes `allowedHosts` from `blockedHosts` and is not a
@@ -68,6 +79,8 @@ pub(crate) struct CreatedResources {
     v6_chain: bool,
     v4_hook: bool,
     v6_hook: bool,
+    v4_physdev_hook: bool,
+    v6_physdev_hook: bool,
 }
 
 /// Flush and delete the chain, reporting whether it is still owned afterward.
@@ -106,7 +119,12 @@ impl CreatedResources {
     /// compiled on every target so Windows and macOS CI still type-check it.
     #[cfg_attr(not(target_os = "linux"), allow(dead_code))]
     fn is_empty(&self) -> bool {
-        !self.v4_chain && !self.v6_chain && !self.v4_hook && !self.v6_hook
+        !self.v4_chain
+            && !self.v6_chain
+            && !self.v4_hook
+            && !self.v6_hook
+            && !self.v4_physdev_hook
+            && !self.v6_physdev_hook
     }
 
     /// Test-only constructor so `signal_cleanup`'s tests can build a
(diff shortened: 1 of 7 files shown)


dhoehna:
Superseded by #798.

PR #796, opened 2026-08-09
[LXC] Make deny rules win over allow rules and fail closed on an unresolvable block

dhoehna:
## Problem

`iptables` evaluates a chain top to bottom and stops at the first matching
rule, so precedence in a chain is decided entirely by emission order.
`build_policy_rules_logged` emitted every `allowedHosts` ACCEPT rule before
every `blockedHosts` DROP rule.

A destination named in both lists was therefore reachable, and any allow entry
broad enough to cover a blocked address — `0.0.0.0/0`, or a CIDR containing it
— silently defeated the block.  The chain looked fully populated and filtered
nothing on that destination.

Measured on the previous commit's binary, same script and same host:

```
Programmed iptables rule: -A MXC-CLI-LXC-Net-DenyWins -d 0.0.0.0/0 -j ACCEPT
Programmed iptables rule: -A MXC-CLI-LXC-Net-DenyWins -d 0.0.0.0/0 -j DROP
MXC_NET_ALLOWED
```

A second defect sits in the same function.  An entry that resolves to nothing
produces no rule, and the code warned and continued in every case.  Whether
that is a hole depends on the closing default rule:

- `defaultPolicy: block` — the closing DROP already denies the destination.
  The unwritten rule was **redundant**.  Warning is correct.
- `defaultPolicy: allow` — the chain ends in ACCEPT, so the unwritten DROP was
  the only thing that would have denied that destination.  Skipping it
  **silently converts a deny into an allow**.

## Change

Emit `blockedHosts` rules before `allowedHosts` rules, so a deny always wins
over an overlapping allow.

Return `Result` from `build_policy_rules_logged` and fail closed when — and
only when — the default policy is `Allow` and a `blockedHosts` entry resolved
to nothing.  `install_firewall_rules` propagates with `?`, so the chains
created so far are rolled back and the caller does not receive a container it
believes is confined.

## Why not error on every unresolvable deny entry

That was the first design, and it is wrong.  `tests/configs/lxc_network_test.json:19`
carries `blockedHosts: ["evil.example.com"]` under `defaultPolicy: block`, and
that hostname is NXDOMAIN — verified by resolving it rather than by reasoning
about it.  The blunt predicate would have broken the flagship LXC network
end-to-end config for a case that has no security hole in it, because the
closing DROP already covers it.

## Residual gap, deliberately left open

Under a DROP default, an allow entry broad enough to cover a destination whose
deny rule went unwritten still reaches that destination.  Detecting it
requires the address the failed entry was *meant* to resolve to, which is by
definition unavailable.  No predicate over the policy text can be complete,
and a partial one would imply a guarantee this code cannot make.  It is
documented in the source and in `docs/lxc-support/lxc-backend.md` rather than
half-fixed.

## Tests

Twelve unit tests written from the documented contract by an agent that never
opened `network_iptables.rs`.  A test written after reading the
implementation encodes that implementation's bugs as expected behavior.

Nine mutants, each a mistake a person could plausibly make here — restore the
old order, error on every unresolvable block, error on unresolvable allows,
never error, invert the default-policy test, swap the jump targets, drop the
warning, leak IPv6 into the IPv4 bucket, omit the host name from the error:

```
caught=9  survived=0  harness_bugs=0
source restored byte-identical: True
```

Mutant 1 is the load-bearing one.  Two tests pinning the old
allow-before-block order were deleted in the implementation commit, and a
deletion with no replacement drops coverage silently while the suite stays
green.  Killing mutant 1 proves the replacement exists.

## End-to-end guard, and proof that it is a guard

`tests/scripts/run_lxc_network_deny_precedence_test.sh` runs the real binary
against a config whose allowed and blocked lists both contain `0.0.0.0/0` and
`::/0`.  Literal CIDRs rather than a hostname: a hostname is resolved
separately for each list entry, so round-robin DNS could hand back different
addresses for the allow and the deny and make the verdict depend on which
address the fetch picked.

The control config allows the same destination and blocks nothing, so it must
come back reachable.  Without it, a host with no egress would produce the same
blocked verdict on the overlap case and look exactly like a pass.

The guard was run against the previous commit's binary in an isolated
worktree to confirm it discriminates:

| binary | rule order | overlap verdict | guard |
|---|---|---|---|
| `b9946e3` | ACCEPT then DROP | `MXC_NET_ALLOWED` | **FAIL, exit 1** |
| this PR | DROP then ACCEPT | `MXC_NET_BLOCKED` | **PASS** |

The control passed in both runs, so the difference is the ordering and not a
host that lost its network.

## Gates

154 unit tests, clippy `-D warnings` clean, fmt clean, all seven LXC
end-to-end scripts pass.

## Stack

Slice 5 of six re-implementing #632 fresh off `main`.  Sits on #788, #789,
#790, and #792; targets `main` so CI actually runs — a `user/*` base matches
no branch filter in `Build.yml` and yields one check instead of 23.  Review
the top two commits.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/796)

Refs AB#62830341.

### DIFF
--- a/src/backends/lxc/common/src/network_iptables.rs
+++ b/src/backends/lxc/common/src/network_iptables.rs
@@ -8,6 +8,7 @@
 //! interface.
 
 use std::net::{IpAddr, Ipv6Addr, ToSocketAddrs};
+use std::path::Path;
 use std::process::Command;
 
 use wxc_common::logger::Logger;
@@ -18,6 +19,16 @@ enum IpFamily {
     V4,
     V6,
 }
+
+/// Where the kernel reports per-interface attributes. Injectable in tests via
+/// the `_in` form of the lookup below.
+const SYSFS_NET_ROOT: &str = "/sys/class/net";
+
+/// Toggles that decide whether bridged packets are handed to iptables and
+/// ip6tables at all. A bridged container's chain is unreachable unless the
+/// matching one reads `1`.
+const BRIDGE_NF_CALL_IPTABLES: &str = "/proc/sys/net/bridge/bridge-nf-call-iptables";
+const BRIDGE_NF_CALL_IP6TABLES: &str = "/proc/sys/net/bridge/bridge-nf-call-ip6tables";
 
 /// Whether a host-list entry produces an ACCEPT or a DROP rule. Local to this
 /// backend: it distinguishes `allowedHosts` from `blockedHosts` and is not a
@@ -68,6 +79,8 @@ pub(crate) struct CreatedResources {
     v6_chain: bool,
     v4_hook: bool,
     v6_hook: bool,
+    v4_physdev_hook: bool,
+    v6_physdev_hook: bool,
 }
 
 /// Flush and delete the chain, reporting whether it is still owned afterward.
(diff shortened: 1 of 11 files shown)


dhoehna:
Superseded by #798.

PR #797, opened 2026-08-09
[LXC] Correct the network policy docs and make the E2E suite gate in CI

dhoehna:
## Problem

The documentation described a firewall that no longer exists, and the
end-to-end scripts that would have caught the drift were never run by CI.

Four claims in `docs/lxc-support/lxc-backend.md` were false against the code:

| Claim | Reality |
|---|---|
| Policy table left precedence unspecified | Deny now wins; ordering is the entire mechanism |
| Unresolvable entries are always "skipped, leaving the rest of the policy in force" | Under an accepting default, an unresolvable blocked host is now fatal |
| Chains are hooked into `FORWARD` "by matching the host-side veth as the input interface" | Omits the `--physdev-in` bridge-port rule and the `br_netfilter` requirement |
| "If MXC cannot discover the container veth, it skips the `FORWARD` hook with a warning" | That path returns an error and rolls back |

## What an independent review caught in my first draft

All three were mine, and all three were wrong in the comfortable direction —
they made the firewall sound stronger than it is.

- **"A deny always wins" is false.** The base chain accepts UDP and TCP port 53
  unconditionally and is installed *ahead* of the generated policy rules, so
  DNS traffic to a blocked destination is accepted before its DROP is reached.
  Narrowing that rule requires knowing which resolver addresses are
  legitimate, and no schema field carries them.  The honest move is to
  document the exemption, not to imply a guarantee the chain does not provide.
- **A hostname in both lists is resolved twice.** Each entry resolves
  independently, so round-robin DNS can hand an address to the allow that the
  deny never saw.  The guarantee holds for *addresses*, not names.  This was
  already known — it is exactly why the deny-precedence E2E guard uses literal
  CIDRs — and it still did not make it into the prose.
- **"Two rules per family" is not unconditional.** On a directly routed veth a
  missing `physdev` match warns and continues, because the interface rule is
  the one that matches there.  Only on a bridged veth is it fatal.  The IPv6
  bridge toggle is checked separately and went unmentioned.

## CI

`lxc-e2e.yml` runs the suite on a provisioned Ubuntu runner.  **No workflow
executed these scripts at all before this**, which is much of how a firewall
that filtered nothing shipped green: the assertions existed and nothing ran
them.

The workflow enables `br_netfilter` explicitly.  Without it a bridged veth
never reaches `FORWARD`, every rule installs cleanly, nothing fires, and the
network tests pass against a firewall that filters nothing — the exact failure
they exist to detect.

`MXC_LXC_TESTS_REQUIRE_EXECUTION` turns an honest skip into a failure.  A
developer box legitimately lacks ip6tables or LXC and should run what it can,
so a skip stays a warning there.  A runner provisioned specifically to execute
this suite is different: a skip means a prerequisite disappeared, and without
this the gate goes green while testing nothing.

Verified four ways:

| Mode | Prerequisites | Exit |
|---|---|---|
| normal | present | 0 (14 passed) |
| strict | present | 0 (14 passed) |
| strict | binary removed | **1**, naming the skipped tests |
| normal | binary removed | 1 |

## Stack

Slice 6 of six re-implementing #632 fresh off `main`.  Sits on #788, #789,
#790, #792, and #796; targets `main` so CI runs.  Review the top commit.

The new workflow's first run on this PR is itself the experiment — if the
GitHub-hosted runner cannot provide LXC or `br_netfilter`, strict mode will
say so loudly rather than passing silently.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/797)

Refs AB#62830341.

### DIFF
--- a/.github/workflows/lxc-e2e.yml
+++ b/.github/workflows/lxc-e2e.yml
@@ -1,0 +1,99 @@
+name: LXC E2E Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  lxc-e2e:
+    name: LXC-Exec Container and Network Policy
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Rust toolchain
+        run: rustup update stable
+
+      - name: Point cargo at the MxcDependencies feed
+        uses: ./.github/actions/setup-cargo-feed
+
+      - name: Cache Rust build artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src
+
+      - name: Install LXC and firewall tooling
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y \
+            lxc lxc-templates lxc-utils iptables debootstrap uidmap bridge-utils
+
+      # A bridged veth only reaches the FORWARD chain while br_netfilter is
+      # delivering bridged packets to iptables. Without it the firewall rules
+      # install cleanly and never fire, so the network policy tests would pass
+      # against a firewall that filters nothing.
+      - name: Enable bridge netfilter
+        run: |
+          sudo modprobe br_netfilter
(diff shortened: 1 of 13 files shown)


dhoehna:
Superseded by #798.

PR #801, opened 2026-08-10
[WSLC] State-aware sandbox lifecycle over the experimental.wslc surface (PR 2b/3)

SohamDas2021:
## Summary
PR 2a/3 (#767) hardened the per-user WSLc daemon internals. This PR (2b/3) puts the
daemon to work: it wires the **state-aware sandbox lifecycle** (provision → start →
exec → stop → deprovision) through the public `experimental.wslc.*` wire schema, so a
caller can drive a long-lived WSLc container across multiple invocations instead of
one-shot run-to-completion.

No new public SDK types yet — that's PR 3/3. This PR lands the Rust backend, wire/schema
surface, engine dispatch, and the E2E harness.

Changes:

- **State-aware backend** — `wslc/common/state_aware.rs` implements `StatefulSandboxBackend`
  (dispatch prefix `wslc`), translating each lifecycle phase into daemon protocol frames
  over the owner-only named pipe via `daemon_client`. `policy.rs` maps + validates the
  per-phase policy (filesystem/network/host rules honored at provision; rejected where the
  backend can't enforce them).
- **Wire + schema** — per-phase `Wslc*Phase` config added to the wire model
  (`wxc_common::wire`); regenerated dev schema (`schemas/dev/mxc-config.schema.0.8.0-dev.json`)
  and generated TS wire types (`sdk/node/src/generated/wire.ts`) — both codegen artifacts,
  not hand-edited.
- **Engine + dispatch wiring** — `mxc_engine` state-aware arm for WSLc, and the
  `wslc` sandbox-id prefix now resolves to `ContainmentBackend::Wslc` in
  `state_aware_dispatch` (with a unit test).
- **Daemon request handling** — daemon `main.rs` handles the state-aware phase requests.
- **Fixtures + E2E** — 18 `tests/configs/wslc_state_aware_*.json` fixtures (provision/start/
  exec/stop/deprovision plus rejection cases for denied paths, hosts, proxy, and filesystem)
  and a multi-invocation `tests/scripts/run_wslc_state_aware_tests.ps1` harness with
  warm-reuse and idle-teardown assertions.
- **Docs** — `docs/wsl/wslc-state-aware.md` (fixture ordering + sandbox-id substitution)
  and a `copilot-instructions.md` update.

## Testing

- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` clean.
- Daemon + `wslc_common` unit tests green (incl. the `wslc`-prefix dispatch resolution test);
  the G8 `notify_one` idle-wakeup regression test and the daemon `resolve_duration` env-override
  tests both pass.
- `build.bat --with-wslc` succeeds.
- WSLc one-shot corpus (`run_wslc_all_tests.ps1`) passing; the state-aware harness
  (`run_wslc_state_aware_tests.ps1`) drives provision → start → exec → exec → stop → deprovision
  on a WSL2 host.

## Coming in the pipeline

- **PR 2c/3 — live exec output streaming + provision fs-delegation gate:** #806 — streams
  live stdout/stderr from state-aware `exec` and adds the provision-time filesystem
  delegation gate.
- **PR 3/3 — TypeScript SDK:** `Wslc*Config`/`*Result` types + branded `SandboxId<'wslc'>`
  and helper prefix wiring, mirroring the LXC state-aware SDK surface.

Let me clean up the temp file I created.

## 🔗 References

- Stacked on #767 (PR 2a/3, daemon hardening)

## ✅ Checklist

- [ ] Linked to an issue
- [x] Updated documentation (`docs/wsl/wslc-state-aware.md`)
- [x] Updated [Copilot instructions](.github/copilot-instructions.md)

## 📋 Issue Type
- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/801)


src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,477 @@
...
+
+        let client = connect_daemon()?;
+        let result = client
+            .exec(ExecConfig {
+                sandbox_id: sandbox_id.to_string(),
+                script_code: request.script_code.clone(),
+                working_directory: request.working_directory.clone(),
+                env,
+                timeout_ms: request.script_timeout,
+            })
```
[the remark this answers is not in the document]
SohamDas2021: Planned in PR 2c/3

tests/scripts/run_wslc_state_aware_tests.ps1
```diff
@@ -0,0 +1,989 @@
...
+# output path lands and an INFO note until then, keeping the lifecycle-mechanic
+# coverage green in the interim without masking a hard-assert regression.
+function Probe-ExecOutput {
+    param([bool]$Condition, [string]$Message)
+    if ($Condition) {
+        Write-Host "  PASS: $Message" -ForegroundColor Green
+    } else {
+        Write-Host "  INFO (exec-output deferred): $Message" -ForegroundColor DarkYellow
+    }
+}
```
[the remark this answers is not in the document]
SohamDas2021: Planned in PR 2c/3

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,565 @@
...
+            }
+            None => request,
+        };
+
+        let volumes = build_daemon_volumes(request)?;
+        let network = map_network(request);
+
+        let client = connect_daemon()?;
+        let sandbox_id = client
+            .provision(ProvisionConfig {
```
MGudgin: **High (reliability/security) - Bound daemon lifecycle requests so one hung SDK call cannot wedge every sandbox.**

This is the first production call path into the persistent daemon. I verified that `DaemonClient::call()` subsequently blocks in synchronous `read_frame`/`read_exact` with no response deadline, while the byte-identical daemon implementation processes provision/start/exec/stop/deprovision on one apartment-affine worker thread. If any underlying SDK call hangs, this phase process waits forever and every later lifecycle command - including stop, deprovision, the idle watchdog's count, and shutdown - queues behind it.

**Fix:** Add an overall client response deadline and/or phase timeout fields enforced daemon-side. Teardown should remain possible, or at minimum fail with an actionable timeout instead of blocking indefinitely. Add a regression test with a deliberately non-responsive worker/pipe.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,565 @@
...
+        sandbox_id: &str,
+        request: &ExecutionRequest,
+        _config: Option<()>,
+    ) -> Result<ExecHandle, MxcError> {
+        // Cooperative proxy: inject HTTP(S)_PROXY (and scrub caller-supplied
+        // proxy vars). The url-form requirement is enforced in `validate_exec`.
+        let env = if request.policy.network_proxy.is_enabled() {
+            let proxy_url = exec_proxy_url(request).ok_or_else(|| {
+                MxcError::policy_validation(
+                    "WSLc: network.proxy requires the 'url' form (a routable proxy URL)",
```
MGudgin: **Low (maintainability/simplicity) - Do not duplicate the proxy-form validation and error text.**

`validate_exec_policy()` has already rejected non-URL proxy forms before `exec()` is called, using the more complete `ERR_PROXY_URL_FORM` message. This second fallible check is unreachable through both dispatch paths and has already drifted from the canonical explanation in its first commit.

**Fix:** Rely on the validated invariant (with an invariant assertion if desired), or share one helper/constant rather than maintaining a second user-facing error.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -972,6 +972,14 @@ fn convert_wire_config(
     // to `Block` either way).
     policy.network_specified = cfg.network.is_some();
     if let Some(net) = cfg.network {
+        // Presence of any network *mode* field (everything except `proxy`), so
+        // post-provision phases can reject an immutable-posture change by
+        // presence while still accepting a proxy-only network block.
+        policy.network_mode_specified = net.default_policy.is_some()
```
MGudgin, line 978: **Medium (correctness/test coverage) - Enforce the non-provision containment contract before applying one-shot WSLc proxy rules.**

The wire contract says `containment` is provision-only and later phases route by `sandboxId`, but the state-aware parser leaves an explicitly supplied containment on the surrogate passed to `convert_wire_config`. A malformed exec carrying `containment: "wslc"` plus a URL proxy therefore gets contradictory policy behavior: omitting `defaultPolicy` fails the one-shot "proxy requires allow" check, while adding `allow` sets this new presence bit and is later rejected as an immutable post-provision mode change. The request should be rejected once, clearly, as a malformed state-aware envelope.

The new five-way presence calculation also has no parser-level regression test; existing backend tests hand-construct `network_mode_specified: true` and cannot catch a missing operand or accidental inclusion of proxy-only requests.

**Fix:** Reject `containment` for every non-provision phase before calling the shared converter, then add parser tests for each mode field and for a proxy-only block remaining `false`.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,565 @@
...
+                };
+                &normalized_request
+            }
+            None => request,
+        };
+
+        let volumes = build_daemon_volumes(request)?;
+        let network = map_network(request);
+
+        let client = connect_daemon()?;
```
MGudgin, line 71: **Low (testability) - Make daemon startup/readiness timeouts overridable in tests.**

These new production call sites inherit the byte-identical client constants `READY_TIMEOUT = 60s` and `SPAWN_LOCK_TIMEOUT = 75s`. Negative tests for spawn failure, lock contention, or readiness failure therefore take at least a minute, unlike the daemon-side idle timers which already support test overrides.

**Fix:** Resolve the client timeouts through test-only/environment overrides using the same bounded-duration pattern as the daemon, while retaining the current production defaults.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,565 @@
...
+    fn provision(
+        &mut self,
+        request: &ExecutionRequest,
+        config: Option<WslcProvisionPhase>,
+    ) -> Result<ProvisionResult<()>, MxcError> {
+        let image = config
+            .as_ref()
+            .and_then(|c| c.image.clone())
+            .unwrap_or_else(|| DEFAULT_IMAGE.to_string());
+        let image_tar_path = config.and_then(|c| c.image_tar_path);
```
MGudgin: **Medium (test coverage) - Exercise the state-aware `imageTarPath` path end to end.**

This new field is forwarded into `ProvisionConfig`, but none of the added `wslc_state_aware_*` fixtures contains `imageTarPath`. The existing tar-import test uses the one-shot runner, so it cannot catch a state-aware serialization/forwarding regression here.

**Fix:** Add a state-aware provision fixture using the existing tar-import prerequisites and verify the container can start/exec, or add a focused backend test that observes the exact `ProvisionConfig` sent to the daemon.

tests/scripts/run_wslc_state_aware_tests.ps1
```diff
@@ -0,0 +1,1007 @@
...
+
+Run-StateAwareTest "D: provision (host filtering rejected)" {
+    $r = Invoke-StateAware -ConfigFile 'wslc_state_aware_provision_rejected_hosts.json'
+    Assert-True ($r.ExitCode -ne 0) "exit code is non-zero (policy rejected)"
+    $envObj = Parse-Envelope -Stdout $r.Stdout
+    $code = if ($envObj) { $envObj.error.code } else { '<no envelope>' }
+    Assert-True ($code -eq 'policy_validation') "error.code is 'policy_validation' (got '$code')"
+} | Out-Null
+
+Run-StateAwareTest "D: provision (proxy at provision rejected)" {
```
MGudgin, line 624: **Medium (test coverage) - Add a post-provision proxy rejection case.**

This scenario proves proxy rejection at provision, and the harness separately proves that a URL proxy is accepted at exec. It never exercises the start/stop/deprovision path that must reject a supplied proxy, so the real wire parser and dispatcher could silently diverge from the policy unit test and documented honor matrix.

**Fix:** Add one fixture supplying `network.proxy` on start or stop and assert a `policy_validation` result.

docs/wsl/wslc-state-aware.md
```diff
@@ -0,0 +1,148 @@
...
+
+WSLc networking is **all-or-nothing** (`WslcContainerNetworkingMode` `None` vs `Bridged`); there is
+no per-host filtering (the container lacks `CAP_NET_ADMIN`, so `allowedHosts` / `blockedHosts` iptables
+rules do not apply — see the empirical finding in the plan history). Proxy is env-var only.
+
+| Field | provision | start / stop / deprovision | exec |
+|-------|-----------|----------------------------|------|
+| `readwritePaths` / `readonlyPaths` | honored → container volumes | rejected | rejected |
+| `deniedPaths` | rejected if overlapping/nested under a mount (a standalone denied path is accepted); no Deny primitive | rejected | rejected |
+| `network.defaultPolicy` | honored: `Block` → `None`, `Allow` → `Bridged` | rejected | rejected (non-default network) |
```
MGudgin: **Medium (documentation-drift) - Describe post-provision network rejection as presence-based.**

`rejected (non-default network)` reads as though exec rejects only a value different from the provisioned mode. The implementation deliberately rejects any explicit mode field after provision, even an explicit default-valued `defaultPolicy: "block"`, because presence rather than value is the invariant.

**Fix:** State that any explicit network-mode field is rejected post-provision regardless of value; only the exec-phase cooperative proxy is accepted.

src/backends/wslc/common/src/policy.rs
```diff
@@ -0,0 +1,335 @@
...
+pub(crate) fn exec_proxy_url(request: &ExecutionRequest) -> Option<String> {
+    if !request.policy.network_proxy.is_enabled() {
+        return None;
+    }
+    request
+        .policy
+        .network_proxy
+        .address
+        .as_ref()
+        .and_then(|addr| addr.original_url.clone())
```
MGudgin: **Low (performance) - Avoid cloning the proxy URL for presence validation.**

`validate_exec_policy()` calls `exec_proxy_url(request).is_none()`, so this clone allocates a `String` only to discard it. The exec path can clone at the point where it actually needs owned data.

**Fix:** Return `Option<&str>` (for example via `as_deref()`) from the validation helper and clone only when constructing the owned daemon request, if ownership is required there.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,565 @@
...
+}
+
+/// Object-identity normalization (D6) then delegation check (D3) for the
+/// provision phase, mirroring the one-shot runner order. Returns the tightened
+/// policy when aliasing required a change, else `None`; either check maps its
+/// `String` error to a `policy_validation` envelope.
+fn normalize_and_check_delegation(
+    request: &ExecutionRequest,
+) -> Result<Option<ContainerPolicy>, MxcError> {
+    let mut logger = Logger::new(Mode::Buffer);
```
MGudgin, line 348: **Low (performance/diagnostics) - Do not allocate buffered normalization warnings and then silently drop them.**

`normalize_object_conflicts` can write warnings into this `Mode::Buffer` logger, but the logger is discarded immediately after normalization. That both allocates unused strings and hides potentially useful information about policy tightening.

**Fix:** Either propagate/emit the buffered warnings, or provide a deliberate no-op logging path/helper when these messages are intentionally suppressed.

MGudgin (changes requested):
## Summary

Requesting changes for one blocking reliability issue: the new production state-aware calls can wait forever on the daemon's single synchronous worker, allowing one hung SDK operation to wedge every WSLc lifecycle on the host, including teardown. The parser-contract/test gap and the Medium test/documentation findings should also be addressed; Low findings are non-blocking cleanup.

**Range verified:** `aa05d77207476f88a63e2486013307a9e89795ec...e5e4e050b7224e678fe5900d9b5e64766cc73b0b` (36 files, +3094/-439). The merge base exactly matches GitHub's PR base. `git diff` and `gh pr diff` both contain 4,022 lines; their only differences are abbreviated `index` hashes.

**Verified clean, with receipts:** the WSLc state-aware implementation is gated behind experimental dispatch; the new policy validator has direct unit coverage for filesystem, host-filter, proxy-form, and immutable-network rejection; and the E2E harness covers the basic lifecycle, bridged networking, cooperative exec proxy injection, and provision-time rejection cases. No Windows/WSL runtime tests were run as part of this filing.

### Findings kept in the review body

**Medium (proportionality) - Remove the unrelated whole-file line-ending rewrite.** `.github/copilot-instructions.md` reports 379 additions / 375 deletions in the raw diff, but `git diff --ignore-cr-at-eol` reports only 4 additions / 0 deletions. The base blob has 375 CR bytes and the head blob has none. Please restore the existing CRLF convention so this PR contains only its four semantic additions; perform any repository-wide normalization separately with a `.gitattributes` rule.

**Low (documentation-drift) - Update the WSL roadmap entry made obsolete by this PR.** `docs/linux-wsl-roadmap-june-2026.md:458,485` is byte-identical between base and head, so this is not attributed as pre-existing faulty code. However, this PR newly makes its claims that WSLc still needs `StatefulSandboxBackend` and that none of the three Linux/WSL backends implement it false. Please mark the WSLc item addressed and update the summary sentence.

### Verified pre-existing - not independently attributed to this PR

- `src/backends/wslc/common/src/daemon_protocol.rs`, `src/backends/wslc/daemon/src/session_manager.rs`, and `src/backends/wslc/common/src/container_steps.rs` are byte-identical base vs head. They are cited only as supporting evidence for the newly reachable daemon-wedge finding; the new `state_aware.rs` calls are what expose that behavior to production requests.
- `docs/linux-wsl-roadmap-june-2026.md` is byte-identical base vs head; it is raised only because this PR's new implementation makes the roadmap statement newly stale.

### Requested findings not filed after verification

- The global-stdio/null-handle finding was withdrawn: `ExecHandle` sentinel handles are supported by the current dispatcher, whose relay intentionally calls only the waiter until a backend exposes relayable pipe handles. WSLc writes to the executor's stdio by design, which remains capturable by the outer phase process.
- The missing `DaemonClient` injection seam was withdrawn: it is an architectural preference without a demonstrated behavior defect; policy logic and transport behavior already have separate unit/integration seams.
- The `mxc_engine` instruction-list finding was withdrawn: the sentence says "including" rather than claiming an exhaustive list, and this PR adds a dedicated WSLc architecture row that documents the state-aware backend.

src/core/wxc_common/src/config_parser.rs
SohamDas2021: Addressed. convert_wire_state_aware now rejects any containment on non-provision phases before the shared converter, plus parser tests for each mode field and for a proxy-only block staying unspecified.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Done. READY_TIMEOUT/SPAWN_LOCK_TIMEOUT/CALL_TIMEOUT are now env-overridable via unit-tested resolve_duration, mirroring the daemon-side pattern.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Done. Extracted a pure build_provision_config and added a focused backend test asserting the exact image/imageTarPath forwarded into ProvisionConfig (plus the default-image case).

tests/scripts/run_wslc_state_aware_tests.ps1
SohamDas2021: Done. Added start- and stop-phase fixtures supplying network.proxy and asserting policy_validation.

docs/wsl/wslc-state-aware.md
SohamDas2021: Done. docs/wsl/wslc-state-aware.md honor matrix now states any explicit network-mode field is rejected post-provision regardless of value; only the exec-phase cooperative proxy is accepted.

src/backends/wslc/common/src/policy.rs
SohamDas2021: Done. exec_proxy_url now returns Option<&str> (via as_deref()); the validation path no longer allocates, and the exec path clones only where it needs owned data.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Done. normalize_and_check_delegation now emits any buffered normalization notes to stderr (stdout carries the phase envelope) instead of discarding the buffer.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Done. DaemonClient::call() now enforces an overall response deadline, so a wedged daemon returns an actionable timeout instead of blocking teardown forever; added a non-responsive-reader regression test.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Done. Removed the unreachable second url-form check in exec(); it now relies on the invariant already enforced by validate_exec_policy().

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -25,28 +25,52 @@ use wxc_common::state_aware_request::{MxcRequest, ParsedStateAwareRequest, Phase
...
+/// so no entry point can reach an experimental backend without the flag.
+fn require_experimental_optin(
+    backend: &wxc_common::models::ContainmentBackend,
+    parsed: &ParsedStateAwareRequest,
+) -> Result<(), MxcError> {
     if matches!(
         backend,
         wxc_common::models::ContainmentBackend::WindowsSandbox
             | wxc_common::models::ContainmentBackend::IsolationSession
+            | wxc_common::models::ContainmentBackend::Wslc
```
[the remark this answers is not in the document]
SohamDas2021: This isn't specific to WSLC or introduced here, it's a pre-existing, cross-cutting limitation of the SDK surface. WindowsSandbox and IsolationSession already sit in this same gate, and every SDK/FFI path parses via load_mxc_request_from_json, which hard-codes experimental_enabled = false. So, all three experimental state-aware backends are equally unreachable via the SDK today. Won't fixing this.

MGudgin (approved):

PR #802, opened 2026-08-10
Support for passing appId to IsolationSession upon sandbox provision

daamenik:
## 📖 Description
Update sandbox-provision APIs to take in an `appId`. If `""` is passed as the `appId`, attempt to detect the calling process's Package Family Name (PFN) and use `PFN:<pfn>` instead. If calling process is unpackaged, use `appId` verbatim.

Use the `AddUserAsync2` API from the IsoSessionOps Preview API instead of `AddUserAsync()`.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/653)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/802)


src/backends/isolation_session/common/src/app_id.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Returns the current process's Package Family Name, or `None` when the
+/// process has no package identity (unpackaged) or the lookup otherwise fails.
+fn current_package_family_name() -> Option<String> {
+    // Sizing call: a null buffer yields ERROR_INSUFFICIENT_BUFFER plus the
+    // required length (in characters, including the terminator) for a packaged
+    // process, or APPMODEL_ERROR_NO_PACKAGE for an unpackaged one.
+    let mut length: u32 = 0;
+    let rc = unsafe { GetCurrentPackageFamilyName(&mut length, Some(PWSTR::null())) };
+    if rc != ERROR_INSUFFICIENT_BUFFER || length == 0 {
+        return None;
```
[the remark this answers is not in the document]
daamenik: Ignoring this; if there's some kind of error when detecting the PFN, we should just continue.

src/backends/isolation_session/common/src/app_id.rs
```diff
@@ -0,0 +1,134 @@
...
+fn resolve_app_id_with(app_id: Option<&str>, pfn: Option<&str>) -> Option<String> {
+    match app_id {
+        // A caller-chosen, non-empty appId is authoritative — used as-is.
+        Some(app_id) if !app_id.is_empty() => Some(app_id.to_string()),
+        // Empty or absent: prefer the caller's PFN when packaged, otherwise
+        // preserve the original (empty or absent) value verbatim.
+        _ => match pfn {
+            Some(pfn) if !pfn.is_empty() => Some(format!("{PFN_PREFIX}{pfn}")),
+            _ => app_id.map(str::to_string),
+        },
```
adpa-ms: **[ReviewAgent]** These two arms produce **different registration identities for the same app**, and the API uses whichever value it receives verbatim.

The contract: a non-empty `appId` is used as-is; an empty (or omitted) one is what opts into PFN-scoped registration. So the `PFN:` prefix on line 48 is part of the identity, not decoration — and a packaged caller who passes the bare PFN `Contoso.App_8wekyb3d8bbwe` is registered under the literal `Contoso.App_8wekyb3d8bbwe`, which is **a different registration** from the `PFN:Contoso.App_8wekyb3d8bbwe` this line produces for that same app.

That bare form is what every example in this repo shows (`docs/schema.md:251`, `mxc-state-aware-sandbox-api.md:1600,1906,1921`, four SDK tests) and what `state-aware-rust.md:53` instructs packaged callers to pass.

Worth stating explicitly wherever `appId` is documented: to obtain PFN-scoped registration, pass `""` / omit the field, **or** pass the literal `PFN:<pfn>`. A bare PFN is a distinct, non-PFN-scoped identity. The examples should then be updated to whichever form is intended.

docs/isolation-session/state-aware-rust.md
```diff
@@ -50,14 +50,14 @@ without metadata use `()`.
...
...o later phases recover the identity actually used without the caller re-supplying it. **Host gating:** the app-scoped `AddUserAsync2` overload is used only when the host advertises `IsoSessionFeature::AppScopedRegistration` via `GetFeatureLevel` (a positive level); on older hosts — where `GetFeatureLevel` returns a non-positive level or rejects the unknown feature value with an error — MXC falls back to the legacy `AddUserAsync`, which carries no `appId`, so the agent user is provisioned without app association (the resolved value is still recorded in the `sandboxId`). Validated **structurally only** (no control characters; at most 256 characters) — MXC does not judge what a valid application identity looks like, so enforcing a PFN grammar would risk rejecting forms a future OS API accepts. Beyond PFN substitution there is no trimming, case folding, or normalisation. An explicitly-supplied **empty string is a distinct value from absent** on an unpackaged host and round-trips as such; JSON `null` is a second spelling of absent. Rejections surface as `policy_validation` from `validate_provision`, before any OS call. The wire path is `experimental.isolation_session.provision.appId`. |
```
adpa-ms: **[ReviewAgent]** This row tells a packaged caller that `appId` "is the Package Family Name" — i.e. a bare PFN — but a bare PFN and `PFN:<pfn>` are two different registration identities, and only the latter (or an empty/omitted value) is PFN-scoped. Suggest documenting both accepted spellings here.

The same wording appears in `state-aware-typescript.md`, `docs/schema.md:251` and `docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md:1600,1906,1921`.

src/backends/isolation_session/common/src/manager.rs
```diff
@@ -98,19 +112,43 @@ impl IsolationSessionManager {
...
+        ) {
+            ops.AddUserAsync2(
+                &HSTRING::from(app_id.unwrap_or_default()),
+                &HSTRING::new(),
+                &HSTRING::new(),
+            )
+        } else {
+            ops.AddUserAsync(&HSTRING::new(), &HSTRING::new())
+        }
+        .map_err(|e| transport_err(op::ADD_USER, "call failed", &e))?;
```
adpa-ms: **[ReviewAgent]** On this branch the `appId` cannot be passed — the 2-arg `AddUserAsync` has no such parameter — yet `state_aware.rs:93-96` still encodes the resolved value into the `sandboxId`. The id then asserts an app association this path did not make.

Provision has to keep succeeding on these hosts, so this isn't a request to fail. But the recorded value shouldn't disagree with what happened: consider having `add_user` return the identity **actually applied** and encoding that (`None` here), so the `sandboxId` can't over-claim.

Separately, `app_scoped_supported_from` maps **any** `Err` to "not supported", while the doc comment above justifies only the documented unknown-feature rejection. A transport failure on the probe is currently indistinguishable from an old host — worth separating "the host says no" from "the query failed".

src/backends/isolation_session/common/src/error.rs
```diff
@@ -23,7 +23,7 @@ use isolation_session_bindings::bindings::{IsoSessionError, IsoSessionResult};
 /// telemetry.
 pub(super) mod op {
     pub(crate) const ACTIVATE: &str = "IsoSessionOps.ActivateInstance";
-    pub(crate) const ADD_USER: &str = "IsoSessionOps.AddUserAsync";
+    pub(crate) const ADD_USER: &str = "IsoSessionOps.AddUserAsync2";
```
adpa-ms, line 28: **[ReviewAgent]** This constant now reports `AddUserAsync2` for **both** branches, including when the fallback actually invoked the 2-arg `AddUserAsync` — and it covers the whole ladder, since `manager.rs:151-154` and the result-property reads reuse it for `"call failed"` and `"wait failed"`.

The module doc immediately above states the contract this breaks: *"Interface-qualified names of the API operations this backend **invokes** … the values that reach the wire as `error.operation`."* On a legacy host this groups failures under an API that was deliberately not called, which is misleading precisely where the fallback is otherwise invisible. Suggest selecting the operation constant alongside the overload.

src/backends/isolation_session/common/src/state_aware.rs
```diff
@@ -72,13 +72,19 @@ impl StatefulSandboxBackend for IsolationSessionRunner {
...
         // The manager is discarded here — each post-provision phase builds its
         // own from the `sandboxId`. Taking it anyway keeps a single provisioning
         // path with `one_shot`, and proves the service instance that minted the
         // user is live rather than re-activating to find out.
         let (provisioned, _manager) =
-            IsolationSessionManager::add_user().map_err(map_lifecycle_error)?;
+            IsolationSessionManager::add_user(app_id.as_deref()).map_err(map_lifecycle_error)?;
 
         // `appId` rides inside the id so later phases recover it without the
         // caller re-supplying it. Nothing consumes it yet; it is carried for a
```
adpa-ms: **[ReviewAgent]** Both halves of this comment are falsified by this PR. *"Nothing consumes it yet"* — `AddUserAsync2` consumes it now. And the justification continuing on lines 91-92 (*"the caller already has the value it supplied"*) no longer holds when the value was auto-detected: the caller supplied nothing, and has no supported way to learn what was used — the payload after `iso:` is documented as backend-private, and `state-aware-typescript.md:60-63` states the `SandboxId` is opaque and *"nothing in the SDK parses past the `iso:` prefix"*.

Suggest echoing the resolved/applied `appId` in `IsolationSessionProvisionMetadata`, and updating this comment plus the matching justification in `state-aware-rust.md:63-64`.

src/backends/isolation_session/common/src/app_id.rs
```diff
@@ -0,0 +1,134 @@
...
+    let mut buffer = vec![0u16; length as usize];
+    let rc = unsafe { GetCurrentPackageFamilyName(&mut length, Some(PWSTR(buffer.as_mut_ptr()))) };
+    if rc != ERROR_SUCCESS {
+        return None;
+    }
+
+    // `length` counts the null terminator; trim it before decoding.
+    let chars = (length as usize).saturating_sub(1).min(buffer.len());
+    Some(String::from_utf16_lossy(&buffer[..chars]))
+}
```
adpa-ms: **[ReviewAgent]** `current_package_family_name()` returns `None` for **any** failure, not only the no-package case — so a packaged caller hitting a transient lookup error takes the unpackaged path and MXC passes `""` onward.

Since `""` is what opts into PFN-scoped registration, the API then resolves it and registers the caller's PFN — while the `sandboxId` records `""`. The registration is right; MXC's record of it is wrong. Same over/under-claiming theme as the fallback path.

At minimum, log the non-no-package failures so this is diagnosable; ideally distinguish them.

adpa-ms (commented):
**[ReviewAgent]** Independent review by three reviewers on different model families, each covering the whole change; every delegated claim re-verified against code before inclusion. Read-only — no builds or test runs, so the author's validation results are relied upon rather than checked. **Verdict: sign off with findings. Nothing blocking.**

**Findings that can't be anchored inline** (files this PR doesn't touch, but which it makes wrong):

- **`src/core/wxc_common/src/wire.rs:574-577`** — still describes `appId` as merely "Carried inside the `sandboxId`", with no mention of resolution or the OS call. This text is emitted verbatim into `schemas/dev/mxc-config.schema.0.8.0-dev.json`, which is what config authors see via `$schema`. Because `wire.rs` didn't change, the schema doesn't drift from its source — **so `check-schema-codegen.js` still passes and CI cannot catch this one.**
- **`sdk/node/src/state-aware-types.ts:46-59`** — the shipped JSDoc still says "Carried **verbatim**" and "**Nothing consumes it yet**". This PR updated `state-aware-typescript.md`, which *describes* this surface, but not the surface's own documentation.
- **`sandbox_id.rs:74-75` and `:99-104`** — "carried only — nothing in MXC consumes it today", and "MXC is a pass-through carrier here". Note `:81-84` is correct as written and needs no change.

**Verified correct, recorded so it isn't re-litigated:** the `PFN:` colon cannot corrupt the `sandboxId` (base64url payload; only the first colon is structural); the `IsoSessionFeature` constant renames preserve their numeric values and have zero references anywhere in `src/`; `AddUserAsync`'s arity is unchanged, so the fallback call site is unaffected; treating an empty/absent `appId` as opting into PFN detection matches the API's own sentinel semantics; validation ordering is safe in both directions; the Win32 buffer protocol (sizing call, allocation, terminator trim) is correct; and the seven inline resolver tests plus the feature-level test cover the decision table well.

**One product-level note:** on a packaged host there is no way to express "no app association" — an empty value and an omitted one both opt into PFN scoping, so every input except a non-empty literal yields a PFN-scoped registration. Flagging it in case that matters for a caller who wants an unassociated agent user.

src/backends/isolation_session/common/src/app_id.rs
```diff
@@ -0,0 +1,134 @@
...
+/// - A non-empty `appId` is returned verbatim (the caller chose an explicit
+///   application identity).
+/// - An empty (`Some("")`) or absent (`None`) `appId` triggers PFN detection:
+///   a packaged caller resolves to `Some("PFN:<pfn>")`; an unpackaged caller —
+///   or any failure to obtain a PFN — keeps the original value unchanged
+///   (`Some("")` stays `Some("")`, `None` stays `None`), so the empty-vs-absent
+///   distinction is preserved on hosts without package identity.
+pub(crate) fn resolve_app_id(app_id: Option<&str>) -> Option<String> {
+    resolve_app_id_with(app_id, current_package_family_name().as_deref())
+}
```
adpa-ms: **[ReviewAgent]** `current_package_family_name()` is an argument expression, so it is evaluated eagerly — the lookup runs on every provision, including when `app_id` is `Some("explicit")` and the result is discarded at line 44. Harmless today, but branching on presence before consulting the detector would make the code visibly do the lookup only when it's needed, and it's the same restructure as any other edit here.

docs/isolation-session/state-aware-rust.md
daamenik: I'll update the documentation to explicitly state that if the app is packaged, the user should provide the PFN in the format "PFN:\<pfn\>"

src/backends/isolation_session/common/src/app_id.rs
daamenik: I'll update the documentation to explicitly state that if the app is packaged, the user should provide the PFN in the format "PFN:\<pfn\>"

adpa-ms: **[ReviewAgent]** Resolved differently, and better than proposed: after discussion with the repository owner the change was reshaped so MXC no longer resolves `appId` at all. `app_id.rs` is deleted and the caller's value is passed through verbatim, so the two-representations problem is eliminated rather than documented around. The `PFN:<packageFamilyName>` requirement is now stated in `wire.rs`, `models.rs`, the SDK types and both guides.

Closing this thread. The remaining example-consistency work is tracked separately.

docs/isolation-session/state-aware-rust.md
adpa-ms: **[ReviewAgent]** Done — `state-aware-rust.md:53` now states the `PFN:<packageFamilyName>` requirement and warns that a bare PFN will not receive PFN-scoped registration. Verified at `bb9c9f79`.

Two follow-ups are tracked separately rather than here: `.github/copilot-instructions.md:116` still carries the pre-reshape wording, and the examples still show a bare PFN. Closing this thread.

src/backends/isolation_session/common/src/state_aware.rs
adpa-ms: **[ReviewAgent]** The stale comment is fixed — it no longer claims nothing consumes `appId`, and it no longer justifies the metadata omission by saying the caller already supplied the value.

The metadata half of this finding is **withdrawn**: per the repository owner, the `sandboxId` is an opaque identifier for a sandbox, not an information source. That it can be decoded is incidental, so a caller was never entitled to read the resolved value back — no defect. Closing.

adpa-ms (approved):

MGudgin (approved):

PR #803, opened 2026-08-10
fix(logger): deliver security warnings explicitly instead of writing to stderr

caarlos0:
## 📖 Description

`Logger::warning_line` wrote every security warning to the process's stderr unconditionally.

That is fine for `wxc-exec`, which owns its terminal. But `wxc_common` is also linked into `mxc_engine` and therefore `mxc-sdk`, so an *embedding application* inherits the write. For a host that draws its own terminal UI, an unannounced stderr line lands in the middle of its display and corrupts it — with no way to opt out.

The fix is to stop the *library* writing to stderr, and instead deliver the warning explicitly on every surface that needs it.

**1. The library stops writing to stderr.** The `eprintln!` is removed from `Logger::warning_line`. Warnings are still recorded, and still forwarded to the file sink and the diagnostic console. Removing it left `warning_line`'s body byte-identical to `log_diagnostic_line`, so it now delegates to it — matching how `log_line` already does.

**2. Every consumer gets an explicit delivery path:**

| Surface | How the warning is delivered |
|---------|------------------------------|
| Rust | `Sandbox::warnings()` / `Output::warnings` (already existed) |
| C# | `RunResult.Warnings` (**new**) |
| C ABI (`mxc_ffi`) | `MxcRunResult::warnings_json_utf8` (**new**, JSON array of strings) |
| `wxc-exec` | drains `logger.warnings()` and prints to stderr after the run (**new**) |

Without the `wxc-exec` drain, `wxc-exec --audit` would have been **completely silent on a successful run**: the "--audit enabled" line goes into a `Mode::Buffer` logger whose `get_buffer()` flushes all sit on *error* branches, and the direct `eprintln!` requires `--audit-verbose`. The drain sits before the dry-run branch, which calls `process::exit` and would otherwise swallow it. It also covers `captureDenials.mode: "allow"` — both routes that inject `permissiveLearningMode` populate `policy.capabilities` before the run, so one drain catches both.

`warning_line` only writes to the file sink and diagnostic pipe, never to the console/buffer, so the drain cannot double-print under `--debug` (`Mode::Console`).

**3. A regression test that can actually fail.** The pre-existing test only inspected the debug buffer, so reintroducing `eprintln!` would still have passed it. The new test runs `warning_line` in a child process — fd 2 is process-global, so redirecting it in-process would race the parallel suite — and asserts the warning never reaches stderr. `--nocapture` is required or libtest swallows the child's streams and the assertion goes vacuous. Verified by mutation: restoring `eprintln!` makes it fail.

**4. WSLC daemon call site.** Merging `main` brought in a new `warning_line` caller in the WSLC daemon, recording an exec that completed after its client disconnected — commented *"so a completed exec is never silently lost"*. That daemon builds `Logger::new(Mode::Console)` with no file sink and never drains its `warnings`, so after this change the message would have gone nowhere, and its long-lived worker loop would accumulate those strings for the daemon's lifetime. It's an operational diagnostic rather than a security warning, so it now uses `log_line`.

### 🔎 Scope: what this does *not* change

The **captureDenials deliverable** is untouched and was never a gap — denials are written to a JSON file, `wxc-exec` prints a one-line JSON pointer to it, and both `Output::output_metadata` and `MxcRunResult::output_metadata_json_utf8` expose it. Only the security *warning* moved.

### ⚠️ Known limitation

Warnings now print **after the workload finishes** rather than at container spawn. A long-running `--audit` run shows nothing until it completes, and a non-terminating one never shows it. Making this timely again needs either a pre-run capability check in `wxc-exec` or an opt-in warning sink on `Logger`; both were considered and deliberately deferred to keep this change minimal.

## 🔗 References

No tracking issue — reported directly against the `mxc-sdk` embedding path.

Related but deliberately **out of scope**: `Logger::enable_file_sink` opens its log create+append, follows symlinks, and leaves mode `0644`. Untouched here and equally reachable from the CLI's `--log-file`; hardening it deserves its own change.

## 🔍 Validation

- `cargo test -p wxc_common -p mxc_engine -p mxc-sdk -p appcontainer_common -p mxc_ffi` → all green
- `cargo clippy` on the above `--all-targets -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean
- `node scripts/check-dotnet-bindings-codegen.js` → OK, csbindgen emits `warnings_json_utf8` after `output_metadata_json_utf8`, so the `#[repr(C)]` layout stays consistent

`wxc` and the WSLC daemon are Windows-only and do not build on the macOS dev host, so they were verified by cross-target type-check:

- `cargo check -p wxc -p wxc_wslc_daemon --target x86_64-pc-windows-msvc` → clean
- `cargo clippy --target x86_64-pc-windows-msvc -- -D warnings` → clean

The new stderr test was validated by mutation, not just by passing.

**Not runtime-tested end to end** — that needs a Windows host with the AppContainer backend, and the C# SDK needs a .NET SDK that isn't on this machine. The `--audit` console output and `RunResult.Warnings` are worth a manual confirmation before merge.

### ℹ️ Note: flaky `plm` test on `windows / x64`

CI is green on the current head. Earlier runs of this PR failed on `plm::start::tests::inherited_output_handle_cannot_block_control_completion` (panic at `src/host/plm/src/start.rs:600`, `"descendant PID should be retained in partial stdout"`). That failure is **pre-existing and unrelated to this PR** — `main`'s own Build runs for 604ffee (#829) and 2d8bad3 (#834) fail on the same job with the identical panic, `git diff origin/main HEAD -- src/host/plm/` is empty, and `plm` never imports `wxc_common::logger`.

It passes on the latest run, which confirms it is timing-sensitive rather than a real break: the test spawns two nested PowerShell processes against a 5s timeout and asserts on *partial* stdout. Flagging it here because it will keep intermittently reddening unrelated PRs and deserves a fix at its source.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task


richiemsft:
For the known consequences, how do you propose to consume the captureDenial logs?

caarlos0:
@richiemsft I don't know either - I don't have any strong opinions other than we should probably not write to stdio directly as it might interfere with the caller 🤔 

I think we can just print them when called from wxc-exec though (investigating)

caarlos0:
Investigated — pushed in 2f92ff2. Short version: yes, `wxc-exec` can just print them, and now does.

@richiemsft I want to separate two things that I'd conflated in the original description, because they have different answers:

**1. The captureDenials deliverable (the denials themselves)** — this was never affected by this PR, and is already consumable everywhere:

- the denials are written to a JSON file (`{ "denials": [...], "summary": {...} }`)
- `wxc-exec` prints a one-line JSON *pointer* to that file on stderr
- Rust callers get it programmatically via `Output::output_metadata` / `Sandbox::output_metadata()`
- FFI/C# callers get it via `MxcRunResult::output_metadata_json_utf8`

So there's no gap there — the structured path already exists on every surface, and nothing here writes the denials to stdio.

**2. The security *warning* string** (`*** SECURITY WARNING *** permissiveLearningMode is ENABLED...`) — this is the only thing this PR moves. It used to be an unconditional `eprintln!` inside `Logger::warning_line`, which is the part that interferes with an embedding caller's terminal. It's now recorded into `Logger::warnings` and `wxc-exec` drains it after the run, so the CLI output is unchanged while library embedders get silence-by-default and read `warnings()` themselves.

Nice side effect: both routes that inject `permissiveLearningMode` (`--audit` and `captureDenials.mode: "allow"`) populate `policy.capabilities` before the run, so the single drain covers both.

Two caveats I'd rather flag than bury:

- **Timing.** The drain runs after `runner.run()` returns, so the warning now prints when the workload finishes rather than at container spawn. For a long `--audit` run you see nothing until the end, and for a non-terminating one, never. Fixing that properly needs either a pre-run capability check in `wxc-exec` or an opt-in warning sink on `Logger` — I left both out to keep this minimal, happy to do either as a follow-up if you think the delay is unacceptable.
- **The warning string still isn't on `MxcRunResult`** — there's no warnings field, only `output_metadata_json_utf8`. So a C#/FFI caller using `captureDenials.mode: "allow"` gets the denials but not the "deny-by-default is off" warning. Rust and CLI callers are fine. Plumbing warnings through the C ABI feels like its own change — want me to open an issue for it?

Also worth noting: `wxc` is Windows-only so I couldn't run this end to end from my Mac. I verified it with `cargo check`/`clippy` against `x86_64-pc-windows-msvc`, but the actual `--audit` console output deserves a manual eyeball on a Windows host before this merges.

richiemsft:
> Investigated — pushed in [2f92ff2](https://github.com/microsoft/mxc/commit/2f92ff23b907fd88ec24023cc544af46d3520d5c). Short version: yes, `wxc-exec` can just print them, and now does.
> 
> Richie Gomez (@richiemsft) I want to separate two things that I'd conflated in the original description, because they have different answers:
> 
> **1. The captureDenials deliverable (the denials themselves)** — this was never affected by this PR, and is already consumable everywhere:
> 
> * the denials are written to a JSON file (`{ "denials": [...], "summary": {...} }`)
> * `wxc-exec` prints a one-line JSON _pointer_ to that file on stderr
> * Rust callers get it programmatically via `Output::output_metadata` / `Sandbox::output_metadata()`
> * FFI/C# callers get it via `MxcRunResult::output_metadata_json_utf8`
> 
> So there's no gap there — the structured path already exists on every surface, and nothing here writes the denials to stdio.
> 
> **2. The security _warning_ string** (`*** SECURITY WARNING *** permissiveLearningMode is ENABLED...`) — this is the only thing this PR moves. It used to be an unconditional `eprintln!` inside `Logger::warning_line`, which is the part that interferes with an embedding caller's terminal. It's now recorded into `Logger::warnings` and `wxc-exec` drains it after the run, so the CLI output is unchanged while library embedders get silence-by-default and read `warnings()` themselves.
> 
> Nice side effect: both routes that inject `permissiveLearningMode` (`--audit` and `captureDenials.mode: "allow"`) populate `policy.capabilities` before the run, so the single drain covers both.
> 
> Two caveats I'd rather flag than bury:
> 
> * **Timing.** The drain runs after `runner.run()` returns, so the warning now prints when the workload finishes rather than at container spawn. For a long `--audit` run you see nothing until the end, and for a non-terminating one, never. Fixing that properly needs either a pre-run capability check in `wxc-exec` or an opt-in warning sink on `Logger` — I left both out to keep this minimal, happy to do either as a follow-up if you think the delay is unacceptable.
> * **The warning string still isn't on `MxcRunResult`** — there's no warnings field, only `output_metadata_json_utf8`. So a C#/FFI caller using `captureDenials.mode: "allow"` gets the denials but not the "deny-by-default is off" warning. Rust and CLI callers are fine. Plumbing warnings through the C ABI feels like its own change — want me to open an issue for it?
> 
> Also worth noting: `wxc` is Windows-only so I couldn't run this end to end from my Mac. I verified it with `cargo check`/`clippy` against `x86_64-pc-windows-msvc`, but the actual `--audit` console output deserves a manual eyeball on a Windows host before this merges.

Yes please, create an issue so we don't lose track of it. Also thanks for explaining the original description.

richiemsft (dismissed):

src/core/wxc_common/src/logger.rs
```diff
@@ -253,20 +253,15 @@ impl Logger {
...
-    /// for in-process callers.
+    /// Record a security warning for in-process callers.
+    ///
+    /// Deliberately not written to stderr: a library must not write to an
+    /// embedding host's terminal behind its back. Callers read warnings back
+    /// through [`warnings`](Self::warnings) / [`take_warnings`](Self::take_warnings)
+    /// and present them however suits their UI.
     pub fn warning_line(&mut self, msg: &str) {
-        eprintln!("{msg}");
         self.warnings.push(msg.to_string());
```
SohamDas2021, line 263: Shall we add a regression test that captures stderr and verifies warning_line writes nothing while retaining the warning? The current test only checks the debug buffer, so reintroducing eprintln! would still pass and regress this PR’s core guarantee?

src/core/wxc_common/src/logger.rs
```diff
@@ -253,20 +253,15 @@ impl Logger {
...
-        if let Some(ref mut f) = self.file {
-            let secs = SystemTime::now()
-                .duration_since(SystemTime::UNIX_EPOCH)
-                .unwrap_or_default()
-                .as_secs();
-            let _ = writeln!(f, "[{}] {}", secs, msg);
-        }
-        self.diag_accumulate(msg);
-        self.diag_accumulate("\n");
+        self.log_diagnostic_line(msg);
```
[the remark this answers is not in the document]
caarlos0: Addressed across 2f92ff2 (CLI) and 9abf75c (FFI).

**CLI:** `wxc-exec` now drains `logger.warnings()` after the run and prints them to stderr — it owns its terminal, so it opts back in explicitly. Placed before the dry-run branch, since `handle_dry_run_exit` calls `process::exit` and would otherwise swallow them. Both routes that inject `permissiveLearningMode` (`--audit` and `captureDenials.mode: "allow"`) populate `policy.capabilities` before the run, so one drain covers both.

**FFI:** you were right that this was the bigger gap. `MxcRunResult` now has `warnings_json_utf8` (JSON array of strings, null when empty), populated from `Output::warnings`, freed in `free_strings`, and surfaced as `RunResult.Warnings` in the C# SDK. csbindgen emits the new field after `output_metadata_json_utf8`, so the `#[repr(C)]` layout stays consistent.

While adding it I restructured the result construction: both JSON payloads are serialized *before* any C string is allocated, so a failure serializing warnings can't leak the already-allocated metadata string.

One caveat worth stating plainly: the CLI drain runs after `runner.run()` returns, so the warning now prints when the workload finishes rather than at container spawn. A long `--audit` run shows nothing until it completes. Making that timely again needs a pre-run capability check in `wxc-exec` or an opt-in warning sink on `Logger`; I left both out to keep this change minimal, happy to follow up if you'd rather have it.

src/core/wxc_common/src/logger.rs
caarlos0: Good catch — you were exactly right that the existing test was vacuous. Added in 9abf75c.

`warning_line_writes_nothing_to_stderr` spawns a child process that calls `warning_line`, then asserts the warning never appears on the child's stderr. It runs out-of-process because fd 2 is process-global — redirecting it in-process would race every other test in the binary, which run in parallel threads.

Two things that turned out to matter:

- `--nocapture` is required. Without it libtest swallows the child's streams, and the assertion would hold even if `warning_line` started writing to stderr again — i.e. the new test would have been vacuous in exactly the way the old one was. There's a comment on the test saying so, so nobody "tidies" it away later.
- The child also prints a canary to stdout, which the parent asserts on, so "stderr was empty" can't be confused with "the child never ran". That distinction actually caught a real bug in my first draft.

Verified by mutation rather than by assertion: restoring `eprintln!("{msg}")` in `warning_line` makes the test fail with `warning_line wrote the warning to stderr: *** SECURITY WARNING *** child canary`, and removing it makes it pass again.

kanismohammed (approved):

SohamDas2021 (approved):

Issue #804, opened 2026-08-11
[WSLC] State-aware exec: forward piped stdin to the container (blocked on WSLC SDK)


SohamDas2021:
### Description of the new feature / enhancement

The state-aware WSLc backend's exec path (`execInSandbox`) streams container **stdout/stderr** live, but it does **not** forward host **stdin** to the container process. Anything piped to a WSLc exec (e.g. `echo data | ... exec`, or `exec < file`) is dropped, and the container child sees immediate EOF on stdin.

The Windows Sandbox (WSB) state-aware backend already supports **piped (non-TTY) stdin** — the guest agent bridges the child's stdin over its IPC transport (`src/backends/windows_sandbox/guest/src/executor.rs` bridges `TCP → child stdin` via `tokio::io::copy`; `src/backends/windows_sandbox/lifecycle/src/state_aware.rs` pumps host stdin into `Stdin` frames). So WSLc lacking stdin forwarding is a backend **parity gap**, not a design choice.

**This is currently blocked on the WSLc SDK.** Live stdout/stderr streaming is implemented by registering the SDK's `onStdOut`/`onStdErr` process callbacks. Per the WSLc SDK header (`wslcsdk.h`):

> Using any callbacks will consume the IO handles, preventing acquisition through `WslcGetProcessIOHandle`.

Because the daemon registers output callbacks (the basis of live output streaming), `WslcGetProcessIOHandle(WSLC_PROCESS_IO_HANDLE_STDIN)` is unavailable — callback-mode output and handle-mode stdin are **mutually exclusive** in the current SDK. There is no way to forward stdin while also using callback-based output streaming.

Scope: this issue covers **piped (non-TTY)** stdin. Interactive TTY/ConPTY stdin is a further step on top (WSB tracks that separately as `h6-pty-plumbing`).

### Proposed technical implementation details

Once the WSLc SDK allows acquiring the stdin handle alongside output (or provides an equivalent stdin-write API that coexists with output delivery), rearchitect the WSLc daemon exec path from callback-mode to **handle-mode**:

- Register **no** output callbacks; instead acquire all three IO handles via `WslcGetProcessIOHandle`.
- Spawn `ReadFile` threads for stdout/stderr feeding the existing streaming sink + capped capture buffers.
- Spawn a `WriteFile` thread draining client `StreamFrame::Stdin` frames into the container process stdin handle; close the handle on EOF.
[...]

PR #806, opened 2026-08-11
[WSLC] State-aware lifecycle: live exec output streaming + provision fs-delegation gate (PR 2c/3)

SohamDas2021:
## 📖 Description

-----

## Summary

Third increment of the WSLc state-aware sandbox lifecycle. Brings `execInSandbox` to **Windows Sandbox streaming parity (Tier 1)**: stdout/stderr are relayed to the caller **live as the container produces them**, instead of being buffered and returned only at exit. Also closes a **filesystem delegation gap** in state-aware `provision` so it enforces the same object-normalization + delegation + overlap checks as the one-shot path.

### Live exec output streaming (Tier 1)
- The daemon relays container stdout/stderr over a per-exec channel; `state_aware.rs` flushes chunks to the caller as they arrive (with a short coalescing gate) rather than accumulating the full capture buffer first.
- New E2E `wslc_state_aware_exec_drip` fixture proves output is streamed, not buffered — the drip liveness check observes an inter-chunk gap of ~2s on the real WSLc host.
- **Piped stdin remains deferred to Tier 2** (blocked on the WSLc SDK's IO-handle/callback exclusivity) — tracked in #804.

### Provision filesystem-delegation gate
State-aware `provision` previously built daemon volume mounts from the **raw** caller policy, skipping the object-identity normalization, D3 delegation check, and denied-path overlap validation that one-shot `start_container` runs. It now mirrors the one-shot order — `normalize_object_conflicts` → `check_delegation` → `validate_denied_path_overlap` **on the normalized policy** — closing delegation-bypass, alias-bypass, and post-normalization overlap holes.

### Review hardening (from 3-model review of the initial change)
- **Bounded live-output channel** — the live relay uses a bounded `mpsc::channel(256)` with `blocking_send` backpressure (runs on the SDK's native IO callback threads, never the async runtime) instead of an unbounded channel, removing a daemon OOM hazard on high-volume output while preserving byte fidelity.
- **Leak-path drain termination** — `write_exec_result` now `close()`s the output receiver before draining on the exit arm, so a leaked/killed container's sender can't keep the drain loop alive indefinitely.
- **Overlap re-check on normalized policy** — `provision` re-runs `validate_denied_path_overlap` after normalization (which only ever grows `deniedPaths`), rejecting denials nested under a mount that only surface post-normalization.

## Testing
- `cargo fmt --all -- --check` clean; `cargo clippy -D warnings` clean on `wslc_common` and the daemon.
- `wslc_common` 166 unit tests + daemon 18 unit tests (incl. new `leak_path_drains_queued_then_terminates`).
- State-aware E2E **55/55** on the real WSLc host; drip liveness gap ~2s confirms live streaming after the `blocking_send` change.

## Coming in the pipeline
- **PR 3/3 — TypeScript SDK:** `Wslc*Config`/`*Result` types + branded `SandboxId<'wslc'>`
  and helper prefix wiring, mirroring the LXC state-aware SDK surface.

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/806)

### DIFF
--- a/src/backends/wslc/common/src/container_steps.rs
+++ b/src/backends/wslc/common/src/container_steps.rs
@@ -109,6 +109,20 @@ fn cstr_bytes(field: &str, value: &str) -> Result<Vec<u8>, ScriptResponse> {
 // after adoption ownership passes to `exit_callback`; the kill path trades a
 // bounded, one-time leak for memory safety.
 
+/// Which standard stream a live-output chunk came from.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum OutStream {
+    Stdout,
+    Stderr,
+}
+
+/// Optional live-output sink invoked from the SDK's stdout/stderr callbacks in
+/// addition to the capped capture buffers. The daemon supplies one to stream a
+/// container's output to the client as bytes arrive; paths that only need the
+/// final captured blob (one-shot, detached init) leave it unset. It receives
+/// the full callback bytes, independent of the capped buffers' truncation.
+pub type OutputSink = Box<dyn Fn(OutStream, &[u8]) + Send + Sync>;
+
 /// Shared buffer for capturing process I/O via SDK callbacks. Fields are
 /// `pub(crate)` so the one-shot runner's wait/collect helpers can read the
 /// captured bytes and exit signal.
@@ -116,6 +130,11 @@ pub struct IoContext {
     pub(crate) stdout: Arc<Mutex<Vec<u8>>>,
     pub(crate) stderr: Arc<Mutex<Vec<u8>>>,
     pub(crate) exited: Arc<(Mutex<bool>, Condvar)>,
+    /// Live sink for streaming output alongside the capped buffers; `None` when
+    /// only the final captured blob is needed. Its owning `Arc<IoContext>` is
+    /// released on the same schedule as the capture buffers, so on the
+    /// deliberate kill-path leak the sink (and its sender) is leaked too.
+    sink: Option<OutputSink>,
 }
 
 /// Per-stream cap on captured stdout/stderr, in bytes. The WSLc SDK streams
@@ -166,12 +185,22 @@ unsafe extern "C" fn io_callback(
     let bytes = std::slice::from_raw_parts(data, data_size as usize);
     match io_handle {
         WslcProcessIOHandle::WSLC_PROCESS_IO_HANDLE_STDOUT => {
-            let mut buf = ctx.stdout.lock().unwrap_or_else(|e| e.into_inner());
-            append_capped(&mut buf, bytes);
+            {
+                let mut buf = ctx.stdout.lock().unwrap_or_else(|e| e.into_inner());
+                append_capped(&mut buf, bytes);
+            }
+            if let Some(sink) = ctx.sink.as_ref() {
+                sink(OutStream::Stdout, bytes);
+            }
         }
         WslcProcessIOHandle::WSLC_PROCESS_IO_HANDLE_STDERR => {
-            let mut buf = ctx.stderr.lock().unwrap_or_else(|e| e.into_inner());
-            append_capped(&mut buf, bytes);
+            {
+                let mut buf = ctx.stderr.lock().unwrap_or_else(|e| e.into_inner());
+                append_capped(&mut buf, bytes);
+            }
+            if let Some(sink) = ctx.sink.as_ref() {
+                sink(OutStream::Stderr, bytes);
+            }
         }
         _ => {}
     }
@@ -274,8 +303,9 @@ impl ProcessSettings {
         script_code: &str,
         env: &[String],
         working_directory: &str,
+        sink: Option<OutputSink>,
     ) -> Result<Self, ScriptResponse> {
-        Self::build_inner(sdk, script_code, env, working_directory, true)
+        Self::build_inner(sdk, script_code, env, working_directory, true, sink)
     }
 
     /// Like [`build`](Self::build) but registers no stdio callbacks and shares no
@@ -290,7 +320,7 @@ impl ProcessSettings {
         env: &[String],
         working_directory: &str,
     ) -> Result<Self, ScriptResponse> {
-        Self::build_inner(sdk, script_code, env, working_directory, false)
+        Self::build_inner(sdk, script_code, env, working_directory, false, None)
     }
 
     unsafe fn build_inner(
@@ -299,6 +329,7 @@ impl ProcessSettings {
         env: &[String],
         working_directory: &str,
         register_callbacks: bool,
+        sink: Option<OutputSink>,
     ) -> Result<Self, ScriptResponse> {
         let mut raw = std::mem::zeroed::<WslcProcessSettings>();
         let hr = sdk.WslcInitProcessSettings(&mut raw);
@@ -310,6 +341,7 @@ impl ProcessSettings {
             stdout: Arc::new(Mutex::new(Vec::new())),
             stderr: Arc::new(Mutex::new(Vec::new())),
             exited: Arc::new((Mutex::new(false), Condvar::new())),
+            sink,
         });
 
         // Callbacks are registered only for the streamed path; a detached init
@@ -858,6 +890,10 @@ pub struct ExecOutcome {
 /// # Safety
 /// `sdk` must hold valid function pointers and `container` must be a live,
 /// started handle.
+// A thin FFI primitive whose parameters mirror the SDK's process inputs plus
+// the optional live-output sink; grouping them into a struct would only add an
+// indirection for a single call site.
+#[allow(clippy::too_many_arguments)]
 pub unsafe fn exec_in_container(
     sdk: &WslcSdk,
     container: WslcContainer,
@@ -865,12 +901,14 @@ pub unsafe fn exec_in_container(
     env: &[String],
     working_directory: &str,
     timeout_ms: u32,
+    sink: Option<OutputSink>,
     logger: &mut Logger,
 ) -> Result<ExecOutcome, ScriptResponse> {
     // `process_settings` owns every buffer the SDK reads at
     // `WslcCreateContainerProcess` time plus the I/O-capture context; it is held
     // as a stationary local until after the process exits below.
-    let mut process_settings = ProcessSettings::build(sdk, script_code, env, working_directory)?;
+    let mut process_settings =
+        ProcessSettings::build(sdk, script_code, env, working_directory, sink)?;
 
     let mut process: WslcProcess = ptr::null_mut();
     let mut err_msg = CoTaskMemPWSTR::null();
--- a/src/backends/wslc/common/src/container_steps.rs
+++ b/src/backends/wslc/common/src/container_steps.rs
@@ -121,6 +121,21 @@ pub enum OutStream {
 /// container's output to the client as bytes arrive; paths that only need the
 /// final captured blob (one-shot, detached init) leave it unset. It receives
 /// the full callback bytes, independent of the capped buffers' truncation.
+///
+/// **Two distinct live-output architectures — why they don't share plumbing.**
+/// The one-shot runner streams via `OutputMode::Stream` (in `wsl_container_runner`)
+/// plus a synchronous in-process `stream_pair`/`StreamReader` (in `stream_buffer`),
+/// drained in the same process (caller pipes, or an `inherit_pump` thread). The
+/// daemon cannot reuse that: it must relay output across a named pipe to a
+/// *separate* client process through an async (tokio) control server, so it
+/// bridges the synchronous SDK callback thread to the async pipe writer with a
+/// bounded mpsc channel (see `session_manager::enqueue_output`). Their
+/// backpressure invariants also differ deliberately: the one-shot `StreamReader`
+/// is drained synchronously in-process, whereas this callback **must never
+/// block** — the same SDK thread also delivers the process-exit callback, so the
+/// daemon path uses a non-blocking `try_send` that drops on overflow rather than
+/// stalling teardown. Keep the two paths separate for those reasons; share only
+/// the leaf primitives ([`OutStream`], the capped capture buffers).
 pub type OutputSink = Box<dyn Fn(OutStream, &[u8]) + Send + Sync>;
 
 /// Shared buffer for capturing process I/O via SDK callbacks. Fields are
--- a/src/backends/wslc/common/src/daemon_client.rs
+++ b/src/backends/wslc/common/src/daemon_client.rs
@@ -38,6 +38,7 @@ use anyhow::{bail, Context, Result};
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 
+use crate::container_steps::OutStream;
 use crate::daemon_protocol::{
     encode_frame, DaemonRequest, DaemonResponse, DeprovisionConfig, ErrKind, ExecConfig,
     ProvisionConfig, StartConfig, StopConfig, StreamFrame, MAX_FRAME_SIZE, PROTOCOL_VERSION,
@@ -122,10 +123,10 @@ const OPEN_RETRY: Duration = Duration::from_millis(20);
 const ERROR_FILE_NOT_FOUND: i32 = 2;
 const ERROR_PIPE_BUSY: i32 = 231;
 
-/// The captured result of an exec run. Live stdout/stderr framing is a daemon
-/// fill-in; today the daemon returns only the terminal exit code, so these
-/// buffers are usually empty, but the client already accumulates any
-/// [`StreamFrame::Stdout`] / [`StreamFrame::Stderr`] the daemon sends.
+/// The captured result of an exec run: the exit code plus fully-buffered
+/// stdout/stderr. Prefer [`DaemonClient::exec_streaming`] to relay output as it
+/// arrives; this buffering variant is for callers (and tests) that want the
+/// whole capture at once.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub struct ExecResult {
     pub exit_code: i32,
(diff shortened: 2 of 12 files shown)


src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -136,6 +140,47 @@ pub enum WorkerCommand {
...
+/// stream before its terminal frame.
+const LIVE_OUTPUT_MAX_CHUNK_BYTES: usize = 64 * 1024;
+
+/// Enqueue an SDK output callback, splitting it into `LIVE_OUTPUT_MAX_CHUNK_BYTES`
+/// pieces so each queue entry and its resulting frame stay bounded regardless of
+/// the callback's buffer size. Stops early once the receiver is gone (client
+/// left / leak-path `close()`), so a leaked callback never parks forever.
+fn enqueue_output(tx: &mpsc::Sender<OutputChunk>, kind: OutStream, bytes: &[u8]) {
+    for chunk in bytes.chunks(LIVE_OUTPUT_MAX_CHUNK_BYTES) {
+        if tx.blocking_send((kind, chunk.to_vec())).is_err() {
```
MGudgin: **High (reliability/security/performance) - Never block the WSLc SDK callback thread for backpressure.**

This `blocking_send` runs synchronously inside the new stdout/stderr callback sink. I verified against the byte-identical existing streaming path that the same SDK callback thread also delivers the process-exit callback; `wsl_container_runner.rs` and `stream_buffer.rs` explicitly prohibit backpressure because stalling this thread can prevent exit delivery and deadlock teardown. A slow/non-reading named-pipe client can fill this queue and recreate exactly that hazard for every sandbox sharing the daemon.

**Fix:** Make the callback-side handoff non-blocking. Use a bounded ring/`try_send` with an explicit truncation or abort signal, or a callback-safe spool whose memory/disk policy is enforced off the SDK thread. Add a regression that fills the handoff while proving the exit callback and terminal frame still complete.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -433,32 +439,86 @@ async fn handle_exec(
...
+    //     the run returns, so `output` closes *before* `done` fires; the biased
+    //     select drains every queued chunk first, then observes the close.
+    //   - Kill/leak path: a container killed without its exit callback leaks the
+    //     `IoContext` (deliberately), so the sink's sender never drops and
+    //     `output` never closes. There `done` is the only completion signal, so
+    //     we drain whatever is already queued and stop.
+    let mut done = done;
+    let terminal = loop {
+        tokio::select! {
+            biased;
```
MGudgin, line 480: **High (correctness/reliability) - Do not let continuous output starve a completed execution's terminal frame.**

On the deliberate kill/leak path the sender may remain alive and continue producing. Because this select is biased toward `output.recv()`, a continuously non-empty queue can prevent the ready `done` branch from ever running; `output.close()` is never reached and the client never receives `Exit`/`Error`. The added `leak_path_drains_queued_then_terminates` test does not cover this: it preloads only two chunks, which this preferred branch drains before `done` is polled.

**Fix:** Prioritize or explicitly latch completion. Once `done` resolves, close the receiver, drain only the already queued bounded tail, and emit the terminal frame. Add a test with a producer that keeps enqueueing while completion is ready, proving the handler terminates.

src/backends/wslc/common/src/daemon_client.rs
```diff
@@ -231,12 +232,39 @@ impl DaemonClient {
...
-    /// [`StreamFrame`] data phase — accumulating stdout/stderr — until the
-    /// terminal [`StreamFrame::Exit`] (or [`StreamFrame::Error`]).
+    /// A convenience wrapper over [`exec_streaming`](Self::exec_streaming) that
+    /// accumulates every stdout/stderr chunk into an [`ExecResult`]. Prefer
+    /// `exec_streaming` when output should be relayed live rather than buffered.
     pub fn exec(&self, config: ExecConfig) -> DaemonResult<ExecResult> {
+        let mut stdout = Vec::new();
+        let mut stderr = Vec::new();
+        let exit_code = self.exec_streaming(config, |stream, data| match stream {
+            OutStream::Stdout => stdout.extend_from_slice(data),
```
MGudgin, line 305: **Low (security/performance) - Bound the whole-capture convenience API.**

This wrapper now accumulates every live chunk into unbounded vectors. The production state-aware runner uses `exec_streaming`, but daemon tests and future internal callers can invoke this public crate API with container-controlled output and exhaust their process memory.

**Fix:** Apply a per-stream cap with a defined truncation/error result, or make the API's unbounded contract explicit and restrict it to trusted test data.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -136,6 +140,47 @@ pub enum WorkerCommand {
...
+/// cannot grow the queue without limit and OOM the persistent daemon, taking
+/// down every sandbox it owns — the same hazard the capture-buffer cap guards
+/// against. When the queue is full the sink applies backpressure (see
+/// [`SessionHandle::exec`]) rather than dropping bytes.
+const LIVE_OUTPUT_CHANNEL_CAPACITY: usize = 256;
+
+/// Max bytes per enqueued live-output chunk. A single SDK callback can deliver
+/// an arbitrarily large buffer; splitting it here bounds each queue entry's
+/// allocation and keeps the resulting `Stdout`/`Stderr` frame well under the
+/// protocol's `MAX_FRAME_SIZE` (a `Vec<u8>` serializes as a JSON number array,
```
MGudgin, line 165: **Medium (performance) - Avoid JSON integer-array encoding on the streaming hot path.**

`StreamFrame` itself is byte-identical from the base, but this PR newly emits one frame per output chunk. As this comment notes, default JSON serialization expands `Vec<u8>` into an integer array at roughly 4x payload size, adding substantial allocation, serialization, pipe bandwidth, and parse CPU for high-volume output.

**Fix:** Use a compact representation such as base64 strings or a binary frame payload, then benchmark representative text and binary output before locking in the protocol.

tests/scripts/run_wslc_state_aware_tests.ps1
```diff
@@ -356,14 +410,36 @@ try {
...
+        Run-StateAwareTest "A: exec (live drip -- incremental streaming)" {
+            $r = Invoke-StateAwareStreaming -ConfigFile 'wslc_state_aware_exec_drip.json' -SandboxId $script:sandboxId
+            Assert-True ($r.ExitCode -eq 0) "drip exec exit code = 0"
+            $p1 = $r.Lines | Where-Object { $_.Text -match 'DRIP-PART1' } | Select-Object -First 1
+            $p2 = $r.Lines | Where-Object { $_.Text -match 'DRIP-PART2' } | Select-Object -First 1
+            Assert-True ($null -ne $p1) "PART1 line observed on stdout"
+            Assert-True ($null -ne $p2) "PART2 line observed on stdout"
+            if ($p1 -and $p2) {
+                $gapSec = ($p2.At - $p1.At).TotalSeconds
+                Assert-True ($gapSec -ge 1.0) `
```
MGudgin: **Medium (testability) - Make the live-streaming E2E proof less scheduler-dependent.**

The test timestamps when PowerShell reads each line and requires a one-second gap from a producer sleeping two seconds. A host/VM/PowerShell stall can allow both lines to accumulate before the reader runs, producing a near-zero observed gap even though the implementation streamed correctly.

**Fix:** Prefer an explicit synchronization/acknowledgment between the first marker and the second emission. If wall time must remain, increase the producer separation and document a tolerance chosen from observed CI variance rather than treating this as an airtight proof.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -186,30 +183,38 @@ impl StatefulSandboxBackend for WslcStateAwareRunner {
...
+                    sandbox_id: sandbox_id.to_string(),
+                    script_code: request.script_code.clone(),
+                    working_directory: request.working_directory.clone(),
+                    env,
+                    timeout_ms: request.script_timeout,
+                },
+                |stream, bytes| match stream {
+                    OutStream::Stdout => {
+                        let _ = stdout.write_all(bytes);
+                        let _ = stdout.flush();
```
MGudgin: **Low (performance) - Avoid forcing an OS flush for every callback chunk.**

Small SDK callback chunks can turn this into a lock/flush syscall per few bytes. That preserves low latency but can severely reduce throughput, particularly after JSON framing already increases the stream volume.

**Fix:** Use a bounded-latency coalescing policy (newline, byte threshold, or short interval), retain a locked writer for the relay duration, and flush once more at completion.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -314,20 +319,45 @@ fn validate_sandbox_id(sandbox_id: &str) -> Result<(), MxcError> {
...
+    wxc_common::filesystem_access::check_delegation(effective)
+        .map_err(MxcError::policy_validation)?;
+    // Re-validate denied-path overlap against the *tightened* policy. Object
+    // normalization only ever grows `denied_paths` (moving an aliased rw/ro
+    // entry to `denied`), so a deny that becomes nested under a still-mounted
+    // parent only after normalization would otherwise reach
+    // `build_daemon_volumes` unenforced — accessible through the parent mount.
+    // Mirrors the one-shot `start_container` order (normalize -> delegation ->
+    // overlap on the normalized policy).
+    crate::policy_mapping::validate_denied_path_overlap(
```
MGudgin: **Medium (maintainability) - Share the normalize/delegate/overlap policy gate between one-shot and state-aware paths.**

This PR fixes a real drift: the one-shot runner already performed this third step, while the state-aware copy did not. Copying the missing call repairs today’s bug but leaves both runners synchronized only by comments, so the next policy check or ordering change can drift again.

**Fix:** Extract the complete three-step WSLc provision policy gate into one helper used by both runners.

src/backends/wslc/common/src/container_steps.rs
```diff
@@ -109,13 +109,32 @@ fn cstr_bytes(field: &str, value: &str) -> Result<Vec<u8>, ScriptResponse> {
...
+    Stdout,
+    Stderr,
+}
+
+/// Optional live-output sink invoked from the SDK's stdout/stderr callbacks in
+/// addition to the capped capture buffers. The daemon supplies one to stream a
+/// container's output to the client as bytes arrive; paths that only need the
+/// final captured blob (one-shot, detached init) leave it unset. It receives
+/// the full callback bytes, independent of the capped buffers' truncation.
+pub type OutputSink = Box<dyn Fn(OutStream, &[u8]) + Send + Sync>;
```
MGudgin, line 139: **Low (maintainability) - Clarify why this second live-output architecture cannot reuse the existing one-shot stream plumbing.**

The crate now has both `OutputMode::Stream`/`StreamReader` and this callback closure feeding an async channel, plus two differently shaped `IoContext` types. The daemon may genuinely need different plumbing, but the relationship and differing callback/backpressure invariants are not documented where maintainers will discover them.

**Fix:** Reuse shared primitives where practical, or cross-reference both implementations and document why the daemon path is intentionally distinct.

src/backends/wslc/daemon/src/control_server.rs
```diff
@@ -417,9 +419,13 @@ async fn handle_client(mut pipe: NamedPipeServer, session: SessionHandle) -> Res
...
-/// TODO(fill-in): bidirectional live stdio (client `Stdin` frames -> process,
-/// process stdout/stderr -> `Stdout`/`Stderr` frames). The skeleton runs the
-/// command to completion and emits only the terminal frame.
+/// Output streaming (process -> `Stdout`/`Stderr`) is live. Client `Stdin`
+/// frames are NOT forwarded: the WSLc SDK consumes all process IO handles once
+/// any `WslcSetProcessSettingsCallbacks` is registered (the callback path this
+/// live output streaming depends on), so `WslcGetProcessIOHandle(STDIN)` is
+/// unavailable. Piped stdin would require a handle-mode rearchitecture (no
+/// callbacks; `ReadFile` threads for stdout/stderr + `WriteFile` for stdin) and
+/// is deferred to Tier 2 (see the 2c plan).
```
MGudgin: **Low (maintainability) - Replace the untraceable “2c plan” reference.**

No checked-in document defines this plan or Tier 2, while the PR description identifies issue #804 as the actual stdin-tracking artifact. Source comments should point to durable repository context rather than review-stack terminology that disappears after merge.

**Fix:** Reference issue #804 or a checked-in design document, or remove the aside.

MGudgin (changes requested):
## Summary

Requesting changes for two blocking live-streaming issues: the bounded queue blocks inside the WSLc SDK callback path even though the existing implementation documents that this callback thread must never stall, and the output-first biased `select!` can prevent a completed/leaked execution from ever sending its terminal frame. These can respectively deadlock exit delivery and leave clients streaming forever.

**Range verified:** `e5e4e050b7224e678fe5900d9b5e64766cc73b0b...9b110a87bf06b1e0239bd13dd24ce4fa6b39c5f6` (7 files, +670/-140). The merge base exactly matches GitHub's PR base. The locally captured `git diff` and `gh pr diff` are byte-for-byte identical at 1,290 lines.

**Verified clean, with receipts:** the provision path now applies normalization, delegation, and denied-overlap validation to the tightened policy in the intended order; the repository already has focused canonical alias/junction tests in `policy_mapping.rs` and one-shot junction E2Es; live-output frame ordering and finite leak-path termination have useful unit tests; Windows-specific modules/tests remain correctly gated. No Windows/WSL runtime tests were run as part of this filing.

### Verified pre-existing - not independently attributed to this PR

- `src/backends/wslc/common/src/daemon_protocol.rs` is byte-identical base vs head. Its JSON `Vec<u8>` representation is cited only because this PR newly makes it the high-volume per-chunk streaming format.
- `src/backends/wslc/common/src/wsl_container_runner.rs` and `src/backends/wslc/common/src/stream_buffer.rs` are byte-identical base vs head. Their explicit non-blocking callback-thread invariant is supporting evidence for the new bounded-channel backpressure finding, not a separate defect charged to this PR.

### Requested findings not filed after verification

- Callback panic/concurrency coverage was withdrawn: the production sink is internally constructed, uses thread-safe `mpsc::Sender`, and has no demonstrated panic path beyond process-fatal allocation failure; testing an arbitrary panicking closure would not validate production behavior.
- Global-stdio coupling and best-effort local write behavior predate this PR; the change moves those writes from post-run buffers to live chunks but does not introduce the architectural coupling itself.
- The junction/symlink coverage finding was withdrawn because the shared overlap validator already has focused symlink/junction/absent-tail tests and the one-shot runner has real junction E2Es; repeating all of those through this orchestration helper would add little signal.
- Racing blocked senders against receiver closure primarily tests Tokio's channel contract rather than repository logic; the material repository defect is blocking the SDK callback in the first place.

Medium and Low comments are non-blocking; the two High findings are the merge conditions for this review.

src/backends/wslc/daemon/src/session_manager.rs
SohamDas2021: Fixed. Replaced blocking_send with non-blocking try_send + an AtomicBool overflow latch so a stalled client can't wedge the exit callback; added a regression that floods the handoff and proves the terminal frame + exit still complete.

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Fixed. Biased the select toward done and latched completion; added continuous_producer_does_not_starve_terminal  with a producer that keeps enqueueing while done is ready.

src/backends/wslc/daemon/src/session_manager.rs
SohamDas2021: Fixed. StreamFrame stdin/stdout/stderr payloads now serialize as base64 strings (~1.33x vs ~4x) via a serde adapter, keeping the single uniform JSON framing; added roundtrip (incl. full 0–255 non-UTF8) + "payload is base64, not integer-array" tests.

tests/scripts/run_wslc_state_aware_tests.ps1
SohamDas2021: Fixed, widened producer separation to 3s and set a documented tolerance band (minGap 1.5s, chosen to sit far above the buffered ~0s floor while absorbing jitter); verified live at a 2.94s observed gap.

src/backends/wslc/common/src/daemon_client.rs
SohamDas2021: Fixed, documented the unbounded-whole-capture contract explicitly as trusted/bounded-output-only, directing any live/untrusted path to exec_streaming.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Fixed, hold locked stdout/stderr writers for the relay's duration and flush only non-newline-terminated chunks (line-buffered writers already flush on newline), with a single final flush, coalescing bulk-output syscalls while preserving progress-output latency.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Fixed. Extracted apply_provision_policy_gate in policy_mapping.rs, now the single gate called by both the one-shot wsl_container_runner and state-aware build_provision_config

src/backends/wslc/common/src/container_steps.rs
SohamDas2021: Expanded the OutputSink doc to cross-reference both architectures and explain why the daemon path is intentionally distinct, while sharing the leaf primitives.

src/backends/wslc/daemon/src/control_server.rs
SohamDas2021: Replaced the "2c plan / Tier 2" aside with a reference to issue #804 (stdin forwarding).

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -73,6 +106,7 @@ pub fn exec_state_aware(
     parsed: ParsedStateAwareRequest,
 ) -> Result<Box<dyn SandboxProcess>, MxcError> {
     let backend = resolve_backend(&parsed)?;
+    require_experimental_optin(&backend, &parsed)?;
```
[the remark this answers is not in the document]
SohamDas2021: Not fixing here, will follow-up.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -0,0 +1,704 @@
...
+            // Drop the locks before mapping the error so error conversion never
+            // contends with the writers we just held.
+            drop((out, err));
+            result.map_err(map_daemon_error)?
+        };
+
+        Ok(ExecHandle {
+            stdout: null_pipe_handle(),
+            stderr: null_pipe_handle(),
+            stdin: null_pipe_handle(),
```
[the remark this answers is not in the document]
SohamDas2021: Not fixing here, will follow-up.

MGudgin (approved):
Verified all nine prior review findings at 7ac1559.

Local verification:
- cargo fmt --all -- --check
- cargo test -p wslc_common: 182 passed
- cargo test -p wxc_wslc_daemon: 22 runnable tests passed; 2 WSL2-host lifecycle tests ignored

The live WSL2 E2E suite was not run locally.

PR #813, opened 2026-08-11
Add guarded WPR captureDenials fallback

richiemsft:
## 📖 Description

Adds a secure `processContainer.captureDenials` fallback for Windows hosts where the complete native PSEC plus V2 Learning Mode capability set cannot honor the requested policy.

Native capture remains preferred when PSEC, V2 Learning Mode, and request-specific capabilities are available. Otherwise MXC selects the highest compatible legacy containment tier—SBOX, AppContainer+BFS, or AppContainer+DACL—and pairs it with the guarded WPR capture session. Proxy and least-privilege policies that PSEC cannot represent route to the compatible legacy path rather than weakening policy.

MXC uses a privilege-separated, brokered elevation model. The main executor and public `plm.exe` process remain unelevated, and a session-scoped UAC-elevated helper is launched only when privileged guarded WPR capture is required. The helper performs a fixed set of WPR control and analysis operations and returns bounded results over an authenticated named pipe.

The guarded fallback builds on the PLM architecture refactored in #760 rather than introducing a separate capture or decoding stack. It reuses PLM's `asInvoker` front end, retained elevated guardian, embedded WPR profile, protected scratch and recovery state, authenticated bounded control channel, and `learning_mode_windows::EtlDenialAnalyzer` canonical decoder. This PR extends that foundation with sandbox job/process-handle attachment, exact descendant lifetime tracking, and a bounded analysis response that feeds the same canonical `DenialsDocument` output as native capture.

The guarded path:

- Uses the same canonical denial analyzer and `DenialsDocument` output as native `captureDenials`.
- Supports block collection through `learningModeLogging` and allow/audit collection through `permissiveLearningMode`.
- Starts WPR before the suspended sandbox root resumes.
- Authenticates the unelevated owner and duplicates its sandbox job and root-process handles.
- Opens, job-validates, and retains a handle for every descendant reported by the job completion port.
- Filters host-wide WPR denial events using exact process creation/exit times from those retained handles, preventing PID reuse from expanding capture scope.
- Checks ordered job notifications and `TotalProcesses` accounting, with a 4096-generation bound.
- Gives the tracker worker independent owned job and completion-port handles. Shutdown posts a sentinel and joins the worker; if bounded waits expire, the guardian fail-stops rather than detaching or releasing handles still in use.
- Keeps raw ETL in protected elevated scratch and returns only bounded canonical denial data.
- Preserves a protected recovery marker on uncertain guardian or pipe failure without issuing blind host-wide WPR cancellation.
- Stops and discards owned traces on definite launch, tracking, or analysis failures.

Runtime resolution locates `plm.exe` relative to `wxc-exec.exe` or the loaded `mxc_ffi` module rather than the caller's working directory. Before elevation, MXC opens the helper with sharing that denies replacement, resolves the pinned object to a stable local DOS path, verifies Authenticode against the pinned handle, requires the embedded signer organization to be Microsoft Corporation, and validates the owner and DACL of the containing directory and its ancestors. The pinned handle remains open across `ShellExecuteExW`, and the resolved path is launched, preventing file replacement and path-alias check/launch races. The elevated child restricts runtime DLL resolution to System32 and fails closed if that hardening cannot be applied.

Windows artifact packaging, ESRP signing/verification, and SDK package assertions are defined in #834. Unsigned local helpers and helpers in user-writable locations are intentionally rejected, so guarded-WPR end-to-end execution requires a signed packaged `plm.exe` in a protected directory. This also applies to Rust SDK source consumers whose module-relative lookup resolves beside a locally compiled executable; native PSEC capture remains available when supported.

## 🔗 References

- #760
- #834

## 🔍 Validation

- `cargo test -p plm --lib -p mxc_engine` — 218 passed, 1 ignored.
- `cargo clippy -p plm -p mxc_engine --all-targets -- -D warnings`
- Release builds of `wxc-exec.exe` and `plm.exe`.
- Windows VM: native PSEC/V2 block capture produced canonical denial output.
- Windows VM guarded fallback:
  - Guarded SBOX block and allow capture passed.
  - Forty short-lived descendants completed without an attestation race.
  - Nonzero sandbox exit code 23 was preserved.
  - A 4100-process workload failed closed at the 4096-generation bound.
  - Successful and bounded-failure paths left no WPR recording, recovery marker, PLM process, or incomplete denial output.
- The runtime signature gate requires an official signed artifact from #834 for signing-dependent VM validation; unofficial PR artifacts are intentionally unsigned and rejected.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task


src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -570,10 +591,35 @@ impl AppContainerScriptRunner {
...
+    pub fn with_guarded_capture_factory(mut self, factory: Arc<dyn GuardedCaptureFactory>) -> Self {
+        self.guarded_capture_factory = Some(factory);
+        self
+    }
+
+    /// Marks `deniedPaths` as enforced by the dispatcher's per-run DACL guard.
+    pub fn with_external_denied_paths(mut self) -> Self {
+        self.denied_paths_enforced_externally = true;
+        self
+    }
```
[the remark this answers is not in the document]
richiemsft: Valid. Restricted `with_external_denied_paths` to `pub(crate)` so only `appcontainer_common` code, specifically the dispatcher that owns the DACL guard, can assert external enforcement. Fixed in the latest commit.

src/host/plm/src/elevated.rs
```diff
@@ -0,0 +1,2558 @@
...
+            notification_sequence: 0,
+            active_process_zero_filetime: None,
+            error: None,
+        }
+    }
+
+    fn fail(&mut self, message: impl Into<String>) {
+        if self.error.is_none() {
+            self.error = Some(message.into());
+        }
```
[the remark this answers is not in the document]
richiemsft: Fixed in 959361dc. A terminal tracker failure now immediately terminates the attested sandbox job. After confirmed termination, the guardian stops and discards the WPR trace and sends a framed terminal response to the authenticated caller. If job termination fails, it reports that failure without claiming termination and preserves the guarded trace state for recovery.

richiemsft:

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1457,10 +1652,65 @@ impl AppContainerSandboxProcess {
...
+
+    fn discard_guarded_capture_after_termination_failure(&mut self) {
+        let Some(mut session) = self.capture_session.take() else {
+            return;
+        };
+        if let Err(error) = session.discard() {
+            capture_output::write_stderr_line_best_effort(format_args!(
+                "failed to discard guarded WPR capture after sandbox termination failure: {error}"
+            ));
+        }
```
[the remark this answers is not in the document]
richiemsft: Fixed in c7620e38, 5f46d424, and 2463a299. AppContainer Drop now relies on a strengthened discard contract that cannot return until guardian handle release is confirmed. Confirmation is bounded to three attempts with backoff; exhaustion aborts the owner before any firewall/BFS/DACL/job guards can drop. Shared blocking-contract and retry-exhaustion tests cover the path.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2421,27 +2708,44 @@ impl BaseContainerSandboxProcess {
...
 
+    fn discard_guarded_capture_after_termination_failure(&mut self) {
+        let Some(mut session) = self.guarded_capture_session.take() else {
+            return;
+        };
+        if let Err(error) = session.discard() {
+            write_stderr_line_best_effort(format_args!(
+                "failed to discard guarded WPR capture after sandbox termination failure: {error}"
+            ));
+        }
```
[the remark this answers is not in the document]
richiemsft: Fixed in c7620e38, 5f46d424, and 2463a299. BaseContainer uses the same shared fail-closed release helper as AppContainer. The PLM adapter confirms guardian process exit before returning, retries with backoff, and aborts on bounded exhaustion so enforcement is never released while the duplicate job handle may remain live.

src/host/plm/src/elevated.rs
```diff
@@ -12,54 +12,114 @@
...
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum GuardControl {
     Stop,
+    StopAndAnalyze,
+    StopAndDiscard,
+    AttachJob,
+}
+
+fn attest_job_process(job: HANDLE, pid: u32) -> Result<(OwnedHandle, u64)> {
```
MGudgin, line 105: **High (reliability) - asynchronous attestation can reject valid short-lived descendants.**

The job completion notification is dequeued asynchronously and only then calls `OpenProcess(pid)`. If the descendant has already exited and its process object is gone, this returns an error that flows into tracker failure and terminates the whole sandbox job. The manual short-lived-child validation demonstrates one timing, but does not cover a delayed/backlogged tracker.

**Fix:** Establish and document a Windows lifetime guarantee that makes this race impossible, or add a regression test that deliberately delays notification processing. If the race is possible, redesign attestation without accepting an unauthenticated PID-only placeholder.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1061,6 +1107,94 @@ impl AppContainerScriptRunner {
...
+        // --- Guarded WPR captureDenials session (legacy-tier fallback) ---
+        //
+        // Started only now — after the job has been created, UI-limited,
+        // and the still-suspended child assigned to it — but before the
+        // caller resumes the child (`SpawnedChild::resume`, called by
+        // `SandboxBackend::spawn` right after this function returns). That
+        // ordering means a guardian-start failure can terminate the
+        // still-suspended child without ever leaving an active host-wide
+        // WPR trace running.
+        let (capture_session, capture_output_path): (
```
MGudgin, line 1119: **High (test coverage) - guarded-capture failure ordering has no runtime regression test.**

This new attach/resume/discard orchestration is security-sensitive, but the fake guarded factory is only used by validation tests and its `start` method is never exercised. There is no test proving that attach failure leaves the child suspended and terminated, that resume failure discards capture, or that compound cleanup failures preserve containment/release ordering.

**Fix:** Add fake-session runtime tests covering attach failure, discard-after-resume failure, and compound errors; assert the child never resumes after attach failure and guardian release is confirmed before containment cleanup.

src/core/mxc_engine/src/guarded_capture.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    fn discard(&mut self) -> Result<(), String> {
+        let discard_error = match self.session.discard() {
+            Ok(()) => return Ok(()),
+            Err(error) => error,
+        };
+
+        match confirm_guardian_release_after_discard_failure(
+            || self.session.cancel().map_err(|error| format!("{error:#}")),
```
MGudgin: **Medium (reliability) - release confirmation can freeze the host for roughly 40 minutes before aborting.**

`discard()` can spend its 10-minute deadline waiting for the guardian, then each of three confirmation attempts re-enters `cancel()`, which repeats a 10-minute `WaitForSingleObject`. Attempts two and three perform no new corrective action after the session is disarmed, so a wedged guardian can block `wxc-exec.exe` or an `mxc_ffi` host for about 40 minutes before `abort()`.

**Fix:** Use a short dedicated confirmation deadline and make retries escalate rather than repeating the same 10-minute wait.

src/backends/appcontainer/common/src/job_object.rs
```diff
@@ -275,16 +295,61 @@ impl UiJobObject {
...
+                    JOB_EMPTY_WAIT_TIMEOUT.as_millis(),
+                    accounting.ActiveProcesses
+                )));
+            }
+            std::thread::sleep(JOB_EMPTY_POLL_INTERVAL);
         }
     }
+
+    /// Terminates the complete process tree and confirms that the job is empty.
+    pub fn terminate_and_wait(&self, exit_code: u32) -> Result<(), WxcError> {
```
MGudgin, line 372: **Medium (proportionality) - strict job-drain errors now affect every ProcessContainer run.**

The runners call `terminate_and_wait` on normal wait/drop paths whether guarded capture is active or not. A non-capture run whose job does not report zero active processes within five seconds can now return a teardown error instead of the child result, expanding this opt-in capture feature's behavior change to all Windows ProcessContainer executions.

**Fix:** Gate strict guardian-driven drain certainty on an active guarded-capture session, or preserve the child result and report non-capture drain failures as warnings.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -867,16 +918,46 @@ impl BaseContainerRunner {
...
+                Self::is_process_security_environment_usable(),
+                psec_supports_deny_paths,
+            )
+        {
             return true;
         }
         crate::fallback_detector::base_container_supports_deny_paths()
     }
 
+    pub(crate) fn uses_native_capture_for_request(request: &ExecutionRequest) -> bool {
```
MGudgin, line 1009: **Medium (maintainability) - native-capture eligibility is independently re-derived.**

The dispatcher-facing static predicate and the runner's instance predicate calculate the same PSEC-plus-capture eligibility using different support sources and test hooks. Their agreement is an unenforced convention, so a future capability change can update one path and silently diverge tier selection from execution.

**Fix:** Factor one eligibility helper parameterized by `CapturePlatformSupport` and route both call paths through it.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -1981,14 +2068,166 @@ impl BaseContainerRunner {
...
+            Some(
+                "the legacy BaseContainer API ignored CREATE_SUSPENDED, so guarded WPR could not \
+                 observe the complete sandbox execution"
+                    .to_string(),
+            )
+        } else {
+            None
+        };
+        if let Some(mut message) = resume_error {
+            let termination_error = job.terminate_and_wait(u32::MAX).err();
```
MGudgin: **Medium (maintainability) - security-sensitive abandonment cleanup is duplicated.**

Terminate-and-wait, session discard, handle closure, legacy cleanup, proxy shutdown, and compound-error construction are repeated across multiple BaseContainer/AppContainer failure branches. Minor control-flow drift already exists, and future cleanup changes must be kept synchronized by inspection.

**Fix:** Extract one guarded-launch abandonment helper that owns cleanup ordering and secondary-error composition.

src/host/plm/src/elevated.rs
```diff
@@ -357,6 +546,522 @@ impl Drop for OwnedHandle {
...
+    if let Err(error) = unsafe { TerminateJobObject(job, u32::MAX) } {
+        if let Ok(mut state) = state.lock() {
+            let termination_error =
+                format!("failed to terminate sandbox job after tracker failure: {error}");
+            state.termination_error = Some(termination_error);
+        }
+    }
+}
+
+fn current_filetime() -> u64 {
```
MGudgin, line 1335: **Medium (testability) - hard-coded time sources prevent deterministic boundary testing.**

The new tracker directly calls precise system time and process-time APIs, so creation/exit boundary cases depend on wall-clock timing and scheduler behavior.

**Fix:** Inject a clock/process-times provider into `JobProcessTracker` and test exact boundary acceptance/rejection with synthetic times.

src/host/plm/src/elevated.rs
```diff
@@ -357,6 +546,522 @@ impl Drop for OwnedHandle {
...
+        }
+    }
+
+    fn fail(&mut self, message: impl Into<String>) {
+        if self.error.is_none() {
+            self.error = Some(message.into());
+        }
+    }
+
+    fn process_started<F>(&mut self, pid: u32, observed_filetime: u64, attest: F)
```
MGudgin, line 641: **Medium (test coverage) - the producer state machine never tests sequential PID reuse.**

Consumer-side lifetime filtering has a synthetic repeated-PID test, but `ProcessTrackerState` is never exercised with start -> exit -> start of the same PID. Existing tracker tests use distinct incrementing PIDs, so they do not prove that two generations are retained distinctly before analysis.

**Fix:** Replay a reused PID through `process_started`/`process_exited` with distinct attestations and assert two separate generations.

src/core/mxc_engine/src/guarded_capture.rs
```diff
@@ -0,0 +1,273 @@
...
+    unreachable!("guardian confirmation attempt range is non-empty")
+}
+
+/// Resolve `plm.exe` next to the module containing `mxc_engine`.
+///
+/// This is the executor directory for `wxc-exec.exe` and the native runtime
+/// asset directory for `mxc_ffi.dll`. `current_exe()` is not sufficient for
+/// library consumers because a framework-dependent .NET app reports
+/// `dotnet.exe`, not the loaded MXC native module.
+fn plm_exe_path() -> Result<std::path::PathBuf, String> {
```
MGudgin, line 89: **Medium (documentation) - document the new `mxc_ffi`-relative runtime resolution.**

This PR newly resolves `plm.exe` relative to the loaded native module so SDK hosts can find it beside `mxc_ffi.dll`, but the PLM documentation still only describes staging beside `wxc-exec.exe`.

**Fix:** Document both runtime resolution locations and clarify that packaging/signing is delivered separately by #834.

src/core/mxc_engine/src/guarded_capture.rs
```diff
@@ -0,0 +1,273 @@
...
+    plm_path: &std::path::Path,
+    owner_pid: u32,
+) -> Result<Box<dyn GuardedCaptureSession>, String> {
+    if !plm_path.exists() {
+        return Err(format!(
+            "plm.exe not found at {} (required for the guarded WPR captureDenials fallback)",
+            plm_path.display()
+        ));
+    }
+    let session = plm::elevated::start_guarded_session_with_executable(plm_path, owner_pid)
```
MGudgin, line 224: **Low (security, defence in depth) - automatic elevation extends an existence-check-only trust pattern to library asset directories.**

The explicit `--audit` path already elevated a neighbouring `plm.exe` after only an existence check, so that pattern is pre-existing. This PR newly makes it automatic for `captureDenials` and resolves from the loaded module directory, which may be a per-user SDK/native-assets location.

**Fix:** Consider Authenticode/catalog verification or requiring a non-admin-writable containing directory before `runas`. This is defence in depth, not a merge condition.

src/host/plm/src/elevated.rs
```diff
@@ -357,6 +546,522 @@ impl Drop for OwnedHandle {
...
+                .take()
+                .context("guarded WPR job tracker failure was not retained")?,
+            termination_error: state.termination_error.take(),
+        })
+    }
+
+    fn stop_worker(&mut self) {
+        if let Some(worker) = self.worker.take() {
+            let _ = unsafe {
+                PostQueuedCompletionStatus(self.completion_port.0, TRACKER_STOP_MESSAGE, 0, None)
```
MGudgin: **Low (reliability) - a failed tracker wake-up can hang the elevated guardian.**

The `PostQueuedCompletionStatus` result is discarded, followed by an unbounded worker join. The worker waits with `INFINITE`, and after the job reaches zero active processes no later job notification is guaranteed to wake it, so a failed post can retain the duplicated job handle until parent-side deadlines expire.

**Fix:** Check the post result and fall back to closing the completion port or another bounded worker-shutdown mechanism.

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -112,6 +177,13 @@ impl<'visitor> Accumulator<'visitor> {
     }
 
     fn add_raw_denial(&mut self, raw: RawDenial) {
+        if self
+            .process_lifetimes
+            .as_ref()
+            .is_some_and(|lifetimes| !lifetimes.contains(raw.pid, raw.filetime))
```
MGudgin, line 192: **Low (security/reliability) - out-of-scope host events consume the shared analysis budget before the new lifetime check.**

The one-million-event budget and its ordering pre-exist, but guarded capture newly analyzes a deliberately host-wide trace. Unrelated host events can therefore consume the budget before this PID/time scope check and leave the sandbox result truncated, with analyze mode only reporting the truncation flag.

**Fix:** Charge the processing budget after lifetime scoping, or maintain separate in-scope/out-of-scope budgets.

src/backends/appcontainer/common/src/capture_output.rs
```diff
@@ -0,0 +1,368 @@
...
+        Err(error) => Err(std::io::Error::other(format!(
+            "captureDenials failed to remove internal capture file {}: {error}",
+            path.display()
+        ))),
+    }
+}
+
+/// Combines a capture result with a best-effort cleanup result, preserving
+/// both failure messages when both operations fail.
+pub fn combine_capture_and_cleanup_results<T>(
```
MGudgin, line 185: **Low (maintainability) - two result-combination helpers duplicate the same four-arm logic.**

`combine_capture_and_cleanup_results` and `combine_process_and_teardown_results` differ mainly in return type and primary-error policy, which makes future error-composition changes easy to apply inconsistently.

**Fix:** Use one generic combinator with an explicit primary-error policy.

src/backends/appcontainer/common/src/dispatcher.rs
```diff
@@ -343,29 +360,44 @@ fn select_backend_with_fallback(
...
+        && (decision.tier != IsolationTier::BaseContainer || !uses_native_capture);
+    if guarded_capture_required && capture_factory.is_none() {
         return Err(DispatchError::CaptureDenialsUnsupported {
             tier: decision.tier,
         });
     }
+    // Only thread the factory into the runner when it will actually be used —
+    // an AppContainer tier honoring `captureDenials`. This keeps T1/T2-without-
+    // capture/T3-without-capture identical to their pre-fallback construction.
+    let capture_factory_for_appcontainer = if request.policy.capture_denials.is_some() {
```
MGudgin: **Low (maintainability) - AppContainer factory gating re-derives `guarded_capture_required`.**

Within the AppContainer tiers, `capture_denials.is_some()` is currently equivalent to `guarded_capture_required`, but that equivalence depends on the tier invariant rather than being expressed directly.

**Fix:** Reuse `guarded_capture_required` or encode the invariant explicitly.

src/backends/appcontainer/common/src/dispatcher.rs
```diff
@@ -374,28 +406,27 @@ fn select_backend_with_fallback(
...
-                        ),
-                    ),
-                    mgr,
-                )
+                let runner = with_capture_factory(
+                    AppContainerScriptRunner::with_filesystem_mode_and_sid_string(
+                        FilesystemMode::Bfs,
+                        sid,
+                    )
+                    .with_external_denied_paths(),
```
MGudgin: **Low (scope) - AppContainer+BFS denied-path behavior changes for non-capture requests too.**

`with_external_denied_paths()` is enabled before checking whether `captureDenials` is requested, relaxing the previous validation gate for ordinary AppContainer+BFS requests as collateral behavior in a capture-fallback PR.

**Fix:** Gate the relaxation on guarded capture, or explicitly document and add tests for the broader denied-path support change.

MGudgin (changes requested):
## Summary

I re-resolved this review against base `b960cf5` and head `2463a29`; the merge-base exactly matches the PR base, and the local/GitHub diffs are line-for-line equivalent apart from abbreviated `index` hashes.

I am requesting changes primarily because:

- the new asynchronous process attestation can fail the whole sandbox when a descendant exits before the tracker opens it;
- the security-sensitive attach/resume/discard/release failure orchestration has no runtime regression coverage;
- a wedged guardian can block an embedding process for roughly 40 minutes before aborting it; and
- strict job-drain failures now affect non-capture ProcessContainer runs too.

- The exact process-lifetime model records creation/exit times, enforces the 4096-generation bound, and has consumer-side PID-reuse tests.
- Manifest changes add only in-tree workspace dependencies and existing Windows API feature gates; there are no new third-party dependencies or unrelated lockfile drift.
- New security-sensitive modules and dependencies are Windows-gated; shared protocol/lifetime types remain portable and non-Windows paths remain fail-closed.

### Body-only findings

**Medium (documentation) - PLM's new production role is undocumented.** `README.md` and `src/host/plm/readme.md` are byte-identical between base and head and still describe PLM strictly as an `--audit` developer workflow. This PR newly makes PLM the automatic production fallback for `processContainer.captureDenials`, so those existing statements become false because of this change. Please update both overviews.

**Medium (testability) - the new job tracker has no deterministic event seam.** `JobProcessTracker` and `ProcessTrackerState` are new, but directly consume live completion-port and process APIs. Extracting an injectable job-event/handle-resolver seam would allow deterministic testing of delayed notifications, rapid exits, and process-open failures. This is not independently a merge blocker, but it compounds the missing runtime coverage noted inline.

**Low (maintainability) - legacy dispatcher wrappers have no production callers.** The old `dispatch_with_fallback` and `spawn_with_fallback` wrappers now only delegate to the capture-aware entrypoints, and their remaining workspace callers are tests. Consider keeping one run and one spawn entrypoint that accept an optional guarded-capture factory.

### Verified pre-existing or withdrawn - not attributed to this PR

- `Global\Mxc_Plm_Audit` test isolation is byte-identical at base and head.
- `run_guarded_stop` replacing an original error when writing the pipe response fails is byte-identical at base and head.
- SDK packaging, ESRP signing, and package assertions are explicitly deferred to #834.
- The original performance claims about retained process handles, 64 MB allocation, and polling were withdrawn: handle retention is a deliberate PID-reuse control, 64 MB is a validated cap rather than an eager allocation, and the cited polling was largely pre-existing.

src/host/plm/src/elevated.rs
richiemsft: Fixed in a1f067bc. A descendant that exits before handle-based attestation is now recorded as an attestation race rather than terminating the sandbox. Execution completes normally, while capture analysis fails closed instead of trusting the PID. Added delayed short-lived-descendant and accounting regressions.

src/backends/appcontainer/common/src/appcontainer_runner.rs
richiemsft: Fixed in a1f067bc. Extracted the guarded attach/cleanup orchestration behind runtime fakes and added regressions for attach failure while suspended, resume/discard failure, and compound containment-before-release ordering.

src/core/mxc_engine/src/guarded_capture.rs
richiemsft: Fixed in a1f067bc. Guardian confirmation now uses three dedicated 10-second attempts instead of re-entering the multi-minute stop deadline. The policy, timeout propagation, retry delay, and successful confirmation path are deterministically tested.

src/backends/appcontainer/common/src/job_object.rs
richiemsft: Fixed in a1f067bc. Strict terminate-and-drain remains limited to guarded capture paths. Ordinary ProcessContainer teardown uses best-effort drain observation, preserves the workload result, and surfaces a warning if the kernel has not reported an empty job yet.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in a1f067bc. Native capture selection now goes through one native_capture_eligible helper parameterized by CapturePlatformSupport, shared by static selection and the runner instance.

src/backends/appcontainer/common/src/base_container_runner.rs
richiemsft: Fixed in a1f067bc. The repeated guarded-launch abandonment sequence is centralized so termination, capture discard, handle closure, containment cleanup, and compound error reporting retain one ordering.

src/host/plm/src/elevated.rs
richiemsft: Fixed in a1f067bc. Lifetime boundary validation is extracted into a pure function with exact boundary tests; the new bounded worker-join and guardian-confirmation paths also inject time/sleep policy for deterministic timeout tests.

src/host/plm/src/elevated.rs
richiemsft: Fixed in a1f067bc. Added producer-state coverage for start -> exit -> start with the same PID and distinct authenticated generations.

src/core/mxc_engine/src/guarded_capture.rs
richiemsft: Fixed in a1f067bc. The PLM documentation now describes module-relative lookup beside both wxc-exec.exe and mxc_ffi.dll and identifies #834 as the release packaging/signing dependency.

src/core/mxc_engine/src/guarded_capture.rs
richiemsft: Disposition documented in a1f067bc. Co-location is explicitly described as discovery, not trust, and the code no longer claims privilege equivalence: a planted plm.exe would self-elevate. Runtime integrity enforcement is not added in this PR; release packaging/signing remains owned by #834, consistent with this defence-in-depth comment being non-blocking.

src/host/plm/src/elevated.rs
richiemsft: Fixed in a1f067bc. PostQueuedCompletionStatus failure is checked and records failure before immediately detaching; successful posts use a bounded join helper with deterministic timeout tests, so guardian shutdown cannot hang indefinitely or incur a guaranteed five-second stall.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Fixed in a1f067bc. Analyze mode now applies process-lifetime scope before charging begin_event against the shared limit. Added a regression proving out-of-scope host events do not consume the analysis budget.

src/backends/appcontainer/common/src/capture_output.rs
richiemsft: Fixed in a1f067bc. The duplicate four-arm combinators now share one generic result-combination helper with preserved primary/cleanup error ordering.

src/backends/appcontainer/common/src/dispatcher.rs
richiemsft: Fixed in a1f067bc. AppContainer factory selection now reuses the already-computed guarded_capture_required decision.

src/backends/appcontainer/common/src/dispatcher.rs
richiemsft: Fixed in a1f067bc. with_external_denied_paths is now enabled only for guarded capture. Added a regression proving non-capture BFS requests still reject deniedPaths.

src/core/mxc_engine/src/guarded_capture.rs
```diff
@@ -0,0 +1,397 @@
...
```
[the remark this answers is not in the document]
richiemsft: Addressed in 8d1dafdc. The launcher now opens the adjacent helper first with read-only sharing, resolves the pinned object to a stable local DOS path, verifies Authenticode against that pinned handle, requires the embedded signer organization to be Microsoft Corporation, and validates the owner/DACL of the containing directory and its ancestors. The pinned handle is retained across ShellExecuteExW and the resolved path is launched, preventing replacement or alias-based check/launch races.

src/host/plm/src/elevated.rs
```diff
@@ -357,6 +550,711 @@ impl Drop for OwnedHandle {
...
+            std::thread::sleep,
+            WORKER_JOIN_TIMEOUT,
+            POLL_INTERVAL,
+        ) {
+            BoundedJoinOutcome::TimedOut => {
+                self.fail_state(
+                    "guarded WPR job tracker did not shut down within the bounded timeout; \
+                     detaching the worker thread"
+                        .to_string(),
+                );
```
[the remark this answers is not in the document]
richiemsft: Addressed in 8d1dafdc. The tracker worker now owns independent duplicated job and completion-port handles for its full lifetime. Shutdown posts a sentinel through the tracker-owned port handle and joins the worker; if bounded waits expire, the guardian fail-stops rather than detaching or closing a handle that the worker may still use. Added duplicate-handle lifetime coverage for both the job and completion port.

src/host/plm/src/trust.rs
```diff
@@ -0,0 +1,1251 @@
...
+    .context("GetAclInformation failed")?;
+
+    let mut entries = Vec::with_capacity(info.AceCount as usize);
+    for index in 0..info.AceCount {
+        let mut ace_ptr: *mut c_void = ptr::null_mut();
+        // A failure to read any ACE is treated as fail-closed corruption.
+        unsafe { GetAce(dacl, index, &mut ace_ptr) }
+            .with_context(|| format!("GetAce({index}) failed"))?;
+        // SAFETY: `ace_ptr` points to a valid ACE within the DACL.
+        let header = unsafe { &*(ace_ptr as *const ACE_HEADER) };
```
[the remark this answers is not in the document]
richiemsft: Addressed in c1268632. Although GetAce guarantees a valid ACE pointer on success for this kernel-returned DACL, the parser now explicitly fails closed on NULL pointers, validates AclBytesInUse and the ACE header/full AceSize against the ACL range, uses unaligned reads only after those checks, and validates the variable SID subauthority length remains within the ACE before conversion.

src/host/plm/src/trust.rs
```diff
@@ -0,0 +1,1251 @@
...
+            Some(kind) => kind,
+            None => bail!(
+                "the DACL contains an unsupported ACE type {:#04x} (object/callback/conditional); \
+                 failing closed because its effect cannot be classified",
+                header.AceType
+            ),
+        };
+        // ACCESS_ALLOWED_ACE and ACCESS_DENIED_ACE share layout through SidStart.
+        let ace = ace_ptr as *const ACCESS_ALLOWED_ACE;
+        let mask = unsafe { (*ace).Mask };
```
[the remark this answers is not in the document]
richiemsft: Addressed in c1268632. Although GetAce guarantees a valid ACE pointer on success for this kernel-returned DACL, the parser now explicitly fails closed on NULL pointers, validates AclBytesInUse and the ACE header/full AceSize against the ACL range, uses unaligned reads only after those checks, and validates the variable SID subauthority length remains within the ACE before conversion.

MGudgin (approved):
Verified the review updates through `4a27b4e`.

- 14 original findings are addressed.
- The three partially addressed items and remaining Low legacy-wrapper item are accepted as non-blocking.
- The newly identified non-ASCII trust-path normalization issue is also accepted as non-blocking for this approval.
- `cargo fmt --all -- --check` passed.
- 675 affected crate tests passed with 0 failures and 1 ignored.

PR #818, opened 2026-08-12
Document stable telemetry privacy contract

RamonArjona4:

-----

## Summary

Defines the approved normative privacy and telemetry contract before implementation: exact consent wording, data inventory and exclusions, fail-closed consent behavior, administrative policy ceiling, localization boundary, and proposed event contract.

This is PR 1 of 5. Review this PR first; implementation follows in the dependent PRs.

## Stack

| Order | Change | PR |
|---|---|---|
| **1** | **Normative privacy and telemetry documentation** | **#818** |

Review only this PR's diff. Later PRs are intentionally based on the preceding branch.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/818)

### DIFF
--- a/docs/telemetry/telemetry-consent-design.md
+++ b/docs/telemetry/telemetry-consent-design.md
@@ -1,0 +1,203 @@
+# Telemetry consent design
+
+This document defines the Windows telemetry-consent contract.
+
+MXC telemetry is Windows-only and fails closed. Collection requires all four
+conditions at the time each event is written:
+
+```text
+per-run telemetry requested
+&& current-version user consent is Granted
+&& administrative policy does not block
+&& the platform/provider is available
+```
+
+No prompt, no response, the literal string "No", dismissal, withdrawal,
+malformed state, an unknown prompt version, or any read/write failure means no
+telemetry. Administrative policy is a deny-only ceiling: it may block
+collection but can never opt a user in.
+
+## Canonical consent resource
+
+MXC owns one immutable, versioned consent resource. Every EXE and SDK presenter
+must show every supplied field verbatim. Hosts control layout, accessibility,
+and native UI, but may not substitute wording.
+
+Current resource:
+
+- Resource version: `1`
+- Locale: `en-US`
+- Mandatory fallback: `en-US`
+- Privacy link: <https://go.microsoft.com/fwlink/?linkid=521839>
+
+### Title
+
+> Help improve Microsoft eXecution Container (MXC)
+
+### Body
+
+> Help improve MXC by sharing optional diagnostic data with Microsoft.
+> If enabled, MXC sends diagnostic information about product usage,
+> performance, and reliability. MXC does not send your commands, file paths,
+> credentials, or other customer content.
+> You can change your choice at any time.
+
+### Actions
+
+- Affirmative: **Yes**
+- Negative: **No**
+- Learn more: **Privacy Statement**
+
+Each field has a stable message ID. Locale lookup normalizes BCP 47 tags and
+falls back to `en-US`; complete messages are translated as units rather than
+assembled from fragments. Adding a translation does not change the public API.
+A material wording or data-inventory change requires a new resource version and
+explicit re-consent.
+
+## SDK presenter requirements
+
+An SDK host owns the native presentation experience, while MXC owns the
+resource, decision validation, and persistence. Implementers must follow this
+contract:
+
+1. Route consent through the MXC-owned request flow. Do not write the consent
+   store directly or create a separate grant path.
+2. Render every supplied field verbatim: title, body, affirmative label,
+   negative label, learn-more label, and learn-more URL. Hosts may control
+   layout, accessibility, and platform-native styling, but must not hardcode,
+   shorten, reorder within a message, or replace the supplied wording.
+3. Map the affirmative control to `Yes` and the negative control to `No`.
+4. Map window close, cancel, timeout, or no response to `Dismissed`. If the UI
+   cannot be presented or otherwise fails, return a presenter error instead.
+   Never infer `Yes` from a default button, policy, prior application
+   preference, or absence of a response.
+5. Make the learn-more control open the supplied URL, normally in the user's
+   default browser. Do not substitute a different privacy destination.
+6. Return only the typed decision to MXC. MXC persists the decision together
+   with the resource version and locale that were actually presented.
+7. Use the typed status API to explain the current stored/effective state and
+   administrative ceiling. Provide an explicit withdrawal control in the
+   host's settings or other appropriate telemetry-controls surface.
+
+MXC does not invoke the presenter when the result is already determined:
+
+- `AlreadyGranted`: the current resource version is already granted.
+- `PolicyBlocked`: administrative policy prevents collection.
+- `NotApplicable`: telemetry is unavailable on this platform.
+
+If the host never calls the request API, telemetry remains off. A presenter
+failure must be surfaced as an error to the host and must not create or alter a
+grant.
+
+## State and persistence
+
+MXC stores consent per user at:
+
+```text
+%LOCALAPPDATA%\mxc\telemetry-consent.json
+```
+
+The persisted consent record is a versioned JSON document. The current schema
+version is 2 and records:
+
+- `consent`: `granted` or `denied`
+- `promptResourceVersion`
+- `promptLocale`
+- MXC version, source, and update timestamp for local audit/support provenance
+
+Stored and effective state are reported separately. A legacy grant without the
+current prompt version remains visible as `storedState: "granted"` but is
+`effectiveState: "undetermined"` with reason `prompt-version-missing` or
+`prompt-version-unsupported`. It never authorizes collection. Legacy denial
+remains denied.
+
+Only two operations mutate the store:
+
+1. A presenter invoked by MXC returns an explicit `yes` or `no` for the prompt
+   supplied in that same invocation.
+2. The explicit withdrawal operation persists `denied`.
+
+Dismissal, EOF, presenter failure, localization fallback failure, or never
+requesting consent does not fabricate or rewrite a decision.
+
+### Atomicity, locking, and recovery
+
+Every consent-store mutation must be atomic across processes. Synchronization
+must use a separate primitive keyed by the consent-store path, such as a named
+mutex or a sibling lock file. It must never keep the live JSON file open
+because that can prevent atomic replacement on Windows.
+
+1. Acquire the cross-process writer lock before reading, validating, or
+   replacing the file.
+2. Write the full replacement document to a sibling temporary file in the same
+   directory.
+3. Flush the temporary file, then replace the live file with a single
+   same-volume atomic rename/replace operation.
+4. Release the lock only after the replacement is durable.
+
+Readers must coordinate through the same separate primitive or use an
+equivalent close-and-retry strategy; they must not hold the live JSON handle
+while a writer replaces it. They must never observe or accept a
+partially-written document. Any malformed file, missing required field,
+version mismatch, unreadable path, or lock/replace failure is treated as
+`undetermined`/`denied` as appropriate and must not authorize collection.
+Recovery is fail-closed: MXC may surface a typed error or status reason to the
+host, but it must not silently recreate a grant from a corrupted file.
+
+## Administrative ceiling
+
+Policy is read from:
+
+```text
+HKLM\SOFTWARE\Policies\Mxc\AllowTelemetry (REG_DWORD)
+```
+
+Only value `3` permits MXC optional diagnostic data. Other values, wrong value
+types, unreadable state, and ambiguity fail closed to `blocked`. An absent key
+is `unrestricted`. `allowed` and `unrestricted` merely leave the user's choice
+in control; neither creates consent.
+
+A consent request made while policy is blocked does not invoke the presenter or
+persist dormant consent. Withdrawal remains available while blocked.
+
+## Per-run enablement
+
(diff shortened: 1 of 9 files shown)


docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+conditions at the time each event is written:
+
+```text
+per-run telemetry requested
+&& current-version user consent is Granted
+&& administrative policy does not block
+&& the platform/provider is available
+```
+
+No prompt, no response, No, dismissal, withdrawal, malformed state, an unknown
```
jsidewhite: >No, [](http://example.com/codeflow?start=24&length=3)

typo?

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+For a terminal request, `wxc-exec` renders the canonical resource and accepts
+only an explicit affirmative or negative choice. Noninteractive input, EOF, or
+unavailable presentation returns a typed non-grant result without launching a
+sandbox.
+
+### Node presenter protocol
+
+Node uses a private same-process stdio handshake:
+
+1. The child emits the canonical prompt and a random 32-byte challenge.
```
jsidewhite: >child [](http://example.com/codeflow?start=7&length=5)

what is `the child`?

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+2. Node invokes the host presenter.
+3. Node returns one typed decision with that challenge and prompt version.
+4. The same child verifies both values before persisting.
+
+Replay, cross-process responses, and prompt-version mismatches are rejected.
+This protocol is internal and does not expose a detached begin/complete API.
```
jsidewhite: The node protocol is kind of opaque

docs/telemetry/telemetry-consent-design.md
jsidewhite: >a random 32-byte challenge [](http://example.com/codeflow?start=44&length=26)

is this overkill?  feels a little like theatre

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+
+let outcome = telemetry::request_consent(Some("en-US"), |prompt| {
+    // Render every field verbatim.
+    Ok(ConsentDecision::Yes)
+})?;
+
+let status = telemetry::get_consent_status();
+let withdrawal = telemetry::withdraw_consent()?;
+# let _ = (outcome, status, withdrawal);
+# Ok::<(), telemetry::ConsentError>(())
```
jsidewhite: is this right?  why are there "#"?

docs/telemetry/telemetry-consent-privacy-review.md
```diff
@@ -0,0 +1,68 @@
+# Telemetry consent privacy review package
```
jsidewhite: what does this title mean?  is this file necessary?  feels like it could just be a blurb in `docs/telemetry/telemetry.md`

docs/telemetry/telemetry-policy.md
```diff
@@ -0,0 +1,235 @@
+# MXC administrative telemetry policy
```
jsidewhite, line 1: >administrative [](http://example.com/codeflow?start=6&length=14)

i'd add "administrative" or "management" to the file name.  "telemetry-policy.md" sounds more like legalese rather than Intune controls.

docs/telemetry/telemetry.md
```diff
@@ -231,6 +242,67 @@ free-form text.
...
+  policy, not Windows'. It is a deny-only ceiling: it can subtract from what
+  the user permitted but can never stand in for the user's consent. See
+  [`docs/telemetry/telemetry-policy.md`](telemetry-policy.md).
+- Consent request/withdraw/status operations use the typed JSON maintenance
+  envelope documented in the consent design. The old grant/revoke/source flags
+  are non-mutating migration tombstones. `--telemetry-consent-status` remains
+  read-only and emits the same typed response as JSON `action: "status"`.
+- Every EXE/SDK presenter receives the exact Rust-owned, versioned `en-US`
+  resource. Only an explicit Yes returned from that invocation may create a
+  current grant. No request, No, dismissal, presenter failure, or legacy grant
```
jsidewhite: >No, [](http://example.com/codeflow?start=29&length=3)

typeo

docs/telemetry/telemetry.md
```diff
@@ -231,6 +242,67 @@ free-form text.
...
+for the full design; the summary:
+
+- Consent state (`granted` / `denied` / `undetermined`) is stored per-user at
+  `%LOCALAPPDATA%\mxc\telemetry-consent.json`, owned entirely by
+  `wxc_common::telemetry::consent`. It is **never** derived from, synced
+  with, or read from any Windows-level setting — not the OS Diagnostic Data
+  level (Settings → Privacy → Diagnostics & feedback), not a UTC opt-in
+  policy, nothing. MXC is an application, not a Windows system component, so
+  it is responsible for its own notice-and-consent experience (see
+  [Provider group vs. consent](#provider-group-vs-consent) below).
```
jsidewhite: Bullets like this are cumbersome to read.  If you slightly restructure as "headline"+"paragraphy sub-bullet", it's much easier to parse.

e.g.
```
- Consent state (`granted` / `denied` / `undetermined`) 
    - Per-user, stored `%LOCALAPPDATA%\mxc\telemetry-consent.json`
    - It's owned entirely by `wxc_common::telemetry::consent`. It is **never** derived from, synced
      with, or read from any Windows-level setting — not the OS Diagnostic Data
      level (Settings → Privacy → Diagnostics & feedback), not a UTC opt-in
      policy, nothing. MXC is an application, not a Windows system component, so
      it is responsible for its own notice-and-consent experience (see
      [Provider group vs. consent](#provider-group-vs-consent) below).
```

docs/telemetry/telemetry.md
```diff
@@ -231,6 +242,67 @@ free-form text.
...
+- `wxc_common::telemetry::is_enabled()` resolves to
+  `config.enabled == Some(true) && policy::get_policy().allows_collection() && consent::get_consent().allows_collection()`:
+  the JSON config field is an **explicit per-invocation opt-in that can only
+  subtract** — omitting it or setting `false` disables collection, and `true`
+  can never bypass a lack of consent. Only `ConsentState::Granted` allows
+  collection; `Undetermined`, `Denied`, and `NotApplicable` (non-Windows) all
+  block it.
```
jsidewhite: This is pretty difficult to read and highly specific to a telemetry.md file.

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+
+```csharp
+var outcome = MxcTelemetry.RequestConsent(prompt =>
+{
+    // Render every field verbatim.
+    return TelemetryConsentDecision.Yes;
+});
+
+TelemetryConsentStatus status = MxcTelemetry.GetConsentStatus();
+MxcTelemetry.WithdrawConsent();
```
jsidewhite: what should the caller do with this status?  do they care if telemetry is off?

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+No prompt, no response, No, dismissal, withdrawal, malformed state, an unknown
+prompt version, or any read/write failure means no telemetry. Administrative
+policy is a deny-only ceiling: it may block collection but can never opt a user
+in.
+
+## Canonical consent resource
+
+Rust owns one immutable, versioned resource in
+`wxc_common::telemetry::consent_prompt`. Every EXE and SDK presenter must show
+every supplied field verbatim. Hosts control layout, accessibility, and native
```
jsidewhite: so claude.exe will call this?  why would they care to enable Windows telemetry?

docs/telemetry/telemetry-consent-privacy-review.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Data sent
+
+- MXC version and channel
+- Containment backend
+- Run outcome and exit code
+- Run duration
+- Bounded failure category
+- Lifecycle phase
+- Random app-session and sandbox-lifecycle correlation identifiers
```
MGudgin: **Medium (security) — The approved inventory promises random lifecycle identifiers, but the relay contract does not enforce that.**

I verified that the correlation relay itself predates this PR, so I am not attributing that implementation to this change. The new privacy-review statement is nevertheless stronger than the contract it approves: later phases accept any syntactically relayable MS-CV, and format validation does not prove that its base is the random value MXC minted during provision. A host can substitute a valid stable pseudonym while the approval record still says only random identifiers are sent.

**Fix:** Bind relayed vectors to the MXC-generated provision value for that sandbox, or disclose the field as caller-supplied/format-validated data and obtain approval for that inventory.

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+const status = queryTelemetryConsent();
+withdrawTelemetryConsent();
+```
+
+The presenter may return a decision directly or a promise. Node never writes
+the consent store itself.
+
+## Live enforcement and event compatibility
+
+Consent and policy are checked immediately before every `MXC.Execution` and
```
MGudgin: **Medium (reliability/performance) — This requirement puts synchronous storage I/O on every event and error path.**

The final stack implements this literally: each authorization check reads/parses the consent file and queries the policy registry, and one logical emission can repeat that check. That adds blocking I/O to terminal and error handling; resource failures can also suppress the event intended to diagnose them.

**Fix:** Specify live authorization semantics without requiring literal per-event storage reads—for example, an in-memory snapshot updated by consent mutations plus file/registry change notifications, evaluated once per logical emission.

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Node.js
+
+```typescript
+const outcome = await requestTelemetryConsent(async (prompt) => {
+  // Render every field verbatim.
+  return 'yes';
+}, 'en-US');
+
+const status = queryTelemetryConsent();
```
MGudgin: **Medium (correctness/API) — The specified Node status and withdrawal APIs synchronously block the event loop.**

The final Node implementation follows these synchronous examples with `execFileSync` and a five-second timeout. A missing, slow, or contended native executable therefore blocks the application's main event loop; reading status, policy, and prompt state separately can spawn the same process multiple times.

**Fix:** Make the primary status and withdrawal APIs Promise-based and use an asynchronous child process. If synchronous variants are retained, name and document them explicitly as blocking.

docs/telemetry/telemetry-consent-design.md
```diff
@@ -0,0 +1,293 @@
...
+
+The public ETW event identities remain:
+
+- `MXC.Execution`
+- `MXC.Error`
+
+Collectors may continue filtering those names; consent stabilization does not
+rename the provider or events.
+
+## Validation and release gate
```
MGudgin, line 195: **Medium (test coverage) — The release gate omits the real Node stdio presenter protocol.**

The protocol above includes framing, a challenge, prompt-version binding, asynchronous presentation, timeout, and child teardown. The final Node unit tests replace the entire protocol runner, so they never exercise the actual subprocess parser/writer; Rust tests cover challenge comparison but not Node's transport behavior.

**Fix:** Add an integration test using a controlled fake executor or debug binary that drives fragmented output, challenge/version mismatch, slow presenter behavior, malformed responses, child exit, timeout phases, and cleanup.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Developers implementing our sandbox would opt to call our telemetry API for the same reason that customers do: out of a shared self-interest in helping make the product better. In theory we should also be able to share silo'ed app data back to them about the performance of Windows and their specific implementation, though that's beyond the scope of our present implementation.

The alternative for an SDK would be to take an "opt out" approach which a handful of other SDKs do, but that's probably not appropriate in our case.

jsidewhite (dismissed):
:shipit:

RamonArjona4:
Correction: at the user's request, the latest local wording changes that expanded the caller-controlled correlation-vector disclosure have been reverted to the prior PR #818 text. The earlier implementation findings and the existing documentation treatment remain unchanged. No follow-up documentation changes have been pushed.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: A caller might get telemetry status to tell the end user that telemetry is on or off.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Addressed in the current stack by `sdk/node/tests/unit/default-consent-protocol-runner.test.ts`, which exercises the actual protocol runner with controlled child and presenter seams, including timeout, malformed-line, close-order, and presenter-failure paths. Resolving as fixed later in the stack.

MGudgin (approved):
Re-verified against current head `3f0df3c`.

The privacy inventory overclaim is fixed in this PR. The Node non-blocking consent APIs and real presenter-protocol integration tests are present downstream in the current stack. The remaining per-emission synchronous authorization I/O is partially improved (one check per logical event) and is a non-blocking reliability/performance follow-up for this documentation PR.

Approved under the stack-aware review.

PR #819, opened 2026-08-12
Add telemetry consent and policy foundation

RamonArjona4:

-----

## Summary

Introduces the privacy foundation behind stable MXC telemetry: versioned consent resources, fail-closed consent persistence and status, withdrawal, the administrative deny-only policy ceiling, and balanced ETW provider lifetime behavior.

This is PR 2 of 5 and depends on the normative contract in PR 1.

## Stack

| Order | Change | PR |
|---|---|---|
| **2** | **Consent and policy foundation** | **#819** |

Review only this PR's diff; prerequisite design context is in the PR above.


src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,1312 @@
...
+    /// This is the single definition of that policy for every MXC consumer
+    /// surface — the Rust SDK, the C ABI, the C# SDK, the Node SDK, and the
+    /// `wxc-exec --telemetry-consent-status` JSON all derive their answer
+    /// from here rather than re-deriving `state == Undetermined` in their own
+    /// language. If the policy ever grows (e.g. re-prompting after a
+    /// materially changed data-collection scope), it changes here once.
+    ///
+    /// Never true for [`NotApplicable`](ConsentState::NotApplicable): MXC
+    /// collects no telemetry off Windows, so there is nothing to consent to
+    /// and a prompt would be asking the user to decide something moot.
```
jsidewhite: Lot of verbiage describing a single line of code.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,1312 @@
...
+    /// "granted" record — undermining the very consent MXC is supposed to
+    /// own. The known-folder API resolves the path registered for the
+    /// process's own user token, independent of environment state.
+    ///
+    /// `SHGetKnownFolderPath` is COM IPC plus a registry read, and — because
+    /// we bind it explicitly to the *process* token (see
+    /// [`resolve_known_folder_local_app_data`]) — the answer cannot change for
+    /// the lifetime of the process, so the result is memoized. A long-lived
+    /// host application querying consent through the SDK/FFI on every
+    /// operation should not pay that cost repeatedly.
```
jsidewhite: Over-explained.

Comment probably doesn't need to say it's calling `SHGetKnownFolderPath` or explain how `SHGetKnownFolderPath` is implemented (com/reg).

I'm still not convinced this should be sold as some kind of immutable consent datastore.  It's just a file in the user profile, which is almost certainly directly editable by whoever launched this process (i.e. no reason to spoof the environment variable).

src/core/wxc_common/src/telemetry/consent_prompt.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Canonical US English consent resource, version 1.
+pub const EN_US_CONSENT_PROMPT: ConsentPrompt = ConsentPrompt {
+    resource_version: CONSENT_RESOURCE_VERSION,
+    locale: FALLBACK_LOCALE,
+    title: ConsentMessage {
+        id: "telemetry-consent-title",
+        text: "Help improve Microsoft eXecution Container (MXC)",
+    },
+    body: ConsentMessage {
```
jsidewhite: >ConsentMessage [](http://example.com/codeflow?start=10&length=14)

how come this doesn't use your resource file idea

jsidewhite (dismissed):
:shipit:

src/core/wxc_common/src/telemetry/consent_prompt.rs
RamonArjona4: Took my eye off copilot. FIxing this now

src/core/wxc_common/src/telemetry/consent.rs
RamonArjona4: Addressed in 8f9fdd4. I shortened this docs block to a concise summary of the shared predicate and removed the line-by-line narration.

src/core/wxc_common/src/telemetry/consent.rs
RamonArjona4: Addressed in 8f9fdd4. I trimmed this to a short note and removed the immutable-datastore framing; it now only states the known-folder lookup rationale and process-level cache.

src/core/wxc_common/src/telemetry/consent_prompt.rs
RamonArjona4: Addressed in 8f9fdd4. The canonical prompt now comes from `src/core/wxc_common/resources/telemetry/consent/en-US.json`, embedded at build time via `build.rs`, and consumed from `consent_prompt.rs`.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,1286 @@
...
+
+fn persist_presented_decision(
+    decision: ConsentDecision,
+    prompt: &ConsentPrompt,
+    policy: super::policy::PolicyState,
+    status: ConsentStatus,
+) -> Result<ConsentActionOutcome, ConsentActionError> {
+    match decision {
+        ConsentDecision::Yes => {
+            platform::write_presented(true, "prompt", prompt)
```
MGudgin: **Medium (security/reliability) — Re-check policy and consent state before persisting the presenter's decision.**

`consent_preflight` snapshots policy/status before invoking a presenter that may remain open indefinitely, and this write uses that stale snapshot. If policy becomes blocked while the UI is open, the code still persists a dormant grant and returns the old policy in `ConsentActionOutcome`; an intervening decision from another surface can likewise be overwritten by the stale dialog.

**Fix:** Re-read policy and status immediately before an affirmative write. Return `PolicyBlocked` or a typed state-changed result when the preflight conditions no longer hold.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,1286 @@
...
+
+    pub(super) fn read_status() -> ConsentStatus {
+        let Some(path) = consent_file_path() else {
+            return ConsentStatus {
+                stored_state: ConsentState::Undetermined,
+                effective_state: ConsentState::Undetermined,
+                reason: Some(ConsentStatusReason::StoreUnreadable),
+            };
+        };
+        let data = match with_io_retry(|| fs::read_to_string(&path)) {
```
MGudgin: **Medium (reliability) — Bound the consent-file read.**

The consent file lives in a user-writable directory, but `read_to_string` allocates for the entire file before parsing the tiny record. A malformed or hostile same-user process can replace it with an arbitrarily large file and force excessive allocation whenever consent is queried.

**Fix:** Read through a small fixed upper bound comfortably above the schema size, and fail closed with `StoreMalformed`/`StoreUnreadable` when the file exceeds it.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,1286 @@
...
+        // (process id + a random suffix, so two concurrent `wxc-exec`
+        // grant/revoke invocations never share — and thus never race on —
+        // the same temp path), then rename over the real path. A crash
+        // mid-write never leaves a torn/corrupt file in place of a
+        // previously-valid one; if the rename itself fails, the temp file is
+        // removed rather than left behind as a leaked, orphaned artifact.
+        let unique = format!("{}-{:x}", std::process::id(), random_suffix());
+        let tmp_path = path.with_extension(format!("json.{unique}.tmp"));
+
+        if let Err(e) = with_io_retry(|| fs::write(&tmp_path, &json)) {
```
MGudgin: **Medium (reliability) — Flush the new record before replacing the old decision.**

The temp file is written and immediately renamed without `sync_data`/`sync_all`. Rename prevents ordinary torn writes, but it does not make the new bytes durable across a crash or power loss. Losing a withdrawal can leave the prior granted record as the effective decision after restart.

**Fix:** Write through an explicit file handle, flush and sync it, then atomically replace the destination; sync the containing directory where the platform supports it.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,1286 @@
...
+            status,
+        } => (prompt, policy, status),
+    };
+
+    let decision = presenter(prompt).map_err(ConsentActionError::Presenter)?;
+    persist_presented_decision(decision, prompt, policy, status)
+}
+
+/// Asynchronous counterpart to [`request_consent`].
+pub async fn request_consent_async<F, Fut>(
```
MGudgin, line 273: **Medium (performance) — The async API performs blocking storage work on the caller's executor thread.**

This function calls synchronous registry/file preflight before its only `await`, then performs synchronous persistence and another status read afterward. Those paths can also execute `thread::sleep` retries. Awaiting this from a single-threaded or latency-sensitive runtime therefore blocks unrelated tasks.

**Fix:** Offload preflight and persistence to a blocking executor, or make the storage operations genuinely asynchronous. If runtime neutrality prevents that here, document the blocking contract explicitly rather than presenting the function as non-blocking.

src/mxc_telemetry/src/lib.rs
```diff
@@ -60,7 +60,7 @@ mod provider {
     /// concurrent `init`/`shutdown` race from leaving the wrapper flag and the
     /// provider's real state out of sync (e.g. flag set to registered after a
     /// `register()` that actually failed, or a double register/unregister).
-    static REGISTERED: Mutex<bool> = Mutex::new(false);
+    static REGISTERED: Mutex<usize> = Mutex::new(0);
```
MGudgin, line 70: **Low (test coverage) — Stress concurrent emission against final provider shutdown.**

The new mutex/refcount tests verify nested `init`/`shutdown`, but every provider test is serialized and no test logs while another thread releases the final registration. That is the concurrency boundary this refcount now owns.

**Fix:** Add a bounded multithreaded test that emits while another thread repeatedly acquires/releases registrations, asserting no panic, deadlock, or inconsistent final `is_active` state.

src/mxc_telemetry/src/lib.rs
```diff
@@ -100,32 +100,35 @@ mod provider {
         });
 
         let mut registered = REGISTERED.lock().unwrap_or_else(|e| e.into_inner());
-        if *registered {
+        if *registered > 0 {
             // Already registered — the tracelogging crate panics on double
-            // register, so we must not call `register()` again.
+            // register, so retain a reference instead.
+            *registered += 1;
```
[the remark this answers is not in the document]
RamonArjona4: Addressed: consent-store reads and writes used by async presentation run on the dedicated blocking worker rather than performing filesystem/lock work inline on the async caller. Resolving as fixed.

src/core/wxc_common/src/telemetry/consent.rs
RamonArjona4: Addressed with concurrent provider initialization, logging, and shutdown coverage, including exactly-once registration lifetime behavior. Resolving as fixed.

src/mxc_telemetry/src/lib.rs
RamonArjona4: Addressed in PR820: error telemetry preserves the originally requested containment kind instead of collapsing failed resolution to `Unknown`. Resolving as fixed later in the stack.

src/core/wxc_common/src/telemetry/mod.rs
```diff
@@ -223,6 +258,7 @@ pub fn emit_completion(
 
     log_execution(&ExecutionEvent {
         backend,
+        sandbox_kind: backend,
         exit_code: response.exit_code,
         outcome,
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in PR820: CI builds the debug executor with `test-support`, and the smoke scripts use authenticated, isolated consent/policy overrides with absolute executor paths. Resolving as fixed later in the stack.

tests/scripts/run_telemetry_etw_smoke_test.ps1
```diff
@@ -152,31 +132,15 @@ Write-Host "ETW session started, writing to $etlFile"
...
-            Remove-Item Env:MXC_TEST_LOCALAPPDATA_OVERRIDE -ErrorAction SilentlyContinue
-        } else {
-            $env:MXC_TEST_LOCALAPPDATA_OVERRIDE = $previousTestOverride
-        }
-    }
+    # Run with --experimental to enable the telemetry section. The provider is
+    # registered during init (before execution); the MXC.Execution / MXC.Error
+    # events are emitted on completion, after the runner returns. The sandbox
+    # itself may fail (e.g. AppContainer prerequisites), but completion
+    # telemetry still fires for the failure, so events should be captured.
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in PR820: emission uses a fresh `EmissionAuthorization` decision at the logical emission boundary, and paired writes cannot continue after consent withdrawal or an administrative-policy change. Resolving as fixed later in the stack.

src/core/wxc_common/src/telemetry/events.rs
```diff
@@ -91,6 +93,7 @@ pub fn log_execution(event: &ExecutionEvent<'_>) {
 
     mxc_telemetry::log_execution(
         event.backend,
+        event.sandbox_kind,
```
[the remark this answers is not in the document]
RamonArjona4: The current executable paths initialize telemetry once per process, and library streaming paths pair registration with handle teardown, so this is not a present leak. I opened #952 to replace the low-level reference counting with an RAII registration guard before future multi-invocation paths make the pairing easier to violate. Resolving in favor of that task.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,2197 @@
...
+                status: current_status,
+                policy: current_policy,
+            });
+        }
+        let is_concurrent_grant_being_withdrawn = matches!(decision, ConsentDecision::No)
+            && current_status.effective_state == ConsentState::Granted
+            && status.effective_state != ConsentState::Granted;
+        if current_policy != policy
+            || (current_status.effective_state != status.effective_state
+                && !is_concurrent_grant_being_withdrawn)
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in `d9fda928`. Consent records now carry a monotonic revision captured with status under the cross-process store lock, and presenter persistence rejects a changed revision even when effective consent remains `Denied`. A deterministic same-state withdrawal versus stale-grant regression covers the race.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,2278 @@
...
+            .ok_or_else(|| "invalid telemetry consent path".to_string())?;
+        fs::create_dir_all(dir).map_err(|e| format!("failed to create {}: {e}", dir.display()))?;
+        let lock_path = dir.join(CONSENT_LOCK_FILE);
+        let mut attempts = 0;
+        let file = loop {
+            let mut options = fs::OpenOptions::new();
+            options
+                .write(true)
+                .create_new(true)
+                .custom_flags(windows::Win32::Storage::FileSystem::FILE_FLAG_DELETE_ON_CLOSE.0);
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in b58d4f81. The lock handle now explicitly requests `GENERIC_WRITE | DELETE` when using `FILE_FLAG_DELETE_ON_CLOSE`, and a Windows regression verifies the lock exists while held and is removed when released.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -0,0 +1,2299 @@
...
+        // mid-write never leaves a torn/corrupt file in place of a
+        // previously-valid one; if the rename itself fails, the temp file is
+        // removed rather than left behind as a leaked, orphaned artifact.
+        let unique = format!("{}-{:x}", std::process::id(), random_suffix());
+        let tmp_path = path.with_extension(format!("json.{unique}.tmp"));
+
+        if let Err(e) = with_io_retry(|| {
+            let mut file = fs::File::create(&tmp_path)?;
+            file.write_all(json.as_bytes())?;
+            file.sync_all()
```
MGudgin: **Medium (reliability/privacy) — Complete crash-durable withdrawal persistence.**

**Attribution:** `introduced_by_change` — this PR introduces the consent store, withdrawal marker, and replacement protocol.

The temporary consent record is now `sync_all`ed, which is good, but the withdrawal marker is created with an unsynced write and neither marker creation/removal nor destination rename is followed by a directory sync or write-through replacement. A power loss can therefore restore the old granted record without a durable marker proving withdrawal was pending.

**Fix:** Durably create the marker, perform a write-through/replace operation (or sync the directory metadata where supported), and durably remove the marker only after the denied record is committed.

src/mxc_telemetry/src/lib.rs
```diff
@@ -391,17 +512,63 @@ mod tests {
...
         let _ = init("", "");
-        log_execution("", 0, "", 0, "", "", "");
-        log_error("", "", 0, "", "");
+        log_execution("", "", 0, "", 0, "", "", "");
+        log_error("", "", "", 0, "", "");
         shutdown();
     }
+
+    #[test]
+    fn concurrent_init_shutdown_is_safe() {
```
MGudgin: **Low (test coverage) — Exercise an active emit racing the final unregister.**

**Attribution:** `introduced_by_change` — this PR introduces the provider registration refcount and its concurrency contract.

`concurrent_init_shutdown_is_safe` retains a registration in each emitter thread until after `log_execution`, so the final unregister cannot occur while an emit is active. It proves nested lifecycle safety, but not the boundary where the last reference disappears concurrently with a write.

**Fix:** Add a controlled stress test or synchronization hook that holds an emit at the provider call while another thread releases the final registration, then assert no panic, deadlock, or inconsistent active state.

src/core/wxc_common/src/telemetry/consent.rs
MGudgin: Thanks—calling `sync_all()` on the temporary consent record addresses the original risk that the replacement file's contents might never reach stable storage. The remaining issue is that the complete withdrawal transaction is still not crash-durable.

The pending-withdrawal marker is created using an unsynced `fs::write`. After the temporary denied record is flushed, it is installed using `fs::rename`, but that does not guarantee the directory metadata recording the replacement has reached stable storage. The marker is then removed without durably committing that removal either.

This leaves a power-loss failure mode where the filesystem recovers the previous `granted` record, while also losing or never committing the pending marker. On restart, MXC then has neither the denied record nor the marker that would force fail-closed recovery, so the attempted withdrawal can be silently undone.

Please durably commit the marker before replacing the record, perform a write-through replacement or sync the relevant directory metadata where supported, and durably remove the marker only after the denied record is known to be committed. The important invariant is that after every crash point, either the denied record is present or a durable marker forces recovery to denied.

bbonaby:
@RamonArjona4 looks like this one's got merge conflicts. Can you merge with main and just verify those suppressed comments from Copilot one more time please. Thanks!

RamonArjona4:
@bbonaby Merged & resolved and verified suppressed comments, now just waiting for checks.

MGudgin (approved):
Re-verified the existing findings against `55ce0a7`. The stale-presenter, bounded-read, async-storage, and crash-durable withdrawal findings are addressed, and the targeted release consent and telemetry-provider tests pass. The remaining emit-vs-final-unregister test-coverage gap is Low and non-blocking.

PR #820, opened 2026-08-12
Complete stable telemetry runtime and lifecycle integration

RamonArjona4:

-----

## Summary

Completes stable telemetry runtime and lifecycle integration across MXC:

- Promotes telemetry from an experimental field to the stable `0.9` configuration contract.
- Updates wire models, exact-version contracts, adapters, generated schemas, and Node types.
- Connects telemetry to the execution engine and all three executors: Windows, Linux, and macOS.
- Covers one-shot, state-aware, attached, streaming, cancellation, timeout, error, panic, and abandoned-handle paths.
- Enforces live consent and administrative-policy authorization.
- Adds typed JSON consent-maintenance requests and retains the read-only status CLI.
- Persists lifecycle correlation across separate state-aware invocations.
- Registers the WINEXT provider contract and required privacy metadata.
- Adds unit, schema, smoke, ETW, and end-to-end coverage.

This is PR 3 of 5 and depends on the consent/policy foundation in PR 2.

## Stack

| Order | Change | PR |
|---|---|---|
| **3** | **Stable config and executor integration** | **#820** |

Review only this PR's diff; prerequisite behavior is in the PR above.


schemas/dev/mxc-telemetry-consent.schema.1.json
```diff
@@ -0,0 +1,79 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
jsidewhite: >$schema [](http://example.com/codeflow?start=3&length=7)

Not sure this file belongs here.  these schemas are effectively **public api surface**.

Your schema here is an implementation detail of an internal file format that no caller should need to care about - at least, I think right?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -73,6 +73,25 @@ where
...
+    let value: serde_json::Value = serde_json::from_str(experimental.get())
+        .map_err(|error| WxcError::ConfigParse(error.to_string()))?;
+    if value
+        .as_object()
+        .is_some_and(|object| object.contains_key("telemetry"))
+    {
+        return Err(WxcError::ConfigParse(
+            "'experimental.telemetry' has moved to the stable section; \
+             use top-level 'telemetry' instead."
+                .to_string(),
```
jsidewhite, line 97: should we only reject legacy telemetry if the passed config's version is < 0.8.0?

schemas/dev/mxc-config.schema.0.9.0-dev.json
```diff
@@ -1042,6 +1032,17 @@
       ],
       "description": "macOS Seatbelt backend configuration. Used when containment is `seatbelt`."
     },
+    "telemetry": {
```
jsidewhite, line 1317: >telemetry [](http://example.com/codeflow?start=5&length=9)

is it weird that telemetry is part of the schema, as opposed to an ancillary property?

jsidewhite (dismissed):
:shipit:

schemas/dev/mxc-telemetry-consent.schema.1.json
RamonArjona4: This is a detail I think I'm unclear about.  Do we expect that we'll ever have to support external callers trying to invoke telemetry config via the JSON config file with the EXE?  If we don't then this is probably not needed.

schemas/dev/mxc-config.schema.0.9.0-dev.json
RamonArjona4: In this context telemetry is a gate that the invocation is requesting telemetry. We're trying to keep execution policy "enable telemetry gate" separate from state mutation ("I consent to telemetry, I withdraw consent, I would like status.")  But as noted earlier exposing the surface of telemetry in the JSON schema might be overkill.

src/core/mxc_engine/src/lib.rs
```diff
@@ -65,20 +67,207 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
 /// [`ErrorCode::UnsupportedContainment`].
 pub fn spawn(request: &SandboxRequest) -> Result<Box<dyn SandboxProcess>, Error> {
     let mut logger = Logger::new(Mode::Buffer);
-    let process = dispatch::spawn_runner(&request.inner, &mut logger).map_err(Error::from)?;
+    let telemetry_active = request
+        .inner
+        .telemetry
+        .as_ref()
+        .map(|config| telemetry::init(config, &mut logger))
```
MGudgin, line 98: **High (security) — Registering ETW from `mxc_engine` invalidates the provider's executable-only unsafe precondition.**

This newly moves provider registration into a library used by `mxc-sdk` and the existing `mxc_ffi` cdylib. `mxc_telemetry` still justifies the unsafe registration by stating MXC is an executable, not a DLL, so unload ordering is irrelevant. A leaked streaming handle or explicit native-library unload can therefore leave ETW holding callbacks into an unmapped module.

**Fix:** Keep registration in process-lifetime executables, or add an explicit FFI/global teardown contract that prevents unloading until all provider references are released, and update the unsafe preconditions accordingly.

src/core/wxc_common/src/telemetry/consent_cli.rs
```diff
@@ -0,0 +1,654 @@
...
+}
+
+#[derive(Deserialize)]
+struct MaintenanceDiscriminator {
+    command: Option<String>,
+}
+
+/// Handle a typed consent-maintenance envelope, if present.
+pub fn handle_maintenance_input(input: &str, is_base64: bool) -> Option<ConsentCliOutcome> {
+    let json = match crate::config_parser::decode_request_input(input, is_base64) {
```
MGudgin: **High (reliability/cross-platform) — The maintenance probe consumes ordinary config streams before normal parsing reads them again.**

Every executor calls this probe first. For a non-maintenance request, `decode_request_input` reads the supplied path here, then the normal loader reads the same path again. Regular files pay duplicate I/O/parsing; named pipes, `/dev/stdin`, and process-substitution paths are drained by the first read and fail or block on the second.

**Fix:** Decode the input once in the executor and pass the loaded JSON to both maintenance discrimination and ordinary request parsing.

src/core/mxc_engine/src/lib.rs
```diff
@@ -65,20 +67,207 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
+                    self.started.elapsed(),
+                );
+            }
+        }
+        self.active = false;
+    }
+}
+
+impl Drop for TelemetryProcess {
+    fn drop(&mut self) {
```
MGudgin, line 399: **Medium (reliability) — Dropping a live telemetry wrapper releases the provider without emitting the promised terminal event.**

`wait()` emits, but `Drop` only shuts down. `kill()` likewise only forwards, and `try_wait()` ignores its error branch. Kill-then-free, drop-without-wait, or a failed poll can therefore lose the invocation's `Execution`/`Error` record while releasing its provider reference, contradicting the wrapper's “exactly one terminal event” contract.

**Fix:** Use an explicit terminal-state machine and emit a bounded killed/cancelled/error outcome on kill, drop, and `try_wait` failure before releasing the provider exactly once.

src/core/wxc_common/src/telemetry/consent_cli.rs
```diff
@@ -0,0 +1,654 @@
...
+    let presentation = maintenance_response(
+        action,
+        wire::TelemetryConsentResult::PresentationRequired,
+        None,
+        Some(prompt_to_wire(prompt)),
+        Some(challenge.clone()),
+    );
+    let json = serde_json::to_string(&presentation)
+        .map_err(|error| format!("failed to serialize consent presentation: {error}"))?;
+    println!("{json}");
```
MGudgin: **Medium (reliability) — A closed presenter pipe causes a Rust panic.**

`println!` panics on stdout write failure. If the SDK host times out, crashes, or closes its pipe before the child emits the presentation envelope, this turns a routine IPC disconnect into an uncontrolled panic before normal executor setup.

**Fix:** Write through `stdout().lock()`/`writeln!` and map `BrokenPipe` or other write failures to a clean transport error and nonzero outcome.

src/core/wxc_common/src/telemetry/mod.rs
```diff
@@ -225,48 +174,53 @@ fn classify_failure(phase: &FailurePhase) -> FailureReason {
...
-        outcome,
-        duration_ms: elapsed.as_millis() as u64,
-        failure_reason,
-        // One-shot execution — no state-aware lifecycle phase.
-        phase: "",
-        // One-shot execution — already correlated by AppSessionGuid, no
-        // cross-phase lifecycle to join.
-        correlation_vector: "",
-    });
+    if invocation_can_emit(active) {
```
MGudgin: **Medium (performance/reliability) — Reuse one authorization result per logical emission.**

The outer completion path checks authorization, then this check rereads the consent file and policy registry before `Execution`, and line 223 does it again before the paired `Error`. Similar repetition exists in early-exit, crash/cancellation, and state-aware paths. This adds blocking retry-capable I/O and can split an intended event pair if authorization changes between reads.

**Fix:** Perform one live check immediately before each logical emission and pass the result through to both event writes.

src/core/mxc_engine/src/lib.rs
```diff
@@ -65,20 +67,207 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
+        .unwrap_or(false);
+    let containment = request.inner.containment.clone();
+    let started = std::time::Instant::now();
+    let process = match dispatch::spawn_runner(&request.inner, &mut logger) {
+        Ok(process) => process,
+        Err(error) => {
+            telemetry::emit_sdk_early_exit(
+                telemetry_active,
+                &containment,
+                telemetry::FailureReason::InitError,
```
MGudgin: **Medium (correctness) — Preserve the actual error category instead of reporting every spawn failure as initialization.**

`dispatch::spawn_runner` returns typed `MxcError` codes, but this path emits every failure as `InitError`. Malformed requests, unsupported containment, and policy validation failures are therefore recorded under the wrong bounded telemetry category.

**Fix:** Share the exhaustive `MxcErrorCode` → `FailureReason` mapping used by the state-aware path and emit the mapped reason here.

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -132,7 +165,37 @@ pub fn run_state_aware_json(request_json: &str, dry_run: bool) -> Result<String,
         )));
     }
 
-    match run_state_aware(parsed, dry_run).map_err(Error::from)? {
+    let phase = parsed.phase;
+    let phase_name = phase.as_str();
+    let incoming_correlation = parsed.correlation_vector.clone();
+    let mut logger = Logger::new(Mode::Buffer);
```
MGudgin: **Medium (correctness/observability) — State-aware telemetry initialization warnings are discarded.**

Both state-aware entry points initialize telemetry with a buffered `Logger`, but never surface `logger.take_warnings()`. Ordinary SDK spawn attaches those warnings to its process wrapper, while state-aware envelope and streaming calls silently lose provider-registration diagnostics.

**Fix:** Include warnings in envelope results and wrap streaming state-aware processes with the same warning propagation used by `spawn`.

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -132,7 +165,37 @@ pub fn run_state_aware_json(request_json: &str, dry_run: bool) -> Result<String,
...
+    let backend = resolve_backend(&parsed)
+        .map(|backend| backend.wire_name().to_string())
+        .unwrap_or_else(|_| "unknown".to_string());
+    let correlation = phase_correlation(telemetry_active, phase, incoming_correlation.as_deref());
+    let started = std::time::Instant::now();
+    let mut outcome = run_state_aware(parsed, dry_run);
+    if phase == Phase::Provision && telemetry_active {
+        inject_correlation_vector(&mut outcome, &correlation);
+    }
+    telemetry::emit_sdk_state_aware(
```
MGudgin: **Medium (test coverage) — Add an engine-level state-aware telemetry execution test.**

The shared crate now tests event planning/emission, and engine tests cover correlation helpers, but no test drives this integration end to end. Phase attribution, provision-vector injection, engine error mapping, and provider release can still regress independently while both component suites stay green.

**Fix:** Add an engine/integration test with a capture sink that executes representative provision and exec outcomes and asserts phase, shared correlation base, error mapping, and final provider release.

src/core/mxc_engine/src/lib.rs
RamonArjona4: Addressed: the provider registration/drop contract is explicit and the owning runtime types pair each successful registration with exactly one release. Resolving as fixed.

src/core/wxc_common/src/telemetry/consent_cli.rs
RamonArjona4: Addressed: the runtime parses the telemetry policy once and carries the typed result through initialization instead of reparsing the same JSON. Resolving as fixed.

src/core/mxc_engine/src/lib.rs
RamonArjona4: Addressed: `TelemetryProcess` emits a synthetic terminal outcome on kill, drop, and polling-error paths, with focused exactly-once tests. Resolving as fixed.

.github/workflows/Build.Windows.Job.yml
```diff
@@ -66,11 +66,32 @@ jobs:
...
+          cargo test --locked --release --target ${{ matrix.target }} -p wxc_common telemetry::policy
+
       - name: Test
         run: cargo test --locked --release --target ${{ matrix.target }}
              --no-default-features
              --features "${{ matrix.features }}"
 
+      - name: Build debug executor for isolated telemetry smoke tests
+        if: matrix.arch == 'x64'
+        run: cargo build --locked -p wxc
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in b04119bf. The telemetry smoke-test executor is built with the debug-only test-support feature, so the consent and policy overrides are compiled in and the scripts exercise the isolated stores.

src/core/mxc_engine/src/lib.rs
```diff
@@ -65,31 +67,349 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
+        self.inner.id()
+    }
+
+    fn kill(&mut self) -> std::io::Result<()> {
+        // Forward the kill first so its I/O outcome is what we return; then
+        // emit the synthesised cancellation event before the provider
+        // reference is released. `emit_synthetic` is idempotent, so a
+        // subsequent `wait` on the (now-killed) child is a no-op.
+        let result = self.inner.kill();
+        self.emit_synthetic(SyntheticTerminal::Killed);
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in 7a69570e. A failed kill leaves the terminal telemetry slot active for a later wait, while a confirmed kill emits the explicit cancelled mapping with exit code 130.

src/core/mxc_engine/src/lib.rs
```diff
@@ -65,31 +67,384 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
-    let process = dispatch::spawn_runner(&request.inner, &mut logger).map_err(Error::from)?;
-    let mut warnings = process.warnings().to_vec();
-    for warning in logger.take_warnings() {
-        if !warnings.contains(&warning) {
-            warnings.push(warning);
+    let telemetry_active = request
+        .inner
+        .telemetry
+        .as_ref()
+        .map(|config| telemetry::init(config, &mut logger))
```
MGudgin: **High (security/reliability) — Own the provider registration immediately with RAII.**

**Attribution:** `introduced_by_change` — this PR newly registers telemetry inside the SDK/FFI engine path.

Registration occurs here, before `dispatch::spawn_runner`, but the owning `TelemetryProcess` is not constructed until spawn succeeds. A panic in that interval is caught at the FFI boundary and can leave the provider reference registered with no owner to release it before `mxc_ffi` is unloaded, recreating the dangling ETW-callback risk the new DLL lifetime contract is meant to prevent.

**Fix:** Acquire registration through an RAII guard immediately; transfer the guard into `TelemetryProcess` only after successful spawn, and let every error/unwind path drop it automatically.

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -279,4 +461,280 @@ mod tests {
...
+    /// state. Two chained failing calls exercise this: if the first one
+    /// leaked, the second would see stale process context (or, worse, would
+    /// double-emit).
+    ///
+    /// We can't directly observe the ETW provider from user-space test code
+    /// on non-Windows platforms, but we can assert the observable contract:
+    /// each call finishes cleanly with the expected error, and the process
+    /// stays healthy across the two.
+    #[test]
+    fn engine_state_aware_provider_released_between_calls() {
```
MGudgin: **Medium (test coverage) — Drive the engine's state-aware telemetry path with telemetry actually enabled.**

**Attribution:** `introduced_by_change` — this PR adds state-aware telemetry initialization, correlation, emission, and release in `mxc_engine`.

`engine_state_aware_provider_released_between_calls` omits telemetry config, so it cannot observe provider acquisition/release or emitted records. Shared-crate tests validate the component emitter, but the engine integration (phase attribution, provision vector injection, error mapping, and release) can still regress independently.

**Fix:** Add an engine/integration capture seam and execute representative provision/exec outcomes with telemetry enabled, asserting phase, correlation, error mapping, emitted records, and final provider release.

docs/schema.md
```diff
@@ -199,6 +199,10 @@ cannot mix both formats in one request.
         "extraMachLookups": []             // Additional Mach service global-names the inner process may resolve
     },
 
+    "telemetry": {                         // Telemetry (Windows only)
+        "enabled": true                    // Emit TraceLogging ETW events via pure Rust tracelogging crate
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 944c6986. The schema documentation now states that telemetry.enabled only requests emission and that MXC user consent plus a permitting administrative policy are also required.

src/core/wxc_common/src/telemetry/events.rs
```diff
@@ -116,7 +116,7 @@ pub fn log_execution(event: &ExecutionEvent<'_>) {
 pub fn log_error(ctx: TelemetryContext<'_>, error_type: FailureReason, exit_code: i32) {
     mxc_telemetry::log_error(
         ctx.backend,
-        ctx.sandbox_kind,
+        super::sandbox_kind_for(ctx.backend, None),
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 944c6986. MXC.Error now receives the request-scoped sandbox kind separately from the concrete backend, the process-global panic/Ctrl-C context preserves both values, and the event contract table documents mxc.sandbox_kind. Regression coverage verifies the Error attribution.

src/core/mxc_engine/src/policy.rs
```diff
@@ -685,6 +685,19 @@ impl SandboxRequest {
...
+    /// Enable or disable telemetry for this invocation.
+    ///
+    /// Enabling this per-request switch is necessary but not sufficient:
+    /// telemetry still requires persisted user consent and an administrative
+    /// policy that permits collection. It is independent of experimental mode.
+    pub fn set_telemetry_enabled(&mut self, enabled: bool) -> &mut Self {
+        self.inner.telemetry = Some(TelemetryConfig {
+            enabled: Some(enabled),
+            ..Default::default()
+        });
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 944c6986. SandboxRequest now retains the original Containment intent before host resolution and set_telemetry_enabled copies its wire name into TelemetryConfig. Tests cover abstract process/vm and explicit containment requests.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1325,6 +1540,10 @@ fn convert_wire_config(
     // Top-level `seatbelt` config. Configs using `experimental.seatbelt` are
     // rejected above.
     let seatbelt = cfg.seatbelt.map(make_seatbelt_config);
+    let telemetry = cfg.telemetry.map(|raw| TelemetryConfig {
+        enabled: raw.enabled,
+        requested_sandbox_kind: Some(requested_sandbox_kind(cfg.containment.as_ref())),
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in bfb0ecb7. A shared versioned-field validator now rejects top-level telemetry with an explicitly pre-0.9 schema version across base64, raw JSON, value, one-shot, and state-aware loaders. Regression tests cover the affected paths.

src/core/mxc_engine/src/lib.rs
```diff
@@ -71,31 +73,384 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
+}
+
+impl Drop for TelemetryProcess {
+    fn drop(&mut self) {
+        // If `wait` / `try_wait(Ok(Some))` / `kill` already emitted the
+        // terminal event, `active` is false and this is a no-op. Otherwise
+        // synthesise one so the "exactly-one terminal event" invariant holds
+        // for the drop-without-wait path as well. `emit` also releases the
+        // provider reference.
+        self.emit_synthetic(SyntheticTerminal::Dropped);
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in bfb0ecb7. TelemetryProcess::drop now polls the child first and emits the observed exit code when it has completed; only a running or unpollable child is classified as abandoned. The three dispositions are covered by unit tests.

sdk/node/src/state-aware-helper.ts
```diff
@@ -20,7 +20,7 @@ export const WSLC_STATE_AWARE_VERSION = '0.8.0-alpha';
 // Wire-format cross-cutting fields that live at the envelope's top level.
 // Anything else on a per-(backend, phase) Config is backend-specific and is
 // nested under `experimental.<backend>.<phase>`.
-export const CROSS_CUTTING_FIELDS = ['filesystem', 'network', 'ui', 'process'] as const;
+export const CROSS_CUTTING_FIELDS = ['filesystem', 'network', 'ui', 'process', 'telemetry'] as const;
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in bfb0ecb7. State-aware Node envelopes containing telemetry now default to 0.9.0-alpha and reject an explicitly older version before spawning. The ordinary per-backend defaults remain unchanged for requests without telemetry.

src/core/wxc_common/src/telemetry/mod.rs
```diff
@@ -245,18 +232,48 @@ pub fn init(config: &TelemetryConfig, logger: &mut Logger) -> bool {
         logger
             .log_line("telemetry: ETW provider registration failed; continuing without telemetry");
     }
+    if activated && !mxc_telemetry::IS_UTC_ROUTED {
+        logger.log_line(
+            "telemetry: events are emitted to local ETW only; this build has no provider group \
+             GUID, so nothing is routed to the Microsoft pipeline (set \
+             MXC_TELEMETRY_PROVIDER_GROUP_GUID at build time for an internal build)",
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in bfb0ecb7. Telemetry registration and non-routed-build messages are now retained in the SDK warning channel while remaining in the executor primary diagnostics. Warning API documentation was broadened to include operational warnings.

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -287,7 +331,8 @@ fn exec_state_aware_attached_with(
     experimental: bool,
     host_is_interactive: impl FnOnce() -> bool,
 ) -> Result<ExecOutcome, Error> {
-    let parsed = parse_state_aware(request_json, experimental)?;
+    let mut logger = Logger::new(Mode::Buffer);
+    let parsed = parse_state_aware(request_json, experimental, &mut logger)?;
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in bfb0ecb7. Attached state-aware exec now performs telemetry initialization, backend/request attribution, correlation lookup, terminal/error emission, and provider shutdown. Because the attached API has no warning-bearing handle, retained initialization warnings are surfaced non-panickingly on the explicitly attached host stderr.

src/core/wxc_common/src/telemetry/consent_cli.rs
```diff
@@ -0,0 +1,844 @@
...
+pub fn handle_consent_flags(flags: &ConsentCliFlags<'_>) -> Option<ConsentCliOutcome> {
+    if !(flags.status || flags.grant || flags.revoke || flags.source.is_some()) {
+        return None;
+    }
+
+    if flags.grant || flags.revoke || flags.source.is_some() {
+        return Some(ConsentCliOutcome::failure(
+            "Error: --telemetry-consent-grant, --telemetry-consent-revoke, and \
+             --telemetry-consent-source no longer change consent. Use the typed \
+             telemetry-consent JSON maintenance envelope instead."
```
jsidewhite: Doesn't this PR introduce these arguments?  Why are they deprecated on day one?

jsidewhite:
Normally a PR that "Promotes telemetry to the stable top-level configuration" should be very minimal - it's promoting to stable because it's baked in experimental.  Typically it doesn't modify behavior of the core feature.

src/core/wxc_common/src/telemetry/correlation_state.rs
```diff
@@ -0,0 +1,784 @@
...
+/// record. Returns an empty string when `active` is false, so an inactive
+/// telemetry provider pays for no RNG/clock/file-system work.
+pub fn pre_dispatch_vector(active: bool, is_provision: bool, sandbox_id: Option<&str>) -> String {
+    if !active {
+        return String::new();
+    }
+    prune_stale_records();
+    match sandbox_id {
+        Some(id) if !is_provision => phase_vector(id),
+        _ => seed(),
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 0ad1b035. The requested lifecycle record is now loaded and refreshed before stale pruning, protected from that sweep, and all correlation-store filesystem operations are serialized with an OS-backed cross-process lock.

src/core/mxc_engine/src/lib.rs
```diff
@@ -71,31 +73,402 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
+        let result = self.inner.try_wait();
+        match &result {
+            // Exit observed: emit the terminal event now.
+            Ok(Some(exit_code)) => self.emit(&Ok(*exit_code)),
+            // Still running: leave the invariant to a later `wait` / `kill` / `Drop`.
+            Ok(None) => {}
+            // Poll-error branch — emit a synthesised terminal event so the
+            // provider reference isn't released without accounting for the
+            // run, then return the error unchanged to the caller.
+            Err(_) => self.emit_synthetic(SyntheticTerminal::PollError),
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 0ad1b035. Non-timeout try_wait errors now preserve the exactly-once telemetry slot for a later confirmed outcome. ErrorKind::TimedOut remains terminal because the affected backends return it only after settling the timeout, and regression tests cover both paths.

src/core/wxc_common/src/telemetry/consent_cli.rs
RamonArjona4: Agreed. Fixed in 0ad1b035 by removing --telemetry-consent-grant, --telemetry-consent-revoke, and --telemetry-consent-source from all three executor CLI surfaces and deleting their compatibility handler/tests. The read-only --telemetry-consent-status command remains.

RamonArjona4:
> Normally a PR that "Promotes telemetry to the stable top-level configuration" should be very minimal - it's promoting to stable because it's baked in experimental. Typically it doesn't modify behavior of the core feature.

 The essential bits of writing events & uploading events remain unchanged but I agree we're definitely doing some heavy things in this middle PR.  Updated the title & the description to be more reflective of this.

src/core/wxc/src/main.rs
```diff
@@ -920,18 +956,27 @@ fn main() {
...
+        Some(Ok(json)) => Some(json),
+        Some(Err(error)) => {
+            eprintln!("Request error");
+            eprintln!("{error}");
+            process::exit(1);
+        }
+        None => {
+            if !cli.delete {
+                eprintln!(
+                    "Error: No config provided. Use a positional path, --config, or --config-base64"
```
jsidewhite, line 1191: >Error: No config provided. Use a positional path, --config, or --config-base64" [](http://example.com/codeflow?start=21&length=79)

Not super keen on overloading the `--config` arg with "telemetry consent" actions

schemas/dev/mxc-telemetry-consent.schema.1.json
jsidewhite: I conferred with @bbonaby about wire contracts, and I don't think this schema makes sense as the vehicle for consent at this point.

Can we simply implement `--telemetry-consent [status | withdraw | request]` on wxc-exec.exe and other executors\?

README.md
```diff
@@ -257,6 +257,17 @@ Telemetry is **off by default**. To keep it off, do not set
...
+Consent is managed separately from execution config:
+
+```powershell
+wxc-exec.exe --telemetry-consent status
+wxc-exec.exe --telemetry-consent request
+wxc-exec.exe --telemetry-consent withdraw
+```
+
+`telemetry.enabled` cannot request, grant, or withdraw consent. Consent
+maintenance is not accepted through `--config` or `--config-base64`.
```
jsidewhite: not necessary

.github/copilot-instructions.md
```diff
@@ -332,6 +332,7 @@ Telemetry is **Windows-only** and is never collected without explicit user conse
 - **Everything fails closed.** Any error, unreadable value, corrupt file, missing native library, or ambiguity must resolve to "no telemetry" (`Undetermined` consent / `Blocked` policy), never to a permissive state.
 - **Policy may restrict, but may never substitute for, consent.** The administrative policy is a deny-only ceiling: it can subtract from what a user permitted, never add to it. An administrator cannot opt a user in — a denied or never-asked user stays opted out even under `AllowTelemetry=3`. Keep the terms combined with `&&`; never add a policy value or config path that grants collection on its own.
 - **MXC owns its own consent state.** It must never read or infer from the Windows system telemetry consent. The consent store is a per-user JSON file; the policy is `HKLM\SOFTWARE\Policies\Mxc` → `AllowTelemetry` (`REG_DWORD`).
+- **Consent is a separate control plane.** Executors expose `--telemetry-consent <status|request|withdraw>`; never route consent maintenance through `--config` / `--config-base64`, add it to `MxcConfig`, or publish a consent JSON schema. `telemetry.enabled` remains only the per-run opt-in.
```
jsidewhite: >never route consent maintenance through `--config` / `--config-base64`, add it to `MxcConfig`, or publish a consent JSON schema. [](http://example.com/codeflow?start=109&length=128)

nit: remove

docs/schema.md
```diff
@@ -210,9 +215,6 @@ cannot mix both formats in one request.
             "portMappings": [              // Host<->container port forwarding. TCP only -- the WSLC SDK runtime returns E_NOTIMPL for UDP, so the parser hard-rejects "udp" entries with a clear message.
                 { "windowsPort": 8080, "containerPort": 80, "protocol": "tcp" }
```
jsidewhite, line 216: why back into experimental?

docs/telemetry/telemetry-consent-design.md
```diff
@@ -96,6 +96,44 @@ If the host never calls the request API, telemetry remains off. A presenter
...
+`lxc-exec` and `mxc-exec-mac` expose the same command and return
+`notApplicable` domain outcomes because telemetry collection is Windows-only.
+`request` optionally accepts `--telemetry-consent-locale <BCP-47>`; missing or
+unsupported locales use the canonical `en-US` fallback. Interactive requests
+write the prompt to stderr, read `Y` or `N` from a terminal, and write one final
+JSON outcome to stdout. A request without terminal stdin and stderr fails
+without changing consent.
+
+The Node SDK uses the hidden
+`--telemetry-consent-protocol stdio-v1` request mode because it owns the host
```
jsidewhite: Let's not do this - we have a meeting yesterday and agreed that the SDK and Runtimes do not need to support independent versioning.  Let's remove this flag.

docs/telemetry/telemetry-consent-design.md
```diff
@@ -96,6 +96,44 @@ If the host never calls the request API, telemetry remains off. A presenter
...
+
+Domain outcomes exit `0`. Invalid command/protocol input exits `64`.
+Operational failures such as a missing terminal, required-decision EOF, broken
+pipe, presenter failure, or persistence failure exit `1`. `status`,
+`withdraw`, and short-circuit request outcomes never invoke a presenter.
+
+Consent maintenance objects are deliberately not accepted through `--config`
+or `--config-base64`, are not part of `MxcConfig`, and have no published JSON
+Schema. The stable `telemetry.enabled` execution setting remains only the
+per-run opt-in and cannot request, grant, or withdraw consent.
```
jsidewhite: remove

docs/telemetry/telemetry.md
```diff
@@ -185,6 +182,18 @@ Telemetry emission is gated by the per-run request, MXC-owned consent,
...
+
+```text
+wxc-exec --telemetry-consent status
+wxc-exec --telemetry-consent request
+wxc-exec --telemetry-consent withdraw
+```
+
+These actions are not MXC execution configurations and are never accepted
+through `--config` or `--config-base64`. The stable `telemetry.enabled`
+setting remains only the additional per-run opt-in.
```
jsidewhite: remove

jsidewhite:
> // Copyright (c) Microsoft Corporation.

Remove all changes to this file?

---
Refers to: scripts/versioning/check-schema-codegen.js:2 in 4169b06. [](commit_id = 4169b06f5d13daef4e94afc0256d6b675aca4366, deletion_comment = False)

docs/schema.md
RamonArjona4: This should result in, which puts telemetry out of experimental but leaves wslc _in_ experimental, which I think is how it should be based on main?

{
  "telemetry": {
    "enabled": true
  },

  "experimental": {
    "wslc": {
      "portMappings": [
        { "windowsPort": 8080, "containerPort": 80, "protocol": "tcp" }
      ]
    }
  }
}

sdk/node/README.md
```diff
@@ -64,7 +64,7 @@ child.on('close', (code) => console.log('exit:', code));
 
 Pick `0.8.0-alpha` for new code on any supported platform.
 
-> **Stable schemas document only the non-experimental surface.** Experimental backends (`windows_sandbox`, `wslc`, `microvm`, `hyperlight`, `isolation_session`), the `experimental.*` block, and state-aware lifecycle live in `0.9.0-dev`. The parser still accepts them when paired with `--experimental` regardless of which schema your config validates against — schema choice affects editor validation, not runtime behavior.
+> **Stable schemas document only the non-experimental surface.** Experimental backends (`windows_sandbox`, `wslc`, `microvm`, `hyperlight`, `isolation_session`), the `experimental.*` block, and state-aware lifecycle live in `0.9.0-dev`. The parser still accepts experimental backend settings when paired with `--experimental`; however, fields introduced in a specific contract, including top-level `telemetry` in `0.9.0-alpha`, are rejected when the request explicitly declares an older schema version.
```
jsidewhite: >s experimental backend settings when paired with `--experimental`; however, fields introduced in a specific contract, including top-level `telemetry` in `0.9.0-alpha`, are rejected when the request explicitly declares an older schema version. [](http://example.com/codeflow?start=260&length=242)

let's revert this.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Will do.

sdk/node/src/state-aware-helper.ts
```diff
@@ -84,23 +86,33 @@ export interface BuildEnvelopeArgs {
...
+  const hasTelemetry = backendSpecific.telemetry !== undefined;
+  const version = suppliedVersion || (hasTelemetry ? TELEMETRY_STATE_AWARE_VERSION : defaultVersion);
+  if (hasTelemetry && suppliedVersion) {
+    const parsed = semverParse(suppliedVersion);
+    if (parsed && parsed.major === 0 && parsed.minor < 9) {
+      throw mxcErrorFromCode(
+        'malformed_request',
+        `telemetry requires schema version ${TELEMETRY_STATE_AWARE_VERSION} or later; got ${suppliedVersion}`,
+      );
+    }
```
jsidewhite, line 114: why is this logic duplicated into all the SDKs?

shouldn't it be in common tight-waist of the executor? (i.e. just before the backend is invoked)

src/core/wxc_common/src/config_parser.rs
```diff
@@ -77,6 +77,89 @@ where
...
+                    RequestKind::OneShot
+                },
+                experimental: discriminator
+                    .experimental
+                    .map(|raw| raw.get().to_string().into_boxed_str()),
+                experimental_span,
+            })
+        })
+}
+
```
jsidewhite: Why is this here - nothing to do with telemetry. Let's remove

jsidewhite:
> // Copyright (c) Microsoft Corporation.

let's revert this file. it's a lot of code to just produce an error message. 

---
Refers to: src/core/lxc/src/main.rs:1 in 4169b06. [](commit_id = 4169b06f5d13daef4e94afc0256d6b675aca4366, deletion_comment = False)

jsidewhite:
> // Copyright (c) Microsoft Corporation.

let's revert this file. it's a lot of code to just produce an error message. 

---
Refers to: src/core/mxc_darwin/src/main.rs:1 in 4169b06. [](commit_id = 4169b06f5d13daef4e94afc0256d6b675aca4366, deletion_comment = False)

src/core/wxc/src/main.rs
```diff
@@ -709,7 +743,27 @@ fn install_dacl_ctrl_handler() {
...
+    if handle_telemetry_consent_flags(&cli) {
+        return;
+    }
+    // Decode the request source (file path / base64) once, up front.
+    let decoded_config: Option<Result<String, wxc_common::error::WxcError>> =
+        decode_config_input_once(&cli);
+    let request_hint = decoded_config
+        .as_ref()
+        .and_then(|result| result.as_ref().ok())
+        .and_then(|json| wxc_common::config_parser::parse_request_hint_from_json(json).ok());
```
jsidewhite: Remove right? this is left over from overloading --config arg

src/core/wxc/src/main.rs
```diff
@@ -709,7 +743,27 @@ fn install_dacl_ctrl_handler() {
...
+
+    // --telemetry-consent: administer the persisted consent state and exit.
+    // Runs BEFORE `force_reclaim` env propagation and
+    // `recover_orphaned_state()` below: this is a read-only/local-file fast path with a 5-second
+    // client-side timeout (Node's consent getter), so it must not be gated
+    // behind unrelated recovery work that scans state files and may restore
+    // host DACLs — that could both time out the query (suppressing the
+    // prompt) and let a plain status read unexpectedly mutate filesystem
+    // ACLs. Matches the Linux/macOS executors, where this fast path also
+    // runs unconditionally immediately after CLI parsing.
```
jsidewhite: remove this

src/core/wxc/src/main.rs
```diff
@@ -709,7 +743,27 @@ fn install_dacl_ctrl_handler() {
...
+    if handle_telemetry_consent_flags(&cli) {
+        return;
+    }
```
jsidewhite, line 953: Don't produce side-effects here.  Pull the exit() call into this main function.

so put something like,

```
 if (cli.telemetry_consent)
{
   // Caller is performing a consent update, not making a sandbox.
// Record their new "consent" value and exit.
    let outcome = telemetry::consent_cli::handle_consent_command(
        action,
        cli.telemetry_consent_locale.as_deref(),
        cli.telemetry_consent_protocol,
    );
    let code = outcome.emit();
    if code != 0 {
        std::process::exit(code);
    }
}
```

sdk/node/src/state-aware-helper.ts
RamonArjona4: Agreed that this policy belongs at the native tight waist. The refactor is substantial and nonblocking for this stack, so I filed #1085 to centralize it: https://github.com/microsoft/mxc/issues/1085

src/core/wxc_common/src/config_contract_adapters/v0_6.rs
```diff
@@ -227,7 +227,7 @@ pub(crate) fn into_wire(request: contract::Request) -> wire::MxcConfig {
         version: Some(convert_version(version).to_owned()),
         phase: None,
         sandbox_id: None,
-        correlation_vector: None,
+        telemetry: None,
```
jsidewhite: >elemetry: None, [](http://example.com/codeflow?start=9&length=15)

this is a frozen contract right?  can't change it.

src/core/wxc_common/src/config_contract_adapters/v0_7.rs
```diff
@@ -257,7 +257,7 @@ pub(crate) fn into_wire(request: contract::Request) -> wire::MxcConfig {
         version: Some(convert_version(version).to_owned()),
         phase: None,
         sandbox_id: None,
-        correlation_vector: None,
+        telemetry: None,
```
jsidewhite: > telemetry: None, [](http://example.com/codeflow?start=7&length=17)

this is a frozen contract right?  can't change it.

src/core/wxc_common/src/config_contract_adapters/v0_8.rs
```diff
@@ -404,6 +403,7 @@ pub(crate) fn into_wire(request: contract::Request) -> wire::MxcConfig {
         ui: ui.into_option().map(convert_ui),
         seatbelt: seatbelt.into_option().map(convert_seatbelt),
         runtime_config: runtime_config.into_option().map(convert_runtime_config),
+        telemetry: None,
```
jsidewhite: >elemetry: None, [](http://example.com/codeflow?start=9&length=15)

this is a frozen contract right?  can't change it.

src/core/wxc_common/src/logger.rs
```diff
@@ -269,6 +269,16 @@ impl Logger {
...
     }
 
+    /// Record a warning for library callers and the primary diagnostic stream.
+    ///
+    /// Use this when executor users must see the message in the ordinary
+    /// buffered/console output while in-process callers also need it through
+    /// [`warnings`](Self::warnings).
+    pub fn retained_warning_line(&mut self, msg: &str) {
+        self.warnings.push(msg.to_string());
+        self.log_line(msg);
```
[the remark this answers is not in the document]
wezell: agent.mjs COPY added; the run.sh syntax error you flagged alongside it is also fixed

src/core/wxc_common/src/telemetry/consent_cli.rs
```diff
@@ -0,0 +1,976 @@
...
+        .flush()
+        .map_err(|error| format!("failed to display telemetry consent prompt: {error}"))?;
+
+    let mut input = String::new();
+    let read = std::io::stdin()
+        .read_line(&mut input)
+        .map_err(|error| format!("failed to read telemetry consent response: {error}"))?;
+    if read == 0 {
+        return Ok(consent::ConsentDecision::Dismissed);
+    }
```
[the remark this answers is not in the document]
RamonArjona4: Fixed: terminal EOF now follows the presenter-failure path and cannot report a successful dismissal.

README.md
RamonArjona4: Removed.

.github/copilot-instructions.md
RamonArjona4: Removed.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Removed the independent protocol-version flag. Node now uses the dedicated consent command without transport versioning.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Removed.

docs/telemetry/telemetry.md
RamonArjona4: Removed.

sdk/node/README.md
RamonArjona4: Reverted this wording.

src/core/wxc_common/src/config_parser.rs
RamonArjona4: Removed the unrelated parser change.

src/core/wxc/src/main.rs
RamonArjona4: Removed the leftover config-based consent path.

src/core/wxc/src/main.rs
RamonArjona4: Removed.

src/core/wxc/src/main.rs
RamonArjona4: Fixed: consent handling returns an outcome and each executor main owns its process exit.

src/core/wxc/src/main.rs
```diff
@@ -335,57 +356,49 @@ fn log_state_aware_dispatch_error(logger: &mut Logger, error: &MxcError) {
...
-    // is active so an inactive provider does no work and provision output is
-    // unchanged.
-    let correlation =
-        compute_phase_correlation(telemetry_active, is_provision, incoming_cv.as_deref());
+    let telemetry_active = parsed
+        .request
+        .telemetry
+        .as_ref()
+        .map(|c| telemetry::init(c, logger))
+        .unwrap_or(false);
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 8a077c27: state-aware warnings are retained and emitted on stderr, leaving stdout as one JSON envelope.

src/core/mxc_engine/src/lib.rs
```diff
@@ -71,31 +73,406 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
+            } => {
+                let outcome = match result {
+                    Ok(exit_code) => Ok(
+                        wxc_common::state_aware_dispatch::DispatchOutcome::ExecCompleted {
+                            exit_code: *exit_code,
+                        },
+                    ),
+                    Err(error) => Err(wxc_common::mxc_error::MxcError::backend_error(
+                        error.to_string(),
+                    )),
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 8a077c27: state-aware streaming preserves timeout classification.

src/core/wxc_common/src/telemetry/mod.rs
```diff
@@ -663,13 +858,13 @@ fn plan_state_aware<'a>(
...
+                    sandbox_kind,
                     exit_code: *exit_code,
                     outcome: if failed { "failure" } else { "success" },
                     duration_ms,
                     // A non-zero guest exit is a faithfully propagated sandbox
                     // exit code, not an MXC infrastructure error — leave the
-                    // reason unset and emit no MXC.Error (mirrors one-shot
+                    // reason unset and emit no Error (mirrors one-shot
                     // emit_completion).
                     failure_reason: None,
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 8a077c27: nonzero workload exits record process_error without emitting an infrastructure error.

src/core/wxc_common/src/logger.rs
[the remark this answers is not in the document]
RamonArjona4: Fixed in 8a077c27: retained warnings no longer write through the stdout logger path.

docs/telemetry/telemetry-consent-design.md
```diff
@@ -96,6 +96,38 @@ If the host never calls the request API, telemetry remains off. A presenter
...
+decision line containing the emitted challenge, prompt resource version, and
+`yes`, `no`, or `dismissed`. Input is limited to 64 KiB; malformed input,
+unknown fields, extra lines, challenge mismatch, and resource-version mismatch
+are rejected without persistence.
+
+Domain outcomes exit `0`. Invalid command/protocol input exits `64`.
+Operational failures such as a missing terminal, required-decision EOF, broken
+pipe, presenter failure, or persistence failure exit `1`. `status`,
+`withdraw`, and short-circuit request outcomes never invoke a presenter.
+
```
jsidewhite: remove

docs/telemetry/telemetry-consent-design.md
```diff
@@ -96,6 +96,38 @@ If the host never calls the request API, telemetry remains off. A presenter
...
+execution configuration:
+
+```text
+wxc-exec --telemetry-consent status
+wxc-exec --telemetry-consent withdraw
+wxc-exec --telemetry-consent request
+```
+
+`lxc-exec` and `mxc-exec-mac` expose the same command and return
+`notApplicable` domain outcomes because telemetry collection is Windows-only.
```
jsidewhite: Remove

jsidewhite:
> // Copyright (c) Microsoft Corporation.

revert please

---
In reply to: [5515676524](https://github.com/microsoft/mxc/pull/820#issuecomment-5515676524) [](http://example.com/codeflow?ancestors=5515676524)

---
Refers to: src/core/lxc/src/main.rs:1 in 4169b06. [](commit_id = 4169b06f5d13daef4e94afc0256d6b675aca4366, deletion_comment = False)

jsidewhite:
> // Copyright (c) Microsoft Corporation.

rvert

---
In reply to: [5515680362](https://github.com/microsoft/mxc/pull/820#issuecomment-5515680362) [](http://example.com/codeflow?ancestors=5515680362)

---
Refers to: src/core/mxc_darwin/src/main.rs:1 in 4169b06. [](commit_id = 4169b06f5d13daef4e94afc0256d6b675aca4366, deletion_comment = False)

src/core/wxc_common/src/config_parser.rs
```diff
@@ -6492,36 +6721,169 @@ mod tests {
...
+        });
+        let mut logger = test_logger();
+        let error = load_request_from_value(config, &mut logger, false).unwrap_err();
+        assert!(
+            error
+                .to_string()
+                .contains("'experimental.telemetry' has moved to the stable section"),
+            "got {error:?}"
+        );
+    }
```
jsidewhite, line 6887: wait, is this right?  we just nuke it from experimental from all past versions

jsidewhite: nvm!

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Removed in c4aa355a, including the non-Windows executor detail.

src/core/wxc_common/src/config_parser.rs
RamonArjona4: Acknowledged; the experimental compatibility handling remains intentional.

src/core/mxc_engine/src/state_aware.rs
```diff
@@ -363,14 +510,65 @@ pub fn exec_state_aware_json(
...
+        .request
+        .telemetry
+        .as_ref()
+        .and_then(|config| config.requested_sandbox_kind);
+    let telemetry_active = parsed
+        .request
+        .telemetry
+        .as_ref()
+        .map(|config| telemetry::init(config, &mut logger))
+        .unwrap_or(false);
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 55a45d18. The state-aware streaming path now uses the same RAII transfer guard across the pre-wrapper interval.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -1435,20 +1440,20 @@ pub mod test_support {
         Some(path)
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 55a45d18 and 416f6916. Inherited consent-store overrides now require the owner PID to match the direct parent, with regression coverage and corrected smoke/E2E launchers.

src/core/mxc_engine/src/lib.rs
```diff
@@ -71,31 +73,418 @@ use wxc_common::sandbox_process::{SandboxProcess, StreamCloser};
...
-    let mut warnings = process.warnings().to_vec();
-    for warning in logger.take_warnings() {
-        if !warnings.contains(&warning) {
-            warnings.push(warning);
+    let telemetry_active = request
+        .inner
+        .telemetry
+        .as_ref()
+        .map(|config| telemetry::init(config, &mut logger))
+        .unwrap_or(false);
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 55a45d18. An RAII registration guard now owns the provider reference from init until the one-shot streaming wrapper takes ownership, including unwind paths.

src/core/wxc/src/main.rs
RamonArjona4: Addressed. Consent is a dedicated --telemetry-consent command and is not accepted through --config or --config-base64.

docs/telemetry/telemetry-consent-design.md
RamonArjona4: Removed in c4aa355a. The executor consent command section is no longer in this design document.

RamonArjona4:
Addressed [macOS file revert request](https://github.com/microsoft/mxc/pull/820#issuecomment-5515680362): c4aa355a removes the consent CLI and parsing refactors; only the stable telemetry promotion wiring remains.

RamonArjona4:
Confirmed the repeated [LXC revert request](https://github.com/microsoft/mxc/pull/820#issuecomment-5517370571) is addressed by c4aa355a.

RamonArjona4:
Addressed [schema-codegen removal request](https://github.com/microsoft/mxc/pull/820#issuecomment-5515358299): PR #820 has no changes to `scripts/versioning/check-schema-codegen.js`.

RamonArjona4:
Addressed [LXC file revert request](https://github.com/microsoft/mxc/pull/820#issuecomment-5515676524): c4aa355a removes the consent CLI and parsing refactors; only the stable telemetry promotion wiring remains.

RamonArjona4:
Confirmed the repeated [macOS revert request](https://github.com/microsoft/mxc/pull/820#issuecomment-5517381062) is addressed by c4aa355a.

jsidewhite (dismissed):

src/core/mxc_engine/src/lib.rs
MGudgin: Thanks—the original `spawn` location is now fixed: it constructs `TelemetryRegistration` immediately after `telemetry::init`, so failure or unwind before the process wrapper is returned releases the provider automatically.

The same lifetime gap remains in two state-aware entry points: `exec_state_aware_attached_with` and `run_state_aware_json`. Both call `telemetry::init`, but neither creates a `TelemetryRegistration`; they rely on eventually reaching `emit_sdk_state_aware_with_kind` to call `shutdown`. If dispatch or correlation-state processing panics first, the FFI catches that panic and returns while the provider reference remains registered, so a later `FreeLibrary` can still leave ETW callbacks targeting unloaded code.

Please apply the existing guard pattern to both paths: construct `TelemetryRegistration` immediately after initialization, then transfer it into the terminal emission call. This should ensure every return and unwind path either transfers or drops the provider reference exactly once.

src/core/mxc_engine/src/state_aware.rs
MGudgin: This test still invokes `run_state_aware_json` without a `telemetry` section. As a result, `telemetry_active` is false, the provider is never initialized, no event is emitted, and there is no provider reference to release. Repeating the expected `BackendUnavailable` result therefore does not exercise the behavior named by the test.

Please drive this through a telemetry-enabled request under deterministic test consent/policy state and add a test seam that can observe provider acquisition, emission, and release. At minimum, assert that an error-path call acquires exactly one reference, emits the expected phase/failure classification, and returns the provider count to zero before a second call. Representative provision/exec coverage should also verify the shared correlation base and final release.

bbonaby (approved):

Pjrich1313 (commented):
Load all diff put copilot tools together merge and run

PR #821, opened 2026-08-12
Expose telemetry through Rust SDK, C ABI, and .NET binding

RamonArjona4:

-----

## Summary

Exposes stable telemetry consent and policy through the public Rust SDK, panic-safe C ABI, and managed .NET binding. Adds typed presenter/status/withdrawal APIs, native binding entry points, streaming and state-aware propagation, and binding-codegen coverage.

This is PR 4 of 5 and depends on the stable runtime integration in PR 3.

## Stack

| Order | Change | PR |
|---|---|---|
| **4** | **Rust SDK, C ABI, and .NET binding** | **#821** |
| 5 | Node.js and .NET SDK refinements | #822 |

Review only this PR's diff; prerequisite behavior is in the PR above.


src/core/mxc-sdk/README.md
```diff
@@ -304,6 +308,85 @@ stdin (`Sandbox::take_stdin()` returns `None`), and its process has no host
...
+
+let outcome = telemetry::request_consent(Some("en-US"), |prompt| {
+    // Render title, body, labels, and privacy link verbatim in host UI.
+    assert_eq!(prompt.resource_version, 1);
+    Ok(telemetry::ConsentDecision::Yes)
+})?;
+
+let status = telemetry::get_consent_status();
+let withdrawal = telemetry::withdraw_consent()?;
+# let _ = (outcome, status, withdrawal);
```
jsidewhite: ># [](http://example.com/codeflow?start=0&length=1)

what does the # do here?

src/core/mxc-sdk/README.md
```diff
@@ -304,6 +308,85 @@ stdin (`Sandbox::take_stdin()` returns `None`), and its process has no host
...
+`ConsentActionResult::NotApplicable` — MXC neither collects nor offers consent
+for telemetry there, so a host can call these unconditionally without
+special-casing the platform.
+
+`ConsentState` and `PolicyState` are SDK-owned types, so the public API never
+leaks the internal `wxc_common` foundation crate. The *decision logic* behind
+them is not duplicated: every function here delegates to
+`wxc_common::telemetry`, the same code the `wxc-exec` CLI flags, the C# SDK
+(via `mxc_ffi`), and the Node SDK all resolve consent through. There is
+deliberately no Rust-SDK-specific consent logic to drift.
```
jsidewhite: This file is supposed to be a welcome to *consumers* that want to use the SDK - not a note to *developers* of the SDK.

Amusingly, this comment is touting the public/private separations, but then immediately explaining Rust impl details.

src/core/mxc-sdk/src/telemetry.rs
```diff
@@ -0,0 +1,568 @@
...
+//!     ConsentState::NotApplicable => {}
+//! }
+//! # let _ = outcome;
+//! # Ok::<(), telemetry::ConsentError>(())
+//! ```
+//!
+//! Off Windows, [`get_consent`] always returns
+//! [`ConsentState::NotApplicable`] without touching disk,
+//! [`needs_consent_prompt`] is always `false`, and consent requests return
+//! [`ConsentActionResult::NotApplicable`].
```
jsidewhite: I feel like I've read this language before

src/core/mxc-sdk/src/telemetry.rs
```diff
@@ -0,0 +1,568 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConsentStatusReason {
+    NoRecord,
+    StoreUnreadable,
+    StoreMalformed,
+    ConsentSchemaUnsupported,
+    PromptVersionMissing,
+    PromptVersionUnsupported,
+    NotApplicable,
+}
```
jsidewhite, line 158: Does these types need to be duplicated from wire.rs?

src/ffi/mxc_ffi/src/streaming.rs
```diff
@@ -201,16 +201,15 @@ fn spawn_inner(
...
-            format!("failed to parse policy JSON: {e}"),
-        )
-    })?;
+    let (policy, telemetry_enabled) =
+        parse_policy_json(policy_json).map_err(|error| (MXC_STATUS_MALFORMED_REQUEST, error))?;
 
     let mut request =
         build_request(&policy, None).map_err(|e| (status_from_error_code(e.code), e.message))?;
     request.set_script(command);
+    if let Some(enabled) = telemetry_enabled {
```
jsidewhite: >telemetry_enabled [](http://example.com/codeflow?start=27&length=17)

should this read something more like 'telemetry_requested'?

jsidewhite (dismissed):
:shipit:

src/core/mxc-sdk/src/telemetry.rs
RamonArjona4: This is intentional to keep from exposing the internals of the of wire.rs in the API.  We could instead re-export them from wxc_common if that's preferable?

src/core/mxc-sdk/src/telemetry.rs
```diff
@@ -0,0 +1,512 @@
...
+
+impl ConsentActionResult {
+    /// Stable wire string used by the maintenance and binding contracts.
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Granted => "granted",
+            Self::Denied => "denied",
+            Self::Dismissed => "dismissed",
+            Self::Withdrawn => "withdrawn",
+            Self::AlreadyGranted => "already-granted",
```
MGudgin: **Medium (compatibility/maintainability) — Use the canonical maintenance result strings.**

This facade emits kebab-case (`already-granted`, `policy-blocked`, `not-applicable`) and the FFI serializes these values, while the canonical `TelemetryConsentResult` maintenance contract and generated Node types use camelCase (`alreadyGranted`, `policyBlocked`, `notApplicable`). The method comment says these are the stable maintenance and binding strings, but there are now two contracts.

**Fix:** Define the values once in the shared model and delegate here, preferably using the canonical camelCase representation; otherwise explicitly version and document the FFI contract as distinct.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -373,6 +432,282 @@ pub extern "C" fn mxc_version() -> *const c_char {
...
+}
+
+/// Request telemetry consent through a host presenter callback.
+///
+/// # Safety
+/// `locale_utf8` must be null or valid NUL-terminated UTF-8. `presenter` must
+/// be a valid callback when non-null. `context` is passed through untouched.
+/// `out_utf8` must point to writable pointer-sized storage.
+#[no_mangle]
+pub unsafe extern "C" fn mxc_telemetry_request_consent(
```
MGudgin, line 641: **Medium (reliability) — Clear JSON out-pointers before any failure path.**

`mxc_telemetry_request_consent`, `mxc_telemetry_withdraw_consent`, and `mxc_telemetry_get_consent_status` validate `out_utf8` but leave its prior value untouched when locale validation, presentation, persistence, serialization, or panic handling fails. Existing tests initialize the pointer to null, so they do not catch a stale non-null caller value being freed after an error.

**Fix:** Immediately write `null_mut()` after validating each non-null out-pointer, matching the defensive convention used by the streaming/state-aware exports, and add a sentinel-pointer regression test.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -56,13 +48,51 @@ mod streaming;
...
+/// Return code from an FFI telemetry-consent presenter callback.
+pub const MXC_TELEMETRY_CONSENT_DECISION_DISMISSED: i32 = 2;
+/// Return code indicating that the host presenter failed.
+pub const MXC_TELEMETRY_CONSENT_PRESENTER_ERROR: i32 = -1;
+
+/// Host callback invoked synchronously with the canonical consent prompt JSON.
+///
+/// The JSON pointer remains valid only for the duration of the callback. The
+/// callback must return one of the `MXC_TELEMETRY_CONSENT_DECISION_*` values.
+pub type MxcTelemetryConsentPresenter =
```
MGudgin, line 89: **Medium (reliability) — State that presenter callbacks must not unwind across the C ABI.**

Rust's `catch_unwind` catches Rust panics only. A C++ or managed exception escaping this callback into Rust is undefined behavior. The callback docs currently specify pointer lifetime and return values but omit this load-bearing precondition.

**Fix:** Document that callbacks must catch all foreign exceptions and translate them to `MXC_TELEMETRY_CONSENT_PRESENTER_ERROR`; provide binding trampolines/examples that enforce it.

src/core/mxc-sdk/src/telemetry.rs
```diff
@@ -0,0 +1,512 @@
...
+{
+    inner_consent::request_consent(locale, |prompt| {
+        presenter(&ConsentPrompt::from(prompt)).map(ConsentDecision::to_inner)
+    })
+    .map(Into::into)
+    .map_err(Into::into)
+}
+
+/// Asynchronous counterpart to [`request_consent`].
+pub async fn request_consent_async<F, Fut>(
```
MGudgin, line 376: **Medium (performance/API) — This async facade still performs blocking storage operations on the caller's executor thread.**

The inner async function performs synchronous registry/file preflight before awaiting the presenter, then synchronous persistence/status reads afterward, including sleep-based retry paths. Exposing it publicly as `async` can stall unrelated tasks on a single-threaded or latency-sensitive runtime.

**Fix:** Offload the storage phases to a blocking executor or provide genuinely asynchronous storage. If runtime neutrality prevents that, document explicitly that only the presenter is asynchronous and callers must isolate the operation from their executor thread.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -373,6 +432,282 @@ pub extern "C" fn mxc_version() -> *const c_char {
...
+        }
+    }
+}
+
+/// Return the typed persisted/effective consent and policy snapshot as JSON.
+///
+/// # Safety
+/// `out_utf8` must point to writable pointer-sized storage.
+#[no_mangle]
+pub unsafe extern "C" fn mxc_telemetry_get_consent_status(out_utf8: *mut *mut c_char) -> i32 {
```
MGudgin, line 760: **Medium (test coverage) — Complete the native telemetry-export failure and release-mode matrix.**

The new commits add good withdrawal success/failure coverage, but `mxc_telemetry_get_consent_status` still has no direct test; invalid UTF-8 locale, presenter-error/invalid-decision mapping, and stale out-pointer behavior are also uncovered. In addition, the consent tests are `debug_assertions`-gated: the release FFI run executes 26 tests versus 35 in debug, so CI-like release testing skips these paths.

**Fix:** Add direct success/null/error/ownership tests for every telemetry export, including sentinel out-pointers and presenter failures, and run the consent suite in a CI configuration where the isolated test-store hooks are available.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -373,6 +430,279 @@ pub extern "C" fn mxc_version() -> *const c_char {
...
+    match catch_unwind(mxc_sdk::telemetry::withdraw_consent) {
+        Ok(Ok(outcome)) => {
+            // SAFETY: validated above.
+            unsafe { write_json_out(consent_outcome_json(outcome), out_utf8) }
+        }
+        Ok(Err(error)) => {
+            report_to_stderr(format_args!(
+                "mxc: failed to withdraw telemetry consent: {error}"
+            ));
+            MXC_STATUS_CONSENT_WRITE_FAILED
```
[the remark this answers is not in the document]
RamonArjona4: Addressed: the presenter callback is invoked behind the FFI panic/unwind boundary, and the ownership and foreign-callback contract is documented. Resolving as fixed.

src/ffi/mxc_ffi/src/lib.rs
RamonArjona4: Addressed with FFI coverage for consent withdrawal, status queries, policy-blocked behavior, persistence failures, and malformed/unknown outcomes. Resolving as fixed.

src/ffi/mxc_ffi/src/lib.rs
RamonArjona4: Addressed: every JSON-returning telemetry export clears its out pointer before work and has sentinel regression coverage for failure paths. Resolving as fixed.

src/core/mxc-sdk/src/telemetry.rs
RamonArjona4: Addressed: `mxc_telemetry_needs_consent_prompt` now follows the documented fail-closed query contract and has failure-path coverage. Resolving as fixed.

src/ffi/mxc_ffi/src/lib.rs
RamonArjona4: Addressed: JSON-returning consent exports consistently null the caller output before validation/allocation and tests verify the sentinel is cleared on error. Resolving as fixed.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
```diff
@@ -56,6 +56,33 @@ public sealed class SandboxPolicy
     /// <summary>Execution timeout in milliseconds (<c>null</c> = no timeout).</summary>
     [JsonPropertyName("timeoutMs")]
     public uint? TimeoutMs { get; set; }
+
+    /// <summary>
+    /// Canonical per-invocation telemetry settings serialized as
+    /// <c>{"telemetry":{"enabled":...}}</c>.
+    /// </summary>
+    [JsonPropertyName("telemetry")]
+    public TelemetrySettings? Telemetry { get; set; }
```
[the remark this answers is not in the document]
RamonArjona4: Fixed: the compatibility clone now preserves canonical telemetry, with combined-policy coverage.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -249,6 +355,98 @@ pub(crate) unsafe fn cstr_to_str<'a>(p: *const c_char) -> Option<&'a str> {
...
+        .map_err(|error| format!("failed to parse policy JSON: {error}"))?;
+    let telemetry_enabled = telemetry_enabled_from_policy(&envelope)?;
+    Ok((envelope.policy, telemetry_enabled))
+}
+
+pub(crate) fn build_run_request(
+    policy_json: &str,
+    command: &str,
+) -> Result<SandboxRequest, (i32, MxcErrorDetail)> {
+    build_ffi_request(policy_json, command)
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: complete run/spawn requests now parse and apply canonical telemetry, with propagation tests.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,476 @@
...
+{
+    public TelemetryConsentResult Result { get; init; }
+    public TelemetryConsentState StoredState { get; init; }
+    public TelemetryConsentState EffectiveState { get; init; }
+    public TelemetryConsentStatusReason? Reason { get; init; }
+    public TelemetryPolicyState Policy { get; init; }
+}
+
+/// <summary>Telemetry consent helpers over the native <c>mxc_ffi</c> surface.</summary>
+public static class MxcTelemetry
```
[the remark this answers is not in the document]
RamonArjona4: Updated the PR title, summary, and stack table to state the managed .NET binding scope explicitly.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,476 @@
...
+    private sealed class PresenterContext
+    {
+        public required Func<TelemetryConsentPrompt, TelemetryConsentDecision> Presenter { get; init; }
+    }
+
+    /// <summary>
+    /// Return the user's recorded consent state.
+    /// Fail-closed native-load failures return <see cref="TelemetryConsentState.Undetermined"/>.
+    /// </summary>
+    public static TelemetryConsentState GetConsent()
```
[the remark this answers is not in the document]
RamonArjona4: Updated the .NET README with the actual presenter, query, status, and withdrawal APIs.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -249,6 +355,140 @@ pub(crate) unsafe fn cstr_to_str<'a>(p: *const c_char) -> Option<&'a str> {
...
+
+#[derive(serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct FfiPolicyEnvelope {
+    #[serde(flatten)]
+    policy: SandboxPolicy,
+    #[serde(default)]
+    telemetry: Option<FfiTelemetrySection>,
+    #[serde(default, deserialize_with = "deserialize_legacy_telemetry_enabled")]
+    telemetry_enabled: Option<bool>,
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: removed the unreleased telemetryEnabled wire alias; only telemetry.enabled is accepted.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcTelemetryTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        Converters = { new JsonStringEnumConverter(JsonNamingPolicy.CamelCase) },
+    };
+
+    [Fact]
+    public void SandboxPolicy_TelemetryEnabledSerializesCanonically()
+    {
+        var policy = new SandboxPolicy
+        {
+            Version = "0.8.0-alpha",
+            TelemetryEnabled = true,
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: the canonical telemetry policy test now uses the maximum supported 0.9 schema.

sdk/dotnet/Microsoft.Mxc.Sdk/Microsoft.Mxc.Sdk.csproj
```diff
@@ -34,6 +34,7 @@
     <MxcWithWslc Condition="'$(MxcWithWslc)' == ''">false</MxcWithWslc>
 
     <MxcCargoFeatures>dotnetsdk</MxcCargoFeatures>
+    <MxcCargoFeatures Condition="'$(Configuration)' == 'Debug'">$(MxcCargoFeatures) wxc_common/test-support</MxcCargoFeatures>
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: ordinary Debug SDK builds no longer enable test support; the test project builds an isolated test-support native library.

src/core/wxc_common/src/telemetry/consent.rs
```diff
@@ -173,10 +173,13 @@ impl ConsentState {
...
-    /// This is deliberately private and policy-blind. Host applications must
-    /// use [`needs_consent_prompt`], which also suppresses the prompt under an
-    /// administrative block.
-    fn needs_prompt(&self) -> bool {
+    /// Policy-blind: this only reflects the stored consent state and does not
+    /// consult administrative policy. Host applications should prefer
+    /// [`needs_consent_prompt`] (or [`ConsentStatus::needs_prompt`]), which
+    /// additionally suppresses the prompt under an administrative block; the
+    /// public SDK surface re-exposes this predicate for parity checks and
+    /// diagnostics.
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: the guidance now directs hosts only to the policy-aware needs_consent_prompt API.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -249,6 +355,140 @@ pub(crate) unsafe fn cstr_to_str<'a>(p: *const c_char) -> Option<&'a str> {
...
+}
+
+pub(crate) fn parse_policy_json(
+    policy_json: &str,
+) -> Result<(SandboxPolicy, Option<bool>), String> {
+    let value: serde_json::Value = serde_json::from_str(policy_json)
+        .map_err(|error| format!("failed to parse policy JSON: {error}"))?;
+    validate_canonical_telemetry_version(&value)?;
+    let envelope: FfiPolicyEnvelope = serde_json::from_value(value)
+        .map_err(|error| format!("failed to parse policy JSON: {error}"))?;
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: direct policy parsing is duplicate-aware and rejects duplicate version, telemetry, and enabled fields.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -249,6 +355,140 @@ pub(crate) unsafe fn cstr_to_str<'a>(p: *const c_char) -> Option<&'a str> {
     CStr::from_ptr(p).to_str().ok()
 }
 
+#[derive(serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+struct FfiTelemetrySection {
+    enabled: Option<bool>,
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: explicit null telemetry and enabled values are rejected while omission remains valid.

src/ffi/mxc_ffi/src/request.rs
```diff
@@ -15,10 +15,37 @@ use mxc_sdk::{
...
+
+impl<'de> serde::Deserialize<'de> for RequestPolicy {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = <Value as serde::Deserialize>::deserialize(deserializer)?;
+        let policy_json = serde_json::to_string(&value).map_err(serde::de::Error::custom)?;
+        let (policy, telemetry_enabled) =
+            crate::parse_policy_json(&policy_json).map_err(serde::de::Error::custom)?;
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 68262767: complete requests retain raw policy JSON through parsing, preserving duplicate-key rejection.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/Microsoft.Mxc.Sdk.Tests.csproj
```diff
@@ -12,12 +12,31 @@
...
 
   <ItemGroup>
     <PackageReference Include="xunit.v3" Version="4.0.0" />
   </ItemGroup>
 
+  <Target Name="BuildMxcTestSupportNativeLibrary"
+          AfterTargets="Build"
+          Condition="'$(DesignTimeBuild)' != 'true'">
+    <Message Importance="high" Text="mxc: building isolated test-support native library" />
+    <Exec Command="cargo build -p mxc_ffi --features test-support $(MxcTestCargoProfileArg) --target-dir &quot;$(MxcTestCargoTargetDir)&quot;" WorkingDirectory="$(MxcSrcDir)" />
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 47217f9d. The isolated test-support build now mirrors the enabled isolation_session and wslc features; the combined optional-backend build succeeds locally.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
```diff
@@ -56,6 +56,33 @@ public sealed class SandboxPolicy
...
+
+    /// <summary>
+    /// Convenience projection for <see cref="Telemetry"/>. Serializes through
+    /// the canonical nested <c>telemetry.enabled</c> field, not the legacy
+    /// top-level alias accepted by the native FFI for compatibility.
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 47217f9d. The documentation now states that the legacy top-level alias is rejected.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/Microsoft.Mxc.Sdk.Tests.csproj
```diff
@@ -12,12 +12,34 @@
...
     <PackageReference Include="xunit.v3" Version="4.0.0" />
   </ItemGroup>
 
+  <Target Name="BuildMxcTestSupportNativeLibrary"
+          AfterTargets="Build"
+          Condition="'$(DesignTimeBuild)' != 'true'">
+    <Message Importance="high" Text="mxc: building isolated test-support native library" />
+    <Exec Command="cargo build -p mxc_ffi --features &quot;$(MxcTestCargoFeatures)&quot; $(MxcTestCargoProfileArg) --target-dir &quot;$(MxcTestCargoTargetDir)&quot;" WorkingDirectory="$(MxcSrcDir)" />
+    <Error Condition="!Exists('$(MxcTestNativeOutput)')"
+           Text="The test-support native build did not produce $(MxcTestNativeOutput)." />
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 77ab5bbd. The isolated test-support build now derives the rustc host triple, passes it through `--target`, and reads the artifact from the target-specific directory. I validated this with `CARGO_BUILD_TARGET=aarch64-pc-windows-msvc` plus both optional Windows backend features; the build still selected and produced the x64 host library.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,476 @@
...
+    public static Task<TelemetryConsentOutcome> RequestConsentAsync(
+        Func<TelemetryConsentPrompt, Task<TelemetryConsentDecision>> presenter,
+        string? locale = null,
+        CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(presenter);
+        return Task.Run(
+            () => RequestConsent(
+                prompt => presenter(prompt).ConfigureAwait(false).GetAwaiter().GetResult(),
+                locale),
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 77ab5bbd with regression coverage. `RequestConsentAsync` now captures the caller synchronization context, keeps the blocking native call on the thread pool, and posts only the presenter back to the captured context. The test exercises an asynchronous continuation on a pumped caller context; the full .NET test executable passes.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -249,6 +353,140 @@ pub(crate) unsafe fn cstr_to_str<'a>(p: *const c_char) -> Option<&'a str> {
...
+
+fn telemetry_enabled_from_policy(policy: &FfiPolicyEnvelope) -> Option<bool> {
+    policy
+        .telemetry
+        .0
+        .as_ref()
+        .map(|telemetry| telemetry.enabled)
+}
+
+fn validate_canonical_telemetry_version(policy: &FfiPolicyEnvelope) -> Result<(), String> {
```
RamonArjona4: The FFI currently maintains a parallel policy envelope and telemetry-version validator instead of reusing the shared native parsing boundary. The follow-up refactor is tracked in #1090; it is intentionally separate from #1085, which covers state-aware Node/.NET version handling.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,516 @@
...
+        catch
+        {
+            return TelemetryConsentState.Undetermined;
+        }
+    }
+
+    /// <summary>
+    /// Read stored and effective consent, plus the current administrative policy.
+    /// </summary>
+    public static TelemetryConsentStatus GetConsentStatus()
```
RamonArjona4, line 307: `GetConsentStatus()` can still throw from this read-only path in standalone PR #821. The fail-closed behavior is implemented in downstream PR #822: failures return an `Undetermined` / `Blocked` sentinel and emit the deduplicated diagnostic instead of escaping to the caller.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -70,13 +70,102 @@ pub use error_detail::*;
...
+///
+/// **The callback must not unwind across the FFI boundary.** Only Rust panics
+/// are caught by [`std::panic::catch_unwind`] on the Rust side; a C++
+/// exception, a .NET/CLR exception, an Objective-C exception, a Go panic, or
+/// any other foreign unwind mechanism escaping this callback into Rust is
+/// undefined behavior. Binding authors wrapping this callback (C#, Node
+/// native addons, C++ hosts, and so on) must wrap the trampoline that invokes
+/// the caller's presenter in a `try` / `catch (...)` (or the runtime's
+/// equivalent) that catches every foreign exception and returns
+/// [`MXC_TELEMETRY_CONSENT_PRESENTER_ERROR`] instead of allowing the exception
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in c2c39c6d. The safety documentation now requires every callback or trampoline, including Rust implementations, to contain failures internally and return `MXC_TELEMETRY_CONSENT_PRESENTER_ERROR`; it no longer claims the outer `catch_unwind` can catch a callback panic.

sdk/dotnet/README.md
```diff
@@ -484,11 +484,16 @@ MXC telemetry is Windows-only and remains off until both of these are true:
...
 
-Telemetry remains off by default unless the caller opts in with `SandboxPolicy.TelemetryEnabled = true` (or the equivalent phase-level `TelemetryEnabled` setting for state-aware requests) and applicable Windows consent/policy gates permit collection.
-
-Any .NET consent surface should stay UI-agnostic, present the canonical
-resource verbatim through a host callback, persist only explicit yes/no
-decisions, treat dismissal and failures as non-grants, and follow the rules in
+Telemetry remains off by default unless the caller opts in with
+`SandboxPolicy.Telemetry = new TelemetrySettings { Enabled = true }` (or the
+equivalent phase-level `TelemetryEnabled` setting for state-aware requests)
+and applicable Windows consent/policy gates permit collection.
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in c2c39c6d. The README now documents only the one-shot `SandboxPolicy.Telemetry` API available in PR #821 and no longer claims that state-aware .NET telemetry options exist on this branch.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,516 @@
...
+
+    private sealed class PresenterContext
+    {
+        public required Func<TelemetryConsentPrompt, TelemetryConsentDecision> Presenter { get; init; }
+    }
+
+    /// <summary>
+    /// Return the user's recorded consent state.
+    /// Fail-closed native-load failures return <see cref="TelemetryConsentState.Undetermined"/>.
+    /// </summary>
```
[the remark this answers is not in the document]
RamonArjona4: Confirmed. Tracked in #1111, assigned to @RamonArjona4.

src/core/mxc-sdk/src/telemetry.rs
```diff
@@ -0,0 +1,523 @@
...
+            inner_policy::PolicyState::Blocked => Self::Blocked,
+            inner_policy::PolicyState::NotApplicable => Self::NotApplicable,
+        }
+    }
+}
+
+/// Return the user's recorded consent decision.
+pub fn get_consent() -> ConsentState {
+    inner_consent::get_consent().into()
+}
```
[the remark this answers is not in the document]
RamonArjona4: Confirmed. Tracked in #1112, assigned to @RamonArjona4.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -507,6 +735,333 @@ pub extern "C" fn mxc_version() -> *const c_char {
     .unwrap_or(c"".as_ptr())
 }
 
+// ---------------------------------------------------------------------------
+// Telemetry consent
+// ---------------------------------------------------------------------------
+
+/// Read the persisted telemetry consent state.
```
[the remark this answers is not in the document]
RamonArjona4: Confirmed. Tracked in #1113, assigned to @RamonArjona4.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,516 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(presenter);
+        var synchronizationContext = SynchronizationContext.Current;
+        return Task.Run(
+            () => RequestConsent(
+                prompt => InvokePresenterAsync(presenter, prompt, synchronizationContext)
+                    .GetAwaiter()
+                    .GetResult(),
+                locale),
+            cancellationToken);
```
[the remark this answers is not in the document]
RamonArjona4: Confirmed. Tracked in #1114, assigned to @RamonArjona4.

src/ffi/mxc_ffi/src/lib.rs
bbonaby: I was going to comment some concerns about this file  and `FfiPolicyEnvelope` in particular. Also about how we're handling  deserialization and validation in this file rather than in the mxc_engine.

Would these issues you filed help move the ffi crate back to its original purpose which is to just translate and pass data back and forth?

RamonArjona4: Yes—issue #1090, “Reuse shared native policy parsing in  mxc_ffi ”, is meant to address this. It calls for removing the parallel FfiPolicyEnvelope & validate_canonical_teleemtry_version logic & moving interprertation back to teh execution boundary.  We can also fix this now if the current state is too odious to defer.

MGudgin (dismissed):
Re-reviewed the current `b3b659a` diff over merged #820. The consent wire contract, FFI out-pointer handling, callback unwind contract, and asynchronous storage behavior are addressed, and no new merge-blocking findings were found. The remaining debug-only native FFI test coverage gap is non-blocking.

src/core/mxc-sdk/README.md
```diff
@@ -24,13 +25,13 @@ let policy = SandboxPolicy {
     timeout_ms: Some(10_000),
 };
 let mut request = build_request(&policy, None)?;
-request.set_script("echo hello");
+request.set_script("echo hello").set_telemetry_enabled(true);
```
jsidewhite: >enabled(true [](http://example.com/codeflow?start=47&length=12)

Is this a normal idiom (double positive)?  "set enabled to true"?

src/core/mxc_engine/src/binding.rs
```diff
@@ -0,0 +1,165 @@
...
+    };
+    if major == 0 && minor < 9 {
+        return Err(
+            "failed to parse policy JSON: top-level 'telemetry' requires config schema version \
+             0.9.0-alpha or later"
+                .to_string(),
+        );
+    }
+    Ok(())
+}
```
jsidewhite: Why are we parsing the schema version of the json here?  That cannot be right.

jsidewhite:
>     pub timeout_ms: Option<u32>,

Why doesn't this have the TelemetrySettings field like you added in sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs?

---
Refers to: src/core/mxc_engine/src/policy.rs:604 in ef7a78b. [](commit_id = ef7a78b54fb3c4b6573b63b3dcf10c370a82c61e, deletion_comment = False)

src/core/mxc_engine/src/binding.rs
```diff
@@ -0,0 +1,165 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
```
jsidewhite: I don't think this file should exist.

1. it introduces a hack type `ParsedPolicy`
1. it parses semver(??) to specifically bypass the config_parser.rs?

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -249,6 +343,43 @@ pub(crate) unsafe fn cstr_to_str<'a>(p: *const c_char) -> Option<&'a str> {
...
+) -> Result<SandboxRequest, (i32, MxcErrorDetail)> {
+    build_ffi_request(policy_json, command)
+}
+
+pub(crate) fn build_spawn_request(
+    policy_json: &str,
+    command: &str,
+) -> Result<SandboxRequest, (i32, MxcErrorDetail)> {
+    build_ffi_request(policy_json, command)
+}
```
jsidewhite: are these two functions identical?

jsidewhite (changes requested):
:clock1:

src/core/mxc-sdk/README.md
```diff
@@ -413,26 +426,104 @@ go through the same parser the executor uses — so a rejected value (e.g. a por
...
+canonical prompt supplied to its presenter callback verbatim. MXC persists a
+grant only from the typed decision returned by that callback. If the host
+never requests consent, telemetry remains off. See the normative
+[SDK presenter requirements](../../../docs/telemetry/telemetry-consent-design.md#sdk-presenter-requirements)
+for control mappings, dismissal behavior, learn-more handling, status, and
+withdrawal.
+
+Telemetry is also off per invocation unless the request explicitly enables it
+with `SandboxRequest::set_telemetry_enabled(true)`. This stable switch does not
+require `set_experimental(true)` and cannot bypass consent or administrative
```
jsidewhite: remove

src/core/mxc-sdk/README.md
RamonArjona4: I agree this is semantically awkward in a "Task failed successfully" sense. I will rename this set_telemetry_opt_in to be less awkward.

src/core/mxc_engine/src/binding.rs
RamonArjona4: Short answer is I went down a wrong path & it compounded. I'd filed this as 1090 to sort out the duplicated parsing, but rather than defer I'll fix it now so it doesn't compound further.

src/core/mxc-sdk/README.md
RamonArjona4: Renamed this API to `SandboxRequest::set_telemetry_opt_in(bool)` in 933f3202. The old name was a normal setter idiom, but it overstated the result because this switch only opts the invocation in; persisted user consent and administrative policy still independently gate emission.

src/core/mxc_engine/src/binding.rs
```diff
@@ -0,0 +1,165 @@
...
+//! Internal policy parsing shared by language-binding adapters.
+
+use crate::policy::SandboxPolicy;
+
+#[derive(serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+struct TelemetrySection {
+    enabled: bool,
```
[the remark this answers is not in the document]
MGudgin: **Medium (compatibility/correctness) — Keep the binding parser aligned with the canonical telemetry contract.**

**Attribution:** `introduced_by_change` — this PR introduces `mxc_engine::binding` as the policy entry point used by the C ABI.

Confirmed against the current head: `TelemetrySection.enabled` is a required `bool`, while the canonical `wire::Telemetry.enabled` is optional and the native parser explicitly accepts `"telemetry": {}` as telemetry being off. Because the binding entry points parse through this file first, the C ABI rejects schema-valid input that the native executors accept.

Please remove the duplicated telemetry/version parsing and route the request through the canonical parser. If this adapter must remain, at minimum make omitted `enabled` map to `None` and add a cross-surface conformance test proving the same policy is accepted by both native executors and FFI entry points.

src/core/mxc-sdk/README.md
```diff
@@ -1,460 +1,551 @@
...
-`mxc_engine` crate; this crate is a thin streaming facade that re-exports the
-curated engine surface and wraps the engine's streaming handle in [`Sandbox`].
-
-The `wxc-exec`, `lxc-exec`, and `mxc-exec-mac` binaries do not (yet) depend on
-this crate. The engine reuses the same backend crates they do; on Windows both
-the streaming and the run-to-completion paths share
-`appcontainer_common::dispatcher`'s tier selection (`select_backend_with_fallback`),
-so they agree on the BaseContainer / AppContainer + BFS / AppContainer + DACL
-tier and spawn the appropriate handle.
+# `mxc-sdk`
```
MGudgin: **Low (review integrity/documentation) — Normalize the README line endings.**

**Attribution:** `introduced_by_change` — the current head rewrites this file with malformed terminators.

The committed blob contains 550 `CR CR LF` line endings (1,101 carriage returns for 551 line feeds); the previously approved blob contained no carriage returns. This leaves an extra `^M` on every line and turns an ordinary documentation edit into an apparent whole-file rewrite.

Please rewrite the file with normal LF line endings while preserving its content, and verify the repository's `.gitattributes`/editor settings prevent this form from recurring.

MGudgin (commented):
Two findings from re-reviewing the current head.

src/core/mxc_engine/src/binding.rs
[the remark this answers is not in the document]
RamonArjona4: Addressed by the canonical request-boundary redesign now present at a701a256. The binding-specific ParsedPolicy/telemetry bool model and binding.rs were removed; mxc_ffi::request::build_request_from_json forwards the original complete JSON string to mxc_sdk::ffi_internals::build_request_from_json and the central parser, which accepts an omitted telemetry.enabled in the canonical telemetry object while preserving canonical null/unknown/duplicate-field validation.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,835 @@
...
+            if (handle.Target is not PresenterContext presenterContext)
+            {
+                return NativeConsentPresenterError;
+            }
+
+            var prompt = ParseConsentPrompt(ReadRequiredJson(promptJsonUtf8, "telemetry consent prompt"));
+            var decision = presenterContext.Presenter(prompt);
+            if (presenterContext.CancellationToken.IsCancellationRequested)
+            {
+                return NativeConsentDecisionDismissed;
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in 8ca74588. The public XML documentation and .NET README now explicitly define RequestConsentAsync cancellation as best-effort relative to native persistence: an unfinished presenter can be dismissed, while a completed decision that wins the race and begins persistence remains authoritative. The persistence-wins regression test was renamed to describe concurrent cancellation. The affected MxcTelemetryTests passed.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxRequest.cs
```diff
@@ -21,39 +23,53 @@ public SandboxRequest(SandboxPolicy policy, string command)
...
-[JsonDerivedType(typeof(WslcContainment), "wslc")]
+/// <summary>Serializes requests through MXC's canonical wire projection.</summary>
+public sealed class SandboxRequestJsonConverter : JsonConverter<SandboxRequest>
+{
+    /// <inheritdoc />
+    public override SandboxRequest Read(
+        ref Utf8JsonReader reader,
+        Type typeToConvert,
+        JsonSerializerOptions options) =>
+        throw new NotSupportedException("SandboxRequest deserialization is not supported.");
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in 8ca74588. SandboxRequestJsonConverter.Read now strictly deserializes the current canonical request shape emitted by CanonicalRequestBuilder, reconstructing filesystem/lifecycle, process/environment, UI, legacy and directional networking, telemetry, ProcessContainer, and WSLC options. Unsupported legacy envelopes and unrepresentable/malformed values fail with JsonException rather than being silently dropped. Shared ProcessContainer, directional-network, and WSLC golden fixtures now round-trip, with additional malformed-input coverage. The affected MxcSandboxTests passed.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcSandboxTests.cs
```diff
@@ -112,7 +113,11 @@ public void FullRequestSerialization_MatchesCrossLanguageGoldens()
                     },
                 },
             },
-            "echo network");
+            "echo network")
+        {
+            Containment = new ProcessContainerContainment(),
+            ContainerName = "golden-directional-network",
+        };
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in e82def39. The ProcessContainer serialization golden assertions now run only on Windows, while the WSLC golden remains cross-platform. The canonical-deserialization theory likewise skips only its two ProcessContainer fixtures on non-Windows. The full .NET suite passed locally (187 total, 159 passed, 28 host/platform skips).

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
```diff
@@ -0,0 +1,897 @@
...
+        };
+        if (request.WorkingDirectory is not null)
+        {
+            process["cwd"] = request.WorkingDirectory;
+        }
+        if (request.Environment.Count != 0)
+        {
+            process["env"] = JsonSerializer.SerializeToNode(
+                request.Environment.Select(pair => $"{pair.Key}={pair.Value}").ToArray());
+        }
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in e82def39. Canonical request projection now rejects empty environment variable names and names containing `=` before constructing the `NAME=VALUE` array, preventing a key from being reinterpreted as a different variable. Regression coverage was added for both invalid forms; the full .NET suite passed locally.

scripts/check-dotnet-api-parity.js
```diff
@@ -258,66 +253,94 @@ const managedPolicy = read(
...
+);
+
+const managedSandboxPolicyFields = managedJsonFields(
+  managedPolicy,
+  "SandboxPolicy"
+);
+
+// `captureDenials` is the one existing managed compatibility alias (MXC0001,
+// removed in 1.0). MxcSandbox.PrepareRequest migrates it to
+// containment.captureDenials and strips it before native serialization.
```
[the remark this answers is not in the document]
RamonArjona4: Addressed in e82def39. The parity-gate compatibility note now names the canonical destination `processContainer.captureDenials`, matching CanonicalRequestBuilder and the request tests. The parity gate passes.

src/core/mxc_engine/src/policy.rs
```diff
@@ -781,13 +794,75 @@ pub fn build_request_with_containment(
...
+    if policy.version.is_empty() {
+        return Err(MxcError::malformed_request("Policy version is required").into());
+    }
+    Ok(())
+}
+
+/// Build a request from a complete canonical MXC configuration document.
+#[cfg(feature = "ffi-internals")]
+#[doc(hidden)]
+pub fn build_request_from_json(request_json: &str) -> Result<SandboxRequest, crate::Error> {
```
jsidewhite: >build_request_from_json [](http://example.com/codeflow?start=7&length=23)

why did this function move from the FFI layer down into the engine?

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
```diff
@@ -0,0 +1,900 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
jsidewhite: >Licensed [](http://example.com/codeflow?start=3&length=8)

Please add a short comment at the top of this file that explains how it fits into the world.

jsidewhite: i.e. what is a CanonicalRequestBuilder?

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
```diff
@@ -0,0 +1,900 @@
...
+        PropertyNameCaseInsensitive = false,
+        UnmappedMemberHandling = JsonUnmappedMemberHandling.Disallow,
+        Converters =
+        {
+            new JsonStringEnumConverter(JsonNamingPolicy.CamelCase),
+            new NetworkProxyPolicyJsonConverter(),
+        },
+    };
+
+    internal static string Serialize(SandboxRequest request, JsonSerializerOptions options)
```
jsidewhite: >Serialize [](http://example.com/codeflow?start=27&length=9)

why are we making yet another serializer, specifically for C#, and in a telemetry pr at that?

jsidewhite: I don't truly understand, but my gut says this whole file needs to go.

src/core/mxc_engine/src/policy.rs
RamonArjona4: We moved  build_request_from_json  into  mxc_engine  because converting a complete canonical MXC document into an executable  SandboxRequest  belongs to the execution engine, rather than to any particular language-binding layer.

This establishes a single parsing boundary for all callers. The engine invokes the existing  wxc_common  configuration parser, which already owns schema validation, duplicate-property detection, normalization, and conversion into the internal execution model. Keeping that logic in the engine prevents the FFI layer from developing a parallel parser whose behavior could drift from the CLI, Rust SDK, or other bindings.

The FFI code is now intentionally thin: it validates the C boundary, forwards the canonical JSON to the engine, and applies execution context that is deliberately not encoded as policy—most notably the caller’s experimental-feature authorization. This separation keeps policy interpretation centralized while preserving the distinction between what the request asks for and what the caller is authorized to execute.

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
RamonArjona4: Added a top-of-file comment explaining that this class adapts the ergonomic .NET authoring model to MXC's canonical cross-language wire contract, while the central Rust parser remains responsible for validation and executability.

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
RamonArjona4: CanonicalRequestBuilder  adapts the ergonomic C# authoring model to MXC’s canonical cross-language JSON contract. The models are not structurally identical, so direct  System.Text.Json  serialization would produce the wrong wire shape. This class performs that projection while leaving validation and executability decisions to the central Rust parser.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandbox.cs
```diff
@@ -121,7 +121,10 @@ public static RunResult Run(SandboxRequest request)
             fixed (byte* requestPtr = requestBuf)
             {
                 MxcRunResult result = default;
-                var status = NativeMethods.mxc_run_request(requestPtr, &result);
+                var status = NativeMethods.mxc_run_request(
+                    requestPtr,
+                    request.Experimental ? 1 : 0,
```
jsidewhite: >request.Experimental ? 1 : 0, [](http://example.com/codeflow?start=20&length=29)

Looking again, is this required?  We promoted telemetry to stable in 820

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
```diff
@@ -0,0 +1,900 @@
...
+        }
+        if (RequiresOutboundForHostRules(request.Containment)
+            && policy.Network is { AllowOutbound: false } networkWithRules
+            && (networkWithRules.AllowedHosts.Count != 0
+                || networkWithRules.BlockedHosts.Count != 0))
+        {
+            throw new MxcException(
+                ErrorCode.MalformedRequest,
+                "allowedHosts/blockedHosts require allowOutbound to be true.");
+        }
```
bbonaby: issue: we actually do this validation in the mxc engine layer. We likely don't need to do it here.

src/core/mxc_engine/src/configs/process_container.rs
```diff
@@ -173,19 +173,6 @@ pub(crate) fn apply(
...
-                "processContainer.learningMode requires schema version 0.8 or later",
-            ));
-        }
-        if process_container.capture_denials.is_some() {
-            return Err(wxc_common::mxc_error::MxcError::malformed_request(
-                "processContainer.captureDenials requires schema version 0.8 or later",
-            ));
-        }
-    }
-
```
bbonaby, line 188: question: do you know why this would need to be removed?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -918,6 +918,26 @@ fn convert_wire_config(
...
+        if cfg
+            .process_container
+            .as_ref()
+            .is_some_and(|section| section.capture_denials.is_some())
+        {
+            return Err(WxcError::ConfigParse(
+                "processContainer.captureDenials requires schema version 0.8 or later".to_string(),
+            ));
+        }
+    }
```
bbonaby: issue: Oh I think this might be related to the removal in process_container.rs. We can probably keep the changes in there and remove these additions

src/ffi/mxc_ffi/src/request.rs
```diff
@@ -1,534 +1,72 @@
...
-#[derive(serde::Deserialize)]
-#[serde(rename_all = "camelCase", deny_unknown_fields)]
-struct RequestSpec {
-    policy: SandboxPolicy,
-    command: String,
-    #[serde(default)]
-    containment: RequestContainment,
-    #[serde(default)]
-    container_name: Option<String>,
-    #[serde(default)]
```
jsidewhite, line 27: where did this all go?

RamonArjona4: This is part of the bit that was doing parsing for C# in the FFI layer.  I believe our options are to keep it here and put C# parsing for telemetry into the FFI layer with it, or move this parsing out of the FFI layer and into the SDK layer (which is reflected in this PR). Please let me know if there's something I'm missing.

src/core/mxc_engine/src/policy.rs
RamonArjona4: This has been reverted. `RequestSpec` parsing and binding-envelope adaptation now live in `mxc_ffi/src/request.rs` again, and the canonical JSON bridge in `mxc_engine` was removed. Resolving as addressed.

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
RamonArjona4: Agreed. `CanonicalRequestBuilder` has been removed. The C# SDK now serializes its established binding request envelope, and `mxc_ffi` parses/adapts that envelope into `SandboxRequest`. Resolving as addressed.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandbox.cs
RamonArjona4: The separate native `experimental` argument shown here was removed when the `RequestSpec` ABI was restored. `experimental` remains in the request envelope because it still gates experimental backends/features; telemetry itself does not depend on it. Resolving this outdated thread.

sdk/dotnet/Microsoft.Mxc.Sdk/CanonicalRequestBuilder.cs
RamonArjona4: Agreed and addressed by the architectural revert: `CanonicalRequestBuilder` no longer exists. Binding-envelope validation/adaptation is back in the FFI `RequestSpec` parser, while canonical policy validation remains in the shared parser. Resolving.

src/ffi/mxc_ffi/src/request.rs
RamonArjona4: This has been restored here. `RequestSpec` and its binding-envelope parsing are back in the FFI layer, including parsing `policy.telemetry.enabled` and applying it as invocation metadata. Resolving as addressed.

RamonArjona4:
Following up on the telemetry placement question in https://github.com/microsoft/mxc/pull/821#issuecomment-5592269493: telemetry is intentionally not a field on the engine `SandboxPolicy` because it is invocation metadata rather than a sandbox restriction. With the restored architecture, the C# binding envelope carries `policy.telemetry.enabled`; the FFI `RequestSpec` parser extracts it and applies it through `SandboxRequest::set_telemetry_opt_in`.

src/core/mxc_engine/src/configs/process_container.rs
RamonArjona4: Agreed. I restored these schema-version checks to `process_container::apply` and restored the unit test to exercise `build_wire_config` directly in 6e2b278f. Resolving as addressed.

src/core/wxc_common/src/config_parser.rs
RamonArjona4: Agreed. The duplicate checks have been removed from `config_parser.rs`; the original validation remains in `process_container.rs` in 6e2b278f. Resolving as addressed.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcTelemetry.cs
```diff
@@ -0,0 +1,840 @@
...
+            LearnMoreLabel = ParseConsentMessage(root.GetProperty("learnMoreLabel")),
+            LearnMoreUrl = root.GetProperty("learnMoreUrl").GetString() ?? string.Empty,
+        };
+    }
+
+    private static TelemetryConsentMessage ParseConsentMessage(JsonElement value) => new()
+    {
+        Id = value.GetProperty("id").GetString() ?? string.Empty,
+        Text = value.GetProperty("text").GetString() ?? string.Empty,
+    };
```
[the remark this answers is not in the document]
RamonArjona4: Fixed in 7a5aa623. Required top-level prompt strings and every nested message `id`/`text` now must be JSON strings; `null` or another type throws `JsonException`, causing the native presenter bridge to return an error without persisting consent. Added coverage for all 12 required string fields.

src/core/mxc_engine/src/policy.rs
```diff
@@ -781,13 +790,31 @@ pub fn build_request_with_containment(
...
+    build_wire_config_with_network_format(policy, containment, container_name)
+        .map(|(config, _)| config)
+}
+
+fn build_wire_config_with_network_format(
+    policy: &SandboxPolicy,
+    containment: &Containment,
+    container_name: Option<&str>,
+) -> Result<(serde_json::Value, NetworkFormat), MxcError> {
     use serde_json::json;
```
jsidewhite, line 818: i don't really understand this subtle change, but good enough.

jsidewhite (approved):
:shipit:

PR #835, opened 2026-08-12
Add 0.6 configuration contract adapter

MGudgin:
This PR adds the typed adapter from exact 0.6.0-alpha contract requests to the current wxc_common wire model.

Details

* Adds the contract-crate dependency and a version-scoped adapter with exhaustive source and destination mappings.
* Preserves optional-field presence, canonical compatibility aliases, and explicit absence for later-version fields.
* Adds expected-wire and current-wire equivalence coverage for representative requests, proxies, enums, and aliases.

Tests

* `cargo check -p wxc_common`
* `cargo clippy -p wxc_common --all-targets -- -D warnings`
* `cargo test -p wxc_common`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/835)

### DIFF
--- a/src/core/wxc_common/src/config_contract_adapters/v0_6.rs
+++ b/src/core/wxc_common/src/config_contract_adapters/v0_6.rs
@@ -1,0 +1,949 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::wire;
+use mxc_config_contract::published::v0_6_0_alpha as contract;
+use mxc_config_contract::ContractVersion;
+
+fn convert_version(value: contract::Version) -> &'static str {
+    match value {
+        contract::Version::V0_6_0Alpha => ContractVersion::V0_6_0Alpha.as_str(),
+    }
+}
+
+fn convert_containment(value: contract::Containment) -> wire::Containment {
+    match value {
+        contract::Containment::Process => wire::Containment::Process,
+        contract::Containment::ProcessContainer => wire::Containment::ProcessContainer,
+        contract::Containment::Lxc => wire::Containment::Lxc,
+        contract::Containment::Bubblewrap => wire::Containment::Bubblewrap,
+    }
+}
+
+fn convert_process(value: contract::Process) -> wire::Process {
+    let contract::Process {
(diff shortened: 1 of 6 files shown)


src/core/wxc_common/src/config_contract_adapters/v0_6.rs
```diff
@@ -0,0 +1,949 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::wire;
+use mxc_config_contract::published::v0_6_0_alpha as contract;
+use mxc_config_contract::ContractVersion;
+
+fn convert_version(value: contract::Version) -> &'static str {
```
bbonaby, line 8: suggestion: I saw these conversion helpers and immediately thought they might be a good fit for the [From/Into traits](https://doc.rust-lang.org/rust-by-example/conversion/from_into.html). Since these are infallible conversions between the contract and wire types, we could implement `From<contract::X> for wire::X`, which would simplify the nested conversions quite a bit when converting from request to MxcConfig:

```rust
Self {
    // ...
    containment: containment.into_option().map(Into::into),
    process: Some(process.into()),
    lifecycle: lifecycle.into_option().map(Into::into),
    process_container: process_container.into_option().map(Into::into),
    lxc: lxc.into_option().map(Into::into),
    filesystem: filesystem.into_option().map(Into::into),
    fallback: fallback.into_option().map(Into::into),
    network: network.into_option().map(Into::into),
    ui: ui.into_option().map(Into::into),
    // ...
}
```

So if we implement `From<contract::Request> for wire::MxcConfig`, I think we could remove `into_wire` entirely aswell and have callers just use `request.into()`.  What do you think?

bbonaby (approved):

src/core/wxc_common/src/config_contract_adapters/v0_6.rs
MGudgin: Sorry, only saw this comment after I'd merged. Looking into this now.

PR #838, opened 2026-08-13
Add 0.7 configuration contract and adapter

MGudgin:
This PR adds the exact published `0.7.0-alpha` contract and its adapter to the current `wxc_common` wire model.

Details

* Adds recursively closed `0.7` request types for annotations and the stable Seatbelt surface.
* Preserves AppContainer and macOS Sandbox compatibility aliases while enforcing string-only enum representations in both published contracts.
* Adds exhaustive adapter mappings, version-boundary coverage, fixtures, and current-wire equivalence tests.

Tests

* `cargo check -p mxc_config_contract -p wxc_common`
* `cargo clippy -p mxc_config_contract --all-targets -- -D warnings`
* `cargo clippy -p wxc_common --all-targets -- -D warnings`
* `cargo test -p mxc_config_contract --no-fail-fast` (175 tests passed)
* `cargo test -p wxc_common` (628 tests passed)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/838)


src/core/mxc_config_contract/src/published/v0_6_0_alpha/mod.rs
```diff
@@ -7,6 +7,92 @@
 //! published contract. They preserve omitted optional fields for a later
 //! adapter to default and normalize.
 
+// Serde's default fieldless-enum deserializer also accepts externally
+// tagged object forms such as {"process": null}. Published contracts accept
+// only string values. This macro generates a string-only deserializer while
+// supporting explicit compatibility aliases.
+macro_rules! string_enum {
```
bbonaby, line 16: note: this macro is duplicated between 0.6 and 0.7. Since they're the same and these configs are stable I think we can probably put them in a common module under `src/common/macros`. Unless of course you plan to update them later.

src/core/mxc_config_contract/src/published/v0_7_0_alpha/primitives.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl<'de> Deserialize<'de> for NonEmptyString {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+        Self::new(value).map_err(de::Error::custom)
+    }
+}
```
bbonaby, line 96: suggestion: We can probably still use serde here for this one too:
```rust
#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
#[serde(try_from = "String")]
pub struct NonEmptyString(String);

impl TryFrom<String> for NonEmptyString {
    type Error = String;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        Self::new(value)
    }
}
```

src/core/mxc_config_contract/src/published/v0_7_0_alpha/primitives.rs
```diff
@@ -0,0 +1,113 @@
...
+    where
+        D: Deserializer<'de>,
+    {
+        if bool::deserialize(deserializer)? {
+            Ok(Self)
+        } else {
+            Err(de::Error::custom("expected true"))
+        }
+    }
+}
```
bbonaby, line 21: suggestion: Since this is a type and has no  fields we can probably use serde here:
```rust
#[derive(Deserialize)]
#[serde(try_from = "bool")]
pub struct True;

impl TryFrom<bool> for True {
    type Error = &'static str;

    fn try_from(value: bool) -> Result<Self, Self::Error> {
        if value {
            Ok(True)
        } else {
            Err("expected true")
        }
    }
}
```

src/core/mxc_config_contract/src/published/v0_7_0_alpha/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+//!
+//! These types validate the JSON structure and value constraints of the
+//! published contract. They preserve omitted optional fields for a later
+//! adapter to default and normalize.
+
+// Serde's default fieldless-enum deserializer also accepts externally
+// tagged object forms such as {"process": null}. Published contracts accept
+// only string values. This macro generates a string-only deserializer while
+// supporting explicit compatibility aliases.
+macro_rules! string_enum {
```
bbonaby, line 16: note: I noticed on the top of where these are used we have `#[rustfmt::skip]`, its up to you might might help future folks if we added a comment to this macro here about why we want to add this. I think it's so rustfmt doesn't mess up the content of string_enum  right?

bbonaby (dismissed):

src/core/mxc_config_contract/src/published/v0_6_0_alpha/mod.rs
MGudgin: This is deliberate. I want the contracts to be entirely self-contained so that any changes to one cannot affect another. While it is *very* unlikely the macro will change, I want to hold this line for a while until I'm more confident about what constructs will and will not change over time.

src/core/mxc_config_contract/src/published/v0_7_0_alpha/mod.rs
MGudgin: It's mainly because rustfmt doesn't seem to know how to format the content of a macro, and I didn't want people to think rustfmt wasn't working. I'll add a comment here.

src/core/mxc_config_contract/src/published/v0_7_0_alpha/primitives.rs
MGudgin: I did consider using try_from support here. Doing so requires public TryFrom<String> implementations on the exported contract type, which would make general-purpose conversions and their associated error types part of the crate API. These types are intended specifically as immutable JSON wire-contract types, so I'm  deliberately keeping validation inside their Deserialize implementations rather than exposing additional construction paths. The implementations are small, preserve the exact accepted JSON shape, and keep the public surface minimal.

src/core/mxc_config_contract/src/published/v0_7_0_alpha/primitives.rs
MGudgin: See above.

src/core/mxc_config_contract/src/published/v0_7_0_alpha/primitives.rs
MGudgin: Note: I have removed pub from new as its not needed.

bbonaby (dismissed):

src/core/mxc_config_contract/src/published/v0_6_0_alpha/primitives.rs
```diff
@@ -66,7 +66,7 @@ impl NonEmptyString {
     /// Creates a validated string.
     ///
     /// Returns an error when `value` is empty.
-    pub fn new(value: String) -> Result<Self, String> {
+    fn new(value: String) -> Result<Self, String> {
```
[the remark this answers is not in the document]
MGudgin: This contract class is not yet wired up.

bbonaby (dismissed):

src/core/mxc_config_contract/src/published/v0_7_0_alpha/request.rs
```diff
@@ -0,0 +1,257 @@
...
+    #[serde(default)]
+    pub keychain_access: OptionalField<bool>,
+    /// Additional Mach service global names the process may resolve.
+    #[serde(default)]
+    pub extra_mach_lookups: OptionalField<Vec<String>>,
+}
+
+/// A complete one-shot `0.7.0-alpha` configuration request.
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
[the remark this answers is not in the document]
MGudgin: This difference is intentional at this layer. The version-specific Request types perform structural deserialization; the planned parse flow adapts them into the runtime wire model and then applies the existing semantic and cross-field validation in config_parser. Moving backend/section compatibility into custom deserialization would duplicate and freeze semantic rules inside each immutable contract module, including rules involving the host-dependent process intent.

I have updated the test examples though, so that they don't contain JSON that will eventually fail in the latest parts of the flow.

bbonaby (approved):

PR #847, opened 2026-08-13
Unify audit with capture denial routing

richiemsft:
## 📖 Description

Unifies `wxc-exec --audit` with the `processContainer.captureDenials`
execution pipeline.

`--audit` synthesizes allow-mode capture with ETL retention and
`permissiveLearningMode`, then delegates backend selection and execution to
`mxc_engine`. Native PSEC/V2 capture remains preferred when available and
policy-compatible; legacy SBOX and AppContainer tiers use guarded WPR with
exact handle-attested process scope.

WPR records a host-wide source ETL inside protected elevated scratch. After the
sandbox process tree terminates, the guardian uses normalized `FILETIME`
selection and Windows Trace Relogger to create a second ETL containing only
supported Learning Mode events from the attested process generations. Analysis
and explicit retention both consume that filtered ETL. The host-wide source
never crosses the privilege boundary, and relogging failure transfers no trace.

Both native and guarded providers honor explicit `retainEtl` after a terminal
wait. Abandoned process handles discard the trace rather than exposing
incomplete output.

Audit post-processing consumes the canonical `DenialsDocument` returned by the
selected provider, relocates outputs to `denials.json` and `trace.etl`,
snapshots file-based source policies, and generates `Adjusted_*.json` without
decoding the ETL again. Truncated analysis preserves JSON, ETL, and the source
snapshot but does not generate an adjusted policy.

## 🔗 References

- Builds on #813
- Reuses the canonical PLM analyzer from #760
- Reuses native ETL retention from #774

## 🔍 Validation

- `cargo test -p learning_mode_windows -p plm --lib`
- `cargo clippy --workspace --all-targets -- -D warnings`
- Release builds of `wxc-exec.exe` and `plm.exe`
- Multi-axis review with no High or Medium findings
- Windows VM guarded-WPR capture with a concurrent unrelated sandbox:
  - retained ETL decoded successfully to the same 11 canonical denials;
  - the target denial was present in canonical JSON and the retained ETL;
  - the unrelated sandbox denial was absent from both;
  - WPR stopped cleanly with no recovery marker or PLM process remaining.

## ✅ Checklist

- [x] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [ ] Linked to an issue
- [x] Updated documentation
- [x] Updated [Copilot instructions](.github/copilot-instructions.md)
[...]

PR #916, opened 2026-08-14
Add verbose Learning Mode denial logging

### LINKED ISSUE #828: Add a new captureDenial_verbose file
### Description of the task

captureDenials stores actionable logs that can be modified in the mxc config. generate a second file that has all denial information regardless if the mxc config currently supports them.

### Additional context

_No response_

richiemsft:
## 📖 Description

Adds deterministic verbose logging for every successful Learning Mode denial analysis while preserving the canonical policy-output contract and expanding canonical decoding for newly validated ETL shapes.

Each analysis now produces a transactional pair:

- `denials.json` retains canonical, actionable denials, including complete file paths.
- `denials.verbose.json` retains a bounded, deduplicated diagnostic superset. Complete file paths are replaced with `<REDACTED>`; standalone account names and username components in non-file identifiers are also redacted before bounding or hashing.

The verbose logging output includes canonical denial occurrences and closed classifications for outcomes omitted from canonical policy generation, including unusable resource paths, unsupported object types and event schemas, malformed payloads, decoder limits, and unresolved capability events. Stable non-file identifiers such as SIDs, capability names, provider GUIDs, PIDs, and named-object details remain available for diagnosis.

The guarded WPR path now preserves known-provider events through process-scoped selection and replay. Brokered event 28 is scoped using its payload `ProcessId`, including partial-decode recovery, rather than the broker's ETW header PID.

Native capture and `plm stop` share a transactional paired-output writer. Both JSON files are staged and synchronized before no-clobber promotion, with ownership- and digest-aware rollback. Audit mode consumes these capture artifacts directly and relocates both the JSON pair and retained ETL through no-clobber staging with identity- and digest-verified source cleanup, keeping metadata truthful without deleting files owned by another process.

### Additional ETL findings

Validation across eight new namespace scenarios (16 retained ETLs: block and allow) identified additional canonical decoder shapes:

- Event 28 is schema-dependent. UI-shaped payloads containing `Category` and `Detail` now emit canonical UI denials such as `resource: "Handles"`, `resourceType: "ui"`, rather than incorrectly treating the payload's package SID as a capability.
- Named `Section`, `SymbolicLink`, and `Timer` access checks now emit canonical `other` resources with object-specific access-mask classification.
- The UI Handles scenario represents `JOB_OBJECT_UILIMIT_HANDLES`, primarily cross-process HWND/USER-object handle access.
- SCM, ETW provider/session, and power-policy scenario-specific checks still expose empty `ObjectType` and `ObjectName`. They remain bounded verbose logging diagnostics because the ETL provides no stable canonical resource identifier.

The updated decoder increased canonical coverage in 15 of the 16 supplied ETLs. The remaining ETW Session block trace was an early process-initialization failure and contained none of the newly supported named shapes.

## 🔗 References

Builds on the Learning Mode capture hardening from #847.
This is for issue #828

## 🔍 Validation

- `learning_mode_core`: 43 tests passed
- `learning_mode_windows`: 224 tests passed
- `appcontainer_common`: 247 tests passed
- `plm`: 226 tests passed, 1 ignored
- `wxc`: 40 tests passed
- Clippy passed with warnings denied for `learning_mode_core`, `learning_mode_windows`, and `plm`
- Release `wxc-exec.exe` and `plm.exe` build
- Guarded-WPR VM validation with canonical and typed verbose logging groups, zero overflow, and no username leakage
- Production `lm_analyze` decoder run against all 16 supplied retained ETLs
- Confirmed canonical `ui/Handles`, Timer, Section, and SymbolicLink output; confirmed event 28 UI no longer emits a bogus package-SID capability
- Fixture accounting now follows the production handler exactly; out-of-scope events do not consume the processing budget
- Create-new output pairs no longer create persistent per-run lock files; relocation publishes committed destinations and surfaces source-cleanup warnings
- Malformed event 28 diagnostics classify UI/capability resources only when the schema name identifies the payload shape
- GitHub Actions build, lint, integration, CodeQL, CLA, and dependency-feed checks passed before the latest ETL decoder commit; replacement checks run automatically

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/916)


src/backends/learning_mode/windows/src/etl_filter.rs
```diff
@@ -0,0 +1,279 @@
...
+        // valid for the duration of this callback.
+        let record = unsafe { event.GetEventRecord()? };
+        let Some(record) = (unsafe { record.as_ref() }) else {
+            return Err(windows::core::Error::from_hresult(
+                windows::Win32::Foundation::E_POINTER,
+            ));
+        };
+        let header = &record.EventHeader;
+        if !is_learning_mode_event(header.ProviderId, header.EventDescriptor.Id) {
+            return Ok(());
```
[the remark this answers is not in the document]
richiemsft: Fixed in a2208623. Both relogging passes now count every event from a known Learning Mode provider via `provider_category`, rather than filtering to supported IDs. In-scope unknown IDs survive into the filtered ETL and are classified as `unsupportedEventSchema`; `relog_selection_tracks_known_provider_ordinals_and_exact_lifetimes` covers the unknown-ID ordinal.

src/host/plm/src/analysis.rs
```diff
@@ -46,7 +50,7 @@ pub fn write_denials(output_path: &Path, analysis: &AnalysisResult, exit_code: i
     temp.persist(output_path)
         .map_err(|error| error.error)
         .with_context(|| format!("failed to replace {}", output_path.display()))?;
-    Ok(())
+    Ok(document)
```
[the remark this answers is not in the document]
richiemsft: Fixed in a2208623 and hardened by the later paired-output commits. `plm stop` now derives the deterministic `.data-loop.json` sibling and writes the canonical/Data Loop documents through shared `write_paired_output_files`. Tests verify both documents and rollback cleanup on promotion failure.

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -542,52 +747,151 @@ unsafe fn process_event_record(event_record: *mut EVENT_RECORD, acc: &mut Accumu
...
+            acc.stop_requested = true;
             return;
         };
-        if !acc.in_analysis_scope(header.ProcessId, filetime) {
+        acc.relog_event_count = next_event_count;
+        let Some(filetime) = normalized_filetime(header.TimeStamp, acc) else {
             return;
+        };
+        if acc.event_in_scope(header.ProcessId, filetime) {
+            if acc.relog_selected_event_indices.len() >= MAX_PROCESSED_EVENTS {
```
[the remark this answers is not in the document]
richiemsft: Fixed in a2208623, with malformed/partial-decode coverage refined in 5ab71790, b1711724, and a261e425. Guarded relog selection decodes event 28 `ProcessId` and scopes by that effective workload PID; it falls back only for an in-scope emitter when the payload cannot be attributed, and fails closed on schema errors. The relog-selection PID tests pass at current HEAD.

src/core/wxc/src/audit.rs
```diff
@@ -186,14 +218,26 @@ fn move_new_file(source: &Path, destination: &Path) -> Result<(), String> {
...
-                    "failed to move audit artifact {} to {}: {rename_error}; \
-                     copy fallback failed: {copy_error}",
-                    source.display(),
-                    destination.display()
-                )
-            })?;
+            if let Err(copy_error) = std::fs::copy(source, destination) {
+                let cleanup_error = std::fs::remove_file(destination)
+                    .err()
+                    .filter(|error| error.kind() != std::io::ErrorKind::NotFound);
```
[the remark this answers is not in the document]
richiemsft: Fixed in 8d11809f. The fallback now opens the destination with `create_new(true)` and copies through that owned handle. An `AlreadyExists` failure performs no cleanup; partial-file cleanup occurs only after this invocation successfully created the destination. Added a regression test proving an existing destination is never clobbered or removed.

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,7 +207,292 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+    for (name, raw_value) in props {
+        if is_timestamp_like_property(name) {
+            continue;
+        }
+        let value = raw_value.trim_matches('"');
+        let sanitized_value = if is_identity_property(name) {
+            REDACTED_USER.to_string()
+        } else {
+            redact_known_username_components(&redact_username_in_path(value), &usernames)
+        };
```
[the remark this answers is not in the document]
richiemsft: Addressed in 469d125c. Data Loop file resources are now type-aware and the complete value is replaced with `<REDACTED>` before bounding; path-shaped unknown-event properties such as `D:\Profiles\alice\secret` receive the same whole-path redaction. Canonical denials retain their actionable path.

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,7 +207,285 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+}
+
+fn looks_like_file_path_property(name: &str, value: &str, object_type: Option<&str>) -> bool {
+    let normalized_name = name.to_ascii_lowercase().replace(['_', '-'], "");
+    if normalized_name.contains("filepath")
+        || normalized_name == "path"
+        || normalized_name.ends_with("filename")
+    {
+        return true;
+    }
```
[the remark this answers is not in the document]
richiemsft: Fixed in 262cc6b2. The sanitizer now treats every property name ending in `Path` (including `AppPath` and `ApplicationPath`) as a complete file path and replaces it with `<REDACTED>` before bounding or hashing. Regression coverage exercises both fields without relying on canonical denial recording.

src/core/wxc/src/audit.rs
```diff
@@ -143,38 +174,45 @@ fn validate_metadata(
...
 
     // Primary denials JSON first: publish its final path the instant it moves so
     // a later ETL failure can never leave `output_path` referencing a file that
     // has already been renamed away.
-    move_file(source_denials, final_denials)?;
-    capture.output_path = final_denials.to_string_lossy().into_owned();
+    move_file(paths.source_denials, paths.final_denials)?;
+    capture.output_path = paths.final_denials.to_string_lossy().into_owned();
+
+    move_file(paths.source_data_loop, paths.final_data_loop)?;
```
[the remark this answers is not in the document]
richiemsft: Fixed in 9ebe3a11. Audit now relocates the canonical/Data Loop JSON files through shared `relocate_paired_output_files`, which stages both no-clobber destinations and rolls the first promotion back if the second fails. `capture.output_path` is published only after the complete pair commits, and `relocate_rolls_back_json_pair_before_publishing_metadata` covers the second-promotion failure.

src/core/wxc/src/audit.rs
```diff
@@ -206,6 +255,37 @@ fn move_new_file(source: &Path, destination: &Path) -> Result<(), String> {
...
+        .open(destination)
+        .map_err(|error| CopyNewFileError {
+            error,
+            cleanup_error: None,
+        })?;
+    if let Err(error) = std::io::copy(&mut source_file, &mut destination_file) {
+        drop(destination_file);
+        let cleanup_error = std::fs::remove_file(destination)
+            .err()
+            .filter(|cleanup_error| cleanup_error.kind() != std::io::ErrorKind::NotFound);
```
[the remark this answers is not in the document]
richiemsft: Fixed in 9ebe3a11 and extended to retained ETL in 46e4f3b1. Promoted files retain a `same_file` identity handle plus SHA-256 digest; cleanup first quarantines the current path and deletes it only when identity and contents still match. Concurrent replacements are restored rather than removed. Audit ETL relocation now uses the same verified no-clobber primitive.

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,7 +207,328 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+
+    if !normalized.equals("objectname") && !normalized.equals("resource") {
+        return false;
+    }
+    if let Some(object_type) = object_type {
+        return object_type.eq_ignore_ascii_case("File");
+    }
+
```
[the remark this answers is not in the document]
richiemsft: Fixed in e2c3aca8. A declared non-`File` object type no longer short-circuits lexical detection for `ObjectName`/`resource`; unambiguous absolute path forms are still replaced with `<REDACTED>`. Non-path object-manager and registry identifiers remain available. Added classifier and full sanitizer regressions for `ObjectType=Section` with `C:\Users\alice\secret.txt`.

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,7 +207,328 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+    let drive_absolute = bytes.len() >= 3
+        && bytes[0].is_ascii_alphabetic()
+        && bytes[1] == b':'
+        && matches!(bytes[2], b'\\' | b'/');
+    drive_absolute
+        || value.starts_with(r"\\")
+        || value.starts_with(r"\??\")
+        || value.starts_with(r"\\?\")
+        || value.starts_with(r"\\.\")
+}
```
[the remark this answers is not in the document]
richiemsft: Fixed in b6cbbe1b. Data Loop sanitization now recognizes case-insensitive NT filesystem forms for `\Device\HarddiskVolumeN\...`, `\Device\Mup\...`, and redirector device components before bounding or hashing, even when `ObjectType` is non-file. Regression coverage includes volume, MUP, and LanmanRedirector paths and confirms `\Device\MountPointManager` remains diagnostic rather than being over-redacted.

src/core/learning_mode_core/src/analyze.rs
```diff
@@ -61,8 +65,129 @@ impl AnalysisResult {
...
+            return Ok(true);
+        }
+
+        let original = std::mem::take(&mut self.data_loop.signatures);
+        let mut retained_bytes = 0usize;
+        const ENVELOPE_HEADROOM: usize = 4 * 1024;
+        let base_len = serde_json::to_vec(self)?.len();
+        if base_len > max_bytes {
+            return Ok(false);
+        }
```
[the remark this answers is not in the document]
richiemsft: Fixed in c2e2d721. The empty-envelope fit check now serializes a borrowed view with all current signatures accounted as overflow before `mem::take`, so a `false` result leaves `AnalysisResult` untouched. The regression now includes a retained signature, asserts exact equality after a failed one-byte fit, and verifies a retry with the original serialized limit preserves the group.

docs/learning-mode/capabilities.md
```diff
@@ -105,9 +105,10 @@ stays enforced:
    hosts use native capture without PLM or UAC, while legacy or incompatible
    tiers use the session-scoped guarded-WPR fallback and elevate only its
    fixed-operation guardian. The CLI consumes the returned JSON and ETL paths,
-   relocates them to `denials.json` and `trace.etl`, and generates the source
-   snapshot and `Adjusted_*.json` from canonical denials without decoding ETL
-   again. Truncated analysis skips the adjusted config.
+   relocates the canonical output, its Data Loop sibling, and the trace to
+   `denials.json`, `denials.data-loop.json`, and `trace.etl`, and generates the
```
sosharm: I don't think we want to call anything data loop. There are denials that are actionable and denials that are not actionable for the caller but would be good for them to know about still.

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -377,26 +564,15 @@ fn resources_from_events_for_process_lifetimes(
...
     };
-    for raw in raws {
-        accumulator.add_raw_denial(raw);
-        if accumulator.stop_requested {
+    for event in events {
+        if !accumulator.begin_event() {
             break;
         }
+        handle_decoded_event(&event.parts, event.pid, event.filetime, &mut accumulator);
     }
```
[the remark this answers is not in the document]
richiemsft: Fixed in daf8f563. The fixture loop no longer pre-calls `begin_event()`; each fixture goes directly through `handle_decoded_event`, then iteration stops if the production handler sets `stop_requested`. The strengthened regression seeds the budget at its boundary and proves an unrelated out-of-scope event does not consume it while the following in-scope event is processed exactly once. All 223 `learning_mode_windows` tests and strict package clippy pass.

docs/learning-mode/capabilities.md
richiemsft: Addressed in afaf37af and 4151adce. The public and internal Data Loop naming is replaced with verbose logging throughout the Rust APIs, serialized field (`verboseLogging`), documentation, diagnostics, tests, and sibling artifact name (`*.verbose.json`).

richiemsft:
Addressed the four findings from Copilot review 4974861785 in 7cd698c2: CreateNew paired writes no longer leave run-specific lock files; relocation now returns a committed outcome with explicit source-cleanup warnings so audit metadata always points to the valid destination; quarantine deletion failures restore the source path; malformed event 28 records are classified as UI/capability only when the schema name identifies the shape; and `.github/copilot-instructions.md` now documents the canonical + verbose output pair. Validation: 43 learning_mode_core tests, 224 learning_mode_windows tests, 40 wxc tests, formatting, and strict targeted clippy.

docs/learning-mode/capabilities.md
```diff
@@ -203,14 +204,18 @@ regenerate its sandbox policy:
 - `denials` is already de-duplicated per `(resource, accessType)`, so
   `summary.totalDenials` equals `denials.length`.
 - Analysis retains at most 10,000 unique denials and processes at most
-  1,000,000 ETW events. Reaching either bound stops further analysis and sets
-  `summary.deniedResourcesTruncated` to `true`.
+  1,000,000 ETW events. Reaching the unique-denial bound stops adding canonical
```
dhoehna: `canonical`, `canonical`, `canonical`.  What is this word supposed to mean in this context and does it have to be placed everywhere?

docs/learning-mode/capabilities.md
```diff
@@ -219,6 +224,114 @@ regenerate its sandbox policy:
...
+Per-event TDH failures use closed diagnostic reasons:
+`eventPayloadMalformed` means the payload conflicts with its declared schema,
+`decoderLimitReached` means a nesting/element/work safety bound stopped
+decoding, and `unsupportedPropertyEncoding` means the decoder cannot consume
+that property shape. When TDH exposes it, the schema-declared name is retained
+as the bounded `EventName` signature property. Free-form decoder errors are
+never serialized. Failure to obtain the event schema remains a fatal analysis
+error rather than being represented as a verbose logging signature.
+
+Verbose logging retains at most 4,096 distinct signatures, 24 sorted properties per
```
dhoehna: Why does a developer need to know this information?  IT looks like CoPilot trying to prove that logging has enough combinations to...to...uh...do something?

docs/learning-mode/capabilities.md
```diff
@@ -232,8 +345,8 @@ so CLI callers can locate the deliverable without scanning the filesystem:
 {"type":"captureDenials","outputPath":"C:\\logs\\denials.4321_0123456789abcdef0123456789abcdef.json","exitCode":0,"totalDenials":2,"deniedResourcesTruncated":false}
 ```
 
-The pointer echoes the file's `summary`; the authoritative record is the file
-itself. In-process Rust callers receive the same information through
+The pointer echoes the canonical file's `summary`; the authoritative policy
+record is that file. In-process Rust callers receive the same information through
```
dhoehna: What is `that`?

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -609,64 +799,257 @@ unsafe fn process_event_record(event_record: *mut EVENT_RECORD, acc: &mut Accumu
...
+    event_index: usize,
+    pid: u32,
+    filetime: u64,
+) {
+    if !acc.event_in_scope(pid, filetime) {
         return;
     }
+    if acc.relog_selected_event_indices.len() >= MAX_PROCESSED_EVENTS {
+        acc.decode_error = Some(format!(
+            "trace exceeded the {MAX_PROCESSED_EVENTS}-event process-scoped relogging limit"
```
dhoehna: What can the developer do to recover and re-run if they run into this issue?

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -609,64 +799,257 @@ unsafe fn process_event_record(event_record: *mut EVENT_RECORD, acc: &mut Accumu
...
+            );
+        }
+        return;
+    };
+    if !acc.event_in_scope(pid, filetime) || !acc.begin_event() {
+        return;
+    }
+
+    let primary = extract_denial(parts, pid, filetime);
+    // Some permissive capability checks leave the primary access-check event
```
dhoehna: Can this list be put into the comments?

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -754,14 +1136,143 @@ mod tests {
...
+        }]);
+        select_capability_event_for_relogging(&mut accumulator, 0, None, 42, 150);
+
+        assert_eq!(accumulator.relog_selected_event_indices, [0]);
+    }
+
+    #[test]
+    fn scoped_analysis_classifies_malformed_capability_from_in_scope_emitter() {
+        let mut accumulator = Accumulator::analyze_for_process_lifetimes(&[ProcessLifetime {
+            pid: 42,
```
dhoehna: Consider making these constants.

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -888,10 +1413,25 @@ mod tests {
             ),
         ];
 
-        let out = dedup_to_resources(denials).denials;
+        let analysis = dedup_to_resources(denials);
 
-        assert_eq!(out.len(), 1);
-        assert_eq!(out[0].resource, r"\\server\share\file.txt");
+        assert_eq!(analysis.denials.len(), 1);
+        assert_eq!(analysis.denials[0].resource, r"\\server\share\file.txt");
```
dhoehna: Crash is no entries.

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -1292,6 +2203,449 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn capability_denial_recovered_from_dacl_records_no_exclusion() {
+        // Same shape as `allow_shape_recovers_capability_from_dacl`, but
+        // asserting the verbose logging side: a successfully DACL-recovered
+        // capability candidate must not also surface an `UnresolvedCapability`
+        // exclusion for the same event.
+        let dacl = "hex:000000000000000001000000010200000000000F0300000001000000";
```
dhoehna, line 2372: Can you add a comment to what this is in decimal?  Or is this long hex string enough for troubleshooting this test?

src/backends/learning_mode/windows/src/etl_decode.rs
```diff
@@ -1292,6 +2203,449 @@ mod tests {
...
+        let signature = &out.verbose_logging.signatures[0].signature;
+        assert_eq!(
+            signature.reason,
+            VerboseLoggingExclusionReason::CanonicalDenial
+        );
+        assert_eq!(signature.access_type, Some(AccessType::Unknown));
+        assert_eq!(signature.resource_type, Some(ResourceType::Capability));
+    }
+
+    // ---- verbose logging signature retention (identifiers + redaction + dedup) --
```
dhoehna: Remove this.  No need to declare a region with a comment.  Now future developers have to understand what methods are "verbose logging signature retention" and remember to put them under this line.

src/backends/learning_mode/windows/src/etl_filter.rs
```diff
@@ -94,7 +103,9 @@ impl RelogSelectionState {
         // a foreign process's event at the same ordinal. Do not advance the
         // selected cursor on mismatch: the final count reconciliation then
```
dhoehna: I'm confused.  What cursor?

src/backends/learning_mode/windows/src/etl_filter.rs
```diff
@@ -145,10 +156,34 @@ impl ITraceEventCallback_Impl for ProcessScopedTraceFilter_Impl {
             ));
         };
         let header = &record.EventHeader;
-        if !is_learning_mode_event(header.ProviderId, header.EventDescriptor.Id) {
+        if provider_category(header.ProviderId).is_none() {
             return Ok(());
         }
-        if self.selection.observe_supported_event(header.ProcessId) {
+        let effective_pid = if header.EventDescriptor.Id
+            == crate::extractors::CAPABILITY_DENIAL_EVENT_ID
```
dhoehna, line 163: Please use () to keep the test understandable since it spans two lines.

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,46 +226,449 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+pub(crate) fn effective_capability_event_pid(process_id: Option<&str>) -> Option<u32> {
+    process_id.and_then(parse_u32)
+}
+
+/// Maps a raw ETW provider GUID to its symbolic verbose logging category.
+///
+/// Returns `None` for providers outside the Learning Mode vocabulary; those
+/// events are ignored entirely (not aggregated), since they are unrelated
+/// host traffic rather than an excluded Learning Mode outcome.
+pub(crate) fn provider_category(provider: GUID) -> Option<VerboseLoggingProvider> {
```
dhoehna: Rename function to etw_GUID_to_logging_directory.  That would be much more descriptive.  OR something similar.

Same for `provider_guid_string`

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,46 +226,449 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+pub(crate) fn provider_guid_string(provider: VerboseLoggingProvider) -> String {
+    match provider {
+        VerboseLoggingProvider::KernelGeneral => format_guid(KERNEL_GENERAL_PROVIDER),
+        VerboseLoggingProvider::PrivacyAuditingPermissiveLearningMode => {
+            format_guid(PRIVACY_LEARNING_MODE_PROVIDER)
+        }
+    }
+}
+
+fn format_guid(guid: GUID) -> String {
```
dhoehna: does rust not have a create for converting strings to GUIDs?

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,46 +226,449 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+    }
+}
+
+/// Returns whether `name` looks like it carries a timestamp, so it is
+/// dropped from a verbose logging signature entirely (never redacted or
+/// truncated) — exact timestamps must not prevent otherwise-identical
+/// events from deduplicating.
+fn is_timestamp_like_property(name: &str) -> bool {
+    let normalized = NormalizedPropertyName(name);
+    [
```
dhoehna: `71m3`, `+|/\/\3` although the second is a bit harder to see as "Time"

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,46 +226,449 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+fn is_identity_property(name: &str) -> bool {
+    let normalized = NormalizedPropertyName(name);
+    if normalized.contains("sid")
+        || normalized.contains("guid")
+        || normalized.contains("identifier")
+        || normalized.ends_with("id")
+    {
+        return false;
+    }
+    normalized.equals("user")
```
dhoehna: I feel like something exists besides hardcoding these string.  Unfortunately I don't know of such a thing.

dhoehna: However, I prefer having a tighter filter over a looser one.  Maybe this can pattern match and accidently omit words we did not want removed?

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,46 +226,449 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+}
+
+fn strip_prefix_ignore_ascii_case<'a>(value: &'a str, prefix: &str) -> Option<&'a str> {
+    value
+        .get(..prefix.len())
+        .is_some_and(|candidate| candidate.eq_ignore_ascii_case(prefix))
+        .then(|| &value[prefix.len()..])
+}
+
+fn ends_with_ignore_ascii_case(value: &str, suffix: &str) -> bool {
```
dhoehna, line 480: What is the boundary for adding more checks?  If a developer in 2 years adds more fields.  What notifies the developer they need to add a method here to normalize the field?

src/backends/learning_mode/windows/src/extractors.rs
```diff
@@ -142,46 +226,449 @@ pub(crate) fn is_learning_mode_event(provider: GUID, event_id: u16) -> bool {
...
+/// entries and [`MAX_SIGNATURE_VALUE_LEN`] characters per value.
+///
+/// Overlong values retain prefix and suffix context plus a digest of the full
+/// sanitized value. The digest prevents distinct values with a long shared
+/// prefix from collapsing into one verbose logging signature.
+///
+/// Callers must sanitize (redact) *before* calling this so neither the retained
+/// context nor the digest is derived from sensitive identity content.
+pub(crate) fn bound_properties(mut properties: Vec<(String, String)>) -> Vec<(String, String)> {
+    properties.truncate(MAX_SIGNATURE_PROPERTIES);
```
dhoehna, line 497: Is this correct?  Bound to MAX_SIGNATURE_PROPERTIES and discard any extra?

src/backends/learning_mode/windows/src/tdh_decode.rs
```diff
@@ -292,6 +451,62 @@ fn decode_properties(
...
 }
 
+fn decode_named_property(
+    info_buf: &[u8],
+    info: &TRACE_EVENT_INFO,
+    event_record: *mut EVENT_RECORD,
+    pointer_size: usize,
+    property_name: &str,
+) -> Result<Option<String>, PropertyDecodeError> {
+    // SAFETY: caller passes a valid EVENT_RECORD; the field accesses
```
dhoehna, line 462: What does `SAFETY` mean in this context?  `Pre-Assert`?

src/core/learning_mode_core/src/analyze.rs
```diff
@@ -61,8 +65,131 @@ impl AnalysisResult {
         Self {
             denials,
             denied_resources_truncated: false,
+            verbose_logging: VerboseLoggingSummary::default(),
         }
     }
+
+    /// Trims verbose logging signatures so the complete compact JSON result fits.
```
dhoehna: Fits in what?

src/core/learning_mode_core/src/analyze.rs
```diff
@@ -132,6 +263,14 @@ mod tests {
         let got = analyzer.analyze(Path::new("ignored.etl")).unwrap();
         assert_eq!(got.denials, denials);
         assert!(!got.denied_resources_truncated);
+        assert!(got.verbose_logging.is_empty());
+    }
+
+    #[test]
+    fn legacy_guarded_payload_defaults_verbose_logging_summary() {
```
dhoehna: `Legacy`?  Does the code this is testing exist?  If so, is the logic "legacy"?

Why legacy?

src/core/learning_mode_core/src/analyze.rs
```diff
@@ -160,4 +299,127 @@ mod tests {
         assert!(!lifetime.contains(42, 201));
         assert!(!lifetime.contains(43, 150));
     }
+
+    #[test]
+    fn serialized_size_limit_moves_verbose_logging_groups_to_overflow() {
```
dhoehna, line 305: Can overflow be overflowed?

src/core/learning_mode_core/src/paired_output.rs
```diff
@@ -0,0 +1,998 @@
...
+/// Copies an existing canonical/verbose logging pair to new no-clobber destinations,
+/// then removes only source files whose identity and contents still match the
+/// files that were copied.
+///
+/// The destination pair uses the same transactional promotion and rollback as
+/// [`write_paired_output_files`]. Source cleanup quarantines and verifies each
+/// file before deletion, so a concurrent replacement is restored rather than
+/// removed.
+///
+/// # Errors
```
dhoehna, line 245: I see `# Errors` and another `# ` string.  What are these for?

src/host/plm/src/stop.rs
```diff
@@ -659,6 +666,32 @@ mod tests {
         assert!(error.to_string().contains("would be overwritten"));
     }
 
+    #[test]
+    fn trace_input_cannot_collide_with_verbose_logging_output() {
```
dhoehna: Is this a genuine scenario?  This seems more like an exception case.

dhoehna (dismissed):
`Canonical`

Does copilot really need to use this every where?

docs/learning-mode/capabilities.md
richiemsft: Updated the user-facing documentation to use “policy output” and “policy denials.” The remaining `canonicalDenial` and `canonicalOverflowOccurrences` spellings are serialized JSON contract names, so those remain unchanged.

docs/learning-mode/capabilities.md
richiemsft: Condensed this section to the limits and truncation fields a caller needs to interpret the output. The internal transport-budget implementation details were removed.

docs/learning-mode/capabilities.md
richiemsft: Reworded this explicitly: the policy file is the authoritative denial record, while the in-process APIs receive the same summary information.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Added recovery guidance to both processing-limit errors: rerun a smaller workload or split it into multiple `captureDenials` runs.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Clarified that these are permissive Event 14 checks and pointed directly to the centralized `capability_dacl::KNOWN_CAPABILITIES` catalog. I avoided duplicating the full catalog here so the comment cannot drift from the recovery implementation.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Replaced the PID and FILETIME literals in this test with named scope constants.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Changed the indexing assertion to use `first().map(...)`, so an empty result produces an assertion failure instead of indexing the vector.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Added the meaningful decimal representation: this hex payload is a DWORD-padded allow ACE containing SID `S-1-15-3-1`, which resolves to `internetClient`.

src/backends/learning_mode/windows/src/etl_decode.rs
richiemsft: Removed the region-style comment.

src/backends/learning_mode/windows/src/etl_filter.rs
richiemsft: Changed “selected cursor” to “selected-event index,” which is the actual value being held when a mismatch occurs.

src/backends/learning_mode/windows/src/etl_filter.rs
richiemsft: Rust denies unnecessary parentheses here under `-D warnings`, so I extracted the multiline expression into the named boolean `is_supported_capability_event`. This makes the condition readable without suppressing the lint.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: Renamed these to `verbose_logging_provider_for_guid` and `verbose_logging_provider_guid` to state both the input/output direction and their purpose.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: The Windows GUID type can be formatted by crate traits, but this serialized field requires a stable uppercase, braced spelling. I retained the small formatter as `format_guid_braced_uppercase` and documented why it is intentionally independent of crate formatting behavior.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: These are trusted ETW schema property names, not free-form attacker spellings. I documented that matching is deliberately restricted to known normalized names and suffixes rather than fuzzy matching such as leetspeak, which could discard unrelated properties.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: Tightened the identity checks from substring matching to exact-name/suffix matching. Added a regression proving `AccountNamePolicy` is retained while a true `...UserName` field is redacted.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: Documented the boundary: sanitization is schema-independent, so every newly decoded property automatically passes through the same name/value checks. A new branch is needed only when a new schema introduces a genuinely new sensitive naming/value shape; the regression tests cover those classifiers.

src/backends/learning_mode/windows/src/extractors.rs
richiemsft: Yes, intentionally. Properties are sorted before this function, then entries after the 24-property bound are discarded deterministically. This bounds one signature; it does not discard the event occurrence or aggregate count. I updated the function documentation to say this directly.

src/backends/learning_mode/windows/src/tdh_decode.rs
richiemsft: `SAFETY:` is the standard Rust comment documenting why the immediately following `unsafe` operation satisfies its invariants; it is not a pre-assert. I made the invariant concrete here: the ETW callback supplies a non-null `EVENT_RECORD` valid for the synchronous decode, and only fixed-size header fields are borrowed.

src/core/learning_mode_core/src/analyze.rs
richiemsft: Clarified the rustdoc: the complete compact JSON serialization must be no larger than the method’s `max_bytes` argument.

src/core/learning_mode_core/src/analyze.rs
richiemsft: Agreed that “legacy” was misleading. Renamed the test to `analysis_payload_without_verbose_logging_defaults_summary`; it tests the current `#[serde(default)]` omitted-field behavior without claiming the protocol itself is legacy.

src/core/learning_mode_core/src/analyze.rs
richiemsft: The counters cannot wrap: all overflow accumulation uses `saturating_add`, so it clamps at `u64::MAX`. Added a focused serialized-size regression test that starts near `u64::MAX`, moves a group to overflow, and verifies both overflow counters saturate.

src/core/learning_mode_core/src/paired_output.rs
richiemsft: These are standard rustdoc Markdown headings. `# Errors` renders an Errors section in generated API documentation and records the conditions under which this `Result`-returning function fails, following the Rust API documentation convention. The other `#` heading serves the same rustdoc structure; no runtime behavior is involved.

src/host/plm/src/stop.rs
richiemsft: This is an exceptional but genuine caller-controlled collision: `--trace-file` can be set to the verbose sibling derived from `denials.json`. Publishing the verbose output would then overwrite the retained ETL before post-processing completed. I renamed the test and added a comment describing that exact scenario.

src/core/learning_mode_core/src/paired_output.rs
```diff
@@ -0,0 +1,998 @@
...
+        return Ok(None);
+    }
+
+    let backup_path = vacant_sibling_path(operation, final_path, kind)?;
+    persist_existing_file_noclobber(final_path, &backup_path).map_err(|error| {
+        std::io::Error::other(format!(
+            "{operation} failed to back up {kind} output file {}: {error}",
+            final_path.display()
+        ))
+    })?;
```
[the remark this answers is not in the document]
richiemsft: Fixed in `57c678c0`. `backup_existing_output` now captures the opened file identity and SHA-256 digest before moving the path, verifies the quarantined backup against that identity, and aborts while restoring the moved file when they differ. Added a race-injection regression proving a concurrent replacement is restored rather than claimed or deleted.

richiemsft:
@bbonaby, could you take a look?

bbonaby (approved):

src/core/learning_mode_core/src/paired_output.rs
```diff
@@ -0,0 +1,1068 @@
...
+    destination_path: &Path,
+) -> std::io::Result<RelocationOutcome> {
+    if source_path == destination_path {
+        return Ok(RelocationOutcome::default());
+    }
+
+    let mut source = std::fs::File::open(source_path)?;
+    let source_identity = PromotedOutput::from_file(source.try_clone()?)?;
+    ensure_output_absent(operation, destination_path, kind)?;
+    let temp = stage_output_file(operation, destination_path, kind, |writer| {
```
bbonaby, line 331: issue (non-blocking): Could we preserve a no-clobber rename-first fast path here when the platform supports it, then fall back to this copy-and-verify flow? Retained ETLs can be large, and this now copies the whole trace and hashes it multiple times even for a same-volume move that used to be an O(1) rename.

src/core/learning_mode_core/src/verbose_logging.rs
```diff
@@ -0,0 +1,529 @@
...
+    pub aggregate_groups_truncated: bool,
+    /// Whether the ETL processed-event bound prevented complete accounting.
+    pub processed_events_truncated: bool,
+    /// Whether the canonical unique-denial bound was reached.
+    pub canonical_denial_limit_reached: bool,
+}
+
+impl VerboseLoggingSummary {
+    /// Records one event outcome while preserving deterministic group order.
+    pub fn record(&mut self, signature: VerboseLoggingSignature) {
```
bbonaby, line 122: issue (non-blocking): Could we remove this second recording path or make it a thin wrapper around `record_with_byte_budget`? Production only uses the byte-bounded method, while this public version duplicates the state machine with weaker limits, which makes it easy for a future caller to bypass the WPR frame-size guard by accident.

bbonaby: @Richiemsft lets follow up via GH issues

src/core/learning_mode_core/src/paired_output.rs
bbonaby: @Richiemsft lets follow up via GH issues

PR #925, opened 2026-08-15
Carry the failing API's detail through the in-process SDKs

### LINKED ISSUE #924: In-process SDKs lose the structured failure detail the backend already produces
### Relevant area(s)

Windows

*(The conversion that drops the detail is cross-platform, in `src/core/mxc_engine/src/error.rs`. The observable impact is Windows today because IsolationSession is currently the only backend that produces the detail — every `ApiFailure` producer lives in `src/backends/isolation_session/common/src/error.rs`.)*

When a platform API call fails, the backend records **which** call failed, its **native status**, and sometimes a **remediation hint**. These reach the TypeScript SDK, which documents them as part of its error envelope along with the invariant that `nativeCode` implies `operation` (`sdk/node/src/errors.ts:27-30`, `:54-56`, `:83-84`).

The **in-process** path does not get them. `wxc_common::mxc_error::MxcError` carries them in `api_failure: Option<Box<ApiFailure>>`, but the public SDK error type is a facade over it — `mxc_engine::error::Error` — which holds only `{ code, message }`. Its `From<MxcError>` conversion (`src/core/mxc_engine/src/error.rs`) copies the code and message and drops `api_failure` and `details` on the floor.

Everything downstream of that conversion is therefore blind to the detail: the Rust SDK (`mxc-sdk`), the C ABI over it (`mxc_ffi`, which mentions none of the three field names), and the C# SDK's `MxcException`, which exposes only `Code` and `Message`.

The result is that two callers observing the **same** failure get materially different diagnoses depending on which SDK they use, with the in-process callers getting strictly less.

### Steps to reproduce

1. On a Windows host running the OS-side IsolationSession service, drive a state-aware lifecycle call that fails **inside** the OS API rather than in MXC's own validation — for example, stopping a session that was never provisioned.
2. Do it through the **TypeScript** SDK and read the thrown error's `operation` / `nativeCode` fields.
3. Do the same through the **Rust** SDK (`mxc_sdk::Error`) or the **C#** SDK (`MxcException`), and try to read the equivalent.

### Expected behavior

Both paths identify the failing call and its platform status — e.g. operation `IsoSessionOps.StopSessionAsync` with native code `0x80070490` — so an in-process caller can log, group, or branch on the same information the TypeScript SDK already exposes.

### Actual behavior

Only the TypeScript SDK reports them. The Rust, C ABI and C# surfaces report a bare code and message — e.g. `backend_error: The provision was not found.` — with no way to learn which call failed or why. The information exists at the point of failure and is discarded one layer later, so it cannot be recovered by the caller.

adpa-ms:
## 📖 Description

When a platform API call fails, the backend records **which** call failed, its **native status**, and sometimes a **remediation hint**. The TypeScript SDK already surfaces all three. In-process callers get none of it.

`wxc_common::mxc_error::MxcError` carries that detail in `api_failure`, but the public SDK facade `mxc_engine::error::Error` holds only `{ code, message }` and its `From<MxcError>` drops `api_failure` on the floor. Everything below is blind: the Rust SDK, the C ABI, and the C# `MxcException`. Two callers observing the *same* failure get materially different diagnoses — in-process callers see `backend_error: The provision was not found.` and no way to learn which call failed.

This carries the detail through all three layers.

- **The engine facade** carries the three flat on `Error`, beside `code` and `message`, matching the wire envelope, the C ABI, the C# binding and the TypeScript SDK. Nesting them behind a sub-struct would make the Rust SDK the one surface of five shaped differently.
- **The C ABI** gains one `MxcErrorDetail` used by every failing surface — `MxcRunResult`, `MxcStateAwareResult`, and the `out_error` parameter — so a binding learns the same things from any of them and frees them all the same way.
- **The C# binding** surfaces `Operation` / `NativeCode` / `Remediation` on `MxcException`. A null pointer becomes `null`, never `""`; null means the layer below supplied nothing.

The Rust SDK gains structured errors too, not only C#.

### Notes for reviewers

- **`out_error` changes ownership shape**, from a library-allocated string to caller-provided storage for one detail. The detail owns heap strings with no destructor, so a caller passing null would leak them; `finish_spawn` frees it on that path.
- **The field-coupling claims are removed.** The docs said `nativeCode` and `remediation` do not appear without `operation`. This decouples `remediation`, so the claims no longer hold — including in the TypeScript SDK and both wire docs.
- **`Error` is `#[non_exhaustive]`**, mirroring the wire envelope and `MxcError`. Adding it later is a breaking change; removing it is not. `ErrorCode` stays closed, mirroring `MxcErrorCode`.
- **The codegen gate gains a parity assertion.** csbindgen's sources and the C# project's MSBuild `Inputs` name the same files; adding to one and not the other compiles against stale bindings.
- **`mxc_version` is the one C ABI entry point not `catch_unwind`-wrapped**, contradicting the module header. Now wrapped, 23 of 23.
- Growing `REQUIRED_ENTRY_POINTS` to the full P/Invoke surface belongs with the C# workstream: the compiler already catches rename and removal.
- The C ABI is **not** a stable external contract — native and binding are co-versioned and generated together — so reshaping the `#[repr(C)]` types is expected.

## 🔗 References

Resolves #924

## 🔍 Validation

- **Local gauntlet, 23 checks green**, plus `wxc_host_prep` (16 tests) run elevated: fmt, clippy `-D warnings`, workspace tests with the isolation-session feature on and off, arm64 cross-build, six versioning gates, both C# SDK gates, `dotnet test` (47), and the Node build, unit, pack and integration legs.
- **Isolation-session E2E** on a VM running the OS-side service, from a package built at this commit: **90 passed, 0 failed, 0 skipped** — 16 one-shot, 62 state-aware, 12 SDK-integration — confirmed against the JUnit output. The agent-account leak check came back empty against a pre-run baseline, and `IsoSessionCli list-users` agreed at `Found 0 agent user(s)`.
- **The three interactive TTY tests** were run by an operator at the VM console and passed.
- `ErrorCode` parity is unaffected; this adds no new `MXC_STATUS_*` codes.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/925)


sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -125,16 +125,13 @@ public static MxcSandboxProcess ExecInSandbox(SandboxId id, string command)
...
                 {
-                    var message = PtrToString(error) ?? "unknown error";
-                    if (error is not null)
-                    {
-                        NativeMethods.mxc_string_free(error);
-                    }
-                    throw new MxcException((ErrorCode)status, message);
+                    var ex = NativeError.ToException(status, error, "unknown error");
+                    NativeMethods.mxc_error_detail_free(&error);
+                    throw ex;
```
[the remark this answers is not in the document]
adpa-ms: Fixed in a4fd7b4a, same shape as `MxcSandbox.Spawn`: the release moved into a `finally` so a throw during marshalling or exception construction cannot strand the detail's strings.

sdk/dotnet/Microsoft.Mxc.Sdk/Native/NativeError.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Mxc.Sdk.Native;
+
+/// <summary>
+/// Turns the native failure detail into an <see cref="MxcException"/>.
+/// </summary>
+/// <remarks>
+/// Shared rather than repeated per call site so every entry point reports a
+/// failure the same way: the three API fields either all cross or all stay
+/// null, and absence stays distinguishable from the empty string.
```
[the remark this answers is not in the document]
adpa-ms: Fixed in a4fd7b4a. The remark was not merely misleading, it was contradicted by this change's own test: `absent_optional_fields_stay_null_rather_than_empty` pins the operation-only shape as supported. It now describes per-field preservation, which is what `ToException` actually does -- each field marshalled independently through `ToStringOrNull`.

scripts/check-dotnet-bindings-codegen.js
```diff
@@ -72,6 +72,118 @@ if (missing.length > 0) {
...
+      "\n\nAdd each to the GenerateNativeBindings `Inputs` in\n" +
+      "  sdk/dotnet/Microsoft.Mxc.Sdk/Microsoft.Mxc.Sdk.csproj"
+  );
+  process.exit(1);
+}
+
 console.log(
-  `C# bindings codegen OK: generated with ${REQUIRED_ENTRY_POINTS.length} expected entry points`
+  `C# bindings codegen OK: generated with ${REQUIRED_ENTRY_POINTS.length} expected entry points; ` +
+    `${csbindgenInputs.length} csbindgen source(s) all declared as MSBuild inputs`
```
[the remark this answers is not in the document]
adpa-ms: Fixed in a4fd7b4a -- `mxc_error_detail_free` is now asserted, and the gate reports 5 entry points. One correction for the record: the list is a smoke-test subset rather than the full P/Invoke surface, and the C# compiler is the real backstop for rename and removal, because csbindgen emits a binding only for a fn carrying `#[no_mangle]` or `#[export_name]` and takes the `EntryPoint` from whichever determines the export. Growing it to the full set belongs with the dedicated C# workstream.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandbox.cs
```diff
@@ -132,16 +131,13 @@ public static MxcSandboxProcess Spawn(SandboxPolicy policy, string command)
...
                 {
-                    var message = PtrToString(error) ?? "unknown error";
-                    if (error is not null)
-                    {
-                        NativeMethods.mxc_string_free(error);
-                    }
-                    throw new MxcException((ErrorCode)status, message);
+                    var ex = NativeError.ToException(status, error, "unknown error");
+                    NativeMethods.mxc_error_detail_free(&error);
+                    throw ex;
```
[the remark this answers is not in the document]
adpa-ms: Fixed in a4fd7b4a. The release is in a `finally` now, so ownership is discharged however the block is left. Worth noting the earlier shape was right about the question two independent reviewers had asked of these lines -- copy-before-free ordering -- and wrong about the one you asked, which is what happens if `ToException` throws before reaching the free. Different question, same code.

src/core/mxc_engine/src/error.rs
```diff
@@ -77,11 +107,61 @@ pub struct Error {
...
     /// A human-readable message.
     pub message: String,
+    /// The failing API call and its platform status, when the failure came
+    /// from one. Absent for errors raised before any API was reached — a
+    /// malformed request, say.
+    ///
+    /// Boxed for the same reason the internal type boxes it: this rides in the
+    /// `Err` arm of every fallible SDK call, and inlining ~72 bytes of mostly
+    /// absent detail would widen every `Result` in the API.
+    pub api_failure: Option<Box<ApiFailure>>,
```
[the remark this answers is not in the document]
adpa-ms: Declining this one, with the evidence for why.

The concern is right in general: adding a field to a struct with public fields breaks any downstream `Error { .. }` literal. It does not apply here yet, because there is no downstream to break.

Neither `mxc-sdk` nor `mxc_engine` has ever shipped. `git ls-tree -r --name-only v0.7.0`, and the same against `v0.7.0-rc1`, both return zero files under either crate — both were added after the most recent release tag, and no published artifact carries this type. Within the repo, the only cross-crate sites that construct `Error` by literal are the tests this change itself adds in `src/ffi/mxc_ffi/src/error_detail.rs`; every other construction goes through `Error::new`. So the set of code this can currently break is empty.

Worth stating explicitly, since it is the obvious follow-up: `#[non_exhaustive]` is the standard way to stop this becoming breaking later, and it was considered in detail rather than overlooked. The audit found the repository has no established convention to follow — three hand-written uses in the entire tree, from two unrelated PRs, neither commit message mentioning it. It is close to free on output-only types like `Error` and `ApiFailure`, but would remove struct-literal construction (including `..Default::default()`) on input types like `SandboxPolicy`, so a blanket sweep would be wrong. Since the crate is pre-first-publish the decision is cheap now and expensive later, which makes it a call worth putting to a human reviewer rather than making as a drive-by. It is now called out in the PR description for exactly that.

If a reviewer wants `#[non_exhaustive]` on `Error` and `ApiFailure`, say so and it goes in — it is a small change plus a test update, and strictly easier now than after the first release.

ionvasilian (dismissed):

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/NativeErrorTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            operation_utf8 = Utf8(operation),
+            native_code_utf8 = Utf8(nativeCode),
+            remediation_utf8 = Utf8(remediation),
+        };
+
+    /// <summary>
+    /// Releases what <see cref="Utf8"/> allocated. Freeing a null pointer is a
+    /// no-op, so every field can be released unconditionally.
+    /// </summary>
+    private static void Free(MxcErrorDetail detail)
```
jeffstall, line 60: I'm concerned that the C# caller would never need to call Marshal.FreeCoTaskMem. Normally, the Marshalling should be automatically declared so this can happen by the P/Invoke machinery. Why is this needed?

jeffstall (changes requested):

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/NativeErrorTests.cs
adpa-ms: These pointers were allocated by the test, so the test frees them. They never reach the native free.

The test fabricates the struct because the native layer only fills `operation`/`nativeCode`/`remediation` when a real platform API fails, which needs a prepared host and backend that a unit test has no way to obtain.

Declared marshalling would not help: `MxcErrorDetail` is generated by csbindgen from the Rust FFI (gitignored, regenerated on every build), so its fields are not ours to attribute, and it must stay unmanaged for its address to be taken — which managed `string` fields would prevent.

Product code never calls `FreeCoTaskMem`. The read is automatic (`Marshal.PtrToStringUTF8`, which copies and takes no ownership); the release crosses back — `MxcSandbox.Spawn` and `MxcLifecycle.ExecInSandbox` pass `&error` to `mxc_error_detail_free`, since Rust allocated with `CString::into_raw` and only its allocator can free it.

jeffstall (dismissed):

sdk/dotnet/Microsoft.Mxc.Sdk/MxcException.cs
```diff
@@ -8,18 +8,90 @@ namespace Microsoft.Mxc.Sdk;
...
+        string message,
+        string? operation,
+        string? nativeCode,
+        string? remediation)
         : base(message)
     {
         Code = code;
+        Operation = operation;
+        NativeCode = nativeCode;
+        Remediation = remediation;
```
bbonaby, line 62: issue (blocking): Preserve the standard exception diagnostics when adding the MXC fields.

This override replaces `Exception.ToString()` with only the code/message/API fields. Any normal exception logger now loses the managed stack trace, inner exception, and exception type; it also drops the new `Remediation` hint. The Rust `Error::fmt` implementation has the same remediation omission, and its `a_remediation_without_an_operation_renders_without_brackets` test currently locks that loss in.

Could we keep `base.ToString()` as the standard representation and append a separate MXC detail/remediation line, then also append remediation in Rust's `Display` implementation? The tests should verify that a thrown exception still renders its stack/inner exception and that remediation survives formatted logging on both SDK surfaces.

sdk/dotnet/Microsoft.Mxc.Sdk/Microsoft.Mxc.Sdk.csproj
```diff
@@ -36,7 +36,7 @@
 
   <Target Name="GenerateNativeBindings"
           BeforeTargets="CoreCompile"
-          Inputs="$(MxcSrcDir)/ffi/mxc_ffi/src/lib.rs;$(MxcSrcDir)/ffi/mxc_ffi/src/streaming.rs;$(MxcSrcDir)/ffi/mxc_ffi/src/state_aware.rs;$(MxcSrcDir)/ffi/mxc_ffi/build.rs"
+          Inputs="$(MxcSrcDir)/ffi/mxc_ffi/src/lib.rs;$(MxcSrcDir)/ffi/mxc_ffi/src/error_detail.rs;$(MxcSrcDir)/ffi/mxc_ffi/src/streaming.rs;$(MxcSrcDir)/ffi/mxc_ffi/src/state_aware.rs;$(MxcSrcDir)/ffi/mxc_ffi/build.rs"
```
bbonaby, line 39: issue (blocking): Ensure the managed bindings cannot load a stale pre-change native ABI.

This PR changes the C layouts from one `error_utf8` pointer to an embedded four-pointer `MxcErrorDetail` (and changes the streaming `out_error` type). `GenerateNativeBindings` rebuilds `src/target/debug/mxc_ffi.dll`, but it does not refresh a DLL previously staged by `build.bat` under `runtimes/<rid>/native`. `NativeLibraryResolver` checks the copied runtime asset before the fresh Cargo target, so after a branch switch an incremental `dotnet test` can compile the new managed layout and load the old native layout. Reading/freeing the result then uses incorrect offsets and can corrupt the heap.

Could this target also refresh the staged runtime DLL, or could the resolver prefer the Cargo target for repository dev layouts? An ABI/version check or a test with a deliberately stale runtime asset would also make this fail safely instead of loading incompatible layouts.

bbonaby (commented):
Two serious issues found in the Rust/adversarial review.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcException.cs
```diff
@@ -8,18 +8,90 @@ namespace Microsoft.Mxc.Sdk;
 /// <see cref="ErrorCode"/> from the native layer; <see cref="Exception.Message"/>
 /// carries the human-readable detail.
 /// </summary>
+/// <remarks>
+/// <para>
+/// When the failure came from an underlying platform API, <see cref="Operation"/>
+/// names the call and <see cref="NativeCode"/> carries its status. Both are
+/// <see langword="null"/> for failures raised before any API call was reached —
```
shschaefer: This is not quite right.  When looking through the code, there seem to be paths which can violate this and produce MxcErrorDetail...

src/ffi/mxc_ffi/src/error_detail.rs
```diff
@@ -0,0 +1,252 @@
...
+            message_utf8: alloc_cstring(message.into().as_bytes()),
+            ..Self::none()
+        }
+    }
+
+    /// The full detail from an SDK error, carrying the API call across when the
+    /// error names one.
+    pub(crate) fn from_error(error: &Error) -> Self {
+        Self {
+            message_utf8: alloc_cstring(error.message.as_bytes()),
```
shschaefer, line 70: What happens on failure here?  Not RAII.

src/ffi/mxc_ffi/src/lib.rs
```diff
@@ -296,7 +315,7 @@ fn run_inner(policy_json_utf8: *const c_char, command_utf8: *const c_char) -> Mx
 
     let mut request = match build_request(&policy, None) {
         Ok(r) => r,
-        Err(e) => return MxcRunResult::error(status_from_error_code(e.code), e.message),
+        Err(e) => return MxcRunResult::from_sdk_error(&e),
```
shschaefer, line 318: A righteous change.  No tests prove it.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcException.cs
```diff
@@ -8,18 +8,90 @@ namespace Microsoft.Mxc.Sdk;
...
+    internal MxcException(
+        ErrorCode code,
+        string message,
+        string? operation,
+        string? nativeCode,
+        string? remediation)
         : base(message)
     {
         Code = code;
+        Operation = operation;
```
shschaefer, line 60: No validation or normalization?

shschaefer (changes requested):

sdk/dotnet/Microsoft.Mxc.Sdk/MxcException.cs
adpa-ms: Correct, nothing enforces it. The claims should not have been documented and are removed.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcException.cs
adpa-ms: Both are the backends' responsibility. These strings only ever come from Rust constants, never user input, and `operation_constants_are_qualified_and_parameter_free` asserts their shape.

src/ffi/mxc_ffi/src/error_detail.rs
adpa-ms: No failure path: `alloc_cstring` sanitizes interior NULs, the only input `CString::new` rejects, so it always returns non-null; allocation failure aborts rather than unwinds. The struct is one `Self { .. }` expression, so no partial value escapes.

`Drop` would be wrong here: `MxcErrorDetail` is `#[repr(C)]`, crosses by value in `MxcRunResult` and via `out_error`, and the caller owns and frees the strings after that. Freeing on the Rust side too is a double free.

Added a test pinning the non-null guarantee, which was untested.

src/ffi/mxc_ffi/src/lib.rs
adpa-ms: Added three tests: `from_sdk_error` carries the API detail, `mxc_run` reaches this arm on a failing `build_request`, and the same for `MxcStateAwareResult::from_sdk_error`.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcException.cs
adpa-ms: Deferring. `main` already rendered `{Code}: {Message}`, so the stack trace and type name predate this change, and `MxcException` has no inner exception constructor. Remediation is a public property here and a public field on the Rust error; it is folded into the string only where the string is the sole carrier, as `describe()` does on the one-shot path.

sdk/dotnet/Microsoft.Mxc.Sdk/Microsoft.Mxc.Sdk.csproj
adpa-ms: Confirmed, and taking the second option: the resolver now searches the Cargo target before the staged `runtimes/` asset. Its class doc already says the packaged case is the default loader's job and this resolver exists for dev/test, so the previous order was inverted.

adpa-ms: Reverted this. Reordering only moves the hazard: a staged dev asset and a packaged asset resolve to the same `runtimes/<rid>/native` path, so probe order cannot tell a stale one from a correct one. Preferring the Cargo target means a packaged consumer under a checkout loads that checkout's build instead of its own.

The fix belongs at the staging layer, having the build output carry the library the bindings were generated from, rather than in probe order. That is larger than this PR. The mechanism is pre-existing and not introduced here.

sdk/dotnet/Microsoft.Mxc.Sdk/Native/NativeLibraryResolver.cs
```diff
@@ -63,14 +63,16 @@ private static IEnumerable<string> CandidatePaths()
 
         var baseDir = AppContext.BaseDirectory;
         yield return Path.Combine(baseDir, file);
-        yield return Path.Combine(baseDir, "runtimes", RuntimeInformation.RuntimeIdentifier, "native", file);
 
-        // Dev layout: walk up looking for the Cargo target dir.
+        // Dev layout: walk up looking for the Cargo target dir. Searched before
+        // the staged runtimes/ asset, which a dev build does not refresh.
```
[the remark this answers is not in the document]
adpa-ms: Agreed, and reverted. The file is now unchanged from main in this PR. Probe order cannot distinguish a stale dev-staged asset from a correct packaged one, so preferring the Cargo target trades one mismatch for the other. Deferred to a build-layer fix.

shschaefer (approved):

PR #930, opened 2026-08-18
[Bubblewrap] Run proxy mode in a private network namespace via rootless slirp4netns (schema 0.8.0+)

SohamDas2021:
## 📖 Description

Moves Bubblewrap's proxy mode out of the host network namespace and into a **private network namespace**, routed by rootless `slirp4netns`, on schema `0.8.0-alpha`+. This is the foundation that per-flow egress enforcement is built on; enforcement itself lands in a follow-up PR.

Today `network.proxy` on Bubblewrap runs the sandbox in the **host's** network namespace. Even setting aside egress policy, that means the sandbox shares the host's interfaces, loopback, and localhost-bound services — there is no network boundary to enforce anything against. This PR creates that boundary, without requiring root or elevation.

### Structure

- **`f910778f` — centralize network mode selection.** Pure refactor, no behavior change. Bubblewrap previously decided "shared netns vs `--unshare-net`", "does this need iptables", and "is the cooperative proxy active" independently at three call sites. This collapses them into one `ResolvedNetworkMode` computed once per request (`LegacyProxy` / `ProxyOnly` / …) and removes the standalone `needs_iptables_rules()` helper. This is what makes the schema gate below a single decision point instead of three correlated booleans.
- **`229405ae` — private network namespace + slirp lifecycle**, behind the 0.8 gate.
- **`17f09e3d` — PR feedback.** Clears `FD_CLOEXEC` in the forked child rather than the parent, so the descriptors are not inheritable by concurrent spawns; adds legacy-path, capability-drop and orphan-reaping E2E cases; wires the proxy E2E into CI.

### How it works

For schema 0.8+ proxy requests:

1. MXC spawns a **supervisor** (`unshare --user --map-current-user --keep-caps`) — a helper process holding a user namespace in which it has `CAP_NET_ADMIN`, even though `lxc-exec` itself does not. This is what makes the design work unprivileged.
2. The supervisor starts **`slirp4netns`** to provide userspace routing into a fresh network namespace, and signals readiness only once routing is up.
3. `bwrap` joins the supervisor's user namespace (rather than creating its own with `--unshare-user`) and drops capabilities.
4. The configured proxy endpoint is translated for in-namespace reachability: a loopback proxy on the host becomes slirp's host gateway (`10.0.2.2`), preserving the original URL's scheme and credentials. Non-loopback endpoints are passed through unchanged.
5. Teardown stops slirp and reaps the supervisor; `Drop` covers the error paths.

### Limitations

- **Egress is not yet restricted.** Inside the namespace, `slirp4netns` routes all outbound traffic, so a workload that ignores `HTTP_PROXY`/`HTTPS_PROXY` can still reach the network directly. The proxy remains cooperative at this commit; the namespace is the boundary that makes enforcement *possible*. Deny-all-except-proxy (AB#62864253 §2) follows in the next PR.
- New host prerequisite for this mode: `slirp4netns`, plus a util-linux `unshare` supporting `--map-current-user` and `--keep-caps`. Both are probed at validate time and fail fast with an actionable message. Neither requires root. CI installs them and runs the E2E.
- Per-host firewall filtering (`network.hostRules`) is not wired into this mode; the existing proxy/firewall mutual exclusivity in the config parser still applies.

## 🔍 Validation

Automated:
- `cargo test -p bwrap_common` — pass, incl. 12 new tests (network-mode resolution across schema versions, proxy-address translation for loopback / non-loopback / URL-form endpoints, and supervisor lifecycle).
- `cargo clippy -p bwrap_common --all-targets -- -D warnings`, `cargo fmt --check` — clean.
- `node scripts/versioning/validate-configs.js`.

E2E — `tests/scripts/run_bwrap_network_proxy_test.sh`, **7/7 pass**. Runs in GitHub Actions (`Build.Linux.Job.yml`) against the real binary and real dependencies, and locally on Ubuntu 24.04 (WSL2) unprivileged, no sudo:
- builtin proxy, proxy allowlist, proxy blocklist.
- **private proxy namespace** — reads `/proc/self/ns/net` inside the sandbox and **fails if it matches the host's** identity, then confirms a proxied request completes. A silent fallback to the host netns fails the test rather than passes it.
- **capability drop** — asserts the workload does not retain the supervisor's user-namespace capabilities.
- **legacy (schema 0.7) compatibility** — exercises the legacy path with `slirp4netns` unavailable, proving 0.6/0.7 neither probe nor use it.
- **supervisor orphan reaping** — no leaked `slirp4netns` / `bwrap` processes after teardown.
- `run_bwrap_basic_test.sh`, `run_bwrap_network_test.sh` — pass.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/930)


src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -0,0 +1,433 @@
...
+set -eu
+state_dir="$1"
+ready_fd="$2"
+exit_fd="$3"
+printf ready > "$state_dir/userns.ready"
+while [ ! -s "$state_dir/child.pid" ]; do
+    sleep 0.01
+done
+child_pid="$(cat "$state_dir/child.pid")"
+exec slirp4netns --configure --mtu=65520 \
```
bbonaby, line 43: **issue (non-blocking):** Can we either restrict the slirp host gateway to the configured proxy or document the broader host-loopback exposure?

Because slirp runs without `--disable-host-loopback`, the workload can connect to any host service bound to loopback through `10.0.2.2:<port>`, not only the configured proxy. This is an important exception to the new private-network boundary and is more sensitive than generic direct internet egress.

**Suggested fix:** Prefer `--disable-host-loopback` plus a single-port forwarding mechanism for the proxy. If that belongs to the follow-up enforcement PR, explicitly document arbitrary host-loopback reachability and add a test that defines the intended contract.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -213,13 +223,65 @@ impl BubblewrapScriptRunner {
...
+            request,
+            proxy_address,
+            denied_files,
+            network_mode,
+        );
+        let mut network_startup = match proxy_network.as_ref() {
+            Some(network) => match network.configure_bwrap(&mut args) {
+                Ok(startup) => Some(startup),
+                Err(error) => {
+                    proxy_network.take();
```
bbonaby: **issue (non-blocking):** Could these startup error paths call `network.stop(logger)` instead of relying on `Drop`?

`proxy_network.take()` immediately drops the namespace object. Its `Drop` implementation uses a throwaway in-memory logger, so warnings about slirp requiring forced termination or failing shutdown inspection are lost exactly when startup is already failing. The same pattern occurs at the other startup cleanup sites below.

**Suggested fix:** Match `BwrapChild::cleanup`:

```rust
if let Some(mut network) = proxy_network.take() {
    network.stop(logger);
}
```

bbonaby (commented):
Adversarial Rust review: posting the two unique serious findings. Existing Copilot comments already cover inheritable FDs, the orphaned supervisor startup window, and automated lifecycle coverage, so I did not duplicate them.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -0,0 +1,433 @@
...
+            .map_err(|error| format!("Bubblewrap: failed to release workload startup: {error}"))
+    }
+}
+
+/// A same-UID user-namespace supervisor and its `slirp4netns` process.
+pub(crate) struct ProxyNetworkNamespace {
+    state_dir: TempDir,
+    supervisor: Child,
+    exit_writer: Option<OwnedFd>,
+    userns: File,
```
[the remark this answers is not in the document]
SohamDas2021: Fixed. Parent-side descriptors now stay `FD_CLOEXEC` throughout; `clear_cloexec` is deleted. `inherit_descriptors` registers a `pre_exec` closure that clears the flag inside the forked child, so only the intended child ever sees them. Applied at both spawn sites (the supervisor and bwrap). The parent's user-namespace handle is also dropped immediately after bwrap spawns, via `userns_handed_off()`, rather than being held for the sandbox lifetime.

Covered by a new unit test `inherited_descriptors_reach_only_the_prepared_child`: the prepared child sees the descriptor, an unrelated concurrently spawned child does not, and the parent's copy still has `FD_CLOEXEC`. Restoring the parent-side clear fails it.

One residual worth naming: bwrap's own inherited copy survives its `fork`/`exec` into the workload and upstream offers no way to close it. That is inert only because bwrap empties the capability bounding set, which the E2E suite now asserts; both are documented in the backend doc.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -0,0 +1,433 @@
...
+        let userns = match File::open(&userns_path) {
+            Ok(file) => file,
+            Err(error) => {
+                terminate_child(&mut supervisor);
+                return Err(format!(
+                    "Bubblewrap: failed to open proxy user namespace {userns_path}: {error}"
+                ));
+            }
+        };
+        clear_cloexec(userns.as_raw_fd())?;
```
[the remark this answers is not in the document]
SohamDas2021: Fixed structurally. `clear_cloexec` no longer exists, so that particular failure point is gone, and there is now no fallible `?` between the supervisor spawn and `Ok(Self)`. The two remaining fallible steps in that window (`wait_for_file` for the userns rendezvous, and `File::open` on `/proc/<pid>/ns/user`) each call `terminate_child`, which kills and reaps, before returning the error.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -0,0 +1,433 @@
...
+const SLIRP_HOST_GATEWAY: &str = "10.0.2.2";
+const SUPERVISOR_SCRIPT: &str = r#"
+set -eu
+state_dir="$1"
+ready_fd="$2"
+exit_fd="$3"
+printf ready > "$state_dir/userns.ready"
+while [ ! -s "$state_dir/child.pid" ]; do
+    sleep 0.01
+done
```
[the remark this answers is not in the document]
SohamDas2021: Fixed. The supervisor now blocks on a parent-owned pipe instead of polling for `child.pid`, exactly as suggested: the parent holds the write end, so if it dies the read hits EOF and the supervisor exits 1 with a diagnostic instead of spinning as an orphan. The `child.pid` file is gone entirely (it also had a partial-read race between `[ -s ]` and `cat`).

Covered by a new end-to-end case `supervisor orphan-reaping`: a stub `bwrap` holds the executor in its startup wait to widen the window from microseconds to seconds, the executor is SIGKILLed inside it, and the test asserts the supervisor is gone. Negative-controlled by making the supervisor spin on EOF, which fails the case.

tests/scripts/run_bwrap_network_proxy_test.sh
```diff
@@ -41,4 +41,30 @@ run_one "builtin proxy"    "bubblewrap_network_proxy_builtin.json"    "PROXY_OK"
 run_one "proxy allowlist"  "bubblewrap_network_proxy_allowlist.json"  "BLOCKED_OK"
 run_one "proxy blocklist"  "bubblewrap_network_proxy_blocklist.json"  "BLOCKED_OK"
 
+echo "Running Bubblewrap private proxy namespace test..."
+HOST_NETNS="$(readlink /proc/self/ns/net)"
+if ! NAMESPACE_OUT=$("$LXC_EXEC" --experimental --allow-testing-features \
+    "$REPO_DIR/tests/configs/bubblewrap_network_proxy_namespace.json" 2>&1); then
```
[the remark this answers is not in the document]
SohamDas2021: Fixed. `.github/workflows/Build.Linux.Job.yml` now runs `tests/scripts/run_bwrap_network_proxy_test.sh` as a real CI step (`iptables` added alongside the existing `bwrap`/`slirp4netns` installs), so the schema-0.8 supervisor/slirp attach and readiness lifecycle is exercised on every run rather than only by hand.

The suite grew from 3 cases to 7 in the process: private-namespace identity, capability drop, schema-0.7 legacy compatibility (with a deliberately poisoned `slirp4netns` to prove the legacy path neither probes nor uses it), and supervisor orphan reaping.

src/backends/bubblewrap/common/src/proxy_network.rs
SohamDas2021: fixed

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: fixed

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -0,0 +1,600 @@
...
+
+use nix::fcntl::{fcntl, FcntlArg, FdFlag, OFlag};
+use nix::unistd::pipe2;
+use tempfile::TempDir;
+use wxc_common::logger::Logger;
+use wxc_common::models::ProxyAddress;
+
+const STARTUP_TIMEOUT: Duration = Duration::from_secs(5);
+const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(2);
+const SLIRP_HOST_GATEWAY: &str = "10.0.2.2";
```
bbonaby, line 23: question (non-blocking): Just so I completely understand the network traffic:

```
1. An app inside the container sends a request to 10.0.2.2:8080.
2. The Slirp engine intercepts these raw network packets in user space
3. Slirp strips away the virtual 10.0.2.2 IP header.
4. Slirp uses standard unprivileged system calls to re-transmit that data directly to 127.0.0.1:8080 on the host
```

Does this look like what should be happening?

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -0,0 +1,600 @@
...
+    info_reader: File,
+    info_writer: Option<OwnedFd>,
+    gate_reader: Option<OwnedFd>,
+    gate_writer: Option<OwnedFd>,
+    /// Descriptors bwrap must inherit, cleared of `FD_CLOEXEC` in the child
+    /// only. See [`inherit_descriptors`].
+    inheritable: Vec<RawFd>,
+}
+
+impl BwrapStartup {
```
bbonaby, line 59: Note: Now that the Rust SDK is in-proc, we should consider whether operations like polling need to be asynchronous. This is likely fine if public APIs such as `SpawnSandbox` and `Run` expose async variants, allowing callers to avoid blocking while our internal implementation remains synchronous.

.github/workflows/Build.Linux.Job.yml
```diff
@@ -119,6 +119,17 @@ jobs:
...
+      # slirp4netns supervisor, joins its user namespace, and asserts the
+      # sandbox lands in a private network namespace. Needs unix-test-proxy
+      # (builtinTestServer) alongside lxc-exec, which the two build steps
+      # above place in the same target directory.
+      - name: Test Bubblewrap proxy networking (end-to-end)
+        working-directory: ${{ github.workspace }}
+        env:
+          LXC_EXEC: ${{ github.workspace }}/src/target/${{ matrix.target }}/release/lxc-exec
+        run: bash tests/scripts/run_bwrap_network_proxy_test.sh
+
```
bbonaby, line 132: note: FYI @theelliotm you might need to add slirp4netns  to your prepare linux host script.

bbonaby (approved):

src/backends/bubblewrap/common/src/proxy_network.rs
SohamDas2021: Yes, your understanding is mostly correct, few things to note though.

1. MXC arranges this: sandbox_proxy_address rewrites a loopback/unspecified proxy host (localhost, 127.0.0.1,  0.0.0.0, [::]) to 10.0.2.2, preserving port and URL. Remote proxies pass through untouched.
2. Not interception, slirp owns the other end. slirp4netns creates tap0 in the sandbox netns and reads frames off the tap fd; the default route points at 10.0.2.2 .
3. Stronger than header-stripping: libslirp is a userspace TCP/IP stack that terminates the connection and originates a new one host-side. Original packets are never forwarded, which is why no elevation is needed.

src/backends/bubblewrap/common/src/proxy_network.rs
SohamDas2021: Agreed, how should we track this?

PR #931, opened 2026-08-18
[Bubblewrap] Enforce proxy-only egress for private networking

SohamDas2021:
## 📖 Description

> **Stacked on #930** (`user/sodas/bwrap-network-model2`).

The base PR moves schema-0.8+ Bubblewrap proxy sandboxes into a private network namespace, but `slirp4netns` routes *everything* out of it — so the proxy is still only cooperative. A workload that ignores `HTTP_PROXY`/`HTTPS_PROXY`, or opens a raw socket, reaches the network directly. This PR makes the boundary real: inside the namespace, **the proxy endpoint is the only reachable destination**. Everything else is dropped, IPv4 and IPv6.

### How it works

The supervisor (already introduced in the base PR) gains firewall responsibility:
1. `slirp4netns` moves from `exec` to a backgrounded child with its own **internal** readiness fd, so the supervisor regains control once routing is up.
2. The supervisor installs an `MXC_EGRESS` chain via `nsenter --net=/proc/<pid>/ns/net`, hooked from `OUTPUT`: accept `lo`, accept `tcp -d <proxy-ip> --dport <proxy-port>`, **DROP everything else**. The IPv6 chain accepts `lo` and drops the rest.
3. Only **after** every rule is installed does it write the external ready signal that releases the workload. **The ordering is the security property** — the sandbox can never observe an unenforced network. `set -e` means a failed rule kills the supervisor rather than proceeding.

`bwrap` joins the supervisor's user namespace and drops capabilities, so the workload sits inside the namespace that owns its own firewall but cannot modify it. That assumption is load-bearing, so it is now both documented and asserted by a test (below).

### DNS is deliberately closed

No `:53` accept rule is installed. An unscoped port-53 allowance is a standing DNS-tunnel exfiltration path, and this mirrors the policy already shipped for LXC model 2. Name resolution happens through the proxy.

### Limitations

- Proxy endpoints must be a **literal IPv4 address + port**. A firewall rule needs an address, and resolving a hostname at rule-install time is a TOCTOU hole. Hostname and routable IPv6 endpoints are rejected at validation with an actionable error; `localhost` / `127.0.0.0/8` / `0.0.0.0` / `::` are rewritten to slirp's gateway, while `::1` is rejected (an IPv6-loopback listener cannot accept the gateway's IPv4 connection). **This is a narrowing of what the base PR accepts** — hostname support requires porting LXC's `ProxyHostPin` (already present in
  `wxc_common::models`) and is the immediate follow-up.
- New host prerequisites for this mode: `nsenter`, `iptables`, `ip6tables`, added to the existing validate-time probe. None require root. `iptables`/`ip6tables` must also resolve to the **`nf_tables` backend** unless `/run/xtables.lock` is writable — the legacy backend opens that lock before touching any table, which an unprivileged same-uid supervisor cannot do on a stock host. `validate` refuses such a host rather than letting the supervisor die at the first rule.
- The 9 rule installations are serialized `nsenter` + `iptables` spawns, each passing `-w`, under a dedicated rule-installation budget rather than the 5s startup timeout. Batching them via `iptables-restore` was considered and dropped: `-w` plus the separate budget closes the correctness half, leaving a latency refactor that would trade per-rule failure attribution for one opaque batch failure.

Follow-ups filed: #935 (slirp liveness during the run), #936 (`AF_VSOCK` and other non-IP families), #937 (`network.proxy` enforced on LXC/bwrap-0.8 but cooperative on Seatbelt/WSLc).

## 🔍 Validation

Automated:
- `cargo test -p bwrap_common` — 112 pass, incl. 37 new tests covering endpoint validation (translation, `::1`, hostname/IPv6, port-0, legacy-schema untouched, `builtinTestServer` exemption), iptables backend selection, and supervisor behaviour. The supervisor tests **execute** the script under a real `sh` with stubbed `nsenter`/`slirp4netns` — rule ordering, readiness-written-last, every rule position failing ⇒ no readiness, slirp dying early, and the exact ACCEPT set (which fails if a `:53` rule is ever added).
- `cargo clippy -p bwrap_common --all-targets -- -D warnings`, `cargo fmt --check` — clean.
- `node scripts/versioning/validate-configs.js` — 233 configs.

Manual E2E on Ubuntu 24.04 (WSL2), unprivileged, no sudo:
- `tests/scripts/run_bwrap_network_proxy_test.sh` — 8/8 pass. The `bubblewrap_network_proxy_egress_denied.json` case asserts seven properties in one run: the proxy is reachable over a raw socket **first** (a positive control, so a blocked direct connection is attributable to the DROP rule rather than a dead network), direct egress blocked, loopback still exempt, `CAP_NET_ADMIN` dropped, an in-sandbox flush refused with a non-zero status, tampering ineffective, and the proxy still reachable at the end (so an offline host fails rather than passing vacuously).
- No leaked `slirp4netns` / `bwrap` processes after teardown.

Every new invariant was checked non-vacuous by injecting the matching regression (`|| true` on a rule, a swallowed rule failure, readiness moved ahead of the rules, the backend check stubbed to pass) and confirming the test fails. The first two are caught only by the executing supervisor tests — the text-based ones stayed green against an unenforced sandbox.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
---

for the steps.


docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -285,8 +315,19 @@ request fails if its private namespace cannot be configured.
...
+   DROP-only chain). The workload is released only after every rule is
+   installed, so it can never run with egress open. A failure to program any
+   rule aborts the supervisor rather than starting an unenforced sandbox.
+
+   Bubblewrap joins the supervisor's user namespace (`--userns`) rather than
+   creating its own, so the sandbox lives in the namespace that owns the
+   rule-bearing network namespace. This relies on Bubblewrap dropping
+   capabilities in the sandboxed process — the runner passes no `--cap-add` —
+   which is what prevents the workload from holding the `CAP_NET_ADMIN` needed
+   to flush the chain.
```
bbonaby, line 330: thought: nice! I do something similar for Windows in AppInfoSvc. That said, I'd be interested in when we clean up things. In Windows we clean up the Windows Filtering Platform rules I add once all processes in the Job object has gone away.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -111,15 +111,37 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+        // learns at policy time instead of after a namespace, a slirp instance
+        // and a rule set have already been built and torn down.
+        //
+        // The builtin test server has no address until it is started (the
+        // parser leaves a port-0 placeholder), so its endpoint stays on the
+        // runtime check in `run`; this only covers an operator-supplied proxy.
+        let proxy_only =
+            ResolvedNetworkMode::from_request(request, request.policy.network_proxy.is_enabled())
+                == ResolvedNetworkMode::ProxyOnly;
+        if proxy_only && !request.policy.network_proxy.builtin_test_server {
```
bbonaby, line 129: note: In 0.8 networking policy we no longer have the `builtin_test_server` property. It was only used by us for testing, so it being in the schema was bad on my part. Also now that we're going into a world of in proc SDKs the `builtin_test_server` won't actually do anything since there won't be a test proxy exe.

What this means for us is that for our testing we'd spin up the exe first either manually or via a test script, then run our test scenario.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -345,13 +490,28 @@ impl Drop for ProxyNetworkNamespace {
...
 pub(crate) fn sandbox_proxy_address(address: &ProxyAddress) -> Result<ProxyAddress, String> {
     let host = address.host().trim_matches(['[', ']']);
+    let parsed = host.parse::<std::net::IpAddr>().ok();
+
+    // slirp's gateway reaches the host's *IPv4* loopback only. A proxy bound
+    // to `::1` listens on the IPv6 loopback exclusively, so rewriting it to
+    // the gateway would hand the sandbox an address nothing answers on --
+    // failing at connect time rather than at policy time. Reject it instead.
+    if matches!(parsed, Some(std::net::IpAddr::V6(ip)) if ip.is_loopback()) {
+        return Err(format!(
```
bbonaby, line 505: thought: thinking about this some more should we have specific errors related to this proxy_network.rs file? Or is the string in the ScriptResponse enough. Also wondering if any of this code can be reused for lxc or if we're doing something different in there.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -98,8 +98,8 @@ pub(crate) enum ResolvedNetworkMode {
     LegacyProxy,
     /// The host network namespace with per-destination iptables filtering.
     FirewallFiltered,
-    /// Cooperative proxy routing inside a slirp-backed private namespace.
-    /// Later work adds proxy-only egress enforcement.
+    /// Cooperative proxy routing inside a slirp-backed private namespace, with
+    /// egress closed to everything but the proxy.
```
bbonaby, line 102: issue: The existing config warning now contradicts this schema 0.8+ guarantee.

`config_parser.rs:1314-1329` still warns every Bubblewrap proxy + `defaultPolicy='block'` request that raw-socket clients bypass the proxy and reach the host network. Its condition does not check the schema version, while schema 0.8+ resolves this case to `ProxyOnly` and installs the default-DROP chain described here.

Could we scope that warning to the legacy proxy path (schema 0.6/0.7), or otherwise suppress it for schema 0.8+ `ProxyOnly` requests?

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -3,35 +3,84 @@
...
+/// Number of `iptables`/`ip6tables` calls the supervisor makes. Only used to
+/// size the rule-installation budget, so it is asserted against the script.
+const RULE_COMMAND_COUNT: u32 = 9;
+/// Budget for the rule-installation phase, which begins once slirp is up.
+///
+/// Sized so that lock contention -- the expected reason a rule stalls -- cannot
+/// exhaust the budget before `-w` has had a chance to succeed. A single
+/// [`STARTUP_TIMEOUT`] cannot cover this: it would reject a viable sandbox the
+/// moment one call waited on a busy host.
+const RULE_INSTALL_TIMEOUT: Duration =
```
bbonaby: issue: The rule-installation deadline has no margin for the rest of this phase.

`RULE_INSTALL_TIMEOUT` is exactly 9 × the 5-second xtables lock wait, but the same deadline also covers slirp readiness, nine `nsenter` process launches, and the rule execution time. On a contended host, commands can legitimately consume most of their `-w` allowance while the supervisor is still making progress, so this can reject a viable sandbox just before completion.

Could we add explicit headroom for startup/process overhead, or give slirp readiness and rule installation separate budgets? The budget test should assert that margin rather than only equality with the lock-wait sum.

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -213,28 +213,56 @@ reach in. Runs fully unprivileged.
...
-> **IPv4 only.** Host names are resolved to IPv4 addresses only; AAAA
-> records and IPv6 literals are silently dropped because `iptables` (the
-> IPv4 tool) cannot accept IPv6 destinations. A host with only AAAA
-> records is effectively unreachable under firewall mode. For dual-stack
-> hosts, use proxy mode (below) instead.
+**Per-host filtering** (`allowedHosts`/`blockedHosts`) — the behavior
+depends on the schema version, because 0.8 replaced a path that did not
+actually filter.
+
+**Schema 0.8+ — enforced.** `enforcementMode: "firewall"` puts the sandbox in
```
bbonaby: note: in 1.0 enforcementMode won't exist (it's still in the 0.8 schema I'm adding in my 0.8 PR), but when we remove the legacy network schema   it won't exist. Firewall will be implicit. Just giving you an FYI for the future, that said, I don't think that changes your messaging below .

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -252,13 +280,16 @@ namespace choice alone decides the outcome:
...
+| `false` (default) | private (`--unshare-net`; isolated, plus 0.8 proxy and firewall modes) | Honored at the sandbox boundary — nothing outside can reach in. `bind()`/`listen()` still succeed on the sandbox's own loopback, so its processes can talk to each other; that is already inside the caller's trust boundary |
 | `false` | shared with host | **Not honored** — the process can bind/listen on host-local addresses |
 | `true` | private (`--unshare-net`) | **Partially honored** — the listener is reachable only from inside the sandbox |
 | `true` | shared with host | Honored |
 
-Rows 2 and 3 emit a `WARNING:` line to the runner log at preflight rather
-than failing silently. Windows (AppContainer's `privateNetworkClientServer`
+Rows 2 and 3 are rejected on schema `0.8.0-alpha` and later — at config parse,
+and again in the runner for programmatic callers that bypass the parser — and
+emit a
```
bbonaby: note: because we're doing this legacy + new networking schema together for 0.8  they should technically be able to still do them when using 0.8 with the legacy network details. I guess unless you're changing the legacy behavior as well?

src/backends/bubblewrap/common/src/proxy_network.rs
SohamDas2021: On errors: the policy-time rejections already reach callers as config errors, so those are typed enough to act on. The distinction the string currently flattens is host-capability vs user-config — "this host has no slirp4netns/nsenter" is an operator problem, "your proxy is bound to `::1`" is a caller problem, and an SDK consumer cannot tell them apart today. I would rather do that as a follow-up so it covers the dependency probes too, instead of just this file.

On LXC: reuse runs the other way. LXC has real `CAP_NET_ADMIN`, so it already has the richer rule layer — `NetworkIptablesManager` plus `network_ingress.rs`, with proxy-aware closing DROP and deny-precedence. What is Bubblewrap-specific is manufacturing the authority at all (userns + slirp4netns + nsenter), because it is rootless. Longer term Bubblewrap should adopt LXC's rule builder rather than the reverse; the namespace-bringup half is not reusable in either direction.

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: Agreed, and useful to know. It is still in the 0.8 contract today (`src/core/mxc_config_contract/src/dev/network.rs`), so this PR keeps the carve-out.

When it goes away the change is a simplification rather than a redesign: the only reason `builtin_test_server` is special here is that it has no address until it is started, so its endpoint has to be validated at runtime in `run` instead of at policy time. If every proxy is operator-supplied — spun up by a test script beforehand, as you describe — that runtime check disappears and validation collapses into the single policy-time path. Happy to fold that into whichever PR removes the property.

docs/bwrap-support/bubblewrap-backend.md
SohamDas2021: Cleanup here is structural rather than bookkeeping: the rules live inside the sandbox's *own* network namespace, not the host's tables. When the last process in that namespace exits, the kernel tears the namespace down and every rule goes with it — so there is nothing to reconcile, and no way for a rule to outlive the sandbox or affect anything else on the host. `Drop for ProxyNetworkNamespace` still explicitly reaps the supervisor, slirp4netns and the temp state dir.

The closer analogue to your WFP case is the legacy `firewall` mode, which *does* write host-global rules and needs explicit removal — one of the reasons 0.8 moved the rules inside the namespace.

src/backends/bubblewrap/common/src/bwrap_command.rs
SohamDas2021: Good catch, and agreed the warning is wrong under 0.8. Tracked in #970 (Priority3) rather than fixed here.

bbonaby (approved):

PR #933, opened 2026-08-18
Use FlatBuffers Object API for OS Policy Serialization

bbonaby:
## 📖 Description

Currently, we use the raw FlatBuffers APIs to construct the FlatBuffer blob that we send to the OS, even though we used the `--gen-object-api` flag in the `flatc` compiler when generating the code. This PR switches to using the generated object API (structs, enums, etc.) to build the FlatBuffer. This makes the code much more readable and the construction and packing of the FlatBuffer much simpler.

## 🔍 Validation

ran the networking scripts for base container and filesystem
## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/933)

### DIFF
--- a/src/backends/appcontainer/common/src/base_container_runner.rs
+++ b/src/backends/appcontainer/common/src/base_container_runner.rs
@@ -909,7 +909,13 @@ impl BaseContainerRunner {
         // Match legacy AppContainer behaviour: when network enforcement uses
         // capabilities and the default policy is Allow, ensure internetClient
         // is present so the sandboxed process has network access.
-        let mut caps = request.policy.capabilities.clone();
+        let mut caps: Vec<_> = request
+            .policy
+            .capabilities
+            .iter()
+            .filter(|capability| !capability.is_empty())
+            .cloned()
+            .collect();
         if Self::needs_internet_client(request) {
             caps.push("internetClient".to_string());
         }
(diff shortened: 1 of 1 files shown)


src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -993,76 +809,39 @@ impl BaseContainerRunner {
...
-                ui_restrictions,
-                fs_read_write,
-                fs_read_only,
-                fs_deny,
-                network_policy,
-            },
-        );
+        let mut spec = PsecProcessSecurityEnvironment::default();
+        spec.version = SchemaVersionT { major: 1, minor: 0 };
+        spec.capabilities = (!capabilities.is_empty()).then(|| capabilities.join(","));
```
SohamDas2021, line 835: [Non-blocking] The PSEC path previously guarded against a blank capability token: it joined policy.capabilities, then appended internetClient only with a separating comma when the joined string was non-empty. effective_capabilities now pushes internetClient onto the vector before joining, so a policy containing an empty capability entry serializes as ",internetClient" where it previously produced "internetClient"?

config_parser rejects capability entries containing commas and reserved names but does not reject empty strings, so "capabilities": [""] reaches this code and sends a malformed token to the OS?

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -993,76 +809,39 @@ impl BaseContainerRunner {
...
+        spec.disallow_win32k_system_calls = request.policy.ui.disable;
```
SohamDas2021, line 836: build_process_security_environment_spec_produces_valid_psec never reads disallow_win32k_system_calls or  ui_restrictions back out of the packed buffer, so a regression that drops or mispacks either field would still pass?

src/backends/appcontainer/common/src/base_container_runner.rs
bbonaby: Fixed in 031f6364 by omitting empty capability entries before appending `internetClient`, with a regression test for the PSEC payload.

src/backends/appcontainer/common/src/base_container_runner.rs
bbonaby: Fixed in 031f6364. The PSEC round-trip test now asserts both `disallow_win32k_system_calls` and the complete `ui_restrictions` mask.

SohamDas2021 (approved):

PR #934, opened 2026-08-18
fix(seatbelt): allow signaling sandbox children

caarlos0:
## Why

Seatbelt currently permits a sandboxed process to signal only itself. Tools that manage child worker pools can spawn workers but receive `EPERM` when terminating them; `oxfmt` completes its work and then crashes recursively while Tinypool tries to shut down its child processes.

`same-sandbox` is the narrow Seatbelt target for this behavior: descendants inheriting the profile can signal one another, while processes outside the sandbox remain unavailable.

## What changed

- replace the self-only signal rule with `target same-sandbox`
- cover the generated profile and a real child `SIGTERM` in the macOS integration suite
- document process-tree management in the Seatbelt backend

## Validation

- `cargo test -p seatbelt_common`
- integration test TypeScript type-check
- freshly built `mxc-exec-mac` running `sleep 30 & child=$!; kill -TERM $child; wait $child; test $? -eq 143` inside Seatbelt
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/934)

### DIFF
--- a/sdk/node/tests/integration/macos-seatbelt.test.ts
+++ b/sdk/node/tests/integration/macos-seatbelt.test.ts
@@ -81,14 +81,27 @@ describe('macOS Seatbelt Container', {
   });
 
   it('should allow a process to signal its child', async () => {
+    // The EXIT trap matters on the failing path this test exists to catch: if
+    // `kill -TERM` is denied, `exit 1` would otherwise skip `wait` and orphan
+    // the child. The runner deliberately won't group-kill it afterwards (a
+    // reaped pid/pgid can be recycled), so it would outlive the test.
+    const script = [
+      'sleep 5 </dev/null >/dev/null 2>&1 & child=$!',
+      `trap 'kill -KILL "$child" 2>/dev/null || true; wait "$child" 2>/dev/null || true' 0`,
+      'kill -TERM "$child" || exit 1',
+      'wait "$child"; status=$?',
+      'trap - 0',
+      'test "$status" -eq 143',
+    ].join('\n');
     const result = await sdk.spawnSandboxAsync(
-      'sleep 30 & child=$!; kill -TERM "$child" || exit 1; wait "$child"; test "$?" -eq 143',
+      script,
       { version: schemaVersion },
       seatbeltSpawnOptions,
       undefined,
       'seatbelt-child-signal',
     );
-    assert.strictEqual(result.exitCode, 0, `Expected exit 0: ${result.stderr}`);
+    // `spawnSandboxAsync` merges PTY output into `stdout`; `stderr` is always ''.
+    assert.strictEqual(result.exitCode, 0, `Expected exit 0: ${result.stdout}`);
   });
 
   it('should deny filesystem access by default', async () => {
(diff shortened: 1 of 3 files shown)


sdk/node/tests/integration/macos-seatbelt.test.ts
```diff
@@ -80,6 +80,17 @@ describe('macOS Seatbelt Container', {
     assert.strictEqual(result.exitCode, 42);
   });
 
+  it('should allow a process to signal its child', async () => {
+    const result = await sdk.spawnSandboxAsync(
+      'sleep 30 & child=$!; kill -TERM "$child" || exit 1; wait "$child"; test "$?" -eq 143',
+      { version: schemaVersion },
```
bbonaby, line 98: issue (non-blocking): Ensure the regression path cannot leave the background child running.

If `kill -TERM "$child"` fails, which is the exact regression this test is supposed to detect, `|| exit 1` skips `wait` and leaves `sleep 30` alive with the inherited PTY descriptors. Once the direct shell has exited, the Seatbelt runner intentionally cannot group-kill it during `Drop`, so repeated or parallel failures can leave work running and delay PTY teardown for up to 30 seconds. The assertion also prints `result.stderr`, but `spawnSandboxAsync()` always merges PTY output into `stdout`.

Could we install an EXIT trap immediately after starting the child that SIGKILLs and reaps it, redirect the child's stdio, shorten the sleep, and use `result.stdout` in the assertion? For example:

```sh
sleep 5 </dev/null >/dev/null 2>&1 & child=$!
trap 'kill -KILL "$child" 2>/dev/null || true; wait "$child" 2>/dev/null || true' 0
kill -TERM "$child" || exit 1
wait "$child"; status=$?
trap - 0
test "$status" -eq 143
```

bbonaby (commented):
One non-blocking test-reliability finding from the Rust/adversarial review.

sdk/node/tests/integration/macos-seatbelt.test.ts
caarlos0: Good catch — both halves are right, fixed in cce3b9c6.

I confirmed the orphan mechanism before changing anything: `SeatbeltSandboxProcess::kill()` returns early once `try_wait()` reports the shell reaped, so after `exit 1` the group-kill genuinely never fires and `sleep 30` survives. And `spawnSandboxAsync` hardcodes `stderr: ''` (sandbox.ts:711) with all PTY output merged into `stdout`, so the assertion message was guaranteed empty on the one path where it mattered.

Took your snippet essentially as-is. One thing worth recording, because it makes the sleep duration load-bearing rather than cosmetic: under the old policy the trap's `kill -KILL` is *also* denied, so it can't do the cleanup — `wait` is what actually reaps the child, by blocking until it exits on its own. That makes the sleep the teardown bound on the failure path, which is the real argument for 5s over 30s.

Measured under `sandbox-exec` with the two profiles:

| policy | exit | wall |
| --- | --- | --- |
| `(target same-sandbox)` | 0 | 0.019s |
| `(target self)` | 1 | 5.03s |

No orphan left behind in either case.

Left the pre-existing `result.stderr` in `'should execute hello world'` alone — same empty-message bug, but it's not this PR's change and I'd rather not widen the diff. Happy to send it separately if you want it.

bbonaby (approved):

Issue #896, opened 2026-08-14
[LXC/Bubblewrap] Network policy: egress enforcement, hook scoping, and proxy pinning


dhoehna:
## Status audit — 2026-09-17

Verified against `origin/main` @ `be51dc4cb`.

| # | Scope | Status on `main` | Evidence | Tracker |
|---|---|---|---|---|
| **755** LXC half | LXC | **Moot** | Veth model deleted by #1041; invariant now in `enforce_netns_discovery` (`lxc_runner.rs:224-247`) | — |
| **755** bwrap half | bwrap | **Fixed** | 0.8+ resolves to `FirewallEnforced` in the sandbox's own netns | — |
| **764** start→install window | LXC | **LIVE** | `lxc_runner.rs:376` → `:390` (30s) → `:406` → `:428` | ADO 64113943 |
| **766** shared CIDR extraction | Cross | **LIVE** | Duplicated in LXC, bwrap, `base_container_helpers.rs` | ADO 64113943 + 63911316 |
| **777** `-m state` RHEL 10 | LXC+WSLc | **LIVE** | 4 sites; **`bubblewrap/network_rules.rs:1047-48` untracked** | ADO 64113943 |
| **851** `lxc.net.N` N≠0 | LXC | **Postponed** | `state_aware.rs` absent; may dissolve on rebase | ADO 64154043 |
| **858** `lxc.net.0` not veth-checked | LXC | **Postponed** | Same | ADO 64154043 |
| **869** FORWARD return path | LXC | **Moot — struck** | Zero `forward_return`/`physdev` in `src/backends/lxc` | removed 2026-09-17 |
| **870** INPUT path to gateway | LXC | **Fixed** | #1041 — `OUTPUT` via `nsenter` (`:1581`) | — |
| **875** `/etc/hosts` proxy pin | LXC | **LIVE** | `lxc_runner.rs:596-640` | ADO 64113943 |
| **877** unresolvable `blockedHosts` | LXC | **Fixed / by design** | Two-tier (`:1019-27` / `:1028-31`) | — |
| **878** failed probe = no interface | LXC | **Moot** | Same deletion as #755 | — |
| **992** `allowedHosts` host-resolved IPs | LXC | Moved | — | ADO 63911310 |
| **1010** `except` shadows later rules | LXC | **LIVE** | `lower_rule:894-929` emits carve-outs ahead of parent | **none** |
[...]

PR #939, opened 2026-08-18
CI Validation Infra: fixes for WSLC test prereqs, Windows log collection, LXC bridge NAT, and rate limits

theelliotm:
## 📖 Description

A batch of validation-matrix (E2E) CI fixes. Four independent threads, all
confined to the matrix workflow, its resolver, and the CI-facing test entry
points.

**1. LXC host preparation — container networking.**

`lxc-net` skips its firewall setup when it believes another manager owns the
ruleset, leaving a bridge that hands out DHCP leases the container cannot
actually use. `prepare-linux-host.sh` now enables bridge netfilter 
(a hard prerequisite — MXC refuses to report success
for a policy it cannot enforce without it) and ensures the bridge's MASQUERADE
rule exists.

**2. Log collection — artifacts were arriving nearly empty.**

`$env:TEMP` is not `$RUNNER_TEMP`. The Windows suites write scratch trees,
transcripts, and results files to the former; `upload-artifact` reads the
latter, so almost nothing was collected.

`run_ci_backend_tests.ps1` now points `TEMP`/`TMP` at `$RUNNER_TEMP` before
dispatching. 

This is a step towards the goal of having an internal dashboard where scheduled tests and logs are easily navigable.

**3. `backendDelayedStart` — staggering network-heavy job starts.**

Every runner in a 1ES pool shares one egress address, so a backend that pulls a
large runtime or several container images concentrates that traffic into a
burst when its jobs start together, drawing rate limiting and stalled
downloads. The matrix config .json file now accepts:

```json
"backendDelayedStart": [
  { "backend": "wslc", "seconds": 45 }
]
```

The resolver validates it and emits `startup_delay_seconds` on every matrix
entry; each job sleeps that long before its first network step. In this example,
the first WSLC job starts immediately, the second WSLC job starts in 45 seconds, the third
in 90 seconds, and so on.

**4. WSLC read-only mount test — missing fixture.**

Removed Strict mode from the main dispatch script. Also added test.txt to the environment since a few tests required it, but I removed it during the merge conflict since it became obvious it was very recently fixed. 

### Limitations

- There is some evidence that idle runners are more likely to be reclaimed by 
  the pool, so keep stagger times short until a better long-term solution is reached. 
- This does not address the remaining LXC network failures (#896) or the
  process-container ACL/UI failures, which are product-side.

## 🔗 References

Related product-side work for remaining LXC failures: #896.

Documentation: `docs/ci-validation-infrastructure.md` is updated for all of the
above (new "Log collection" section, `backendDelayedStart` reference, and
matching runbook entries).

## 🔍 Validation

- `node scripts/ci/resolve-validation-test-matrix.mjs --plan <pr|nightly|weekly|enabled>` — all four exit 0.
- WSLC entries stagger 0/45/90/135 seconds; every other entry carries `startup_delay_seconds: 0`.
- Workflow YAML parses; step order verified in all three jobs.
- Confirmed that files written to `$TEMP` by a child process land in `$RUNNER_TEMP`.

## ✅ Checklist

## 📋 Issue Type

- [x] Bug fix
- [x] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/939)


.github/workflows/Validation.Tests.Matrix.Job.yml
```diff
@@ -48,6 +48,17 @@ jobs:
...
+      - name: Delayed start
+        if: matrix.startup_delay_seconds > 0
+        shell: pwsh
+        run: |
+          # Spreads network-heavy setup across the pool's shared egress address
+          # instead of issuing every job's downloads at once. Configured by
+          # backendDelayedStart in scripts/ci/validation-test-matrix.json.
+          $seconds = ${{ matrix.startup_delay_seconds }}
+          Write-Host "Waiting $seconds second(s) before downloading prerequisites."
+          Start-Sleep -Seconds $seconds
```
bbonaby, line 60: thought: I guess you're trying to find the sweet spot here with timing? Question though, I forgot to ask. Do we get emails about failures anywhere?

.github/workflows/Validation.Tests.Matrix.Job.yml
```diff
@@ -71,16 +80,25 @@ jobs:
...
+            ${{ runner.temp }}
+            !${{ runner.temp }}/**/*.vhd
+            !${{ runner.temp }}/**/*.vhdx
+            !${{ runner.temp }}/**/*.tar
+            !${{ runner.temp }}/**/*.msi
+            !${{ runner.temp }}/**/*.msix
+            !${{ runner.temp }}/**/*.appx
+            !${{ runner.temp }}/**/*.zip
+            !${{ runner.temp }}/**/*.exe
+            !${{ runner.temp }}/**/*.dll
```
bbonaby, line 101: question:  Why use the whole temp here again? perhaps I'm misunderstanding the comment above but I'm still not sure why we wouldn't want to specify a specific folder inside type or file.

tests/scripts/run_wslc_all_tests.ps1
```diff
@@ -125,12 +114,9 @@ function Run-WslcTest {
         return @{ Name = $ConfigFile; Pass = $true; Skipped = $true; Reason = "File not found" }
     }
 
-    # Skip if the config references a tar file that doesn't exist locally.
-    # Read the chain defensively: this suite inherits Set-StrictMode -Version
-    # Latest from run_ci_backend_tests.ps1, under which touching a missing
-    # property is a terminating error, and most configs have no wslc.imageTarPath.
+    # Skip if the config references a tar file that doesn't exist locally
     $configJson = Get-Content $configPath -Raw | ConvertFrom-Json
-    $tarPath = Get-JsonProperty (Get-JsonProperty (Get-JsonProperty $configJson 'experimental') 'wslc') 'imageTarPath'
```
bbonaby, line 133: issue: @SohamDas2021  are we good with this removal?

scripts/ci/prepare-linux-host.sh
```diff
@@ -129,57 +129,56 @@ start_lxc_bridge() {
...
-# diagnostic: never fails the job, so a networking problem still surfaces as
-# the backend test failure rather than as a prerequisite error.
-report_lxc_network_diagnostics() {
+# Outbound container traffic leaves the bridge subnet with a private source
+# address, so it needs a MASQUERADE rule to reach anything off-host. lxc-net
+# normally installs one, but it skips its firewall setup when it believes
+# another manager owns the ruleset, leaving a bridge that hands out leases the
+# container cannot use. The symptom is a name-resolution failure inside the
+# guest, which reads like a policy problem and is not one.
+ensure_bridge_nat() {
```
bbonaby, line 138: note: You should check with Soham as I think we might need to install another package to test bubblewrap's new proxy stuff. @sohamdas2021

.github/workflows/Validation.Tests.Matrix.Job.yml
theelliotm: No, not currently. Great idea though. I'll add it to the list :)

tests/scripts/run_wslc_all_tests.ps1
theelliotm: I added it during my prior PR. Strict mode was unnecessary and I've stripped it back out.

.github/workflows/Validation.Tests.Matrix.Job.yml
theelliotm: Making it a catch-all for now since T1 tests create a ton of different logs in a couple folders. I'm willing to narrow the scope of it in the future.

bbonaby (approved):

PR #940, opened 2026-08-18
[Bubblewrap] Support hostname proxy endpoints under proxy-only egress

SohamDas2021:
## Description

Stacked on #931. Lets `network.proxy` name a **hostname** (not just an IP literal) under Bubblewrap proxy-only egress, by porting the `ProxyHostPin` approach already used by LXC.

`sandbox_proxy_address` is replaced by `SandboxProxy::resolve`, which returns both the URL handed to the workload and the `ProxyEgress { ip, port, pin }` the firewall authorizes — from a **single** lookup. Two lookups can disagree under round-robin or short TTLs, which would pin the sandbox to an address the chain never opened.

- **Hostnames** keep their configured URL and are pinned via a private `/etc/hosts` bind mount, so `Host` headers and proxy-auth realms still match. With DNS closed, the pin is the only resolution path, so only the first IPv4 answer is opened.
- **IP literals** are rewritten to the slirp gateway as before. The `0.0.0.0`/`::` wildcard rewrites and the specific `::1` rejection from #931 are preserved.
- `resolve_with` takes an injected resolver, so pin behaviour is unit-testable without DNS.

Behaviour is unchanged on schema 0.6/0.7, which stay cooperative-only.

### Validation

`cargo test -p bwrap_common` 122 pass (112 existing + 10 new) · `cargo fmt --check` · `cargo clippy -D warnings` · `validate-configs.js` 234 configs · full Bubblewrap E2E suite (`run_bwrap_all_tests.sh`), including a new hostname-pin case (`tests/configs/bubblewrap_network_proxy_hostname.json`) asserting the pin resolves, egress via the proxy succeeds, and direct egress stays blocked.

### Notes for reviewers

- The LXC and Bubblewrap resolution helpers are **deliberately not shared** for now; rationale and the two verified divergences are recorded in https://github.com/microsoft/mxc/issues/896#issuecomment-5335268879. Notably `::` is rejected by LXC's proxy path but rewritten by Bubblewrap — that looks like an LXC bug, tracked there rather than fixed here.
- `validate` now performs a DNS lookup that `run` deliberately repeats, so a late DNS change is caught at execution rather than trusted from validation time.
- `docs/sandbox-policy/0.8.0/networking/networking.md` D5 ("remote proxies are out of scope for GA") is left untouched: it is a GA scope decision that LXC already contradicts today, independent of this branch.

Related Issues

- AB#62864253

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/940)

### DIFF
--- a/src/backends/bubblewrap/common/src/bwrap_command.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_command.rs
@@ -13,6 +13,13 @@ use wxc_common::filesystem_resolve::FsIntent;
 use wxc_common::models::{ExecutionRequest, NetworkEnforcementMode, NetworkPolicy, ProxyAddress};
 use wxc_common::proxy_env::{is_managed_proxy_key, PROXY_SET_KEYS};
 
+/// The fixed prefix of the command bwrap is asked to run.
+///
+/// `build_args` appends this last, so its position identifies where the
+/// options end -- unlike a bare `--` token, which a caller-supplied
+/// environment value can also be. Shared so the two cannot drift apart.
+pub(crate) const COMMAND_TAIL: [&str; 3] = ["--", "sh", "-c"];
+
 /// Read-only host paths bind-mounted into every Bubblewrap sandbox as the
 /// deny-by-default baseline. Mirrors the seatbelt backend's
 /// `SYSTEM_READ_ALLOW` (`src/backends/seatbelt/common/src/profile_builder.rs`):
@@ -378,9 +385,7 @@ pub(crate) fn build_args_classified_with_mode(
     }
 
     // -- Command -----------------------------------------------------------
-    args.push("--".into());
-    args.push("sh".into());
-    args.push("-c".into());
+    args.extend(COMMAND_TAIL.iter().map(|arg| arg.to_string()));
     args.push(request.script_code.clone());
 
     args
--- a/src/backends/bubblewrap/common/src/bwrap_runner.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_runner.rs
@@ -112,13 +112,19 @@ impl SandboxBackend for BubblewrapScriptRunner {
             }
         }
 
-        // Proxy-only networking rewrites the configured endpoint to slirp's
-        // host gateway and then opens exactly that address in the egress
-        // chain. Both steps can reject an endpoint (an IPv6 loopback the
-        // gateway cannot reach, a routable IPv6 literal the IPv4-only rules
-        // cannot express, a zero port). Running them here means the caller
-        // learns at policy time instead of after a namespace, a slirp instance
-        // and a rule set have already been built and torn down.
+        // Proxy-only networking derives the sandbox-visible endpoint from the
+        // configured one and then opens exactly that address in the egress
+        // chain. That step can reject an endpoint (an IPv6 loopback the
+        // gateway cannot reach, a routable or IPv6-only endpoint the IPv4-only
+        // rules cannot express, a zero port, a hostname that does not resolve).
+        // Running it here means the caller learns at policy time instead of
+        // after a namespace, a slirp instance and a rule set have already been
+        // built and torn down.
+        //
+        // A hostname costs a DNS lookup here, which `run` repeats -- the
+        // resolved address is deliberately not carried over, since the one the
+        // egress chain opens must come from the same lookup that produces the
+        // pin the sandbox is given.
         //
         // The builtin test server has no address until it is started (the
         // parser leaves a port-0 placeholder), so its endpoint stays on the
@@ -128,9 +134,7 @@ impl SandboxBackend for BubblewrapScriptRunner {
                 == ResolvedNetworkMode::ProxyOnly;
         if proxy_only && !request.policy.network_proxy.builtin_test_server {
             if let Some(address) = request.policy.network_proxy.address.as_ref() {
-                let egress = proxy_network::sandbox_proxy_address(address)
-                    .and_then(|translated| proxy_network::ProxyEgress::from_address(&translated));
-                if let Err(error) = egress {
+                if let Err(error) = proxy_network::SandboxProxy::resolve(address) {
                     return Err(ScriptResponse::error(&error));
                 }
             }
@@ -247,10 +251,10 @@ impl BubblewrapScriptRunner {
         }
 
         let network_mode = ResolvedNetworkMode::from_request(request, proxy.is_active());
-        let sandbox_proxy_address = if network_mode == ResolvedNetworkMode::ProxyOnly {
+        let sandbox_proxy = if network_mode == ResolvedNetworkMode::ProxyOnly {
             match proxy.address() {
-                Some(address) => match proxy_network::sandbox_proxy_address(address) {
-                    Ok(address) => Some(address),
+                Some(address) => match proxy_network::SandboxProxy::resolve(address) {
+                    Ok(resolved) => Some(resolved),
                     Err(error) => {
                         proxy.stop(logger);
                         return Err(ScriptResponse::error(&error));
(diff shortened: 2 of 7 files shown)


src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -592,6 +811,126 @@ fn run_probe(mut command: Command, label: &str) -> Result<ProbeOutput, String> {
...
+            return Err(format!(
+                "Bubblewrap: failed to read {SANDBOX_HOSTS_PATH} while pinning the proxy host: \
+                 {error}"
+            ))
+        }
+    };
+
+    let contents = format!("{}\n{}", pin.hosts_line(), existing);
+    fs::write(path, contents)
+        .map_err(|error| format!("Bubblewrap: failed to write the proxy hosts pin: {error}"))
```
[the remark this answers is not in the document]
SohamDas2021: Confirmed and fixed in 8410b706 — and the failure is stronger than "does not guarantee". Measured through real glibc, with the pin on line 1 and a stale loopback entry left below it:

```
BEFORE -> ['127.0.0.1', '10.0.2.2', '10.9.9.9']   # loopback returned FIRST
AFTER  -> ['10.0.2.2']                            # localhost still resolves
```

So the doc comment this function carried ("the pin goes first because the first match wins in a hosts file") was simply wrong: glibc's `files` backend collects *every* matching line and `getaddrinfo` then re-sorts them by RFC 6724, which promotes loopback above the pin. File order buys nothing.

`write_pinned_hosts` now strips the pinned name from the preserved entries via `strip_host_from_hosts`:

- removes only the **name**, never the whole line, so `127.0.0.1 localhost <proxy>` keeps serving `localhost`;
- matches **case-insensitively**, since DNS names are — otherwise `Proxy.EXAMPLE` survives and reintroduces the competing mapping;
- passes untouched lines through **byte for byte** (real `/etc/hosts` is tab-aligned; normalizing it would churn the file for no benefit);
- keeps a trailing comment when an entry is fully consumed.

Six tests added, including the requested duplicate-host regression plus an end-to-end check that exactly one mapping for the pinned name survives against the real host file. The now-false rationale in `pinned_hosts_file_puts_the_pin_first` was corrected too.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -592,6 +811,126 @@ fn run_probe(mut command: Command, label: &str) -> Result<ProbeOutput, String> {
...
+/// The pin goes first because the first match wins in a hosts file. A host
+/// whose `/etc/hosts` already maps the proxy name -- to a stale address, or to
+/// one the egress chain never authorized -- would otherwise shadow the pin and
+/// leave the sandbox unable to reach its proxy.
+///
+/// The host's entries are kept rather than discarded so the sandbox retains
+/// the mappings a workload expects (`localhost` above all). They are read
+/// before the file is created, so a read failure cannot leave a half-written
+/// pin behind.
+fn write_pinned_hosts(path: &PathBuf, pin: &ProxyHostPin) -> Result<(), String> {
```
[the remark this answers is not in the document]
SohamDas2021: Changed to `&Path` in 8410b706, but flagging that the stated rationale does not hold: `ptr_arg` does **not** fire here, so there was no `-D warnings` build failure to avoid.

Verified rather than assumed — `touch`ed the file to defeat clippy's cache, confirmed there is no `allow(clippy::ptr_arg)` and no workspace `[lints]` config, and then validated the methodology by injecting a probe function taking **both** `&String` and `&PathBuf`:

```
warning: writing `&String` instead of `&str` ...  <- fired
(no diagnostic for the &PathBuf parameter)        <- did not fire
```

clippy 1.93 flagged the `&String` and ignored the `&PathBuf`. Taking the change anyway since `&Path` is the more idiomatic signature and every caller already coerces, but the CI-failure premise was incorrect.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -13,6 +13,13 @@ use wxc_common::filesystem_resolve::FsIntent;
 use wxc_common::models::{ExecutionRequest, NetworkEnforcementMode, NetworkPolicy, ProxyAddress};
 use wxc_common::proxy_env::{is_managed_proxy_key, PROXY_SET_KEYS};
 
+/// The fixed prefix of the command bwrap is asked to run.
+///
+/// `build_args` appends this last, so its position identifies where the
+/// options end -- unlike a bare `--` token, which a caller-supplied
+/// environment value can also be. Shared so the two cannot drift apart.
+pub(crate) const COMMAND_TAIL: [&str; 3] = ["--", "sh", "-c"];
```
dhoehna, line 21: If this is always last and only 1 place uses this I suggest to move the constant to where it is used.  That'll show the "Has to be on the tail" requirement more clearly.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -972,6 +998,111 @@ mod tests {
         );
     }
 
+    /// A hostname endpoint is reached by pinning it over `/etc/hosts`, and the
+    /// pin outranks every policy mount -- so a policy that denies the file must
+    /// refuse the combination rather than silently hand the file back.
+    #[test]
+    fn validate_rejects_a_hostname_proxy_that_would_defeat_a_denied_hosts_file() {
+        let mut req = base_request();
+        req.schema_version = "0.8.0-alpha".into();
```
dhoehna, line 1008: Can the schema be turned into a constant?  Reduce duplication.

dhoehna (dismissed):
Looks good to me.  I don't know enogh about bubblewrap's network model to comment on logic.  But the code is sound.

src/backends/bubblewrap/common/src/bwrap_command.rs
SohamDas2021: It has two users in different modules, which is why it's shared rather than local: `bwrap_command.rs:388` appends it, and `proxy_network.rs:993-994` reads it back to find where the options end (`args[sep..sep + COMMAND_TAIL.len()] == COMMAND_TAIL[..]`). Moving it to either site would let the writer and the reader drift apart -- the exact failure the shared constant prevents -- so I'd prefer to keep it here.

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: Not doing here.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -144,36 +156,262 @@ wait "$slirp_pid"
...
+const SLIRP_NETWORK_OCTETS: [u8; 3] = [10, 0, 2];
+
+/// Resolve `host` to the address the egress chain will open.
+///
+/// Only the first IPv4 answer is used, and it is the same one the sandbox is
+/// pinned to. Opening the rest would widen the chain to addresses the sandbox
+/// can no longer select: with DNS closed, the pin is its only resolution path.
+fn resolve_ipv4(host: &str, port: u16) -> Result<Ipv4Addr, String> {
+    let resolved: Vec<IpAddr> = (host, port)
+        .to_socket_addrs()
```
bbonaby, line 448: note: This one is a blocking DNS resolution call that could exceed the amount of time the user added to their timeout config. We can probably use the tokio crate for this so we have asynchronous timeouts. That said, it's pretty heavy. 

Thinking about this some more isn't the host always on 127.0.0.1 or [::1] ?

SohamDas2021: Fixed! `resolve_ipv4` now runs the lookup on a thread with a 15s `recv_timeout` bound (`with_deadline`), which is what tokio would give us here anyway since `lookup_host` also delegates to blocking `getaddrinfo`; it bounds the wait, not the query, as getaddrinfo can't be cancelled. 

On loopback: only `builtinTestServer`/`localhost:<port>` bind loopback, and those never reach the resolver; `network.proxy.url` takes an arbitrary host, which is why the pin exists. DNS is closed in the sandbox, so we resolve on the host and inject `/etc/hosts`

bbonaby (approved):

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -144,36 +156,327 @@ wait "$slirp_pid"
...
+    work: impl FnOnce() -> T + Send + 'static,
+) -> Result<Result<T, mpsc::RecvTimeoutError>, std::io::Error> {
+    let (sender, receiver) = mpsc::channel();
+    thread::Builder::new()
+        .name(label.to_string())
+        .spawn(move || {
+            // Abandoned on expiry: the receiver is gone, so this send fails
+            // and the answer is discarded rather than blocking the thread.
+            let _ = sender.send(work());
+        })?;
```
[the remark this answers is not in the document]
bbonaby: FYI @sohamdas2021, searching online looks like we may potentially be able to get around this by using https://lib.rs/crates/rsdns  or https://crates.io/crates/async-std-resolver libraries. But yea even  Tokio would have this issue I think.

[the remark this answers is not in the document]
SohamDas2021: Ty! Keeping a note to revisit in the next week.

PR #941, opened 2026-08-18
Add 0.8 state-aware configuration adapters

MGudgin:
This PR adds exhaustive state-aware adapters for the mutable `0.8.0-alpha` configuration contract.

This PR is stacked on #929.

Details

* Map every state-aware lifecycle and provision request into the current wire model.
* Preserve validated experimental data and source text in a neutral pre-normalization value.
* Destructure contract marker fields by their qualified type pattern rather than discarding them, so widening a marker fails to compile at the adapters that hardcode its wire value.
* Add development-request facade coverage plus comprehensive expected-wire and current-wire equivalence tests.

Tests

* `cargo clippy -p wxc_common -p mxc_config_contract --all-targets -- -D warnings`
* `cargo test -p wxc_common -p mxc_config_contract`

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/941)

### DIFF
--- a/src/core/wxc_common/src/config_contract_adapters/dev/common.rs
+++ b/src/core/wxc_common/src/config_contract_adapters/dev/common.rs
@@ -1,0 +1,107 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::wire;
+use mxc_config_contract::dev as contract;
+use mxc_config_contract::ContractVersion;
+
+pub(super) fn convert_version(value: contract::Version) -> &'static str {
+    match value {
+        contract::Version::V0_8_0Alpha => ContractVersion::V0_8_0Alpha.as_str(),
+    }
+}
+
+pub(super) fn convert_process(value: contract::Process) -> wire::Process {
+    let contract::Process {
+        command_line,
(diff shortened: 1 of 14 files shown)


bbonaby (dismissed):

src/core/mxc_config_contract/src/dev/mod.rs
```diff
@@ -90,17 +90,53 @@ macro_rules! string_enum {
                 deserializer.deserialize_str(StringEnumVisitor)
             }
         }
+
+        #[cfg(test)]
+        #[allow(non_snake_case)]
+        #[test]
+        fn $name() {
```
bbonaby: question (non-blocking): Do we need a comment in here explaining why we need this for future folks?

bbonaby (approved):

src/core/mxc_config_contract/src/dev/mod.rs
MGudgin: Weird, this comment seems to have landed on the wrong PR. The above code is added by #949. Will address.

PR #969, opened 2026-08-20
Resolve the command before the request is parsed

MGudgin:
Resolve the command before the request is parsed

This PR changes CLI command handling so trailing commands are resolved and
inserted before request parsing, and Rust SDK requests receive their command at
construction time. It preserves behavior for valid requests and requests
without a trailing CLI command; command-preparation failures remain entry-point
errors when the command and policy are independently invalid.

Details

* Probe the request phase and reuse one duplicate-preserving raw source pass to
  select backend-specific quoting and edit `process.commandLine`.
* Keep typed validation authoritative for the effective document, route request
  errors through one production helper, and log replacements only after parsing
  succeeds.
* Require commands in request-builder APIs, remove post-build script mutation,
  and migrate SDK and FFI callers with direct validation coverage.
* Document native CLI syntax, state-aware exec restrictions, and the distinction
  between complete schema-valid requests and CLI templates.

Tests

* `cargo clippy --workspace --all-targets -- -D warnings`
* `cargo test --workspace`
* `cargo check -p mxc_engine --tests --target x86_64-apple-darwin`

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/969)

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,18 @@ wxc-exec.exe --config-base64 <base64-encoded-json>
 
 # Debug output
 wxc-exec.exe --debug config.json
+
+# Supply or replace process.commandLine from trailing arguments
+wxc-exec.exe config.json -- python --version
 ```
+
+For `wxc-exec.exe`, arguments after the required `--` separator are rendered
+for the selected backend and spliced into `process.commandLine` before the
+request is parsed. They may supply a missing command or replace the policy's
+command. This form is supported for one-shot requests and state-aware `exec`;
+other state-aware phases reject it. A policy that relies on trailing arguments
+is a CLI template rather than a complete request that can be executed
+independently.
 
 On Linux: `./lxc-exec config.json`
 On macOS: `./mxc-exec-mac --experimental config.json`
--- a/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
+++ b/docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
@@ -1006,11 +1006,14 @@ fields (`filesystem` / `network` / `ui`) on a per-(backend, phase) Config map di
 to top-level wire fields — they are already wire-format-aligned in the Config, so the
 SDK passes them through unchanged. Cross-backend exec fields (`commandLine`, `cwd`,
 `env`, `timeout`) flow through the top-level `process` block, not through
-`experimental`. For non-exec phases the executor emits a single JSON envelope on
-stdout; for exec the script's output streams raw and the SDK constructs the result
-from PTY events. Responses unwrap any `result` envelope at the SDK boundary so the
-caller sees a plain `ProvisionResult` / `StartResult` / `ExecResult` / `StopResult` /
-`DeprovisionResult`.
+`experimental`. The typed SDK requires `commandLine`. The native `wxc-exec.exe`
+entry point may instead complete an `exec` template from arguments after `--`;
+it inserts or replaces `process.commandLine` before parsing. Trailing commands
+are rejected for every non-exec phase. For non-exec phases the executor emits a
+single JSON envelope on stdout; for exec the script's output streams raw and the
+SDK constructs the result from PTY events. Responses unwrap any `result`
+envelope at the SDK boundary so the caller sees a plain `ProvisionResult` /
+`StartResult` / `ExecResult` / `StopResult` / `DeprovisionResult`.
 
 ## 8. Error model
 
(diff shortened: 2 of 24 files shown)


src/core/mxc_engine/src/policy.rs
```diff
@@ -1727,6 +1733,7 @@ mod tests {
         let err = build_request_with_containment(
             &policy,
             &Containment::Wslc(WslcSection::default()),
+            "echo hello",
```
dhoehna: Most of these tests use "echo hello".  I don't know if "echo hello" is a magic string or not.  Can this be put into a const?

src/core/wxc/src/main.rs
```diff
@@ -1588,31 +1527,144 @@ mod tests {
...
+        let (resolved_request, log) = resolve_with_cli(argv, policy);
+
+        let request = resolved_request.unwrap();
 
         assert_eq!(request.script_code, "cli-app.exe --from-cli");
         assert_eq!(request.working_directory, "C:\\workspace");
         assert_eq!(request.policy.readwrite_paths, vec!["C:\\workspace"]);
-        assert!(logger.get_buffer().contains(
+        assert!(log.contains(
             "Overriding policy process.commandLine with CLI command: cli-app.exe --from-cli"
```
dhoehna: Does `cli-app.exe --fromcli` depend on the same strings on line 1520?  If so, can they be put into either variable or a const?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -165,29 +144,37 @@ pub fn load_mxc_request(
...
+            cli_command: &[],
         },
     )
 }
 
 /// Options-aware variant of [`load_mxc_request`]. When
-/// `LoadOptions::allow_missing_command` is set, a missing or empty
-/// `process.commandLine` in the policy is tolerated and `script_code`
-/// is left empty for the driver to fill in from a CLI override.
+/// `LoadOptions::cli_command` is non-empty, it is rendered for the request's
```
dhoehna, line 257: When does upstream guarantee this assumption?

But.  Why say that when line 165  checks if cli_command is not empty?  That comment leads me to believe that, at this point, cli_command is not empty and never should be empty.  If it is, something happened further upstream.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -199,39 +186,101 @@ pub fn load_mxc_request_from_json(
...
+/// Returns the input **unchanged** whenever the override cannot be applied for
+/// a reason the parser will itself report. Those inputs then keep today's error
+/// text and output routing rather than inheriting a probe's stricter, and
+/// differently routed, diagnostic.
+///
+/// Probing the phase first is load-bearing: it selects which [`ParseError`]
+/// variant a later failure uses, and that selects the stdout envelope versus
+/// the stderr diagnostic.
+///
+/// `argv` is expected to be non-empty; an empty one is rejected rather than
```
dhoehna: Where is the code that does the rejection?  I'm guessing `cmdline_from_argv_for_context` returns an error for an empty argv?

src/core/wxc_common/src/lib.rs
```diff
@@ -20,9 +20,14 @@ pub mod logger;
 pub mod microvm_staging;
 pub mod models;
 pub mod mxc_error;
+// Pre-parse request probes. Not yet reachable from production dispatch; see Gudge.
```
dhoehna: If I see you, what should I ask?  When the pre-parse request probes will be reachable from production dispatch?

dhoehna (dismissed):
Besides the comments about de-duplication the code is sound.  I am concerned to see duplicate methods for one-shot and state-aware.  Merging those paths is a future issue.

src/core/wxc_common/src/lib.rs
MGudgin: Ask anything you like.

src/core/mxc-sdk/README.md
```diff
@@ -419,8 +418,13 @@ use mxc_sdk::{
...
-request.set_script("python3 -c 'print(42)'").set_experimental(true);
+let mut request = build_request_with_containment(
+    &policy,
+    &Containment::Wslc(wslc),
+    "python3 -c 'print(42)'",
+    None,
+)?;
+request.set_experimental(true);
 let output = run(request)?;
 # Ok::<(), mxc_sdk::Error>(())
```
bbonaby: note: FYI @caarlos this is a breaking (at compile time) change for GitHub copilot in the Rust SDK. Basically you no longer build the request, and _then_ set the script.

With the new API, you would add in the script to the new build request API directly instead of after. Also @mgudgin, we probably would want to make sure we have parity in the other SDKs right? Same kind of calling convention I mean.

docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md
```diff
@@ -1102,24 +1105,39 @@ domain types for trivial enum/struct conversions). The state-aware path reuses
 this wire model while retaining its per-backend `experimental` subtree for
 dispatch-time typing.
 
+When the native CLI supplies trailing command arguments, the loader first
+splices the rendered command into `process.commandLine` and then parses that
+effective document. Diagnostics and retained state-aware source text are
+therefore relative to the effective document; replacing or inserting the
+command may shift a later same-line column from its position in the caller's
+original bytes.
```
bbonaby: thought: read this a few times but not sure I fully grasp what it's saying. Is this paragraph needed?

src/core/mxc-sdk/README.md
MGudgin: Thanks for flagging the compatibility impact. The other public SDKs already take the command at the operation boundary: Node passes the script to `spawnSandbox` / `spawnSandboxAsync`, and C# passes the command to `MxcSandbox.Run` / `RunAsync`. Requiring the Rust builder to receive the command brings its construction flow into line with those surfaces, so I do not think an additional Node or C# API change is needed here.

README.md
```diff
@@ -147,8 +147,19 @@ wxc-exec.exe --config-base64 <base64-encoded-json>
...
+# Supply or replace process.commandLine from trailing arguments
+wxc-exec.exe config.json -- python --version
 ```
 
+For `wxc-exec.exe`, arguments after the required `--` separator are rendered
+for the selected backend and spliced into `process.commandLine` before the
+request is parsed. They may supply a missing command or replace the policy's
+command. This form is supported for one-shot requests and state-aware `exec`;
+other state-aware phases reject it. A policy that relies on trailing arguments
+is a CLI template rather than a complete request that can be executed
```
bbonaby: question: CLI template? Do you mean a CLI argument?

bbonaby (dismissed):

shschaefer (dismissed):

shschaefer (approved):

PR #961, opened 2026-08-20
Add additive schema 0.8 networking contract

bbonaby:
## 📖 Description

Adds schema 0.8 directional networking wire/domain types and parsing. Preserves schema 0.6/0.7 legacy networking and rejects mixed legacy/directional fields. Regenerates the dev schema and required TypeScript wire parity types.

Does not add backend validation or enforcement.

## 🔗 References

- Stacked validation: #962
- Schema design: `docs/sandbox-policy/0.8.0/networking/`

## 🔍 Validation

- `cargo test -p wxc_common`
- `cargo fmt --all -- --check`
- `npm test` in `sdk/node`
- Schema and SDK wire codegen checks

## ✅ Checklist

- [x] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [ ] Linked to an issue
- [x] Updated documentation (if applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed)
- [ ] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes (see [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md))

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

PR #974, opened 2026-08-20
Realign Seatbelt with the official 0.8 GA networking schema

### LINKED ISSUE #937: [Seatbelt] network.proxy is enforced on LXC/Bubblewrap-0.8 but only cooperative on macOS Seatbelt
### Relevant area(s)

macOS

`network.proxy` means different things on different backends, and on macOS Seatbelt a policy that *looks* enforced is only cooperative. LXC already established the project's position on this and **fails closed** — `apply_firewall_rules` refuses to run rather than report success for an enforcement that did not happen:

> `network.proxy` requires `network.enforcementMode='firewall'` or `'both'`. This policy enables a proxy under 'capabilities', where no iptables rules are installed, so the proxy environment would be injected while direct egress stayed unrestricted -- any client that ignores `HTTP_PROXY` would bypass the proxy entirely. Refusing to apply rather than reporting success for an enforcement that did not happen.

— `src/backends/lxc/common/src/network_iptables.rs:1673-1682`

Seatbelt accepts the same configuration and silently does the cooperative thing instead:

| Backend | Behavior with `network.proxy` |
|---|---|
| LXC | **Enforced**; fails closed if enforcement is unavailable |
| Bubblewrap, schema 0.8+ | **Enforced** (default-DROP + proxy allowlist in a private netns) |
| Bubblewrap, schema 0.6 / 0.7 | Cooperative — env vars only, shared host network |
| Seatbelt (macOS) | Cooperative — injects `HTTP_PROXY`/`HTTPS_PROXY`, and `defaultPolicy: allow` emits an unfiltered `(allow network-outbound)` |

Any client that ignores the proxy environment variables reaches the network directly, and the caller gets no signal that this is what they configured.

Bubblewrap 0.6/0.7 is a **deliberate, documented** concession for existing consumers on those schema versions and should be left as-is; it is listed here only for comparison. Seatbelt is the case that needs a decision.

This issue is about resolving the inconsistency, not about picking a specific answer up front. Reasonable outcomes differ — enforce it, reject the configuration the way LXC does, or accept it explicitly with the limitation documented and surfaced as a warning. What should not persist is the current state, where the same policy is silently enforced on one backend and advisory on another.

Source: adversarial review of #931 (Medium), where it was raised as a macOS-only gap.

### Steps to reproduce

1. Author a policy with `network.proxy` set to a reachable proxy.
2. Run it under Seatbelt on macOS with a workload that makes an outbound connection **without** consulting `HTTP_PROXY` — e.g. a raw TCP connect to a public address on port 443.
3. Run the equivalent policy under LXC with `enforcementMode` left at its default.

### Expected behavior

Consistent, predictable handling across backends. A backend that cannot enforce the proxy should not accept the configuration silently; it should either reject it as LXC does, or surface an explicit warning stating that enforcement is cooperative and that non-cooperating clients will bypass the proxy. Whatever is chosen should be reflected in `docs/schema.md` and `docs/macos-support/seatbelt-backend.md`.

### Actual behavior

The direct connection succeeds under Seatbelt, bypassing the proxy, with no error or warning. The same shape of policy is refused outright by LXC. Callers writing one policy for multiple backends have no way to tell which are enforcing it.

---

The WSLc half of this issue was split out into its own tracking item, since it is blocked on the WSLc SDK rather than on a decision.

Related: #946 — the no-proxy path on the same backend, where `allowedHosts` under `defaultPolicy: "block"` degrades to allow-all outbound.

theelliotm:
## 📖 Description

Brings the macOS Seatbelt backend in line with the 0.8 GA networking schema
(#961 / #962), and closes several ways a network policy could look enforced
but not be.

### Schema 0.8 support

- Declares `network_policy_support()` = `EGRESS_DEFAULT | INGRESS_DEFAULT | HOST_LOOPBACK | RUNTIME_PROXY`.
  No `EGRESS_RULES` / `PROXY_PEER_IDENTITY` — Seatbelt can't filter by destination.
- `profile_builder` reads the directional `network.egress` / `network.ingress`
  when present, else the legacy `defaultPolicy` / `allowLocalNetwork`. Both
  shapes produce the same profile.
- Rejects an `ingress.hostLoopback` that differs from `ingress.default` —
  Seatbelt has no separate loopback posture to enforce it against.

### Fixes

All four were fail-open: the config was accepted, but egress wasn't restricted.

| Config | Before | Now |
|---|---|---|
| `[::1]` proxy + `egress: deny` | `(allow network-outbound)` — fully open | scoped to the proxy port |
| remote proxy + deny | fell back to allow-all | rejected |
| `allowedHosts` + deny | allowlist silently dropped | rejected |
| proxy + `allow` | proxy silently ignored | rejected |

The `[::1]` case is the notable one. Several layers each decided "is this
loopback?" independently, and some used string equality against `"::1"` — which
never matches, since a URL can only contain the bracketed form (RFC 3986). The
mismatch fell through to the remote-proxy branch, which opened all egress.

This PR fixes it on the **0.8 directional** path. The legacy
`network.proxy.url` shape still has the same string-equality bug in the config
parser and is tracked separately in #1004, to keep `config_parser.rs`
untouched here.

### Consolidation

- Seatbelt's four network invariants now live in one place
  (`wxc_common/src/seatbelt_policy.rs`) and run from
  `SeatbeltScriptRunner::validate`. Previously they only ran on the JSON path,
  so `mxc_engine::run(&ExecutionRequest)` skipped all of them.
- `validate` runs on every entry path (`Runner<B>` calls it before `execute`),
  so the Rust SDK and FFI callers are now covered too, not just the CLI.
- `config_parser.rs` is deliberately **unchanged** in this PR. Two of the four
  invariants also still run there, inherited from `main` — duplicated, but the
  messages match, so behaviour is identical.
- Added the loopback predicate used to *admit* a Seatbelt proxy:

  | name | matches | used to |
  |---|---|---|
  | `host_is_loopback` | all of `127.0.0.0/8`, `::1`, `localhost` | **reject** (LXC) |
  | `host_is_canonical_loopback` | `127.0.0.1`, `::1`, `localhost` | **admit** (Seatbelt) |

  The narrow set is exactly what Seatbelt's `(remote ip "localhost:<port>")`
  enforces — verified: `127.0.0.2` gets `EPERM` under that rule.

### Docs

Updated `docs/seatbelt/seatbelt-backend.md` (moved from `docs/macos-support/`),
`docs/examples.md`, the backend table in
`docs/sandbox-policy/0.8.0/networking/schema-updates.md`, and the Seatbelt row
in `.github/copilot-instructions.md`. Removed earlier wording that described
the 0.8 directional shape as Seatbelt-only — it's the cross-backend schema;
Seatbelt is just the first backend to declare enforcement.

## 🔗 References

- Closes #937
- Follow-up: #1004 (same `[::1]` bug on the legacy proxy path)
- [`networking.md`](https://github.com/microsoft/mxc/blob/main/docs/sandbox-policy/0.8.0/networking/networking.md) — cross-backend 0.8 networking design
- [`schema-updates.md`](https://github.com/microsoft/mxc/blob/main/docs/sandbox-policy/0.8.0/networking/schema-updates.md) — field mapping

## 🔍 Validation

- `cargo test` across `wxc_common`, `seatbelt_common`, `mxc_engine`,
  `mxc-sdk`, `mxc_darwin`, `lxc_common`, `bwrap_common` — **1438 passed, 0 failed** (23 binaries).
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all -- --check` — clean.
- `node scripts/versioning/validate-configs.js` — 234 configs OK.
- `./build-mac.sh --rust-only` — builds a working `mxc-exec-mac`.

E2E against the built binary:

- `[::1]` proxy + `egress: deny` → proxied request succeeds; direct HTTPS and
  direct-IP both blocked. (Legacy `network.proxy.url` with `[::1]` is still
  rejected by the parser — pre-existing, see #1004.)
- All four rejections above confirmed, in the legacy and directional shape.
- `30_mac_network_schema_v2.json` now exits 1 with no proxy listening (was 0).
- Confirmed the deny is genuine and not a DNS artifact: a raw IP is blocked
  under `deny` and reachable under `allow`.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/974)


src/core/wxc_common/src/models.rs
```diff
@@ -584,6 +584,28 @@ pub fn unbracket_host(host: &str) -> &str {
...
+/// Use this to *admit* a proxy host. The accept-set is exactly what Seatbelt's
+/// `(remote ip "localhost:<port>")` enforces — measured, not assumed: under
+/// that rule `127.0.0.1` and `::1` connect while `127.0.0.2` gets `EPERM`.
+/// Matching `127.0.0.0/8` here would admit a proxy the sandbox then blocks.
+///
+/// Shared by the parser gate, the Seatbelt policy check, and the profile
+/// builder; if they disagree, a host one layer admits another sends down the
+/// wrong branch. For the reject-side question use
+/// `network_parser::host_is_any_loopback`.
+pub fn host_is_canonical_loopback(host: &str) -> bool {
```
bbonaby: note: This along with the tests below should go in network_parser.rs or another common helper file

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1105,51 +1106,14 @@ fn convert_wire_config(
...
-                       silently weakening the deny for raw-socket clients that ignore \
-                       HTTP_PROXY. Use a loopback proxy (127.0.0.1/::1/localhost) or \
-                       'network.proxy.builtinTestServer: true' for port-scoped reachability \
-                       under deny.";
-            logger.log_line(msg);
-            return Err(WxcError::ConfigParse(msg.to_string()));
+        // All Seatbelt network invariants live in one place so the backend's
+        // own `validate` enforces the same rules for callers that build an
+        // ExecutionRequest directly and never touch this parser.
+        if containment == ContainmentBackend::Seatbelt {
```
bbonaby: issue: can we move new seatbelt specific policy errors to the seatbelt backend? we can leave the current ones here for now as they'll be moved in the future. Just new ones.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1105,51 +1106,14 @@ fn convert_wire_config(
...
+            if let Err(msg) = validate_seatbelt_network_policy(&policy) {
+                logger.log_line(&msg);
+                return Err(WxcError::ConfigParse(msg));
```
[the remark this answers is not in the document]
theelliotm: not true

src/core/wxc_common/src/seatbelt_policy.rs
```diff
@@ -0,0 +1,116 @@
...
+                    network.enforcementMode='firewall' or 'both'. macOS Seatbelt \
+                    enforces network policy through the sandbox profile and has no \
+                    packet-filter layer, so a firewall mode cannot be honored."
+            .to_string());
+    }
+
+    // A remote proxy can't be expressed as a reachability rule, so it would
+    // simply be unreachable. Loopback proxies (including builtinTestServer,
+    // whose address is resolved at runtime and absent here) stay port-scoped.
+    if !outbound_allowed
```
bbonaby, line 58: question: does this change behavior in 0.6 and 0.7?  Wondering what was the case beforehand.

tests/examples/20_mac_combined_restrictions.json
```diff
@@ -15,9 +15,6 @@
         ]
     },
     "network": {
-        "defaultPolicy": "block",
-        "allowedHosts": [
-            "api.github.com"
-        ]
+        "defaultPolicy": "block"
```
bbonaby: suggestion: perhaps we can make this one an allow policy since the other one was block.

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -118,6 +132,23 @@ impl SandboxBackend for SeatbeltScriptRunner {
...
+        if let Some(ingress) = request.policy.network_ingress.as_ref() {
+            if ingress.host_loopback != ingress.default {
+                return Err(error_response(
+                    "macOS Seatbelt has no private loopback and cannot enforce an \
+                     independent network.ingress.hostLoopback posture. Set \
+                     'hostLoopback' equal to 'default', or omit 'hostLoopback' to \
+                     inherit 'default'."
+                        .to_string(),
+                ));
+            }
```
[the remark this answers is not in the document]
bbonaby: issue (blocking): Could we keep this open until Seatbelt either enforces `hostLoopback` in both directions or stops declaring `HOST_LOOPBACK`? The shared capability is explicitly bidirectional, but `egress.default=allow` with `ingress.default/hostLoopback=deny` still emits unrestricted outbound access, so sandbox-to-host loopback remains reachable.

tests/examples/30_mac_network_schema_v2.json
```diff
@@ -0,0 +1,26 @@
+{
+    "$schema": "../../schemas/dev/mxc-config.schema.0.8.0-dev.json",
+    "version": "0.8.0-alpha",
+    "containment": "seatbelt",
+    "process": {
+        "commandLine": "echo 'Testing schema-0.8 network policy...'; echo ''; rc=0; echo 'Test 1: Connecting through the loopback proxy...'; if curl -s --max-time 5 -x http://127.0.0.1:8080 https://api.github.com/zen; then echo ''; echo '  SUCCESS'; else echo '  ERROR: Proxied request failed (is a proxy listening on 127.0.0.1:8080?)'; rc=1; fi; echo ''; echo 'Test 2: Connecting directly (should be blocked, egress.default is deny)...'; if curl -s --max-time 5 --noproxy '*' https://example.com > /dev/null 2>&1; then echo '  ERROR: Should not have connected!'; rc=1; else echo '  SUCCESS: Direct connection blocked as expected'; fi; echo ''; if [ $rc -eq 0 ]; then echo 'Schema-0.8 network policy test complete!'; else echo 'Schema-0.8 network policy test FAILED'; fi; exit $rc",
```
bbonaby, line 6: issue (blocking): Could we promote this scenario into the existing macOS `wxc_e2e_tests` before approval? The macOS CI currently proves proxy environment injection but does not run this enforcement check, and the Seatbelt validation dispatcher is still unwired. Since we cannot validate locally, CI should show that the proxy succeeds while direct HTTPS/raw-IP egress is denied, with legacy 0.7 allow/deny coverage alongside the 0.8 case.

src/core/wxc_common/src/seatbelt_policy.rs
bbonaby: note: I took another look, and this behavior change seems appropriate. Seatbelt isn't available in 0.6, and for 0.7 this closes previous enforcement gaps by rejecting policy combinations Seatbelt can't faithfully honor rather than accepting them and running fail-open. The normal 0.7 allow/block paths remain unchanged, so I don't see this as a blocker. Resolving

bbonaby (approved):

PR #976, opened 2026-08-21
[ProcessContainer] Enforce schema 0.8 networking

bbonaby:
## 📖 Description

Enforces additive schema 0.8 networking for Windows ProcessContainer. Directional egress rules, runtime proxy configuration, proxy peer identity, and host-loopback policy are enforced through PSEC. Compatible default-only requests may use the legacy SBOX/AppContainer tiers, but schema 0.8 runtime proxy and other PSEC-only networking requests are rejected when PSEC is unavailable rather than weakened. Schema 0.6/0.7 behavior and legacy proxy fallback remain intact.

The change also extracts shared network-policy helpers and adds regression coverage for unsupported AppContainer directional rules, policy-owned capabilities, PSEC network serialization, and isolated child-environment construction.

## 🔗 References

- `docs/sandbox-policy/0.8.0/networking/networking.md`
- `docs/process-container/networking.md`

## 🔍 Validation

- `cargo test -p wxc_common -p appcontainer_common`
- `cargo clippy -p wxc_common -p appcontainer_common --all-targets -- -D warnings`

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/976)


src/backends/appcontainer/common/src/network_policy_helpers.rs
```diff
@@ -0,0 +1,92 @@
...
+use wxc_common::models::{ContainerPolicy, NetworkAction, NetworkEnforcementMode, NetworkPolicy};
+
+pub(crate) const INTERNET_CLIENT_CAPABILITY: &str = "internetClient";
+pub(crate) const PRIVATE_NETWORK_CAPABILITY: &str = "privateNetworkClientServer";
+
+pub(crate) fn allows_network_egress(policy: &ContainerPolicy) -> bool {
+    policy.network_egress.as_ref().map_or(
+        policy.default_network_policy == NetworkPolicy::Allow,
+        |egress| egress.default == NetworkAction::Allow,
+    )
```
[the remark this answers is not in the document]
bbonaby: Fixed in c1a5a2c8. `allows_network_egress` now treats a non-empty directional allow list as requiring `internetClient`, even when the default is deny. Added helper coverage and a PSEC serialization assertion for the filtered-egress case.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+            NetworkAction::Deny => SboxFilterAction::deny,
+        };
+        network.egress = Some(Box::new(egress));
+    }
+    network
+}
+
+fn build_psec_network_policy(policy: &ContainerPolicy) -> PsecNetworkPolicy {
+    let mut network = PsecNetworkPolicy::default();
+    if policy.network_proxy.is_enabled() {
```
SohamDas2021, line 138: build_psec_network_policy branches if network_proxy.is_enabled() and only sets network.proxy . The whole network.egress block (default action + allow/deny) is never populated. Its SBOX twin guards the same branch with && !policy.runtime_network_proxy_specified; PSEC omits it. Since psec_policy_compatible selects PSEC precisely when the proxy is runtime-specified, a 0.8 config with a runtime proxy + egress.deny rules serializes a PSEC spec with no filtering at all?

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+    capabilities
+}
+
+fn non_empty_paths(paths: &[String]) -> Option<Vec<String>> {
+    (!paths.is_empty()).then(|| paths.to_vec())
+}
+
+fn build_legacy_sbox_network_policy(policy: &ContainerPolicy) -> SboxNetworkPolicy {
+    let mut network = SboxNetworkPolicy::default();
+    if policy.network_proxy.is_enabled() && !policy.runtime_network_proxy_specified {
```
SohamDas2021: This is the correct form. The PSEC counterpart at line 133 omits the second condition, plz see the comment there.

src/backends/appcontainer/common/src/network_policy_helpers.rs
```diff
@@ -0,0 +1,99 @@
...
+) {
+    let uses_capabilities = matches!(
+        policy.network_enforcement_mode,
+        NetworkEnforcementMode::Capabilities | NetworkEnforcementMode::Both
+    );
+    if uses_capabilities && allows_network_egress(policy) {
+        ensure_capability(capabilities, INTERNET_CLIENT_CAPABILITY);
+    }
+
+    if policy
```
SohamDas2021: internetClient is correctly gated on uses_capabilities at line 35, but privateNetworkClientServer here is granted unconditionally? Under NetworkEnforcementMode::WfpOnly the backend is supposed to rely solely on WFP and grant no token capabilities, so this leaks bidirectional private-network reachability the caller explicitly opted out of? Please fold uses_capabilities && into this condition.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+
+fn effective_capabilities(policy: &ContainerPolicy, include_host_loopback: bool) -> Vec<String> {
+    let mut capabilities: Vec<_> = policy
+        .capabilities
+        .iter()
+        .filter(|capability| !capability.is_empty())
+        .cloned()
+        .collect();
+    add_default_network_capabilities(policy, &mut capabilities);
+    if include_host_loopback && unrestricted_host_loopback_allowed(policy) {
```
SohamDas2021: privateNetworkClientServer :  LOOPBACK_NETWORK_CAPABILITY is granted without consulting network_enforcement_mode? Worth extracting the uses_capabilities check out of add_default_network_capabilities?

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,274 @@
...
+
+const SANDBOX_SPEC_VERSION: &str = "0.1.0";
+
+pub(super) fn requires_psec_networking(policy: &ContainerPolicy) -> bool {
+    policy
+        .network_egress
+        .as_ref()
+        .is_some_and(|egress| !egress.allow.is_empty() || !egress.deny.is_empty())
+        || policy.allowed_proxy_peer.is_some()
+        || unrestricted_host_loopback_allowed(policy)
```
SohamDas2021, line 38: Test gap. This clause is the only thing routing a host-loopback-only request to PSEC instead of legacy SBOX, and nothing asserts it. A regression here would silently land on a tier that can't enforce loopback scoping. Please add a case with only hostLoopback='allow' set, asserting requires_psec_networking is true and legacy_sbox_compatible_with_request is false?

SohamDas2021: Same for has_conflicting_proxy_identity just below (line 39): four input combinations, no direct coverage?

A table-driven test over build_psec_network_policy asserting network.egress.is_some() for every proxy/runtime-proxy combination would have caught the line 133 bug.

docs/process-container/networking.md
SohamDas2021: This row is now contradicted by the code: appcontainer_runner.rs:1533-1545 explicitly rejects ingress.default='allow' combined with denied egress on the AppContainer fallback, because privateNetworkClientServer is bidirectional.

Two other docs drifted in the same direction and aren't in this PR:
•  docs/process-container/os-version-support.md:135-157 — still schema-0.7-only, and claims firewall enforcement on every Windows release; needs 0.8 rows split by PSEC / SBOX / AppContainer.
•  docs/sandbox-policy/0.8.0/networking/networking.md:359-372 — the proxy row says "Per-AppContainer WinHTTP configuration" with no qualification, but schema 0.8 on the SBOX tier is cooperative env-var only.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: `runtimeConfig.networkProxy` is parser-valid only with `egress.default=deny` and empty direct allow/deny lists, so PSEC proxy and direct-egress forms are intentionally mutually exclusive. Commit f0a437f3 documents that invariant and adds serialization coverage for legacy proxy, runtime proxy, and no-proxy forms.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: The SBOX condition is intentionally different: runtime proxy is a schema 0.8 form and must not enter the legacy SBOX proxy field. PSEC accepts runtime proxy, while direct egress is excluded by parser validation. Added an explicit invariant comment and coverage in f0a437f3.

src/backends/appcontainer/common/src/network_policy_helpers.rs
bbonaby: `enforcementMode` is legacy-only; the parser rejects mixing it with schema 0.8 directional fields. Directional ingress always needs the capability gate, even if the compatibility field retains its default value. I made this separation explicit through `uses_network_capabilities` and added a regression test proving directional policy ignores the legacy mode.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: Host loopback is directional-only and therefore must not be gated by legacy `enforcementMode`. I extracted `uses_network_capabilities` and now use it for both default and loopback capability derivation, with a test covering a directional policy alongside legacy `WfpOnly` state.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: Added direct host-loopback-only coverage asserting both `requires_psec_networking` and SBOX incompatibility in f0a437f3.

bbonaby: Added all four peer-identity/host-loopback combinations. The PSEC serialization table asserts the actual contract: either proxy is present and egress is absent, or proxy is absent and egress is present. Runtime proxy plus direct rules is rejected earlier by the parser.

docs/process-container/networking.md
bbonaby: Updated all three documents. The matrix now qualifies the PSEC-only deny-egress/allow-ingress result, the OS support guide splits schema 0.8 by PSEC/SBOX/AppContainer, and the parent spec qualifies proxy enforcement. I also documented that SBOX has no policy-lifetime cleanup handle, so it never receives schema 0.8 WFP filters or runtime proxy posture.

docs/sandbox-policy/0.8.0/networking/networking.md
```diff
@@ -342,9 +342,10 @@ already implemented.
 
 ### Process containers (Windows): GA target and compatibility behavior
 
-The model-2 guarantees below apply to ProcessContainer paths with OS-scoped
-proxy enforcement. The AppContainer compatibility fallback is cooperative and
-does not satisfy model 2; see the implementation doc for its limitations.
+The model-2 guarantees below require the PSEC ProcessContainer path with
+OS-scoped proxy enforcement. Legacy SBOX and the AppContainer fallback reject
+schema 0.8 runtime proxy requests because they cannot preserve the required
+peer-identity or host-loopback posture.
```
SohamDas2021: This correctly says schema 0.8 runtime proxy is rejected on fallback paths, but D5 at line 296 still claims cooperative AppContainer fallback, and the PR description promises cooperative SBOX fallback. Please align both with this PSEC-only behavior.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,282 @@
...
+fn unrestricted_host_loopback_allowed(policy: &ContainerPolicy) -> bool {
+    policy
+        .network_ingress
+        .as_ref()
+        .is_some_and(|ingress| ingress.host_loopback == NetworkAction::Allow)
+}
+
+fn allowed_appcontainer_peer(policy: &ContainerPolicy) -> Option<String> {
+    if unrestricted_host_loopback_allowed(policy) {
+        Some(LOOPBACK_NETWORK_PEER.to_string())
```
MGudgin, line 184: **Medium (correctness/security) - Reserve the unrestricted-loopback peer identity.**

**Attribution:** `introduced_by_change` - this PR introduces `MXC-Loopback` as a reserved PSEC identity while `allowedProxyPeer` remains a caller-controlled non-empty string.

A caller can supply `allowedProxyPeer: "MXC-Loopback"` with `hostLoopback: "deny"`, producing the same OS-facing peer value as unrestricted loopback while the internal posture says loopback is denied. That makes the serialized representation ambiguous and bypasses the intended mutual exclusion if the OS recognizes the reserved identity.

**Fix:** Reject the reserved value for caller-supplied `allowedProxyPeer`, or represent unrestricted loopback outside the caller-controlled peer namespace. Add a regression test for the reserved value.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -110,6 +104,26 @@ fn encode_env_block(env_vars: &[String]) -> Vec<u16> {
...
 
+fn build_runtime_proxy_env_block(request: &ExecutionRequest) -> Result<Option<Vec<u16>>, WxcError> {
+    if !request.policy.runtime_network_proxy_specified {
+        return Ok(None);
+    }
+    let Some(address) = request.policy.network_proxy.address.as_ref() else {
+        return Ok(None);
+    };
+
+    let env = if request.env.is_empty() {
```
MGudgin: **Medium (maintainability/testability/simplicity) - Consolidate runtime-proxy environment construction.**

**Attribution:** `introduced_by_change` - this helper adds a second default-environment lookup and UTF-16 encoding path beside the existing child-environment pipeline.

The common empty-env case calls the live Win32 environment APIs, converts tuples to `KEY=VALUE` strings, and uses a new encoder; the adjacent non-proxy branch independently resolves and encodes the same default environment. The added runtime-proxy tests avoid this branch by supplying a synthetic environment, so the normal input is both drift-prone and difficult to test deterministically.

**Fix:** Resolve the base child environment once, apply proxy variables as an optional overlay, and use one canonical encoder. Keep the proxy merge independently testable from live OS environment acquisition.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,282 @@
...
+    capabilities
+}
+
+fn non_empty_paths(paths: &[String]) -> Option<Vec<String>> {
+    (!paths.is_empty()).then(|| paths.to_vec())
+}
+
+fn build_legacy_sbox_network_policy(policy: &ContainerPolicy) -> SboxNetworkPolicy {
+    let mut network = SboxNetworkPolicy::default();
+    if policy.network_proxy.is_enabled() && !policy.runtime_network_proxy_specified {
```
MGudgin: **Low (maintainability/simplicity) - Remove the duplicate runtime-proxy condition.**

**Attribution:** `introduced_by_change` - this PR adds both this condition and `legacy_sbox_compatible_with_request`, which already rejects every request with `runtime_network_proxy_specified` before the SBOX builder is called.

The second check cannot vary in normal execution and does not fail closed if the precondition is violated; it simply chooses another serialization form. That makes readers trace two layers for an invariant enforced by the selector.

**Fix:** Remove `!policy.runtime_network_proxy_specified`, or make the builder explicitly assert/reject a violated SBOX-selection precondition.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,282 @@
...
+    if ports.is_empty() {
+        return None;
+    }
+
+    let destinations: Vec<_> = rule
+        .to
+        .iter()
+        .filter(|peer| peer.cidr.address.is_ipv4() == ipv4)
+        .collect();
+    if !rule.to.is_empty() && destinations.is_empty() {
```
MGudgin, line 262: **Low (test coverage) - Pin ICMP address-family suppression.**

**Attribution:** `introduced_by_change` - this PR adds the logic that drops an ICMP family when the rule has destinations but none from that address family.

The rich-rule test supplies both IPv4 and IPv6 destinations, so both generated ICMP rules survive. There is no focused test proving that an IPv4-only destination cannot accidentally produce an unconstrained ICMPv6 rule, or vice versa.

**Fix:** Add IPv4-only and IPv6-only ICMP cases and assert that only the matching protocol rule is serialized.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,282 @@
...
+        return None;
+    }
+    Some(psec_endpoint_rule(destinations, ports))
+}
+
+fn psec_endpoint_rules(rules: &[NetworkRule]) -> Vec<PsecEndpointRuleT> {
+    let mut endpoints = Vec::new();
+    for rule in rules {
+        let ports: Vec<_> = rule.ports.iter().filter_map(psec_non_icmp_port).collect();
+        if rule.ports.is_empty() || !ports.is_empty() {
```
MGudgin, line 272: **Low (test coverage) - Cover destination-only all-port rules.**

**Attribution:** `introduced_by_change` - this PR adds the branch that maps `ports: []` to a destination rule with no FlatBuffer ports vector.

Current coverage exercises explicit TCP ranges and ICMP splitting, but not this all-protocol/all-port form. A regression here could either drop the destinations or accidentally constrain the rule.

**Fix:** Add allow and deny cases with destinations plus `ports: []`, then assert that destinations are present and the serialized ports vector is absent.

MGudgin (dismissed):
## Summary

Approved with five non-blocking inline findings. I re-resolved the current head (`504fadb6`) and checked attribution against GitHub's actual PR range from merge base `33f688f`; the branch is behind the current `main` tip (`43a781c`). The local diff matches `gh pr diff` apart from line endings: 10 changed files, +1087/-252. The PR adds 11 unit tests in `base_container_runner.rs` (57 to 68), and no High-severity defect survived source validation.

None. All five comments are on behavior or test gaps introduced by this PR.

src/backends/appcontainer/common/src/appcontainer_runner.rs
```diff
@@ -1529,9 +1518,30 @@ impl AppContainerScriptRunner {
...
 
 impl SandboxBackend for AppContainerScriptRunner {
+    fn network_policy_support(&self) -> NetworkPolicySupport {
+        // AppContainer naturally enforces host-loopback deny. Advertise the
+        // field so shared validation accepts that default posture; `validate`
+        // below rejects the unsupported allow value.
+        NetworkPolicySupport::EGRESS_DEFAULT
+            | NetworkPolicySupport::INGRESS_DEFAULT
+            | NetworkPolicySupport::HOST_LOOPBACK
+            | NetworkPolicySupport::RUNTIME_PROXY
```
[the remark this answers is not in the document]
bbonaby: Fixed. The AppContainer fallback rejects `egress.default="deny"` combined with `ingress.default="allow"` because `privateNetworkClientServer` is bidirectional and that tier has no schema 0.8 WFP filter. Existing validation coverage pins the unsupported combination.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -672,16 +689,20 @@ impl BaseContainerRunner {
...
+    /// host-loopback or peer-identity posture cannot be represented by SBOX.
     fn legacy_sbox_compatible_with_request(
         request: &ExecutionRequest,
         queried_capabilities: Option<u64>,
     ) -> bool {
+        if requires_psec_networking(&request.policy) {
+            return false;
+        }
+        if request.policy.runtime_network_proxy_specified {
+            return false;
```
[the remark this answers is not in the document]
bbonaby: Aligned at the current head. Schema 0.8 runtime proxy is PSEC-only; legacy SBOX and AppContainer fallback reject it. The networking docs, contributor architecture notes, and PR description now all state that behavior. Legacy schema proxy fallback remains unchanged.

src/backends/appcontainer/common/src/base_container_helpers.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+fn effective_capabilities(policy: &ContainerPolicy, include_host_loopback: bool) -> Vec<String> {
+    let mut capabilities: Vec<_> = policy
+        .capabilities
+        .iter()
+        .filter(|capability| !capability.is_empty())
+        .cloned()
+        .collect();
+    add_default_network_capabilities(policy, &mut capabilities);
```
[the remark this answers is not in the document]
bbonaby: Fixed. Directional networking now removes `internetClient`, `internetClientServer`, `privateNetworkClientServer`, and `networkLoopback` case-insensitively from caller capabilities before re-deriving only those selected by policy. The shared helper covers both PSEC/BaseContainer and AppContainer capability paths; legacy policies retain caller capabilities.

src/backends/appcontainer/common/src/network_policy_helpers.rs
```diff
@@ -0,0 +1,135 @@
...
+        )
+}
+
+pub(crate) fn add_default_network_capabilities(
+    policy: &ContainerPolicy,
+    capabilities: &mut Vec<String>,
+) {
+    let uses_capabilities = uses_network_capabilities(policy);
+    if uses_capabilities && allows_network_egress(policy) {
+        ensure_capability(capabilities, INTERNET_CLIENT_CAPABILITY);
```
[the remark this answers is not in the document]
bbonaby: Fixed in the shared capability helper. Directional policy strips all four policy-owned network capabilities before adding back only the required gates, while preserving unrelated capabilities and all legacy behavior. Regression coverage includes mixed-case caller values.

docs/sandbox-policy/0.8.0/networking/networking.md
bbonaby: Fixed. D5 now states that Windows schema 0.8 runtime proxy uses the PSEC BaseContainer path and is rejected by legacy SBOX/AppContainer fallback. The PR description has also been updated to remove the cooperative SBOX claim.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: Fixed. `validate_process_container_proxy_policy` rejects the reserved `MXC-Loopback` identity case-insensitively. The regression now lives beside that validation in `network_parser.rs` and directly asserts the typed configuration error.

src/backends/appcontainer/common/src/base_container_runner.rs
bbonaby: Fixed. BaseContainer now resolves the child environment once, applies the runtime proxy as an optional overlay, and uses `appcontainer_runner::encode_env_block`. The duplicate encoder/default-environment path was removed, and the existing explicit-entry proxy merge remains independently unit-tested.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: Fixed. The redundant `!policy.runtime_network_proxy_specified` branch was removed from the SBOX policy builder. `legacy_sbox_compatible_with_request` remains the single selector that rejects schema 0.8 runtime proxy before serialization.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: Fixed. Added focused IPv4-only and IPv6-only ICMP cases. Each asserts that PSEC serializes exactly one endpoint rule with the matching `icmpv4` or `icmpv6` protocol, preventing an unconstrained opposite-family rule.

src/backends/appcontainer/common/src/base_container_helpers.rs
bbonaby: Fixed. Added both allow and deny destination-only rules with `ports: []`; the test asserts that each destination survives and the FlatBuffer ports vector is absent (all ports).

docs/sandbox-policy/0.8.0/networking/networking.md
```diff
@@ -409,11 +410,11 @@ than strict model-2 enforcement.
 
 | Configuration concept | Enforcement mechanism | Notes |
 |---|---|---|
-| IP/CIDR allow/block | IPv4/IPv6 WFP filters scoped to AppContainer SID | Public and private destinations |
+| IP/CIDR allow/block | PSEC IPv4/IPv6 WFP filters scoped to AppContainer SID | Public and private destinations; unsupported on SBOX and AppContainer fallback |
 | Port filtering | Port filtering via WFP | Port ranges supported. |
 | Protocol filtering | Protocol filtering via WFP | Schema values are `tcp`, `udp`, `icmp`, and `any`; WFP maps ICMP by address family. |
-| Default-deny | WFP block-all baseline filter at lower precedence than explicit allows. AppContainer has no internetClient capability. | |
-| Proxy (HTTP/S only) | Per-AppContainer WinHTTP configuration, endpoint filtering, and optional scoped peer access | Identity-scoped proxies keep `hostLoopback: "deny"`; the identity-less development/testing compatibility path requires `"allow"` |
+| Default-deny | PSEC WFP block-all baseline filter at lower precedence than explicit allows. AppContainer has no `internetClient` capability. | |
```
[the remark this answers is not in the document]
bbonaby: Fixed in `6f8dd279`. The default-deny row now explains that a non-empty allow list grants `internetClient` only as the capability prerequisite while WFP limits egress to explicit allows; with no allows, the capability remains absent.

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -2199,15 +2038,26 @@ struct BaseChild {
 }
 
 impl SandboxBackend for BaseContainerRunner {
+    fn network_policy_support(&self) -> NetworkPolicySupport {
+        NetworkPolicySupport::ALL
+    }
```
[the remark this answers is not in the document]
bbonaby: Fixed in e20d90cc. The runner now revalidates the final resolved networking contract both during `validate()` and immediately after the launch-time PSEC probe. If a transient native-capture/PSEC probe changes to SBOX, the request fails closed with `BackendUnavailable` unless the legacy SBOX contract can fully represent it. `resolved_sbox_contract_rejects_psec_only_networking_after_capture_probe_change` covers this transition, and `resolved_sbox_contract_accepts_compatible_networking` preserves valid fallback.

docs/process-container/networking.md
```diff
@@ -1,6 +1,6 @@
 # Process Container Networking Configuration, GA
 
-The planned schema 0.8.0 ProcessContainer networking uses the shared `network.egress` and `network.ingress` policy plus
+Schema 0.8.0 ProcessContainer networking uses the shared `network.egress` and `network.ingress` policy plus
```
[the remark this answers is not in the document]
bbonaby: Fixed in 35642b6c. The sentence now describes the examples as using the schema 0.8 network shape, consistent with the page's current implemented-status wording.

SohamDas2021 (approved):

Issue #955, opened 2026-08-19
[Bubblewrap] Network policy follow-ups: incomplete parser/validate() pairs, unusable IPv6 rules, and misleading diagnostics


SohamDas2021:
### Relevant area(s)

Linux (Bubblewrap backend)

### Brief description of your issue

Five follow-ups on the Bubblewrap network policy path. Items 1 and 2 are
incomplete parser/`validate()` pairs; items 3-5 are correctness and diagnostic
gaps found in review of the firewall-enforcement change.

**None of these is a fail-open**, and none weakens the enforced posture. Items
1, 2, 4 and 5 are error-reporting or robustness defects; item 3 fails *closed*
(traffic that cannot flow) plus an inaccurate doc claim.

Bubblewrap's network gates are deliberately duplicated across two entry points —
the JSON config parser (`config_parser.rs`) and the runner's `validate()`
(`bwrap_runner.rs`) — because a Rust caller can hand `mxc_engine` a hand-built
`ExecutionRequest` and never touch the parser. `bwrap_runner.rs` states this
outright:

> The parser rejects both of these too, but only sees JSON configs; a Rust
> caller can build an `ExecutionRequest` and reach the runner directly, so the
> checks are repeated here rather than assumed.

**1. The D3 hostname rule-address check exists only in the runner.**

`validate()` rejects non-IP/CIDR rule addresses via `EgressPlan::for_policy`
(`src/backends/bubblewrap/common/src/bwrap_runner.rs`). There is no
`config_parser.rs` twin, unlike the neighbouring unenforced-host-rules and
local-network gates, which exist in both.

Both callers are still rejected, so this is purely about which error the user
sees and when: `backend_error` at validate rather than `config_parse` at parse
time.

**2. The proxy + firewall rejection exists only in the parser.**

`config_parser.rs` rejects `network.proxy` combined with
`network.enforcementMode` of `firewall` or `both` on Bubblewrap. `validate()`
has no twin, and cannot reach one as written:
`ResolvedNetworkMode::from_request` tests `proxy_active` **first**
(`src/backends/bubblewrap/common/src/bwrap_command.rs`) and returns
`ProxyOnly`, so `enforcementMode` is never consulted on that path.

A programmatic caller therefore has an explicitly-set `enforcementMode`
silently discarded. The resulting posture is not weaker — `ProxyOnly` is a
private netns with egress closed to everything but the proxy endpoint — but the
caller is never told their field was ignored.

[...]

PR #979, opened 2026-08-21
[Bubblewrap] Map the 0.8 directional network schema onto Bubblewrap and declare what it enforces

SohamDas2021:

-----

## Summary

Maps the 0.8 directional network schema (`egress` / `ingress`) onto the Bubblewrap backend, and declares in code exactly which parts of that schema Bubblewrap actually enforces.

Third and final PR of the Bubblewrap networking stack. Based on #964; review only the top commit if the earlier two are already reviewed.

### What the backend now honors

| Schema | Bubblewrap behavior |
|---|---|
| `egress.default` | `deny` -> ruleless DROP chain; `allow` -> permissive |
| `egress.allow` / `egress.deny` CIDR + port rules | lowered into the sandbox namespace iptables chains, deny-wins |
| `except` carve-outs | subtracted from their enclosing rule |
| `runtimeConfig.networkProxy` | resolves to the existing proxy-only egress posture |
| `ingress.default` / `ingress.hostLoopback` | `deny` only; `allow` is refused rather than silently ignored |

Any directional section is refused before 0.8, so pre-0.8 configs are bit-for-bit unchanged.

### Declaring capability instead of inferring it

`network_policy_support()` is a hand-written declaration of which policy fields the backend can enforce, and unsupported fields fail closed. That makes a bit that was never added indistinguishable from one deliberately refused - which is exactly what happened to `runtimeConfig.networkProxy`: the 0.8 spelling normalizes into the same `policy.network_proxy` field the legacy `network.proxy` feeds and reaches identical enforcement, but the gate refused it because nobody had declared the bit.

Declaring it surfaced two real defects:

1. **A blocker.** `external_proxy_host_rules_rejection` read `default_network_policy`, a legacy-shape field the directional path never writes, so it sat at its `Block` default on every directional request and refused all of them. That clause is now gated behind the legacy shape, matching the discriminator `EgressPlan::for_request` and `ResolvedNetworkMode::from_request` already use.
2. **A fail-open the first fix introduced.** A programmatic caller could pair directional egress rules with an external proxy, pass the relaxed guard, and have the rules silently dropped - `ProxyOnly` builds its chain from `for_proxy` and never reads `network_egress`. A `directional_conflicts` twin now refuses any directional egress that is not exactly deny-with-no-rules.

To keep this from recurring, `every_network_policy_support_bit_is_a_deliberate_decision` asserts the declared and refused bits union to `NetworkPolicySupport::ALL`. Adding a bit to the enum without recording a decision for it now fails the build.

## Commits

Reviewable one at a time, in order:

| Commit | What it does |
|---|---|
| `ff0ddd1a` | Maps the directional 0.8 network posture onto the Bubblewrap chains |
| `0453077b` | Lowers directional egress rules (CIDR, ports, `except`) into the iptables chains |
| `e26490ea` | Routes the directional posture into a private network namespace and declares what Bubblewrap enforces |
| `bd0b30c0` | Adds the directional E2E tests and documents the mapping |
| `361b18e4` | Declares `RUNTIME_PROXY` and closes the two guard gaps behind it |
## Testing

- `cargo test -p bwrap_common -p wxc_common -p lxc_common` - 1450 pass (280 in `bwrap_common`)
- `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` - clean
- `node scripts/versioning/validate-configs.js` - 250 configs
- `tests/scripts/run_bwrap_directional_test.sh` - 8 new E2E cases against real iptables in a real namespace, including a legacy/directional proxy parity pair anchored against explicit sentinels so it cannot pass by both spellings being identically broken
- `tests/scripts/run_bwrap_all_tests.sh` - 11 passed, 0 failed, 1 skipped (inbound-deny needs root)
- `sudo tests/scripts/run_bwrap_inbound_deny_test.sh` - passes

The new refusal paths were mutation-tested rather than trusted green: undeclaring `RUNTIME_PROXY`, dropping a table entry, and restoring the over-broad guard each fail for the expected reason.

Related Issues

- #63099833, #62864432, #62864253 (internal)


src/backends/bubblewrap/common/src/network_rules.rs
```diff
@@ -270,29 +297,366 @@ fn name_rejected(entry: &str) -> String {
...
+                    );
+                }
+                specs.push(PortSpec {
+                    protocol: Some("icmp"),
+                    range: None,
+                });
+            }
+            // Ports exist only for TCP and UDP, so `any` with a port narrows to
+            // those two rather than widening to every protocol on that number.
+            NetworkProtocol::Any => match range {
```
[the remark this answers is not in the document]
SohamDas2021: Tracked in #980

docs/sandbox-policy/0.8.0/networking/networking.md
```diff
@@ -418,17 +418,28 @@ Model 2 permits only the proxy endpoint.
...
+> chain described above. `ingress.hostLoopback` is bidirectional per this
+> contract, so its deny also closes the container-to-host direction: under
+> slirp that path is the gateway `10.0.2.2`, which maps onto the host's own
+> loopback, and a drop for it is lowered *ahead* of every caller rule so that a
+> broad allow — including a bare `0.0.0.0/0` — cannot reopen it on the
+> first-match chain. An omitted `ingress` section enforces the same deny, since
+> that is the schema's default rather than an absence of policy. Proxy mode
+> needs no equivalent rule: it opens the proxy endpoint alone and drops the
+> rest of the gateway, and a proxy can no longer be combined with directional
+> rules. IPv4 only — slirp gives the sandbox no IPv6 route to the host.
```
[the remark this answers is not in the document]
SohamDas2021: Tracked in #981. Confirmed: `EgressPlan::for_proxy` never reads `host_loopback` -- the gateway DROP is lowered by `for_directional_policy_with_ingress`, which proxy mode does not call -- so the value is accepted and discarded rather than partially honored. Both docs in this PR now describe that plainly instead of claiming enforcement. The remaining question is contract-level and cross-backend (LXC has the same shape), so it belongs in the shared 0.8 spec rather than here.

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -374,6 +374,147 @@ added in the same change that enables it.
...
+**What the deny postures actually do.** `ingress.default` installs the
+`MXC_INGRESS` chain on `INPUT`. `ingress.hostLoopback` is bidirectional per the
+0.8 contract, so its deny also has to close container-to-host: under slirp that
+path is the gateway `10.0.2.2`, which maps onto the host's own loopback. That
+drop is lowered *ahead* of every caller rule, because the chain is first-match
+and a broad allow — a bare `0.0.0.0/0` included — would otherwise win and the
+deny would be decorative. An omitted `ingress` section enforces the same deny,
+since deny is the schema's default rather than an absence of policy. Proxy mode
+needs no equivalent: it opens the proxy endpoint alone and closes on `DROP`, so
+the rest of the gateway is already unreachable. IPv4 only — slirp gives the
```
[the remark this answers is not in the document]
SohamDas2021: Tracked in #981. Confirmed: `EgressPlan::for_proxy` never reads `host_loopback` -- the gateway DROP is lowered by `for_directional_policy_with_ingress`, which proxy mode does not call -- so the value is accepted and discarded rather than partially honored. Both docs in this PR now describe that plainly instead of claiming enforcement. The remaining question is contract-level and cross-backend (LXC has the same shape), so it belongs in the shared 0.8 spec rather than here.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -61,6 +63,34 @@ impl BubblewrapScriptRunner {
...
+    /// the `egress.default='deny'` with no direct rules posture. That is
+    /// exactly `ResolvedNetworkMode::ProxyOnly`, which this backend already
+    /// enforces via the proxy's own private-namespace egress chain, so the 0.8
+    /// spelling reaches the identical enforcement as the 0.7 one.
+    ///
+    /// `PROXY_PEER_IDENTITY` stays undeclared: it is a ProcessContainer concept
+    /// with no Bubblewrap equivalent, so shared validation refuses it here.
+    fn network_policy_support(&self) -> NetworkPolicySupport {
+        NetworkPolicySupport::EGRESS_DEFAULT
+            | NetworkPolicySupport::EGRESS_RULES
```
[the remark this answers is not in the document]
SohamDas2021: The observation is accurate — slirp4netns runs without `--enable-ipv6` (`the_supervisor_leaves_ipv6_disabled_in_the_namespace`), so an allowed IPv6 CIDR is unreachable. But this is a known, documented limitation rather than something this PR introduces:

- **Pre-existing.** `EgressPlan::for_policy` on merged `main` already accepts IPv6 literals/CIDRs through `RuleAddress::parse`. This PR extends an existing accepted-but-inert surface; it does not create it.
- **Already documented,** with this exact caveat, in `docs/bwrap-support/bubblewrap-backend.md` (L43-46, L244-249): "An IPv6 rule programs `ip6tables`, but the sandbox's namespace has no IPv6 connectivity today ... so an allowed IPv6 destination stays unreachable regardless of the rule (see #955)." Enabling IPv6 is tracked in #955.
- **Fails closed in every combination.** `chain_lines` gives the unmatched family its own terminal verdict, asserted by `a_v4_only_policy_still_closes_ipv6` (`ip6tables -j DROP`). Under `deny` an IPv6 allow never fires; under `allow` there is no v6 route to use. This is a reachability surprise, not an escape.

On the two suggested remedies: enabling IPv6 requires the RFC 4890 ICMPv6 carve-outs in the same change (see the rationale at `network_rules.rs` `chain_lines`), which is #955's scope. Rejecting the IPv6 portions would change merged behavior in firewall mode and break configs that are valid today.

Leaving as-is and deferring to #955.

bbonaby (commented):
note: Superseded by the inline comments on the relevant changed lines.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -376,7 +409,7 @@ impl BubblewrapScriptRunner {
             // endpoint. There is no hostname to pin because rule addresses are
             // literals and CIDRs only.
             None if network_mode == ResolvedNetworkMode::FirewallEnforced => {
-                let plan = match network_rules::EgressPlan::for_policy(request) {
+                let plan = match network_rules::EgressPlan::for_request(request) {
```
bbonaby, line 412: suggestion (non-blocking, performance): Reuse the plan already derived during validation.

`validate()` calls `EgressPlan::for_request` at line 213, and this path immediately repeats CIDR subtraction and rule expansion for the same request. Returning/preparing the validated plan once would remove duplicate synchronous launch work and ensure installation uses the exact derived result that validation accepted.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -158,6 +160,44 @@ impl ResolvedNetworkMode {
...
+        // host-loopback drop. A missing egress defaults to deny, failing closed.
+        //
+        // Directional never selects `Shared`: the host namespace has no inbound
+        // primitive, and `network_ingress` has no `_specified` twin, so a
+        // defaulted `ingress.default='deny'` is indistinguishable from one the
+        // caller wrote and could not be refused without refusing every
+        // "allow all outbound" policy with it. The private namespace honors
+        // both directions, so an open egress posture becomes an accept-all
+        // chain rather than an unfiltered namespace.
+        if request.policy.network_egress.is_some() || request.policy.network_ingress.is_some() {
```
bbonaby, line 182: suggestion (non-blocking, maintainability): Centralize this directional-shape predicate.

This same two-field test (or its negation) is repeated in `directional_network_rejection`, `external_proxy_host_rules_rejection`, and `EgressPlan::for_request`. Because checking only one section can route policy into the wrong enforcement path, a shared `is_directional(&ContainerPolicy)` helper would make the fail-closed invariant harder to drift.

tests/scripts/run_bwrap_directional_test.sh
```diff
@@ -0,0 +1,391 @@
...
+    exit 1
+fi
+for sentinel in ALLOWED_PORT_OK DENIED_PORT_BLOCKED_OK; do
+    if ! grep -q "$sentinel" <<<"$PORT_OUT"; then
+        echo "$PORT_OUT"
+        echo "FAIL: directional port narrowing (sentinel '$sentinel' not found)"
+        exit 1
+    fi
+done
+echo "PASS: directional port narrowing"
```
bbonaby, line 318: suggestion (non-blocking, test): Add one live directional case for IPv6 CIDR and ICMP-to-ICMPv6 rendering.

The unit tests pin the rendered strings, but no directional E2E config sends these new rules through real `ip6tables-restore`. IPv6 connectivity is not required here; checking successful installation/debug output would catch an invalid token or restore transaction before release.

tests/configs/bubblewrap_network_directional_proxy.json
```diff
@@ -0,0 +1,20 @@
+{
+  "version": "0.8.0-alpha",
+  "containerId": "CLI-Bubblewrap-Network-Directional-Proxy",
+  "containment": "bubblewrap",
+  "process": {
+    "commandLine": "bash -c 'set -u; pu=\"${HTTP_PROXY:-${http_proxy:-}}\"; if [ -z \"$pu\" ]; then echo PARITY_NO_PROXY_ENV; exit 1; fi; echo PARITY_PROXY_ENV_OK; pa=\"${pu#*://}\"; pa=\"${pa%%/*}\"; ph=\"${pa%%:*}\"; pp=\"${pa##*:}\"; if [ \"$pp\" = \"$ph\" ]; then pp=80; fi; if ! timeout 6 bash -c \"exec 3<>/dev/tcp/$ph/$pp\" >/dev/null 2>&1; then echo PARITY_PROXY_UNREACHABLE; exit 1; fi; echo PARITY_PROXY_REACHABLE_OK; timeout 6 bash -c \"exec 3<>/dev/tcp/1.1.1.1/443\" >/dev/null 2>&1; if [ $? = 0 ]; then echo PARITY_DIRECT_LEAKED; exit 1; fi; echo PARITY_DIRECT_BLOCKED_OK; if ! curl -fsSL --max-time 15 https://api.github.com/zen >/dev/null; then echo PARITY_PROXIED_FETCH_FAILED; exit 1; fi; echo PARITY_PROXIED_FETCH_OK'"
```
bbonaby, line 6: issue (non-blocking, test): This parity test hard-depends on `api.github.com` without the preflight/skip treatment used for the other external anchors.

A GitHub outage, rate limit, DNS failure, or restricted CI network becomes indistinguishable from a proxy regression. Could we pre-probe and skip consistently when the endpoint is unavailable, or use a deterministic local upstream fixture that still proves end-to-end forwarding through `unix-test-proxy`?

SohamDas2021:
Thanks — all five evaluated, and the blocking one is fixed in the latest push.

**1. Bound cumulative peer expansion (blocking, performance) — fixed.**

The mechanism is real: `lower_peers` reset `MAX_BLOCKS_PER_PEER` per peer, the peer count is unbounded, and `lower_rule` consulted the total ceiling only after the full address vector existed.

I measured the residual amplification before fixing, to size it. Worst case I could construct was `2000::/3` minus two `/128`s — the cheap `::/0` shape is already refused by the straddling-mapped-range guard:

| | JSON in | intermediate | factor |
|---|---|---|---|
| before the `MAX_EGRESS_RULES` fix earlier in this PR | 109 KB | 677 MB | ~6,200× |
| as reported here | 136 KB | 7 MB | ~54× |

So the residual was ~54×, meaning the input had to be roughly proportional to the memory consumed — closer to ordinary parser cost than to an amplification DoS. I've fixed it regardless, since the bound is cheap and exact: a rule emits at least one entry per address (`ports.len() >= 1`), so `lower_peers` now takes the remaining budget and stops the moment the address count passes it. That caps the intermediate at `MAX_EGRESS_RULES + MAX_BLOCKS_PER_PEER` entries (~2 MB) for any input.

Covered by `peer_expansion_stops_at_the_budget`, which puts a semantically-refused peer last so that reaching it would prove the expansion ran past the budget — the test asserts the budget refusal is the one reported, not that peer's. Mutation-verified: replacing the guard with `if false` fails the test.

**2–5 — all valid, all confirmed, tracked in #991.**

- **Reuse the validated plan.** Confirmed: `bwrap_runner.rs:213` and `:412` both call `EgressPlan::for_request`.
- **Centralize the directional predicate.** Confirmed at `bwrap_command.rs:182`, `:450` and `network_rules.rs:779` (three sites). Noted on the issue that `wxc_common::validator.rs:66` already has a *related but different* `directional_posture_supplied` which also folds in the proxy, so the two must not be conflated during consolidation.
- **Live IPv6/ICMPv6 directional config.** Confirmed: none of the seven `bubblewrap_network_directional_*.json` configs contain `::` or `icmp`. Worth noting the v6 payload does already reach the real `ip6tables-restore` on every directional run and fails closed before `slirp.ready`, so the tool hop is implicitly proven — what's missing is a content assertion. Same family as #955 item 6, and the issue notes they should land together.
- **`api.github.com/zen` in the parity test.** Confirmed, and the sharpest of the four: `run_parity` turns any nonzero exit into a FAIL while the same script pre-probes `1.1.1.1`/`9.9.9.9` via `probe_anchor` and `exit 77`s. Flagged on the issue that five already-merged proxy configs share the dependency, so a fix should cover them rather than only the new config.

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: Valid — confirmed and tracked in #991 (assigned, deferred as non-blocking). Details in the summary comment on this PR.

src/backends/bubblewrap/common/src/bwrap_command.rs
SohamDas2021: Valid — confirmed and tracked in #991 (assigned, deferred as non-blocking). Details in the summary comment on this PR.

tests/scripts/run_bwrap_directional_test.sh
SohamDas2021: Valid — confirmed and tracked in #991 (assigned, deferred as non-blocking). Details in the summary comment on this PR.

tests/configs/bubblewrap_network_directional_proxy.json
SohamDas2021: Valid — confirmed and tracked in #991 (assigned, deferred as non-blocking). Details in the summary comment on this PR.

bbonaby (approved):

PR #983, opened 2026-08-21
Implement schema 0.8 network policy for the LXC backend

dhoehna:
## 📖 Description

Implements schema 0.8 network policy for the LXC backend and declares GA
network support for it.  LXC is the first backend to claim the directional
posture — `EGRESS_DEFAULT`, `EGRESS_RULES`, `INGRESS_DEFAULT`, and
`HOST_LOOPBACK`.  `RUNTIME_PROXY` and `PROXY_PEER_IDENTITY` stay unclaimed.

`network.egress` is lowered into the container's iptables chain: CIDR peers
with `except` carve-outs, single ports and inclusive ranges, and `tcp`, `udp`,
`icmp`, and `any` protocols.  IPv4 and IPv6 peers go to the `iptables` and
`ip6tables` chains respectively.  `deny[]` rules precede `allow[]` rules, and
`egress.default` becomes the chain's terminal verdict.

Three behaviors are worth calling out.

**A run carries one schema, and the backend acts on the declared version.**
`ExecutionRequest.schema_version` selects the path once, and every downstream
decision reads that selection rather than inspecting what the policy contains.
Schema 0.7 host lists keep their existing behavior.

**0.8 always enforces.**  It states a posture with no mode to opt out of, and a
0.8 request installs a chain.  Under 0.7 the enforcement mode still decides.

**DNS is not exempt under 0.8.**  The unconditional port 53 accept is carried
only on 0.7.  Under 0.8 DNS is governed like any other destination, leaving an
`egress.default: "deny"` container no standing DNS-tunnel path.  This follows
GA decision D3 in `docs/sandbox-policy/0.8.0/networking/networking.md`.  LXC's
bridge resolver is addressed to the bridge gateway and reached through the
host's `INPUT` path, which this `FORWARD` chain never sees, and ordinary name
resolution is unaffected.

### Limitations

Permissive inbound (`allowLocalNetwork: true` under a firewall-using mode) is
refused with a clear not-yet-implemented error rather than silently ignored.
The decision table is in `network_ingress_permissive_spec_tests.rs`.

## 🔗 References

AB#63505947.  GA network contract:
`docs/sandbox-policy/0.8.0/networking/networking.md`.

## 🔍 Validation

- `cargo test -p lxc_common -p wxc_common -p bwrap_common` — 359, 848, and 143
  tests pass with no warnings.
- New black-box specs cover egress lowering and inbound handling against the
  documented contract, driving the real parser end to end from JSON for both
  schema versions.
- `tests/scripts/run_lxc_all_tests.sh` on a live LXC host — 23 passed, 0 failed.
  `run_lxc_network_ga_egress_test.sh` and `run_lxc_network_v07_schema_test.sh`
  are new, covering the 0.8 and 0.7 network paths respectively.
## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/983)


src/core/wxc_common/src/models.rs
```diff
@@ -785,6 +785,54 @@ impl ContainerPolicy {
             &self.blocked_hosts,
         )
     }
+
+    /// True when this policy describes anything the egress firewall has to
```
bbonaby: suggestion: These can actually just be helper functions  in a helper file rather than inside model.rs

src/core/wxc_common/src/config_parser.rs
```diff
@@ -5396,6 +5396,50 @@ mod tests {
         assert!(request.policy.network_egress.is_none());
     }
 
+    // Backends read this one bit to decide which schema's rules to apply, so
+    // it has to follow the format the parser selected. The version string
+    // cannot stand in for it: 0.8 still accepts a legacy `network` block.
+    #[test]
```
SohamDas2021: you may want to move these tests too?

docs/linux-wsl-roadmap-june-2026.md
```diff
@@ -95,23 +95,23 @@ File:line citations reference paths under `src/backends/<backend>/...` and `src/
 
 | # | Item | Status | Description | Effort |
 |---|---|---|---|---|
-| 15 | **(N3) IP/CIDR only, no DNS names** | 🟡 Actionable | Accepts hostnames, resolves to IPv4 only. IPv6 silently dropped — dual-stack bypass. No CIDR range support. New GA schema (`egress.allow[]/deny[]` with CIDR+port+protocol) replaces `allowedHosts`/`blockedHosts`. | L |
+| 15 | **(N3) IP/CIDR only, no DNS names** | ✅ Addressed | Schema 0.8 `network.egress` is lowered into the chain with CIDR peers, `except` carve-outs, ports, and protocols. IPv4 and IPv6 peers are routed to the `iptables` and `ip6tables` chains by address family, which closes the dual-stack bypass for this shape. The legacy `allowedHosts`/`blockedHosts` path is unchanged and still resolves names. Covered end-to-end by `tests/scripts/run_lxc_network_ga_egress_test.sh`. | L |
```
bbonaby, line 98: issue (should be done in followup): Could we add a live schema 0.8 IPv6 allow/deny case before calling this covered end to end? The GA configs currently exercise only IPv4; the existing live IPv6 test uses the legacy host-list shape, so the parser-to-`ip6tables` path for directional peers is still covered only by unit assertions.

tests/configs/lxc_network_ga_egress_allow.json
```diff
@@ -0,0 +1,26 @@
+{
+  "version": "0.8.0-alpha",
+  "containerId": "CLI-LXC-GA-Egress-Allow",
+  "containment": "lxc",
+  "process": {
+    "commandLine": "sh -c \"wget -qO- --timeout=10 https://api.github.com/zen >/dev/null 2>&1 && echo MXC_NET_ALLOWED || echo MXC_NET_BLOCKED\""
```
bbonaby, line 6: issue (non-blocking): Could we use a CI-controlled peer for this positive-path check instead of unauthenticated `api.github.com`? A prior LXC run received a 403 rate-limit response and reported the policy as over-blocking, so remote service health is currently indistinguishable from a firewall regression.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -1230,9 +1372,218 @@ impl NetworkIptablesManager {
...
+        let matches = Self::lower_port_selectors(&rule.ports);
+        let wildcard_peers;
+        let peers = if rule.to.is_empty() {
+            wildcard_peers = Self::every_destination_peers();
+            &wildcard_peers[..]
+        } else {
+            &rule.to[..]
+        };
+
+        for peer in peers {
```
bbonaby, line 1502: issue (non-blocking): Could we either bound the `to`/`except`/`ports` cardinalities or batch these through `iptables-restore`? This expansion multiplies peers by selectors and exclusions, and every emitted entry currently becomes a separate synchronous `iptables`/`ip6tables` process during container startup.

src/backends/lxc/common/src/network_iptables.rs
```diff
@@ -1230,9 +1372,218 @@ impl NetworkIptablesManager {
...
+            for matching in &matches {
+                // The carve-out takes the direction's default verdict rather
+                // than the rule's own — `except` excludes the range from the
+                // rule; it does not reverse it. Nothing is pushed when the
+                // two verdicts already agree.
+                if default_action != action {
```
SohamDas2021, line 1508: (Blocking): This exception ACCEPT can shadow later explicit deny rules. For example, deny 10/8 except 10.10/16 followed by deny 10.10.1/24 permits the /24 because the earlier ACCEPT wins. Please ensure exceptions affect only their parent rule and add this regression.

tests/scripts/run_lxc_network_ga_egress_test.sh
```diff
@@ -0,0 +1,153 @@
...
+assert_allowed() {
+    if echo "$CASE_OUTPUT" | grep -Fq "MXC_NET_BLOCKED"; then
+        fail "$1"
+    fi
+    if ! echo "$CASE_OUTPUT" | grep -Fq "MXC_NET_ALLOWED"; then
+        fail "the case produced no verdict at all; the container command did not run."
+    fi
+}
+
+echo "Running LXC schema 0.8 egress enforcement test..."
```
SohamDas2021, line 123: (Non-blocking) The live suite covers core IPv4 cases, but IPv6, ICMP/ICMPv6, port ranges, and protocol: any remain rule-level only. Please add representative live coverage.

bbonaby (approved):

PR #1029, opened 2026-08-24
[Bubblewrap] Report proxy-enforcement host support from getPlatformSupport

### LINKED ISSUE #994: [Bubblewrap] Report a host network support tier from getPlatformSupport so callers can tell if proxy mode is supported

On Windows, a caller can ask the host what it will actually get *before* launching a sandbox. `PlatformSupport.isolationTier` reports `base-container` / `appcontainer-bfs` / `appcontainer-dacl` plus `isolationWarnings` explaining any degradation, sourced from the native fallback probe.

Bubblewrap has no equivalent. `getPlatformSupport()` on Linux returns only `isSupported` and `availableMethods: ['bubblewrap']`, derived from a `bwrap` version check. A caller cannot discover whether the host can actually satisfy a schema 0.8 network policy — in particular whether **model 2 (loopback proxy / private network namespace)** is supported — until the run fails.

This came up in review of the Bubblewrap proxy-mode private-network-namespace work:

> "Can we also have it return a support tier (like we do on Windows) so the caller can know if proxying is supported in the host? (it has all dependencies)"

**Why the existing surface does not answer this.** `BubblewrapRunner::network_policy_support()` already advertises `RUNTIME_PROXY`:

https://github.com/microsoft/mxc/blob/main/src/backends/bubblewrap/common/src/bwrap_runner.rs#L86-L92

but that is a **static, compile-time declaration** of what the backend *understands*, consumed by `validate_network_policy_support` to reject unhonorable policy. It is identical on every Linux host and never reaches the caller. A host missing the runtime dependencies still advertises `RUNTIME_PROXY` and then fails at launch — the Linux analogue of #785, where the Windows probe reports a launchable tier for runs that fail.

The host dependencies that actually vary are not probed anywhere:

- `iptables` availability and `CAP_NET_ADMIN` — required for per-host filtering (`bwrap_runner.rs:19-21`, `NetworkIptablesManager`); today the failure surfaces only as "failed to apply iptables firewall rules" at `bwrap_runner.rs:214`.
- Unprivileged user-namespace creation being permitted by the host (distro/sysctl/AppArmor dependent).
- Any userspace networking helper the private-namespace proxy mode ends up requiring.

`git grep -rn "slirp4netns\|pasta" -- src` currently returns nothing, so there is no probe for these at all.

**Proposed behavior.** Extend the Linux path of `getPlatformSupport()` (and the Rust `mxc_engine::platform_support`) to report a Bubblewrap network capability/tier alongside `availableMethods`, so callers can branch to model 1 vs model 2 up front instead of discovering failure mid-run. Degradation reasons should be reported the way `isolationWarnings` already are on Windows, so the answer is actionable rather than a bare boolean.

- Add a runtime probe next to `bwrap_common::bwrap_version::probe_bwrap` covering the dependencies above, and surface its result through `mxc_engine::platform::platform_support` (`src/core/mxc_engine/src/platform.rs`) and `sdk/node/src/platform.ts`.
- Keep the reported value distinct from `NetworkPolicySupport`: that bitflag stays the static "what the backend understands" contract used by validation; the new field is "what this host can actually do".
- Follow the existing Windows shape (`IsolationTier` + warnings, `sdk/node/src/types.ts:459-470`) rather than inventing a new reporting convention, so the two platforms stay legible side by side.
- Probe failure must fail closed — report the capability as unavailable rather than assuming it is present. See #975 and #878 for LXC cases where a failed probe was read as permissive.

**Related**

- #769 — CI gate ensuring every tier/backend stays probe-covered (would want to cover this new probe).
- #785 — `wxc-exec --probe` reports a launchable tier for runs that fail (same class of over-reporting, on Windows).
- #935 — Bubblewrap proxy-only egress: the helper dying mid-run leaves a broken network with no error (runtime counterpart to this pre-flight gap).
- #975, #878 — LXC probes failing open.

SohamDas2021:
## 📖 Description

-----

## Summary

Schema `0.8.0-alpha`+ Bubblewrap proxy mode runs the sandbox in a private network namespace and default-drops everything except the proxy endpoint. That needs host tooling (`slirp4netns`, `unshare`, `nsenter`, `iptables`) and there is deliberately **no fallback** — a policy the host can't satisfy fails rather than silently degrading to the weaker shared-host-network model, since silent degradation would reintroduce, the proxy bypass this mode exists to close.

Today a caller can only discover a missing dependency by failing a run. This surfaces the fact up front, mirroring how Windows reports its isolation tier.

### What this adds to the probe path

This is not purely plumbing of an existing check. Alongside exposing the
result through `getPlatformSupport()` / `--available-backends`, the
pre-flight walk gained real probe logic:

- **An active user/network-namespace probe.** `--help` only proves `unshare`
  was compiled with `--map-current-user` and `--keep-caps`; the kernel can
  still refuse the namespaces (unprivileged userns disabled, AppArmor's
  `kernel.apparmor_restrict_unprivileged_userns`, or an exhausted
  `user.max_net_namespaces`). The walk now runs
  `unshare --user --map-current-user --keep-caps --net -- sh -c :` so that
  refusal surfaces here instead of deep inside supervisor startup. `sh -c :`
  rather than `true` because the supervisor really does run as `sh -c`.
- **A new aggregate deadline.** `probe_proxy_enforcement()` bounds the whole
  walk by `PRE_FLIGHT_BUDGET` (3s), tighter than the sum of the per-command
  `PROBE_TIMEOUT`s the launch path allows itself, because a caller asking
  "can this host do it?" before spawning cannot wait for the worst case.
  Timeouts are attributed: a command clamped by the residual budget is
  reported as the *budget* running out, not as a hung tool.
- **Bounded output capture.** Probe output is captured and size-capped rather
  than read unbounded, and a banner that exceeds the cap is refused rather
  than judged truncated.

Deliberately **not** added: an `nf_conntrack` check. The module is
demand-autoloaded, so its absence from `/proc/modules` is not evidence the
ingress chain will fail. It stays a launch-time failure, where the
`iptables-restore` transaction rolls back and aborts before the workload runs.

After merging #723, the bounded-execution mechanics behind both this walk and
the `bwrap --version` gate live in one shared `probe_exec` module.

### What callers get

- **Rust** — `mxc_engine::platform_support()` gains `bubblewrap_network`.
- **TypeScript** — `getPlatformSupport()` gains `bubblewrapNetwork`: `{ proxyEnforcement: 'supported' | 'unsupported', warnings: string[] }`.
- **CLI** — new `lxc-exec --available-backends` emits `available_backends()` as JSON and exits, handled immediately after arg parsing so it mutates no host state.

Bubblewrap now advertises a `proxyEnforcement` capability when supported, and a `warnings` entry naming the missing dependency when not; the backend still appears either way, preserving the API's "absence means unusable" rule.

The result is advisory, not a gate — the runner still probes at launch (TOCTOU, and the SDK probe is a separate process that doesn't share the success cache).

### Note on the flag name

`docs/backend-support-probe-api-plan.md` §2 reserves `--probe` for the per-backend *diagnostics* payload (an object) and §6 proposes `--available-backends` for this API (an array), so they are separate flags — one name must not carry two schemas.
The Windows `--probe` surface is untouched.

### Scope limits

- Linux only; Windows and macOS unchanged.
- This does **not** yet make `getPlatformSupport()` a pure projection of the native probe on Linux — the `lxc-ls` / `bwrap --version` checks still run in TypeScript. The drift risk in the plan doc's §4.2 step 2 remains open.
- Host-fact reporting only — no schema change, no config input, no codegen gates. `PlatformSupport` is SDK-local (absent from `wire.rs` / `generated/`), and it should stay that way.

## Validation

- `cargo fmt --all -- --check`, `cargo clippy … -D warnings` (Linux **and** Windows targets), `cargo test -p mxc_engine` (69 pass)
- `npm run build` + `npm test` in `sdk/node` (284 pass / 0 fail), including 7 new fail-closed parser tests
- `check-schema-versions.js`, `check-version-sync.js` green
- Rust helpers were split into pure functions (`bubblewrap_backend`, `bubblewrap_network_support`) so both branches are testable without a host that has or lacks `slirp4netns`

## Docs

`docs/backend-support-probe-api-plan.md` (struct, testing rules, §6 status),
`docs/bwrap-support/bubblewrap-backend.md` (new "Checking host support before you run"), `sdk/node/README.md`, `.github/copilot-instructions.md`.

Closes #994

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1029)


src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -1229,6 +1229,16 @@ impl PrivateNetworkUse {
...
 }
 
+/// Whether this host can enforce proxy-only egress, and why not when it cannot.
+///
+/// Pre-flight counterpart to the launch-path check in
+/// [`crate::bwrap_runner`]: schema 0.8 proxy policy has no fallback, so a
+/// caller that cannot ask this ahead of time only learns the answer when the
+/// run fails. Advisory — the runner still probes before it launches.
+pub fn probe_proxy_enforcement() -> Result<(), String> {
+    probe_dependencies(PrivateNetworkUse::ProxyOnlyEgress)
```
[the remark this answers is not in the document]
SohamDas2021: Partially addressed in 61baf12 — leaving this open because I deliberately did not implement all of it.

**Landed: the namespace check.** `probe_dependencies` now actually creates the namespace instead of trusting `unshare --help` to prove the flags are compiled in:

`unshare --user --map-current-user --keep-caps -- true`

A host with unprivileged user namespaces disabled now fails the probe with a remedy (`kernel.unprivileged_userns_clone` / `user.max_user_namespaces`) rather than reporting `Ok(())` and dying at startup. Worth noting this gap was backend-wide rather than new here — `probe_bwrap()` is likewise just `bwrap --version` — but you're right that it matters most on a surface explicitly advertised as fail-closed.

**Also landed, related:** the pre-flight walk is now bounded by a 3s aggregate budget, and `bwrap --version` is no longer an unbounded wait. Previously eight dependencies at 5s each could outlive the SDK's own timeout, and on a fail-closed surface being killed from outside reports a perfectly usable host as `unsupported`.

**Not landed: the `nf_conntrack` check.** `nf_conntrack` is demand-autoloaded — it is routinely absent from `/proc/modules` on a host where rule installation will succeed, because the first `conntrack`-matching rule loads it. So absence is not evidence of failure, and since this surface fails closed, keying off that signal would refuse working hosts. That is a worse failure than the one it prevents: a false `unsupported` silently steers callers away from a mode they could have used, whereas the true failure already surfaces at launch with a targeted diagnostic (`proxy_network.rs`, the `conntrack_hint` path names the missing module). I've documented it as a known limitation on `probe_proxy_enforcement` and in the backend guide instead.

**Not landed: installing a representative rule.** Same reasoning, plus cost: a meaningful test means standing up a disposable netns and running a real `iptables` transaction against the xtables lock, which is both slow for a pre-flight call and racy against a concurrent run holding that lock. Contention would read as `unsupported`.

Net: the probe is strictly stronger and still explicitly advisory — the runner re-probes at launch, since the pre-flight answer comes from a different process at an earlier time and a package can disappear in between. Happy to reconsider the rule-installation half if you think the false-negative risk is worth it.

sdk/node/README.md
```diff
@@ -148,6 +148,20 @@ The default `processcontainer`, `bubblewrap`, `lxc`, and `seatbelt` backends wor
...
+const network = getPlatformSupport().bubblewrapNetwork;
+if (!network) {
+  throw new Error('Bubblewrap is not available on this host');
+}
+if (network.proxyEnforcement !== 'supported') {
+  throw new Error(`proxy mode unavailable: ${network.warnings.join('; ')}`);
+}
+```
+
+It is reported **fail closed**: if the probe cannot run, the result is `'unsupported'` with the reason in `warnings`, never absent. The check is advisory — the runner still verifies the dependencies at launch, since the probe runs in a different process at an earlier time. See [the Bubblewrap backend guide](../../docs/bwrap-support/bubblewrap-backend.md#checking-host-support-before-you-run).
```
bbonaby, line 163: note: We'll have to revisit this when this SDK moves to an in proc one.

sdk/node/src/platform.ts
```diff
@@ -173,6 +174,89 @@ function populateIsolationFromProbe(support: PlatformSupport): void {
...
+ * Ask `lxc-exec --available-backends` whether this host can enforce Bubblewrap
+ * proxy-only egress.
+ *
+ * Unlike the Windows isolation probe, this reports **fail closed**: a missing
+ * binary, timeout, or malformed payload yields `'unsupported'` with the reason
+ * as a warning, never an absent field. Reporting "unknown" as "supported"
+ * would let a caller launch a 0.8 proxy policy the host cannot satisfy, and
+ * that policy has no fallback.
+ */
+export function _probeBubblewrapNetwork(): BubblewrapNetworkSupport {
```
bbonaby, line 291: question: do we need to merge any ideas from https://github.com/microsoft/mxc/pull/723 ?

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -662,6 +662,58 @@ request fails if its private namespace cannot be configured.
 > so the same config on `0.6`/`0.7` runs the legacy shared-host-network proxy
 > path instead.
 
+### Checking host support before you run
```
bbonaby, line 701: note: we'll also need this in rust sdk as well. I can see us pass in the containment type we want to some publically exposed probe function and we return the results. We should probably check if Huzaifa's probe stuff can be augmented a bit as well.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -1282,6 +1376,34 @@ fn probe_dependencies_uncached(use_case: PrivateNetworkUse) -> Result<(), String
         ));
     }
 
+    // `--help` only proves the flags are compiled in. The kernel can still
+    // refuse the namespaces (unprivileged user namespaces disabled, or
+    // user.max_net_namespaces exhausted), which would otherwise surface deep
+    // inside supervisor startup. Both are requested together, mirroring the
+    // CLONE_NEWUSER|CLONE_NEWNET bwrap itself asks for.
+    let mut userns_command = Command::new("unshare");
```
bbonaby: issue (blocking): Add regression coverage for namespace refusal

This new check is the fail-closed boundary for hosts that disable or exhaust user/network namespaces, but no test exercises the non-zero `unshare` result. Please add a regression test that makes this command fail and verifies the probe reports `Unsupported` with the actionable namespace-limit guidance.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -1229,8 +1296,32 @@ impl PrivateNetworkUse {
...
+/// rules are installed, so it still surfaces at launch.
+///
+/// Bounded by [`PRE_FLIGHT_BUDGET`] in total, unlike the launch-path walk: a
+/// caller answering "can this host do it?" before spawning cannot afford the
+/// sum of every per-command timeout.
+///
+/// Uncached in both directions. Sharing the launch cache would let one
+/// advisory call report support long after the tooling was removed, and would
+/// let it stand in for the fresh walk the runner promises to do at launch.
+pub fn probe_proxy_enforcement() -> Result<(), String> {
```
bbonaby, line 1405: issue (blocking): Exercise one complete successful pre-flight walk

The deadline tests cover individual timeout and exhausted-budget branches, but none prove that the full dependency sequence can complete successfully within `PRE_FLIGHT_BUDGET`. Please add a test with fast successful probe commands that drives `probe_proxy_enforcement()` through the whole walk and returns `Ok(())`; this will catch future checks that accidentally make supported hosts fail closed.

src/core/lxc/src/main.rs
```diff
@@ -125,6 +129,20 @@ fn main() {
 
     let cli = Cli::parse();
 
+    // Detection-only fast path used by SDK `getPlatformSupport()`; runs before
+    // config handling so no JSON file is needed just to ask what the host can
+    // do, and mutates no host state.
+    if cli.available_backends {
```
bbonaby, line 148: suggestion (non-blocking): Add an end-to-end contract test for this CLI path

The Rust JSON serializer and TypeScript parser are tested separately, but there is no test that invokes `lxc-exec --available-backends` and projects its output into `PlatformSupport.bubblewrapNetwork`. A small integration test would pin the transport contract and catch drift between the native array shape and the SDK parser.

src/backends/bubblewrap/common/src/bwrap_version.rs
```diff
@@ -162,6 +178,48 @@ pub fn probe_bwrap() -> Result<BwrapVersion, BwrapUnavailable> {
     check_version_output(&stdout)
 }
 
+/// Wait for `child`, killing it if it outlives [`VERSION_PROBE_TIMEOUT`].
+///
+/// The pipes are drained only after exit -- safe for a version banner, but it
+/// would deadlock for a command that can fill one.
+fn wait_bounded(mut child: Child) -> Result<Output, String> {
```
bbonaby: suggestion (non-blocking): Share the bounded child-wait implementation

This adds a second `try_wait` / sleep / kill / wait loop alongside `proxy_network::run_probe`, which this PR is also extending with budget attribution. Extracting the common wait/termination core would keep timeout and cleanup semantics from drifting between the two probe paths.

src/backends/bubblewrap/common/src/proxy_network.rs
SohamDas2021: Added. A host where namespaces work can't be made to produce these statuses, so the classifier is now a pure `namespace_probe_error(status, use_case)` and two tests drive it: exit 1 asserts the refusal names every knob that could be responsible (AppArmor, `unprivileged_userns_clone`, both quotas), and 126/127 assert an exec failure is reported as a missing `sh` rather than a refused namespace.

src/backends/bubblewrap/common/src/proxy_network.rs
SohamDas2021: Added `the_whole_dependency_walk_passes_on_a_host_that_meets_every_requirement`. No command seam needed: `prepare-linux-host.sh` installs slirp4netns/util-linux/iptables on every CI Linux host, and `probe_proxy_enforcement()` is uncached, so the test drives the real walk end to end. It verified `Ok(())` locally on a fully equipped host. On a less-equipped one it instead asserts the failure names the specific dependency, so the walk can never fail closed anonymously; it also pins the elapsed time against the budget.

src/core/lxc/src/main.rs
SohamDas2021: Agreed there's a gap, but it needs a built `lxc-exec` plus a cross-language harness, so I'd rather not grow this PR with it. Filing a follow-up issue to pin the transport contract.

src/backends/bubblewrap/common/src/bwrap_version.rs
SohamDas2021: Fair, and the duplication is mine. Holding off deliberately: #723 rewrites this exact loop (adds `ProbeGate`, `MAX_WAIT_POLL_INTERVAL`, process-group termination), so extracting a shared core now would make that conflict materially worse for ~25 lines of dedup. Better done once #723 lands, on top of its version.

docs/bwrap-support/bubblewrap-backend.md
SohamDas2021: This is already in the Rust SDK as of this PR: `mxc_sdk::platform_support().bubblewrap_network` gives the proxy-enforcement tier plus warnings, and `available_backends()` returns the full host-capability set with tiers, which is close to the "ask about a containment type, get results back" shape you describe.

sdk/node/README.md
SohamDas2021: Agreed - the caching wording in particular is specific to spawning the executor, so it'll need revisiting when the SDK goes in-proc.

sdk/node/src/platform.ts
SohamDas2021: Good catch - the overlap is bigger than it looks. #723 (+2668/-138, currently CONFLICTING, last updated Aug 10) independently implements three things this PR touches:

- 64 KiB output caps, which duplicate the `MAX_CAPTURE` added here
- process-group termination and descendant reaping, which is exactly the remaining hardening item on this PR's probe path
- cached advisory / `probe_bwrap_uncached` for validation, which duplicates the cache split done here in an earlier review round

So I don't think this is a merge so much as a sequencing decision - building process groups here too would guarantee a painful conflict in `bwrap_version.rs`. 

Following up with @gudge directly on landing order.

SohamDas2021:

SohamDas2021:
Blocked on https://github.com/microsoft/mxc/pull/723

bbonaby:
issue (blocking): hey looks like there is a merge conflict on this one

@SohamDas2021

SohamDas2021:
yes, this is on hold, will merge this after Gudge's https://github.com/microsoft/mxc/pull/723

> issue (blocking): hey looks like there is a merge conflict on this one
> 
> Soham Das (@SohamDas2021)



> issue (blocking): hey looks like there is a merge conflict on this one
> 
> Soham Das (@SohamDas2021)

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -682,6 +683,60 @@ request fails if its private namespace cannot be configured.
...
+So that callers do not have to discover a missing dependency by failing a run,
+the host requirements are reported ahead of time.
+
+From the command line:
+
+```bash
+lxc-exec --available-backends
+```
+
+This emits the available backends as JSON and exits without launching anything.
```
bbonaby: thought: I think this one should probably say something like "without running the workload" right?

src/core/lxc/src/main.rs
SohamDas2021: This landed rather than being deferred — the contract test now exists as `lxc-exec --available-backends contract` in `sdk/node/tests/integration/linux-bubblewrap.test.ts`.

It is a CLI-level transport test, not a unit test of the probe: it spawns the real `lxc-exec` binary, captures its actual stdout, and feeds those bytes to the real SDK parser. Three cases:

1. **`emits the array shape the SDK parser consumes`** — pins the wire shape the parser depends on: a JSON array, every entry carrying a string `backend` wire name, and `capabilities` / `warnings` as arrays of strings whenever present. The parser reads those two as string arrays and would silently fall back to `unsupported` if either became an object or a bare string.

2. **`agrees with getPlatformSupport on whether bubblewrap is usable`** — asserts the native payload and `getPlatformSupport()` reach the same verdict. `MIN_BWRAP_VERSION` is mirrored between `bwrap_version.rs` and `platform.ts`, so this catches drift between the two version floors instead of hiding behind it.

3. **`projects the native payload into PlatformSupport.bubblewrapNetwork`** — the projection you asked for: it takes the capability list the CLI actually emitted and asserts `bubblewrapNetwork.proxyEnforcement` matches it, with an unsupported host always carrying a reason. The captured bytes are injected rather than re-probed, so a second live walk exhausting its pre-flight budget on a loaded machine can't make it flaky.

Verified against a real Linux host rather than reasoned about: built `lxc-exec` under WSL (bwrap 0.9.0) and ran the real CLI, which reports

```json
[
  { "backend": "bubblewrap", "capabilities": ["proxyEnforcement"] },
  { "backend": "lxc" }
]
```

The integration file passes 6/6 there.

That covers both halves of the ask — invoking the CLI path, and projecting its output into `PlatformSupport.bubblewrapNetwork` — so the Rust serializer and TypeScript parser can no longer drift apart unnoticed.

docs/bwrap-support/bubblewrap-backend.md
jsidewhite: @bbonaby and C#?

SohamDas2021: jfyi, this was an ask from Carlos and I don't think they are using the C# one yet, right? If that is the case, I would prefer not to block this one, I would like to do the C# integration in another PR.

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -695,6 +698,60 @@ request fails if its private namespace cannot be configured.
 > so the same config on `0.6`/`0.7` runs the legacy shared-host-network proxy
 > path instead.
 
+### Checking host support before you run
+
```
bbonaby, line 702: question: Do you have access to Github copilots internal codebase? I'd like you to see what Carlos is saying about bubblewrap. This is what they are passing when they create an MXC today:

Image

Notice for hostloopback they always pass deny. I Created a bug on their side about this for the windows side and in general. For Linux (bubblewrap) is it expected that they can't use the localhost? I see the text below that states  that we cannot configure private networking but want to make sure if this includes localhost.

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -695,6 +698,60 @@ request fails if its private namespace cannot be configured.
...
+per-tool timeout the launch path allows itself, so `getPlatformSupport()`
+answers promptly. A host slow enough to exhaust that budget is reported
+`unsupported` with a warning naming the timeout — the run itself is not
+subject to that budget, so such a host may still launch successfully.
+
+It is also not exhaustive. It confirms `slirp4netns`, the `iptables` tooling and
+a usable backend, and that the kernel actually grants the unprivileged user and
+network namespaces `bwrap` will ask for — but a missing `nf_conntrack` module is
+only detectable once the rules are installed, so that case still surfaces at
+launch with a targeted error message rather than here.
```
bbonaby: thought: we probably should align on how the probing works. I also have probing work that probably needs to get done as well. I see https://github.com/microsoft/mxc/blob/3a22a2b94a81da081d260b087c4c7faeb7be7f14/docs/backend-support-probe-api-plan.md#L4 and I'm wondering if this PR effectively updates mechanism with a capabilities section?

docs/backend-support-probe-api-plan.md
```diff
@@ -27,6 +27,25 @@ pub struct AvailableBackend {
...
     pub tier: Option<String>,
+    /// Optional backend features usable on this host. Omitted when empty.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub capabilities: Vec<BackendCapability>,
+    /// Why an optional capability is absent from `capabilities`. Omitted when
+    /// empty. This is the only place the API says why something is *not*
+    /// available, and it is scoped to optional capabilities — a backend that is
+    /// unusable outright is still simply absent (§2).
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub warnings: Vec<String>,
```
bbonaby, line 38: praise: nice, the backend capability is actually what I want to use for basecontainer. That said, can you think of a way to make this so we don't have a BackendCapability struct with all flavors of backend specific options? For example, it currently has "CaptureDenials" but that's process container specific. Anything we can do with traits here? Although I guess we'd have to think about how that gets represented in the other SDK layers.

docs/backend-support-probe-api-plan.md
```diff
@@ -27,6 +27,25 @@ pub struct AvailableBackend {
...
+}
+
+/// Optional feature supported by a containment backend on the current host.
+#[derive(Debug, Clone, Copy, serde::Serialize)]
+#[non_exhaustive]
+pub enum BackendCapability {
+    /// Windows ProcessContainer denial capture.
+    CaptureDenials,
+    /// Bubblewrap proxy-only egress in a private network namespace.
+    ProxyEnforcement,
```
bbonaby, line 48: nice: definitely need a loopback one too I think.

docs/backend-support-probe-api-plan.md
```diff
@@ -150,6 +182,18 @@ side-effect-free transport (e.g. a new `wxc-exec --available-backends` mode hand
...
+   `available_backends()` and exits, and it is handled immediately after argument parsing —
+   before any config load or container work — so it is read-only with respect to host state.
+   The SDK uses it on Linux to populate `PlatformSupport.bubblewrapNetwork`.
+
+   Two deliberate scope limits: the flag is **not** yet the transport for the rest of
+   `getPlatformSupport()` on Linux (the `lxc-ls` and `bwrap --version` checks still run in
+   TypeScript, so the drift risk in step 2 is not yet closed), and the Windows side is
+   untouched. The name `--available-backends` is used rather than `--probe` for the reason
+   above *and* because §2 reserves `--probe` for the per-backend **diagnostics** payload,
+   which has a different (object, not array) shape; one flag name must not carry two schemas.
```
bbonaby: issue: I think we should be careful of going into too much specifics or in the weeds for the backends. In this case for bubblewrap this all might be a big too mush detail. I guess how I see the docs is that, someone outside coming into the repo, would they care about the nitty gritty details or just "ok on this platform if this thing doesn't exist X should happen".

sdk/dotnet/README.md
```diff
@@ -132,13 +132,57 @@ every backend the host can run, including lifecycle-only backends such as
...
+    Console.Error.WriteLine(string.Join(Environment.NewLine, bubblewrap.Warnings));
+}
+```
+
+Reported **fail closed**: when the probe cannot run, the capability is absent
+with a warning, never reported as supported. The check is advisory rather than
+a gate — the runner probes again at launch, because a package can be removed in
+between — and not exhaustive, since a missing `nf_conntrack` module is only
+detectable once the rules are installed and so still surfaces at launch.
+
```
bbonaby, line 182: issue: same thing here, we probably shouldn't go into too many backend specific things in the SDK readme, and rather point to the probe API doc if need be. Having an example usage in the readme is still good though. I just don't want us to get into too much implementation detail in the .mds.

In the future I'm going to propose we keep the files in docs/ as public facing docs for usage/high level design, and then for each Backend or anything that requires it we put a README.md in the module which can go into implementation details. CC: @jsidewhite

src/backends/bubblewrap/common/src/probe_exec.rs
```diff
@@ -0,0 +1,626 @@
...
+//!   descendant that inherited it blocks the read past the deadline, and killing
+//!   the direct child does not help.
+//! * A file fixes that but bounds only how much output is *read* — a wedged
+//!   descendant keeps writing to the unlinked inode.
+//!
+//! So the child gets its own process group, its pipes are drained by readers
+//! that `poll()` against the deadline, and the group is `SIGKILL`ed on timeout
+//! *and* once the leader exits — bounding what the child can write, not just
+//! what we agree to read. The leader stays unreaped until both readers finish so
+//! its PID cannot be recycled out from under the group sweep.
```
bbonaby, line 20: issue: hmm for the rust sdk that distributes by source, how would it use the probe API? This is something we'll have to figure out. Lets say we were to get rid of the probe executable, is there anything else we can do?

bbonaby (dismissed):

docs/bwrap-support/bubblewrap-backend.md
SohamDas2021: I don't think I have access. 

Confirming Bubblewrap 0.8 onwards refuses `ingress.hostLoopback: "allow"`, so deny is the only reachable posture and it's bidirectional (also dropping container->host egress to `10.0.2.2`, except the proxy endpoint). *Host* loopback is therefore unreachable, because slirp installs no route in and the schema has no port list to configure one; the spec marks this "not yet covered."

docs/backend-support-probe-api-plan.md
SohamDas2021: Shall we do this in a follow-up?

docs/backend-support-probe-api-plan.md
SohamDas2021: Prefer to scope this pr to proxy only, filed file a follow-up for hostloopback-https://github.com/microsoft/mxc/issues/1152

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -1229,8 +1382,78 @@ impl PrivateNetworkUse {
...
+    })
+}
+
+/// Run `walk` on a worker and return by `budget` whatever it is doing.
+///
+/// The walk's own budget bounds each command it *waits* on, but not the spawn
+/// itself: a PATH lookup or exec can block on a stalled filesystem before any
+/// deadline is consulted.
+///
+/// On timeout the worker is **abandoned, not joined**. Joining would
```
MGudgin, line 1420: **High (reliability) - Bound the number of abandoned preflight workers.**

**Attribution:** `introduced_by_change` - this PR adds the uncached worker-per-call supervisor and explicitly drops the worker on timeout.

If `Command::spawn` blocks, the three-second timeout returns but this thread can remain alive indefinitely. Because every advisory call starts another worker, a long-lived process can accumulate stuck threads until it exhausts memory or thread capacity.

**Fix:** use a killable helper process, or add a single-flight/stuck-probe gate so at most one abandoned worker can exist and later callers fail fast until it completes.

src/backends/bubblewrap/common/src/proxy_network.rs
```diff
@@ -1243,16 +1466,91 @@ pub(crate) fn probe_dependencies(use_case: PrivateNetworkUse) -> Result<(), Stri
     if PROBED.get().is_some() {
         return Ok(());
     }
-    probe_dependencies_uncached(use_case)?;
+    probe_dependencies_uncached(use_case, ProbeBudget::unbounded())?;
```
MGudgin, line 1469: **Medium (correctness/test coverage) - Keep the active namespace check out of the success cache.**

**Attribution:** `newly_exposed_by_change` - the cache predates this PR, but the PR adds a volatile `unshare` allocation check to the walk it caches.

A successful first launch now caches the namespace result along with stable tool/version checks. Later launches skip the active check even if `user.max_net_namespaces` is exhausted or AppArmor/sysctl policy changes, so they lose the early diagnostic this PR is adding.

**Fix:** cache only stable dependency facts and run the namespace-allocation check on every launch. Please also extract an injectable cache helper and test success reuse, failure retry, and cross-use-case reuse.

sdk/node/src/platform.ts
```diff
@@ -221,6 +364,7 @@ function computeSupport(): PlatformSupport {
     const bubblewrap = _probeBubblewrap();
     if (bubblewrap.available) {
       methods.push('bubblewrap');
+      support.bubblewrapNetwork = _probeBubblewrapNetwork();
```
MGudgin, line 367: **Medium (reliability/performance) - Avoid putting this volatile, expensive result into the existing lifetime cache.**

**Attribution:** `introduced_by_change` - this new assignment runs after the TypeScript LXC/Bubblewrap checks, then becomes part of `cachedSupport` for the rest of the module lifetime.

The first call synchronously performs partly duplicated native checks, while later calls can keep reporting proxy enforcement as supported after tooling or namespace policy changes. This gives callers the worst trade-off: a relatively expensive cold call and a stale warm result.

**Fix:** preferably make the native probe authoritative for all Linux backend facts in one invocation, and refresh the network readiness separately (per call or with a short TTL). If synchronous probing is required, document the cost and consider an async API.

sdk/node/src/platform.ts
```diff
@@ -188,6 +196,141 @@ function populateIsolationFromProbe(support: PlatformSupport): void {
...
+
+let linuxProbeRunner: LinuxProbeRunner = defaultLinuxProbeRunner;
+
+/**
+ * Worst case for the native `--available-backends` walk on Linux, whose probes
+ * run **sequentially**: 5s `bwrap --version` (`BWRAP_VERSION_TIMEOUT`) + 3s
+ * proxy dependency walk (`PRE_FLIGHT_BUDGET`) + 3s `lxc-ls --version` (LXC
+ * `PROBE_TIMEOUT`). Keep in sync with those Rust constants.
+ */
+const NATIVE_PROBE_WORST_CASE_MS = 5_000 + 3_000 + 3_000;
```
MGudgin, line 213: **Medium (maintainability) - Do not mirror three Rust timeout constants by hand.**

**Attribution:** `introduced_by_change` - this PR adds a TypeScript total derived from constants owned by separate Rust crates.

The current test only proves the TypeScript backstop exceeds the TypeScript mirror. It cannot detect a Rust timeout increase, after which Node could kill a valid probe and report the host as unsupported.

**Fix:** let the native command own one authoritative whole-operation deadline and keep Node's safety timeout independent, or add a repository parity/codegen check that verifies this value against the Rust constants.

MGudgin (changes requested):
## Summary

I found one merge-blocking resource-containment issue and three non-blocking follow-ups. I verified all four against the current PR head (`0d75d3d`) and GitHub's 26-file PR diff. The branch merge-base is `1ced007c`, behind the current `main` OID `69e5d09`, so anchors were derived from the actual PR diff.

Clean checks: the new command arguments are constant and fail closed; Linux gating and the Rust/Node/.NET discovery shapes are consistent; the documentation matches the implementation; dependencies and lockfiles are unchanged.

I did not file three lower-signal review notes from the broader sweep: continuing to drain capped output is still deadline-bounded, the suggested process-runtime test seam is a design preference rather than a demonstrated defect, and the supervisor coverage-table concern is subjective maintenance cleanup. None is a merge condition.

None.

sdk/node/src/platform.ts
SohamDas2021: Documented this. 

Making the native probe authoritative for all Linux facts is the step-2 work `docs/backend-support-probe-api-plan.md` already defers- bigger than this PR. The narrower TTL option isn't narrow either: `getPlatformSupport()` documents a module-lifetime cache and a test pins object identity, so a TTL changes public behavior on every platform to fix one Linux field, right?

The JSDoc now states the synchronous cost, names `bubblewrapNetwork` as the one field whose subject changes under a running process and says why that's survivable- the runner re-probes at launch, so a stale value costs a clear error.

I can do the restructure as a follow-up if you'd rather not leave it.

MGudgin: Thanks — the documentation is sufficient for this PR. I agree that changing the module-lifetime cache globally, or making the native probe authoritative for every Linux fact, is broader than this change. The remaining transport consolidation is already recorded in `docs/backend-support-probe-api-plan.md`; I am treating this comment as partially addressed and the documented tradeoff as non-blocking.

MGudgin (approved):
## Re-review disposition

Re-verified against current head `d051c4c` and base `43d579b`.

- **Addressed:** the advisory preflight now has a process-wide single-flight gate, so a stuck worker cannot accumulate another worker per call. Tests cover a stranded walk, overlapping healthy calls, and panic release.
- **Addressed:** stable tooling checks are success-cached separately, while the volatile namespace-allocation check runs on every launch. Tests cover warm-cache namespace rechecks, failure retry, and cross-use-case reuse.
- **Addressed:** `scripts/versioning/check-linux-probe-timeouts.js` now checks the TypeScript total against all three Rust constants and is wired into the versioning CI job.
- **Partially addressed / accepted:** Node still caches `bubblewrapNetwork` for the module lifetime, but the synchronous cost and possible staleness are now explicit in the public JSDoc, and launch re-probes remain fail-closed. The broader native-authoritative Linux transport remains documented plan work and is not a merge condition for this PR.

Verification: the timeout parity gate passed locally with an 11,000 ms native total. Current GitHub checks are green, including Linux build/lint, LXC E2E, versioning checks, and Linux SDK unit tests (345 passed, 0 failed). I could not run the four targeted Rust tests locally because the Debian WSL image lacks a C linker and installing it requires unavailable sudo credentials; I inspected the test implementations and relied on the green Linux CI for platform validation.

PR #1030, opened 2026-08-24
Add Apple Container samples and qualification

richiemsft:
## 📖 Description

Adds the samples and local qualification layer for the experimental Apple
Container one-shot backend introduced by #972.

- adds raw allow-network and default-block `0.8.0-alpha` configurations;
- updates the Node SDK example and runtime guidance;
- expands real-runtime coverage for streams, mounts, timeout cleanup,
  concurrency, network isolation, firewall tamper resistance, and hard-exit
  recovery;
- adds a local network-boundary harness that reproducibly builds and verifies
  the pinned MXC init image before qualification;
- hardens mount overlap validation, recovery record creation, and bounded
  cleanup uncovered by the qualification suite.

The Apple Container qualification remains local-only and does not add a CI
workflow.

## 🔗 References

- #972

## 🔍 Validation

- Validated 234 example/config files against the dev schema.
- Built the Node SDK and passed the three targeted Apple Container unit tests.
- Passed all 25 `apple_container_common` unit tests.
- Passed all eight ignored real-runtime Apple Container qualification tests.
- Passed the local Apple Container network-boundary qualification.
- Ran both raw sample configurations through `mxc-exec-mac`.
- Passed the exact macOS release build/test package matrix and Clippy with
  warnings denied.
- Passed Rust formatting and diff whitespace checks.

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [x] Task

---

with write access can trigger it on a PR with `/azp run`.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1030)

### DIFF
--- a/sdk/node/README.md
+++ b/sdk/node/README.md
@@ -73,7 +73,7 @@ Pick `0.7.0-alpha` for new code on any supported platform.
 | --- | --- | --- | --- |
 | Windows 11 24H2+ (verified on 25H2) | `processcontainer` | `windows_sandbox`, `wslc`, `microvm`, `isolation_session` | `processcontainer`: 26100 (24H2)<br>`isolation_session`: 26300.8553 ([Insider Preview](https://learn.microsoft.com/en-us/windows-insider/release-notes/experimental/preview-build-26300-8553)) |
 | Linux x64 / ARM64 | `bubblewrap` | `lxc` | — |
-| macOS ARM64 (schema `0.7.0-alpha`+) | `seatbelt` | `apple_container` (config only) | — |
+| macOS ARM64 (schema `0.7.0-alpha`+) | `seatbelt` | `apple_container` | `apple_container`: macOS 26 on Apple silicon |
 
 The default `processcontainer`, `bubblewrap`, `lxc`, and `seatbelt` backends work out of the box. **Experimental backends** (`windows_sandbox`, `wslc`, `microvm`, `isolation_session`, `hyperlight`, `apple_container`) require `{ experimental: true }` in `SandboxSpawnOptions` when you spawn — see [Choosing a Backend](#choosing-a-backend).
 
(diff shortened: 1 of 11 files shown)


tests/scripts/apple_container_network_qualification.mjs
```diff
@@ -0,0 +1,607 @@
...
+      throw new Error(`network security gate failed: ${failures.join("; ")}`);
+    }
+    if (!natExternal.probe?.ok || !natHost.probe?.ok) {
+      throw new Error("network security gate is inconclusive because a control failed");
+    }
+
+    console.log(
+      "\nQUALIFICATION PASS: Apple Container networking met the tested MXC boundary.",
+    );
+  } finally {
```
bbonaby, line 578: issue (non-blocking): This `finally` handles ordinary failures, but Node exits immediately on the default `SIGINT`/`SIGTERM` behavior, so Ctrl-C during any awaited container/network operation skips this cleanup and leaves the labeled resources behind. Since this is an interactive, long-running qualification command, could we install signal handlers that run the same cleanup path before exiting (and ideally reconcile resources carrying `OWNERSHIP_LABEL` on the next run)?

sdk/node/README.md
```diff
@@ -216,15 +216,26 @@ schema version `0.8.0-alpha`:
```
bbonaby, line 216: issue (non-blocking): This now documents Apple Container as executable, but `docs/schema.md` still says the backend is config-only and runtime execution fails closed with `unsupported_containment` (the `apple_container` row around line 278). Could we update that canonical schema reference in this PR as well? The two docs currently give contradictory guidance about whether this example can run.

tests/scripts/run_apple_container_network_qualification.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+SKIP_EXIT=77
+INIT_IMAGE="local/mxc-loopback-init:0.2"
```
bbonaby, line 7: issue (non-blocking): This adds a third independent copy of the trusted init-image tag, alongside `src/backends/apple_container/init/build.sh` and `NETWORK_BLOCK_INIT_IMAGE` in `cli.rs`. The wrapper runs `build.sh` but ignores the verified image it reports, then exports this separate literal; after a version bump, a missed update here could either fail qualification or silently exercise an older locally-cached image while production trusts the new tag/digest. Could `build.sh` expose the canonical tag in a sourceable form (or print it for this wrapper to capture) so the qualification and production invariants cannot drift?

Issue #1032, opened 2026-08-24
[Schema 0.8] Contradictory spec and parser semantics for an omitted network.ingress.hostLoopback


theelliotm:
### Relevant area(s)
Linux, macOS, Windows

### Brief description of your issue

The spec and the parser disagree on what an omitted `network.ingress.hostLoopback` means, and the spec contradicts itself.

- **Normative** ([`networking.md:109`](../blob/main/docs/sandbox-policy/0.8.0/networking/networking.md#L109)): "The specific `hostLoopback` value **overrides** `default` for the host-loopback path" — implies absent means inherit `default`.
- **Bubblewrap addendum** (line 481): "An omitted `ingress` section enforces the same deny, since that is the schema's default" — implies absolute deny. (Bubblewrap rejects `allow` for both fields, so both readings coincide there; it never disambiguates.)
- **JSON schema**: both fields optional, no `default` declared.
- **Parser** (`network_parser.rs:300`): `unwrap_or_default()` → `Deny`, independent of `default`.

The readings diverge only when `ingress.default: "allow"` and `hostLoopback` is omitted: inherit → `allow`, parser → `deny`.

Not academic — it changes Windows acceptance. `network_parser.rs:592` requires `hostLoopback: "allow"` for a ProcessContainer proxy without `allowedProxyPeer`; today omission resolves to `deny` and the config is rejected, but under inheritance it would be accepted.

Found while investigating #1021.

### Steps to reproduce
1. Author a 0.8 config with `network.ingress.default: "allow"` and no `hostLoopback`.
2. Observe the parser resolves `host_loopback` to `Deny` instead of inheriting `Allow`.
3. On Windows, add a ProcessContainer `runtimeConfig.networkProxy` without `allowedProxyPeer`; observe rejection with "requires `network.ingress.hostLoopback='allow'`" despite `ingress.default` already being `"allow"`.

### Expected behavior

One documented rule for an omitted `hostLoopback`, applied consistently by the parser and all backends, with the schema stating the default if one exists. Either fix `convert_ingress` to inherit and reword the addendum, or reword the normative "overrides" sentence to state absolute deny.

### Actual behavior

Normative text implies inheritance, an addendum asserts deny, the schema declares nothing, and the parser implements deny. Callers can't tell what omitting the field does.

PR #1033, opened 2026-08-24
[Seatbelt] Enforce hostLoopback, fix ip6 loopback bug, and consolidate network validation

### LINKED ISSUE #1004: Seatbelt: legacy loopback proxy in IPv6 form (http://[::1]:port) is rejected as remote under defaultPolicy='block'
### Relevant area(s)

macOS

On the Seatbelt backend, a **legacy-shape** (schema 0.6/0.7) loopback proxy written in IPv6 form — `http://[::1]:8080` — is rejected at config-parse time as a "remote proxy" when combined with `network.defaultPolicy: "block"`.

The rejection message itself tells the user to *"Use a loopback proxy (127.0.0.1/**::1**/localhost)"*, so the only remedy the error suggests is the exact form that fails.

**Root cause.** The guard in `src/core/wxc_common/src/config_parser.rs:1142` compares the proxy host against unbracketed string literals:

```rust
.is_some_and(|addr| !matches!(addr.host(), "127.0.0.1" | "::1" | "localhost"))
```

`ProxyAddress::host()` returns the value stored by `network_parser.rs::convert_wire_proxy_at`, which comes from `url::Url::host_str()`. For an IPv6 literal that returns the host **with brackets** (`[::1]`), so it never matches `"::1"`. Full-form IPv6 (`[0:0:0:0:0:0:0:1]`) fails for the same reason plus a lack of canonicalization.

**Scope.** Legacy `network.proxy.url` only. The schema-0.8 directional path (`runtimeConfig.networkProxy`) is **not** affected — it routes through `network_parser::host_is_canonical_loopback`, which unbrackets and parses to `IpAddr` before comparing.

| Proxy host | legacy `network.proxy.url` | directional `runtimeConfig.networkProxy` |
|---|---|---|
| `127.0.0.1` | accepted | accepted |
| `localhost` | accepted | accepted |
| `[::1]` | **rejected** | accepted |
| `[0:0:0:0:0:0:0:1]` | **rejected** | accepted |

Because the check runs in the parser, the request is refused before the backend is reached, so no backend-side logic can recover it.

### Steps to reproduce

1. On macOS, build the executor: `./build-mac.sh --rust-only`
2. Save this config as `repro.json`:
   ```json
   {
     "version": "0.7.0-alpha",
     "containment": "seatbelt",
     "process": { "commandLine": "echo hello" },
     "network": {
       "defaultPolicy": "block",
       "proxy": { "url": "http://[::1]:8080" }
     }
   }
   ```
3. Run it:
   ```
   src/target/aarch64-apple-darwin/release/mxc-exec-mac --config repro.json
   ```
4. For contrast, change the host to `127.0.0.1` and re-run — it is accepted.

### Expected behavior

`[::1]` is IPv6 loopback and is documented as a valid loopback proxy host (`docs/seatbelt/seatbelt-backend.md:339` lists `localhost` / `127.0.0.1` / `[::1]`). The config should be accepted, exactly as `127.0.0.1` and `localhost` are, and exactly as the directional shape already accepts it.

### Actual behavior

The run fails at parse time:

```
Configuration parse error: Seatbelt: a remote network.proxy (non-loopback host) cannot be
combined with defaultPolicy='block'. ... Use a loopback proxy (127.0.0.1/::1/localhost) or
'network.proxy.builtinTestServer: true' for port-scoped reachability under deny.
```

---

**Candidate fix.** Canonicalize the proxy host when building `ProxyAddress` in `network_parser.rs::convert_wire_proxy_at` — unbracket, parse to `IpAddr`, and store the canonical rendering. `ProxyAddress::to_url()` already re-brackets via `bracket_if_ipv6`, which suggests unbracketed storage is the intended canonical form. Alternatively, reuse the existing `host_is_canonical_loopback` predicate in the legacy guard instead of the string literals.

Note that the blast radius of changing the stored form should be checked first — `ProxyAddress::host()` also feeds the LXC iptables rules, the Bubblewrap proxy path, and `hosts_line`.

### LINKED ISSUE #1021: [Seatbelt] Enforce bidirectional hostLoopback semantics and deterministic network E2E coverage
### Relevant area(s)
macOS

Seatbelt declares schema 0.8 `HOST_LOOPBACK` support, whose contract is bidirectional, but only enforces the inbound half when outbound is allowed. A request with `network.egress.default: "allow"`, `network.ingress.default: "deny"`, and `network.ingress.hostLoopback: "deny"` passes validation and emits unrestricted `(allow network-outbound)`, so sandbox-to-host loopback remains reachable.

The validation message also incorrectly tells callers to omit `hostLoopback` to inherit `ingress.default`; the parser defaults an omitted value independently to `deny`. Additionally, the new egress-denial E2E tests return successfully without exercising denial when `1.1.1.1:443` is unreachable under the allow-policy precondition.

Found while reviewing #974. This consolidates #1013 and #1017.

### Steps to reproduce
1. Configure Seatbelt with schema 0.8, `egress.default: "allow"`, `ingress.default: "deny"`, and `ingress.hostLoopback: "deny"`.
2. Run a listener on host loopback and connect to it from the sandbox.
3. Observe that validation succeeds and unrestricted outbound permits the connection.
4. Set `ingress.default: "allow"` and omit `hostLoopback` as the validation message recommends; observe that omission becomes `deny` and validation still fails.
5. Run the Seatbelt egress-denial E2E tests on a host where `1.1.1.1:443` is unavailable; observe that they print `SKIPPED`, return normally, and are counted as passed without running the deny assertion.

### Expected behavior

Seatbelt must either enforce `hostLoopback` in both directions or reject combinations it cannot faithfully enforce. Validation guidance must match actual defaulting semantics, and CI must deterministically exercise or explicitly report loss of the enforcement assertion.

### Actual behavior

An accepted `hostLoopback: "deny"` policy can still reach host-loopback services when egress is allowed, the suggested omission remedy is ineffective, and CI can silently pass without testing egress denial.

### LINKED ISSUE #1022: [Seatbelt] Consolidate network validation ownership and remove parser/backend drift
### Relevant area(s)
macOS

Seatbelt-specific network invariants and effective-policy helpers are exposed from the foundational `wxc_common` crate through `src/core/wxc_common/src/seatbelt_policy.rs`, despite all production consumers belonging to `seatbelt_common`. At the same time, `config_parser.rs` retains independent Seatbelt-specific checks, creating two validation implementations that can drift.

That drift has already produced a concrete behavior difference for legacy bracketed IPv6 loopback proxy URLs; the concrete parser bug remains tracked by #1004 and should not be duplicated here. This issue tracks the ownership and deduplication defect that permits such divergence.

Found while reviewing #974. This consolidates #1019.

### Steps to reproduce
1. Inspect `wxc_common::seatbelt_policy::validate_seatbelt_network_policy` and its public export from `wxc_common::lib.rs`.
2. Inspect the independent Seatbelt guards in `wxc_common::config_parser.rs`.
3. Compare their loopback classification and covered invariants.
4. Observe that the foundational crate owns backend-specific policy and the parser/backend implementations are not a single canonical validation path.

### Expected behavior

Seatbelt-specific validation should live in `seatbelt_common`, with generic predicates remaining in `wxc_common`. Parsed and directly constructed `ExecutionRequest` values should reach one canonical implementation of each Seatbelt invariant.

### Actual behavior

Backend-specific policy leaks into `wxc_common`, while duplicated parser and backend checks can and already do diverge.

### LINKED ISSUE #1023: [Seatbelt] Correct schema 0.8 documentation and example drift
### Relevant area(s)
macOS

Several documentation and example statements introduced or retained by #974 do not match the repository:

- `docs/examples.md` says Seatbelt is the only backend with directional network support implemented, although Bubblewrap, LXC, and ProcessContainer implementations also declare directional capabilities.
- `tests/examples/30_mac_network_schema_v2.json` references nonexistent `schemas/dev/mxc-config.schema.0.8.0-dev.json`; the matching committed artifact is `mxc-config.schema.0.8.0-alpha.json`.
- `.github/copilot-instructions.md` says `config_parser.rs` rejects legacy proxy plus allow, while that invariant is enforced by `SeatbeltScriptRunner::validate` through `validate_seatbelt_network_policy`.

This consolidates #1018.

### Steps to reproduce
1. Open the schema reference in `tests/examples/30_mac_network_schema_v2.json` and attempt to resolve it in the repository.
2. Compare the directional-support claim in `docs/examples.md` with each backend's `network_policy_support()` implementation.
3. Compare the enforcement attribution in `.github/copilot-instructions.md` with call sites of `validate_seatbelt_network_policy`.

### Expected behavior

Documentation and examples should reference committed artifacts and accurately describe backend capabilities and enforcement ownership.

### Actual behavior

The example points to a missing schema, and two documentation claims contradict the current implementation.

theelliotm:
## 📖 Description

Four Seatbelt (macOS) fixes. One is a real containment gap; the rest is cleanup around it.

**`hostLoopback: "deny"` was accepted but never enforced** (#1021). Seatbelt advertised support for this schema-0.8 field, but whenever egress was allowed the profile emitted a blanket `(allow network-outbound)` that included the host's own loopback. So sandboxed code could reach services running on the developer's machine while the config said it couldn't. It now emits a `localhost`-scoped deny after the broad allow.

Seatbelt applies the *last* matching rule, so the port-scoped proxy allow must be emitted last. Getting this backwards fails closed silently, which is why it's now structural rather than incidental.

**An IPv6 loopback proxy was rejected** (#1004). `http://[::1]:8080` never had its brackets stripped before being parsed as an IP, so it looked like a remote host and was refused under `defaultPolicy: "block"` — while the IPv4 spelling of the same proxy worked.

**Seatbelt validation lived in two places** (#1022). Two rules were duplicated between the shared `config_parser.rs` and the backend. They now live only in `seatbelt_policy.rs`, invoked from the backend's `validate()`, which both the run-to-completion and streaming paths call.

**Docs and examples had drifted** (#1023). `docs/examples.md` claimed Seatbelt was the only backend with directional network support (five declare it), and 12 macOS examples pointed at schema files that don't exist.

Legacy 0.6/0.7 configs are unaffected for the above.

**GUI access was checked independent of UI being enabled.** `seatbelt_policy::gui_access_effective` is now the single source of truth for GUI being accessible. All configs (0.7 & 0.8) that enable GUI access without explicitly setting `ui.disable = false` are now rejected.

## 🔗 References

Resolves #1004 
Resolves #1021
Resolves #1022
Resolves #1023

Filed while investigating, not addressed here: #1032 — the spec and the parser disagree on what an omitted `hostLoopback` means. This PR preserves the parser's existing "omitted = deny" behavior and documents it.

## 🔍 Validation

- Reproduced the #1021 leak before the fix and confirmed it gone after, verified against the **target server's own access log** rather than the client's exit code.
- Confirmed the deny is loopback-scoped, not blanket — external egress still works.
- Mutation test: removing the new rule turns the new E2E test and two unit tests red.
- The new E2E test needs no internet (its control is a listener the test owns), so it cannot silently skip on an offline runner — which is how the existing egress tests could pass without asserting anything.
- Sweeps: 23 test binaries / 1497 passed; macOS-equivalent 16 / 976. `cargo fmt`, `clippy`, and `validate-configs.js` (266 configs) clean.

## ✅ Checklist

- [x] Linked to an issue
- [x] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes — `Cargo.lock` is unchanged

## 📋 Issue Type

- [x] Bug fix

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1033)


src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -321,9 +315,55 @@ fn write_network_rules(
...
+            // `localhost` matches this machine on any of its addresses, so this
+            // also closes the "reach the host via its LAN IP" path.
+            out.push_str(";; --- network: ingress.hostLoopback=deny — close host loopback,\n");
+            out.push_str(";;     including this host's own non-loopback addresses ---\n");
+            out.push_str("(deny network-outbound (remote ip \"localhost:*\"))\n");
+        }
+        (true, false) => {
+            out.push_str(";; --- network: ingress.hostLoopback=allow — open host loopback\n");
+            out.push_str(";;     under an otherwise-deny egress default ---\n");
+            out.push_str("(allow network-outbound (remote ip \"localhost:*\"))\n");
```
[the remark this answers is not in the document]
theelliotm: It only means the sandbox can send requests to on-device services by using the Mac's LAN IP. Anything off-device is still blocked. Scope hasn't changed from spec at all.

src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -321,9 +315,55 @@ fn write_network_rules(
...
+/// is already covered by the blanket outbound allow.
+fn write_host_loopback_rules(out: &mut String, policy: &ContainerPolicy, allow_outbound: bool) {
+    // The legacy shape has no hostLoopback concept — leave 0.6/0.7 untouched.
+    let Some(loopback_allowed) = seatbelt_policy::host_loopback_allowed(policy) else {
+        return;
+    };
+
+    match (loopback_allowed, allow_outbound) {
+        (false, true) => {
+            // `localhost` matches this machine on any of its addresses, so this
```
SohamDas2021, line 353: This contradicts host_is_canonical_loopback’s measured behavior limiting localhost:* to canonical loopback. The new LAN-address guarantee is therefore unsupported. Remove it or add enforcement and a macOS LAN-address test?

src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -1029,6 +1069,113 @@ mod tests {
...
+        // would silently swallow the proxy exception and fail closed.
+        let mut r = req();
+        r.policy.network_egress = Some(egress(NetworkAction::Allow));
+        r.policy.network_ingress = Some(ingress(NetworkAction::Deny));
+        let addr = ProxyAddress::new("127.0.0.1".into(), 44444);
+        let p = build_profile_with_proxy(&r, Some(&addr)).unwrap();
+
+        let deny = p.find("(deny network-outbound (remote ip \"localhost:*\"))");
+        let proxy = p.find("(allow network-outbound (remote ip \"localhost:44444\"))");
+        if let (Some(deny), Some(proxy)) = (deny, proxy) {
```
SohamDas2021: This test passes vacuously: open egress prevents the proxy rule from being emitted, so this assertion is skipped?

src/backends/seatbelt/common/src/profile_builder.rs
theelliotm: Adjusted documentation

src/backends/seatbelt/common/src/profile_builder.rs
theelliotm: Modified + added tests to address this.

docs/examples.md
```diff
@@ -139,12 +139,20 @@ for the full design and per-backend enforcement matrix), not a
...
 any — it actually enforces; a backend that hasn't declared support for a
-given field rejects a config that sets it. **Seatbelt (macOS) is currently
-the only backend with that support implemented** (`EGRESS_DEFAULT`,
-`INGRESS_DEFAULT`, `HOST_LOOPBACK`, `RUNTIME_PROXY` — see
-[`docs/seatbelt/seatbelt-backend.md`](seatbelt/seatbelt-backend.md#schema-08-network-shape-egress--ingress--runtimeconfignetworkproxy));
-other backends are expected to add their own declarations as separate
-follow-up work. On Seatbelt, `runtimeConfig.networkProxy` covers only the
+given field rejects a config that sets it:
+
+| Backend | Declared schema-0.8 network support |
```
dhoehna: I thought the PR was for seatbelt.  Why is the other backends included?

src/backends/seatbelt/common/src/seatbelt_policy.rs
```diff
@@ -115,7 +124,33 @@ pub fn validate_seatbelt_network_policy(policy: &ContainerPolicy) -> Result<(),
...
 mod tests {
     use super::*;
-    use crate::models::{ProxyAddress, ProxyConfig};
+    use wxc_common::models::{ProxyAddress, ProxyConfig};
+
+    #[test]
+    fn host_loopback_allowed_is_none_for_the_legacy_shape() {
+        // 0.6/0.7 has no hostLoopback concept; the caller must not synthesize
+        // one, or legacy configs would change behavior.
+        let mut p = policy();
```
dhoehna, line 199: Why does making a policy object restrict the test to pre 0.8.0 behavior?

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -138,20 +140,27 @@ impl SandboxBackend for SeatbeltScriptRunner {
...
-                     'hostLoopback' equal to 'default', or omit 'hostLoopback' to \
-                     inherit 'default'."
-                        .to_string(),
-                ));
+                let name = |action: NetworkAction| match action {
+                    NetworkAction::Allow => "allow",
+                    NetworkAction::Deny => "deny",
+                };
+                return Err(error_response(format!(
+                    "macOS Seatbelt cannot enforce a network.ingress.hostLoopback \
```
dhoehna: Is this error seatbelt specific or is this a general configuration issue that would be rejected by all backends?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1105,54 +1105,7 @@ fn convert_wire_config(
...
-                       silently weakening the deny for raw-socket clients that ignore \
-                       HTTP_PROXY. Use a loopback proxy (127.0.0.1/::1/localhost) or \
-                       'network.proxy.builtinTestServer: true' for port-scoped reachability \
-                       under deny.";
-            logger.log_line(msg);
-            return Err(WxcError::ConfigParse(msg.to_string()));
-        }
-
-        // LXC is the inverse of the two guards above: it *does* have a
+        // LXC is the inverse of the guard above: it *does* have a
```
dhoehna, line 1108: `// LXC is the inverse of the guard above: it *does* have a` does it now?

src/core/wxc_common/src/network_parser.rs
```diff
@@ -72,7 +72,7 @@ enum NetworkFormat {
 /// Use this to *reject* a host (LXC treats all of `127/8` as the container's
 /// own namespace loopback). Breadth is fail-safe here: a wider match rejects
 /// more. To *admit* a host, use [`host_is_canonical_loopback`].
-pub(crate) fn host_is_loopback(host: &str) -> bool {
+pub(crate) fn host_is_any_loopback(host: &str) -> bool {
     if host.eq_ignore_ascii_case("localhost") {
```
dhoehna, line 83: NIT: Returning true based on one check.  Can this be changed to `return host.eq_ignore_ascii_case('localhost");`?  Also, how often is this method called?  If it is a low number maybe this is better removed.

src/backends/seatbelt/common/src/seatbelt_runner.rs
theelliotm: Seatbelt specific - seatbelt can't enforce host-specific ingress policy (proxy being the exception).

src/core/wxc_common/src/network_parser.rs
theelliotm: this is an old function i renamed to differentiate it between host_is_canonical_loopback. removing it or modifying it isn't relevant to this PR.

tests/examples/15_mac_hello_world.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "../schemas/stable/mxc-config.schema.0.7.0-alpha.json",
+    "$schema": "../../schemas/stable/mxc-config.schema.0.7.0-alpha.json",
```
dhoehna, line 2: If the string now searched 2 directories up.  Were the tests moved one directory lower?

dhoehna (commented):
Some questions.

tests/examples/15_mac_hello_world.json
theelliotm: At some point before this PR the directories changed, and they were never updated to reflect that.

src/core/wxc_common/src/config_parser.rs
theelliotm: I edited the comment to reflect that the Seatbelt guard above it was removed.

src/backends/seatbelt/common/src/seatbelt_policy.rs
theelliotm: policy() uses the legacy behavior by default, unless you modify the 0.8 Options. req() uses the new schema

theelliotm: If I'm wrong about this please let me know!

docs/examples.md
theelliotm: Deleted that table.

.github/copilot-instructions.md
```diff
@@ -188,7 +188,7 @@ The Rust workspace (`src/`) implements multiple sandboxing backends behind the `
...
...s `localhost` covers the host's non-loopback addresses too, closing the LAN-IP path), while `allow` under a deny egress default emits the matching allow, with the same breadth in reverse — it opens every address bound to this host, not just `127.0.0.1`. Neither case reaches another machine: SBPL's `localhost` means *this machine*, not *this network*, and there is no narrower primitive (a literal `127.0.0.1` is a profile syntax error); the port-scoped proxy allow is always emitted last so it survives. The inbound half is not expressible, so `validate()` still rejects a `hostLoopback` that diverges from `network.ingress.default`. An omitted `hostLoopback` is `deny`, not an inherit of `default` ([#1032](https://github.com/microsoft/mxc/issues/1032)). Every Seatbelt-specific network invariant (proxy vs. `enforcementMode`, remote proxy under deny, proxy combined with an allow default, `allowedHosts`/`blockedHosts` under deny) lives in `backends/seatbelt/common/src/seatbelt_policy.rs` and is enforced from the backend's own `validate`, which both the run-to-completion and streaming paths call — `wxc_common::config_parser` holds no Seatbelt rules. See `docs/seatbelt/seatbelt-backend.md`. |
```
bbonaby: issue (non-blocking): The ownership statement is broader than the implementation. `blockedHosts` rejection and the `hostLoopback != ingress.default` invariant still live directly in `seatbelt_runner.rs::validate`, outside `seatbelt_policy.rs`, even though this says every Seatbelt-specific network invariant (and specifically `blockedHosts`) lives in `seatbelt_policy.rs`. Could we either move those two checks into the policy module as stated or narrow this text to describe which validations were consolidated? This file guides future changes, so the current wording could send contributors to the wrong owner and recreate drift.

theelliotm: further consolidated the validation into seatbelt_policy. good catch

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -122,39 +122,11 @@ impl SandboxBackend for SeatbeltScriptRunner {
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
         validate_network_policy_support(request, self.network_policy_support())?;
 
-        // Shared with the config parser so a caller that builds an
-        // ExecutionRequest directly gets the same rules.
-        wxc_common::seatbelt_policy::validate_seatbelt_network_policy(&request.policy)
+        // Seatbelt's own invariants — the only home for them, so a caller that
+        // builds an ExecutionRequest directly gets the same rules.
+        crate::seatbelt_policy::validate_seatbelt_network_policy(&request.policy)
```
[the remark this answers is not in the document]
theelliotm: Will be putting this in a new PR.

tests/examples/31_mac_network_0_8.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "../../schemas/dev/mxc-config.schema.0.8.0-dev.json",
+    "$schema": "../../schemas/stable/mxc-config.schema.0.8.0-alpha.json",
```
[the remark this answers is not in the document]
theelliotm: That change is already there. Removed promise.

shschaefer (approved):

PR #1042, opened 2026-08-25
[WSLC] Honor portMappings on the state-aware surface and reject unenforceable host lists

### LINKED ISSUE #824: [WSLC] One-shot vs state-aware network-policy parity: fail-open redundant host lists and dropped portMappings
### Relevant area(s)

Windows (WSLc backend)

The one-shot WSLc flow (`WSLContainerRunner::start_container`) and the state-aware WSLc flow (`WslcStateAwareRunner` + daemon) diverge on two networking behaviors that should be identical. A full one-shot-vs-state-aware policy comparison found filesystem handling is identical and the cooperative proxy is enforced identically (same `apply_cooperative_proxy_env` on the workload; only the declaration site differs, which is structural and acceptable). Two real gaps remain:

**Gap 1 — Redundant host lists are handled inconsistently (fail-open in one-shot).**
Both flows already reject *meaningful* per-host filtering at parse time (`config_parser.rs:1064`, `needs_host_filtering()`, added in #681). But a *redundant* host list — `defaultPolicy: "block"` + `blockedHosts`, or `defaultPolicy: "allow"` + `allowedHosts` — passes that gate (`needs_host_filtering()` → false):
- One-shot `validate_runner` (`wsl_container_runner.rs:636`) also uses `needs_host_filtering()`, so it **accepts and silently ignores** the list.
- State-aware `reject_host_filtering` (`policy.rs:115`) rejects on **any** non-empty list.

This is a fail-open footgun: a user who writes `defaultPolicy: "allow"` + `allowedHosts: ["github.com"]` (a classic "I meant an allowlist" typo) gets **all** egress permitted in one-shot, with no error — the opposite of intent.

**Gap 2 — State-aware silently drops `experimental.wslc.portMappings`.**
One-shot forwards ports (`wsl_container_runner.rs:1491` → `WslcSetContainerSettingsPortMappings`). The state-aware provision config `WslcProvisionPhase` (`wire.rs:550`) deliberately omits `portMappings`, and because the `experimental` block is permissive (no `deny_unknown_fields`), a state-aware config carrying `portMappings` is **silently ignored** — no inbound forwarding, no error (`container_steps.rs:803` create_daemon_container passes `&[]`).

### Steps to reproduce

Gap 1:
1. Run a one-shot WSLc config with `network: { "defaultPolicy": "allow", "allowedHosts": ["github.com"] }`.
2. Observe the container starts and all egress is permitted (the `allowedHosts` list is ignored).
3. Run the same policy through the state-aware provision phase — it is rejected with `policy_validation`.

Gap 2:
1. Run a one-shot WSLc config with `experimental.wslc.portMappings` — the host→container port forward is configured.
2. Provide the same `portMappings` to a state-aware provision config.
3. Observe no port forwarding is set up and no error is raised.

### Expected behavior

- Gap 1: Both flows handle host lists identically, failing closed — any non-empty `allowedHosts`/`blockedHosts` on WSLc is rejected (WSLc cannot enforce per-host filtering).
- Gap 2: Both flows support `experimental.wslc.portMappings`, or state-aware explicitly rejects it rather than silently ignoring it.

### Actual behavior

- Gap 1: One-shot accepts and silently ignores redundant host lists (fail-open); state-aware rejects any non-empty list. Behavior differs.
- Gap 2: One-shot forwards ports; state-aware silently drops `portMappings`.

-----

### Proposed actions

**Gap 1 — make one-shot reject (align both flows to fail-closed):**
- Change the shared parser predicate `config_parser.rs:1064` from `policy.needs_host_filtering()` to reject any non-empty `allowed_hosts`/`blocked_hosts` for WSLc; reword the error message (drop the "allowedHosts with defaultPolicy='block'…" framing).
- Mirror the same predicate in one-shot `validate_runner` (`wsl_container_runner.rs:636`) for the parser-bypass path.
- Flip the redundant-list tests (`config_parser.rs` WSLc host-filtering tests, `wsl_container_runner.rs` `validate_runner_*` tests) to expect rejection; keep the bare-defaults-accepted tests.
- State-aware `reject_host_filtering` (`policy.rs:115`) already matches — no change.

**Gap 2 — make state-aware support `portMappings` (align to one-shot):**
- Add `port_mappings` to `WslcProvisionPhase` (`wire.rs:550`) and regenerate the dev schema + SDK wire types (`mxc_schema_gen`).
- Add `port_mappings: Vec<PortMapping>` to the daemon `ProvisionConfig` (`daemon_protocol.rs:75`).
- Thread it: `state_aware.rs::provision` → `ProvisionConfig` → `session_manager.rs` → `create_daemon_container` (`container_steps.rs:803`), passing the slice into `ContainerSettings::build` (already supports `port_mappings`, `container_steps.rs:499`) instead of `&[]`.
- Add the SDK type (`sdk/node/src/state-aware-types.ts` `WslcProvisionConfig`) and an E2E state-aware port-mapping config.

Both are self-contained: Gap 1 is tiny (one predicate + tests); Gap 2 is a field threaded through the daemon wire protocol into container creation.

SohamDas2021:
## 📖 Description

A WSLc sandbox driven through the state-aware lifecycle could not expose a container port to the host. The same `portMappings` block worked on the one-shot surface, so the two surfaces disagreed on what a WSLc sandbox could do.

`experimental.wslc.provision.portMappings` now forwards host TCP ports into the container. Mappings are applied when the container is created and stay fixed for the sandbox's life; `windowsPort` must be unique and non-zero, and `tcp` is the only accepted protocol (the WSLC SDK returns `E_NOTIMPL` for UDP). The one-shot and state-aware surfaces now accept the same entries and validate them identically.

Separately, a WSLc config could ask for per-host egress filtering and be silently ignored. `blockedHosts` under a `block` default, or `allowedHosts` under an `allow` default, was accepted and the list discarded. The `allow` case is the harmful one: it reads as a restriction while leaving all egress open. WSLc cannot enforce per-host rules in any form — the container has no `CAP_NET_ADMIN` for in-container `iptables` — so any non-empty host list is now refused rather than dropped.

### Note on overlap with #1145

The v0.9 directional networking cutover landed while this PR was open and removed legacy network authoring outright, so `network.proxy` and the host lists are now rejected before a state-aware request reaches this code. The `!state_aware` guards here are therefore defensive rather than load-bearing on v0.9; they still apply on the published 0.6–0.8 contracts, which is also where the host-list fix remains live. They are kept deliberately, and they replace a narrower and equally unreachable guard that `main` already carried.

## 🔗 References

Closes #824

## 🔍 Validation

Automated:

- 4,500 Rust tests and 423 Node SDK tests, zero failures
- `cargo fmt --check` and workspace `clippy --all-targets -D warnings` clean
- Schema-version, rolling and exact codegen, and config-validation gates all pass (350 configs)

Live on a Windows 11 + WSL2 host, against a `wslc`-enabled `wxc-exec`:

- `run_wslc_state_aware_tests.ps1` — **81/83**
- `run_wslc_all_tests.ps1` — **24/25 suites**, including all 19 executed one-shot config fixtures (`wslc_port_mapping_tcp.json` and `wslc_port_mapping_multiple.json` among them)
- The port mapping was proved functionally, not just structurally: a listener inside the container answered through the mapped Windows host port (`65268 -> 80`), so an implementation that dropped `portMappings` would fail the round-trip rather than quietly pass

Two pre-existing failures remain, both in the state-aware section D and neither touched by this PR:

- `D: exec (directional network change rejected structurally)` asserts `network` is an unknown field at exec, but `ExecRequest` on `main` declares `pub network: OptionalField<Network>`; #1145 made the field structurally valid and left the assertion stale. The backend now reports `policy_validation` instead.
- `D: provision (deniedPaths nested under mount rejected)` returns no envelope; filesystem policy, unrelated to this change.

## ✅ Checklist

- [x] Linked to an issue
- [x] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes (see [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md)) — no `Cargo.lock` change

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task


docs/wsl/wsl-container-getting-started.md
```diff
@@ -353,18 +353,23 @@ that combine the proxy with a `block` default or host lists are **rejected**.
...
 
-WSLC **cannot** enforce per-host egress filtering. `allowedHosts` with
-`defaultPolicy: "block"` (an allowlist) or `blockedHosts` with
-`defaultPolicy: "allow"` (a blocklist) would require in-container `iptables`
-rules, but a WSLC container runs **without** `CAP_NET_ADMIN` (the SDK's
-`Privileged` flag does not grant it), so those rules cannot be applied — and MXC
-has no VM-level enforcement hook either (WSLC cannot expose one without breaking
-other security promises such as MDE). Rather than fail the run at exec time,
-such configs are **rejected at config-parse time**:
+WSLC has no way to enforce per-host egress filtering, so **MXC refuses any
```
dhoehna: Seems odd that just because WSLc can't enforce something MXC refuses the config.  Does MXC have per-backend parsing?

docs/wsl/wsl-container-getting-started.md
```diff
@@ -353,18 +353,23 @@ that combine the proxy with a `block` default or host lists are **rejected**.
...
+config that asks for it** — the request is never accepted and then quietly
```
dhoehna: An un-accepted request should be an error, really.  Seems like MXC should catch that.  Unless MXC does not and WSLc instead ignored the incorrect configuration.

src/backends/wslc/common/src/daemon_protocol.rs
```diff
@@ -53,6 +53,15 @@ pub struct VolumeMount {
     pub read_only: bool,
 }
 
+/// One host→container port forward, mirroring the one-shot runner's port
```
dhoehna: ->

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -368,9 +372,47 @@ fn build_provision_config(
...
+fn map_provision_port_mappings(
+    config: Option<&WslcProvisionPhase>,
+) -> Result<Vec<PortMapping>, MxcError> {
+    let Some(mappings) = config.and_then(|c| c.port_mappings.as_ref()) else {
+        return Ok(Vec::new());
+    };
+    let mut converted = Vec::with_capacity(mappings.len());
+    let mut seen = std::collections::HashSet::new();
+    for (idx, m) in mappings.iter().enumerate() {
+        if m.windows_port == 0 {
```
dhoehna: This makes the loop exit early if this condition is true.  Is this the expected behavior?

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -368,9 +372,47 @@ fn build_provision_config(
...
+            return Err(MxcError::policy_validation(format!(
+                "experimental.wslc.provision.portMappings[{idx}]: 'windowsPort' must be > 0"
```
dhoehna: Is this error caused by an incorrect configuration, or a developer error?

dhoehna (commented):
Just some questions.

docs/wsl/wsl-container-getting-started.md
SohamDas2021: Yes, parsing is per-backend: config_parser branches on containment, and each backend adds its own validation logic on top.

docs/wsl/wsl-container-getting-started.md
SohamDas2021: It is an error: allowedHosts/blockedHosts are rejected at every WSLC phase. Closing that fail-open is the point of this PR

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Intentional fail-fast: this is validation, so the first bad mapping aborts the whole provision rather than silently truncating the list

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Config error: the values come from experimental.wslc.provision.portMappings[i], which the caller supplies

src/backends/wslc/common/src/daemon_protocol.rs
SohamDas2021: updating

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -368,9 +372,47 @@ fn build_provision_config(
         image_tar_path,
         volumes,
         network,
+        port_mappings,
     })
 }
 
+/// Map + validate the provision phase's port mappings (wire → daemon). Mirrors
```
bbonaby: suggestion (non-blocking): Could we extract the non-zero and duplicate-port validation into a shared helper used by both this state-aware path and the one-shot parser? The conversion is shared, but validation is still implemented twice, and the dedupe keys already differ (`windows_port` here versus `(windows_port, protocol)` in `config_parser.rs`). They behave the same while TCP is the only protocol, but keeping two implementations makes future protocol support and error wording easy to drift.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1040,7 +1041,11 @@ fn convert_wire_config(
         // host lists to it, and a 'block' default (the WSLc default) yields no
         // outbound networking / a drop-floor that can't even reach the proxy.
         // Require an 'allow' default with no host lists so the proxy is reachable.
-        if containment == ContainmentBackend::Wslc
+        //
+        // One-shot only: state-aware binds the network mode at provision, so a
+        // later phase omits `defaultPolicy`; its own per-phase gates cover proxy.
+        if !state_aware
+            && containment == ContainmentBackend::Wslc
```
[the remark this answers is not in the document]
SohamDas2021: Fixed. The earlier WSLc proxy-form check is now gated on `!state_aware` too, so non-URL proxy forms on the state-aware surface fall through to `validate_exec_policy` / `validate_post_provision_policy` and return `policy_validation`. Covered by the parser test `proxy_wslc_non_url_form_deferred_to_backend_on_state_aware` plus the existing backend test `exec_rejects_non_url_proxy`.

src/core/wxc_common/src/config_contract_adapters/dev/state_aware.rs
```diff
@@ -94,10 +95,17 @@ fn convert_wslc_provision(value: contract::WslcProvision) -> wire::WslcProvision
...
+        port_mappings,
     } = value;
     wire::WslcProvisionPhase {
         image: image.into_option(),
         image_tar_path: image_tar_path.into_option(),
+        port_mappings: port_mappings.into_option().map(|mappings| {
+            mappings
+                .into_iter()
+                .map(convert_wslc_port_mapping)
+                .collect()
```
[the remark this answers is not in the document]
SohamDas2021: Fixed. `WSLC_ALL_FIELDS_REQUEST_JSON` now carries two `provision.portMappings` entries (one with an explicit `protocol`, one without, covering both arms of `protocol.into_option().map(convert_protocol)`), and `wslc_request_maps_expected_wire_fields` asserts the converted `provision.port_mappings` values. Since that fixture also feeds `assert_matches_current_wire_deserialization`, exact-vs-rolling equivalence is now covered for port mappings as well.

tests/scripts/run_wslc_state_aware_tests.ps1
```diff
@@ -686,6 +686,136 @@ try {
...
+            $iar = $client.BeginConnect($HostName, $Port, $null, $null)
+            if (-not $iar.AsyncWaitHandle.WaitOne(2000)) { throw "connect timeout" }
+            $client.EndConnect($iar)
+            $stream = $client.GetStream()
+            $stream.ReadTimeout = 3000
+            $buf = New-Object byte[] 256
+            $n = $stream.Read($buf, 0, $buf.Length)
+            if ($n -gt 0) {
+                return [System.Text.Encoding]::ASCII.GetString($buf, 0, $n)
+            }
```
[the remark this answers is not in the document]
SohamDas2021: Fixed. `Get-TcpResponse` now accumulates into a `MemoryStream` until `Read` returns 0 (the `printf | nc` peer closes after sending). Verified both ways against a listener that splits `MXC_PORTMAP_OK` across two segments with a 300 ms gap: the new loop reassembles the full payload, while the previous single `Read` returned only the first 8 bytes (`MXC_PORT`).

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Good catch, and the drift had already started - the two sites disagreed on both the dedupe key and the message text. Extracted `wxc_common::models::validate_wslc_port_mappings`, now used by both the one-shot parser and this state-aware path. It returns a bare message string so each caller keeps its own error type (`ConfigParse` vs `policy_validation`) and JSON path prefix. The state-aware dedupe key widens to `(windowsPort, protocol)` to match the parser; that is unobservable today since the wire model is TCP-only.

dhoehna (dismissed):

docs/wsl/wsl-container-getting-started.md
```diff
@@ -344,27 +344,34 @@ address the container can reach:
...
-other security promises such as MDE). Rather than fail the run at exec time,
-such configs are **rejected at config-parse time**:
+WSLC has no way to enforce per-host egress filtering, so **MXC rejects any
+config that asks for it** rather than accepting the request and silently
+ignoring the lists. Any allowlist (`allowedHosts` with `defaultPolicy: "block"`)
+or blocklist (`blockedHosts` with `defaultPolicy: "allow"`) would require
+in-container `iptables` rules, but a WSLC container runs **without**
+`CAP_NET_ADMIN` (the SDK's `Privileged` flag does not grant it), so those rules
+cannot be applied — and MXC has no VM-level enforcement hook either (WSLC
+cannot expose one without breaking other security promises such as MDE).
```
bbonaby: question (non-blocking): what does MDE mean?

docs/wsl/wslc-state-aware.md
```diff
@@ -90,6 +90,14 @@ whose value equals the default — because an explicit default is indistinguisha
 one by value alone. Only the exec-phase cooperative `proxy` is accepted after provision; a
 proxy-only `network` block (no mode fields) is therefore honored at exec.
 
+## Port forwarding
+
+`experimental.wslc.provision.portMappings` forwards host (Windows) ports to the container, mirroring
+the one-shot `experimental.wslc.portMappings` surface. Each entry is `{ windowsPort, containerPort }`
```
bbonaby, line 102: question: (non-blocking): so WSLc has host to container networking but specifically only related to port redirection? If so I wonder if this should mean that wslc configs that want to use port mappings need to also do `network.ingress.hostloopback : allow`  together. Up to you though if you want to keep portmappings a completely separate exception.

src/backends/wslc/daemon/src/session_manager.rs
```diff
@@ -390,6 +390,17 @@ impl Worker {
...
             .collect();
+        // Daemon port mappings carry only ports (TCP-only today); map to the
+        // domain `PortMapping` that `ContainerSettings::build` consumes.
+        let port_mappings: Vec<PortMapping> = config
+            .port_mappings
+            .iter()
+            .map(|p| PortMapping {
+                windows_port: p.windows_port,
+                container_port: p.container_port,
+                protocol: "tcp".to_string(),
```
bbonaby, line 401: thought: Maybe I read wrong but I figured it was TCP or UDP?

schemas/dev/mxc-config.schema.0.9.0-dev.json
```diff
@@ -710,7 +710,7 @@
       "type": "string"
     },
     "PortMapping": {
```
bbonaby, line 712: question: not related to this file but do we need to add anything for the Rust SDK in terms of the types consumers will use to author a wlsc config?. OR does it look like it's already taken care of?

Here:  https://github.com/microsoft/mxc/blob/main/src/core/mxc_engine/src/policy.rs#L504

docs/wsl/wsl-container-getting-started.md
SohamDas2021: Microsoft Defender for Endpoint, updated doc.

docs/wsl/wslc-state-aware.md
SohamDas2021: Yes, that is correct.

However, coupling it to network.ingress.hostLoopback isn't expressible today: WSLc passes NetworkPolicySupport::LEGACY on every phase, so the whole 0.8 directional network.ingress section is rejected. Requiring it would mean first declaring INGRESS_DEFAULT | HOST_LOOPBACK, claiming a posture WSLC can't actually enforce. They're also different directions: hostLoopback governs whether the container can reach the host's loopback, whereas portMappings is inbound NAT into the container.

schemas/dev/mxc-config.schema.0.9.0-dev.json
SohamDas2021: Types are already covered.

**One-shot:** `WslcSection.port_mappings: Vec<(u16, u16)>` is the authoring surface, and `wire()` emits `protocol: "tcp"`.

**State-aware:** the Rust SDK is JSON-in/JSON-out by design here, `run_state_aware_json` takes the wire-format request, so there's no typed per-phase builder to extend. Callers write `experimental.wslc.provision.portMappings` directly, same as the other state-aware backends.

src/backends/wslc/daemon/src/session_manager.rs
SohamDas2021: You are right, TCP is the only option here (`wire::TransportProtocol` has a single `Tcp` variant), so the hardcoded `"tcp"` is the only reachable value.

Fixed the stale doc on the domain `PortMapping`, which claimed `"tcp" or "udp"`.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -368,9 +374,29 @@ fn build_provision_config(
...
+/// Map and validate the provision phase's port mappings (wire → daemon).
+fn map_provision_port_mappings(
+    config: Option<&WslcProvisionPhase>,
+) -> Result<Vec<PortMapping>, MxcError> {
+    let Some(mappings) = config.and_then(|c| c.port_mappings.as_ref()) else {
+        return Ok(Vec::new());
+    };
+    let validated =
+        validate_wslc_port_mappings(mappings, "experimental.wslc.provision.portMappings")
+            .map_err(MxcError::policy_validation)?;
```
bbonaby, line 390: note: I guess this will conflict with your promotion PR

src/core/mxc_config_contract/src/dev/state_aware/provision/mod.rs
```diff
@@ -13,8 +13,9 @@ pub enum ProvisionRequest {
     IsolationSession(IsolationSessionProvisionRequest),
     /// A Windows Sandbox provision request.
     WindowsSandbox(WindowsSandboxProvisionRequest),
-    /// A WSLC provision request.
-    Wslc(WslcProvisionRequest),
+    /// A WSLC provision request. Boxed: its port-mapping list makes it much
+    /// larger than its siblings.
```
bbonaby, line 17: note: probably don't need the second part of the comment.

src/core/mxc_config_contract/src/dev/state_aware/provision/wslc.rs
```diff
@@ -22,6 +22,10 @@ pub struct WslcProvision {
     /// Optional path to a local container image archive.
     #[serde(default)]
     pub image_tar_path: OptionalField<String>,
+    /// Optional host-to-container TCP port mappings. Per-container, so unlike
+    /// the one-shot sizing knobs it is honored on the shared daemon session.
```
bbonaby: thought: sizing knobs?

src/core/mxc_engine/src/policy.rs
```diff
@@ -1642,6 +1640,44 @@ mod tests {
         );
     }
 
+    #[test]
+    fn wslc_rejects_redundant_host_lists() {
+        // A list matching the default enforces nothing, so before #824 these
+        // two were accepted and dropped. `allow` + allowedHosts is the harmful
+        // one: it reads as a restriction while leaving all egress open.
```
bbonaby, line 1647: note: Jeff brought this up the other day actually. I don't think it's a WSLc issue. Its a greater config issue. I'm thinking the logic for this should go in network_parser.rs  since it's a greater config violation. Thoughts?

src/core/wxc_common/src/config_contract_adapters/dev/one_shot.rs
```diff
@@ -3,6 +3,7 @@
 
 use crate::config_contract_adapters::dev::common::{
     convert_filesystem, convert_network, convert_process, convert_telemetry, convert_version,
+    convert_wslc_port_mapping,
```
bbonaby, line 6: note: I think this addition can be deleted right?

src/core/wxc_common/src/config_contract_adapters/dev/one_shot.rs
```diff
@@ -210,25 +211,6 @@ fn convert_windows_sandbox(value: contract::OneShotWindowsSandbox) -> wire::Wind
...
-        container_port,
-        protocol,
-    } = value;
-    wire::PortMapping {
-        windows_port: windows_port.get(),
-        container_port: container_port.get(),
-        protocol: protocol.into_option().map(convert_protocol),
-    }
-}
-
```
bbonaby, line 231: question: since this PR doesn't remove wslc from experimental, does removing these make sense?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -4523,6 +4512,52 @@ mod tests {
         );
     }
 
+    #[test]
+    fn wslc_rejects_redundant_block_with_blocked_hosts() {
```
bbonaby, line 4516: note: see comments above, I think this isn't a wslc issue and a config issue. In general we shouldn't allow default block + block hosts, and default allow + allow hosts. So we should have the network_parser.rs handle that rather than it being backend specific.

src/core/wxc_common/src/models.rs
```diff
@@ -839,10 +845,47 @@ pub struct PortMapping {
...
-    /// Protocol: "tcp" or "udp". Default: "tcp".
+    /// Transport protocol, lowercase. Always `"tcp"`: this carries a WSLc port
+    /// forward, whose `wire::TransportProtocol` has no `udp` variant.
     pub protocol: String,
 }
 
+/// Validate + convert WSLc wire port mappings. Shared by the one-shot parser
+/// and the state-aware provision phase so the two cannot drift. `path` is the
+/// caller's JSON location; the error is a bare message each caller wraps.
+pub fn validate_wslc_port_mappings(
```
bbonaby, line 856: thought: I'm not sure model.rs is the right place to put validation code tbh. When I think about models I think of just the raw objects themselves. Is there maybe another place we can put this?

src/core/wxc_common/src/wire.rs
```diff
@@ -656,16 +656,19 @@ pub struct Wslc {
...
 /// Filesystem mounts and network mode derive from the top-level `policy`
 /// section (readwrite / readonly paths, network), not from here. The
-/// one-shot-only sizing knobs (`cpuCount` / `memoryMb` / `gpu` / `storagePath`
-/// / `portMappings`) are deliberately absent: the daemon shares a single session
-/// across sandboxes and does not apply per-sandbox sizing. start / exec / stop /
-/// deprovision carry no backend-specific config (the exec command flows through
-/// the top-level `process` section), so they have no phase struct.
+/// one-shot-only sizing knobs (`cpuCount` / `memoryMb` / `gpu` / `storagePath`)
+/// are deliberately absent: the daemon shares a single session across sandboxes
+/// and does not apply per-sandbox sizing. `portMappings`, by contrast, is
```
bbonaby, line 667: thought: Oh this is what "sizing knobs" mean. That said, thinking about it some more 

> The daemon shares a single session across sandboxes and does not apply per-sandbox sizing

Would developers expect this? This seems a bit unexpected that I can do things with the one shot that I can't do with the state aware. This backend might potentially be the only one doing that. Would it be a big lift to re-align the two?

docs/wsl/wslc-state-aware.md
SohamDas2021: #1145 landed and WSLC now declares `EGRESS_DEFAULT | INGRESS_DEFAULT | HOST_LOOPBACK | RUNTIME_PROXY`, so the ingress section *is* expressible and `policy.rs` enforces a uniform posture, meaning any bridged config that can use `portMappings` already carries `ingress.hostLoopback: allow`.

src/core/mxc_engine/src/policy.rs
SohamDas2021: Agreed, and #1145 did exactly that: v0.9 now rejects legacy `allowedHosts`/`blockedHosts` centrally at the version gate, not per-backend.

src/core/wxc_common/src/config_parser.rs
SohamDas2021: v0.9 removed legacy host-list authoring outright, so the general rule you wanted now lives centrally rather than in this backend.

src/backends/wslc/common/src/state_aware.rs
SohamDas2021: Yes, I will check this in first.

src/core/wxc_common/src/config_contract_adapters/dev/one_shot.rs
SohamDas2021: this import is being used in `convert_wslc_port_mapping` further down in `convert_wslc`

src/core/wxc_common/src/config_contract_adapters/dev/one_shot.rs
SohamDas2021: These aren't removed, they moved to `common.rs` so the state-aware adapter can share them

src/core/wxc_common/src/wire.rs
SohamDas2021: Yes, "sizing knobs" is the wrong label, since `gpu` is a capability and `storagePath` a location; I'll reword that.

`cpuCount` / `memoryMb` / `gpu` / `storagePath` are **session-level** properties. State-aware WSLC runs a persistent daemon that owns **one shared WSLC session** across every sandbox on the host. Those values are fixed when that session is created before any particular `provision` call arrives. A per-sandbox value would either be silently ignored or would require tearing down and recreating the shared session, destroying the warm-reuse amortization that state-aware exists for.

I am documenting this and filing an ADO item to track this gap

schemas/dev/mxc-config.schema.0.9.0-dev.json
```diff
@@ -1159,6 +1159,16 @@
             "string",
             "null"
           ]
+        },
+        "portMappings": {
```
jsidewhite, line 1163: >portMappings [](http://example.com/codeflow?start=9&length=12)

Adding `portMappings` to the `provision` step this feels like othe wrong design, no?

Why can't it use the one from the general namespace (parent namespace)

jsidewhite (changes requested):
:clock1:

SohamDas2021:
Closing this, would be moving to a fresh PR once the state-aware redesign is out.

Issue #1002, opened 2026-08-22
LXC E2E: the GA egress allow-path test depends on unauthenticated api.github.com


dhoehna:
Raised on #983 by @bbonaby, marked non-blocking.

## What happens

`tests/configs/lxc_network_ga_egress_allow.json` uses `api.github.com` as the reachable peer for the positive-path check.

A prior LXC run received a 403 rate-limit response from that endpoint and the suite reported the policy as over-blocking.  Remote service health and a firewall regression are currently indistinguishable to this test.

## Suggested fix

Point the positive path at a CI-controlled peer.  The failure mode to remove is a red test caused by something outside the repository.

PR #1044, opened 2026-08-25
[LXC] Stop blaming host loopback for a LAN inbound refusal

### LINKED ISSUE #1024: LXC: ingress.default rejection reports an incorrect host-loopback rationale
> **Security notice:** When reporting BSODs or security issues, **DO NOT** attach memory dumps, logs, or traces to GitHub issues. Instead, send them to secure@microsoft.com referencing the GitHub issue. For application crashes, include a Feedback Hub link if possible (open with Win+F, choose "Share My Feedback" after submission).

### Relevant area(s)
Linux

The LXC rejection message for `network.ingress.default: "allow"` says the field requests "permissive host-loopback inbound" and explains that a host-loopback forwarder is missing. `network.ingress.default` governs LAN/private-network inbound, not host loopback, so the diagnostic gives the user the wrong reason for rejection.

This was identified while verifying review comments on PR #983. The behavior remains fail-closed; the defect is the inaccurate user-facing explanation.

### Steps to reproduce
1. Submit a schema 0.8 LXC request with `network.ingress.default: "allow"` and `network.ingress.hostLoopback: "deny"`.
2. Run the request through the LXC backend.
3. Read the returned inbound network policy error.

### Expected behavior
The error uses a field-neutral explanation or separately explains that permissive LAN/private inbound is not implemented for `network.ingress.default`.

### Actual behavior
The error identifies `network.ingress.default` but describes it as "permissive host-loopback inbound" and gives a host-loopback-specific rationale.

dhoehna:
## 📖 Description

`network.ingress.default` and `allowLocalNetwork` govern LAN/private-network inbound;
`network.ingress.hostLoopback` is the separate host-loopback control
([policy.md](docs/sandbox-policy/0.8.0/policy.md)).  The refusal named whichever field the
operator wrote, then explained all three as "permissive host-loopback inbound" and pointed at a
`loopbackPorts` field and a forwarder that do not exist.  An operator who set `ingress.default`
was sent to a field they had not touched.

The refusal now gives the rationale that covers all three, which the LXC backend doc already
states: LXC has a single inbound chain and no way to scope an accept to particular ports,
sources, or interfaces, so the only rule available today would accept new inbound from every
interface and source.

The path stays fail-closed and the refusal still names the field.  The explanation is the only
observable change.  The module doc carried the same conflation and is corrected with it.

## 🔗 References

Resolves #1024

## 🔍 Validation

- New regression test `a_lan_inbound_refusal_does_not_give_a_host_loopback_rationale` drives
  `apply_firewall_rules` for both LAN-inbound fields and asserts the refusal names the field and
  gives no host-loopback rationale.  It fails on `main` and passes here.
- `cargo test -p lxc_common` — 342 passed, 0 failed.
- `cargo fmt --all -- --check` and `cargo clippy -p lxc_common --all-targets` — clean.

## ✅ Checklist

- [x] Linked to an issue
- [x] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1044)

### DIFF
--- a/src/backends/lxc/common/src/network_ingress.rs
+++ b/src/backends/lxc/common/src/network_ingress.rs
@@ -34,17 +34,19 @@
 //! cannot run, the inbound deny is unenforceable for that family, so the run
 //! fails closed rather than silently leaving IPv6 open.
 //!
-//! **Permissive path is not yet implemented.** `allowLocalNetwork: true` asks
-//! for the sandboxed process to bind, listen, and accept incoming connections
-//! (see `ContainerPolicy::allow_local_network`). The policy carries no way to
-//! narrow that to particular ports or sources, so the only rule available today
-//! is an unscoped `--state NEW -j ACCEPT` accepting inbound from every
-//! interface and source, LAN and WAN included. Rather than install that
-//! silently, [`IngressManager::apply_firewall_rules`] returns a
-//! not-yet-implemented error. Narrowing it needs a `loopbackPorts` policy field
-//! and an MXC-owned host-loopback forwarder, tracked as AB#63505947. The
-//! internal rule *builder* still supports both toggle values so the decision
-//! table is testable.
+//! **Permissive path is not yet implemented.** `allowLocalNetwork: true`, and its
+//! 0.8 successors `network.ingress.default: "allow"` and
+//! `network.ingress.hostLoopback: "allow"`, each ask for the sandboxed process to
+//! bind, listen, and accept incoming connections. LXC has a single inbound chain
+//! and the policy carries no way to narrow an accept to particular ports, sources,
+//! or interfaces, so the only rule available today is an unscoped
+//! `--state NEW -j ACCEPT` accepting inbound from every interface and source, LAN
+//! and WAN included. Rather than install that silently,
+//! [`IngressManager::apply_firewall_rules`] returns a not-yet-implemented error
+//! naming the field the operator wrote. Scoping the host-loopback field on its own
+//! additionally needs a `loopbackPorts` policy field and an MXC-owned forwarder,
+//! tracked as AB#63505947. The internal rule *builder* still supports both toggle
+//! values so the decision table is testable.
 //!
 //! **Why the container netns.** A packet destined to a container socket
 //! traverses the *container's* `INPUT` chain, inside the container's network
@@ -561,12 +563,13 @@ impl IngressManager {
         // is mandatory), so there is no inert path that could safely emit it.
         if let Some(field) = Self::permissive_inbound_field(policy, uses_directional_schema) {
             return Err(format!(
-                "{field} (permissive host-loopback inbound) is not yet implemented for the \
-                 LXC firewall path. Scoping inbound to host loopback requires a loopbackPorts \
-                 policy field and an MXC-owned host-loopback forwarder that do not exist yet; \
-                 the only rule available today would accept new inbound connections from every \
-                 interface and source (LAN and WAN), which is broader than requested. Refusing \
-                 rather than installing an over-broad accept."
+                "{field} asks for permissive inbound, which is not yet implemented for \
+                 the LXC firewall path. LXC has a single inbound chain and the policy \
+                 carries no way to scope an accept to particular ports, sources, or \
+                 interfaces; the only rule available today would accept new inbound \
+                 connections from every interface and source (LAN and WAN), which is \
+                 broader than requested. Refusing rather than installing an over-broad \
+                 accept."
             ));
         }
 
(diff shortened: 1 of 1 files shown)


src/backends/lxc/common/src/network_ingress.rs
```diff
@@ -34,17 +34,19 @@
...
-//! narrow that to particular ports or sources, so the only rule available today
-//! is an unscoped `--state NEW -j ACCEPT` accepting inbound from every
-//! interface and source, LAN and WAN included. Rather than install that
-//! silently, [`IngressManager::apply_firewall_rules`] returns a
-//! not-yet-implemented error. Narrowing it needs a `loopbackPorts` policy field
-//! and an MXC-owned host-loopback forwarder, tracked as AB#63505947. The
-//! internal rule *builder* still supports both toggle values so the decision
-//! table is testable.
+//! **Permissive path is not yet implemented.** `allowLocalNetwork: true`, and its
+//! 0.8 successors `network.ingress.default: "allow"` and
```
SohamDas2021: network.ingress.hostLoopback is not a successor of allowLocalNetwork. The 0.8 migration maps allowLocalNetwork only to network.ingress.default; hostLoopback  is a new, separate control with no 0.7 equivalent. Please reword this to distinguish.

SohamDas2021 (dismissed):

src/backends/lxc/common/src/network_ingress.rs
```diff
@@ -34,17 +34,19 @@
...
+//! `network.ingress.hostLoopback: "allow"`, each ask for the sandboxed process to
+//! bind, listen, and accept incoming connections. LXC has a single inbound chain
```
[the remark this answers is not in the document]
dhoehna: Fixed in 8f798d6.  You're right, and the migration table says so directly: `allowLocalNetwork` maps only to `network.ingress.default`, while `network.ingress.hostLoopback` is listed as `No equivalent` (docs/sandbox-policy/0.8.0/networking/schema-updates.md:18-19).

The header now names `network.ingress.default` as the sole 0.8 successor and calls out `hostLoopback` as new in 0.8 with no 0.7 equivalent.  The paragraph groups the three only by the property that matters here, which is that each one asks for permissive inbound.

Worth noting the regression test this PR adds already had the distinction right; only the module header was wrong.

src/backends/lxc/common/src/network_ingress.rs
dhoehna: Fixed in 8f798d6.  You're right, and the migration table says so directly: `allowLocalNetwork` maps only to `network.ingress.default`, while `network.ingress.hostLoopback` is listed as `No equivalent` (docs/sandbox-policy/0.8.0/networking/schema-updates.md:18-19).

The header now names `network.ingress.default` as the sole 0.8 successor and calls out `hostLoopback` as new in 0.8 with no 0.7 equivalent.  The paragraph groups the three only by the property that matters here, which is that each one asks for permissive inbound.

Worth noting the regression test this PR adds already had the distinction right; only the module header was wrong.

SohamDas2021:
> ## Pull request overview
> Copilot reviewed 1 out of 1 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (1)
> **Previously missed (1)** — in code that hasn't changed since the last review.
> 
> **src/backends/lxc/common/src/network_ingress.rs:50**
> 
> * `loopbackPorts` is not a missing policy field: the GA contract deliberately has no ingress port rules and defines `ingress.default` plus `ingress.hostLoopback` as the complete ingress surface (`docs/sandbox-policy/0.8.0/networking/networking.md:241-242`). Keeping it as a prerequisite still directs maintainers toward the nonexistent field this change is intended to stop recommending. Describe the missing bidirectional host-loopback routing/forwarding implementation instead.
> 
> ```
> //! naming the field the operator wrote. Scoping the host-loopback field on its own
> //! additionally needs a `loopbackPorts` policy field and an MXC-owned forwarder,
> //! tracked as AB#63505947. The internal rule *builder* still supports both toggle
> ```

shall we fix this as well?

dhoehna:
CI is green again - all 26 checks pass.

The earlier red `LXC-Exec Container and Network Policy` was not this diff.  It failed on `LXC Network GA Egress (0.8)`, whose allow case wgets `https://api.github.com/zen` and allows `140.82.112.0/20`, so a rate-limit or outage at GitHub is reported as the firewall over-blocking.  A rerun on the same commit passed.  #1002 tracks replacing that live dependency with a controlled peer.

This PR only touches ingress, and the ingress case passed in the failing run too.  Re-requesting review since the doc-comment push dismissed the earlier approval.

SohamDas2021 (approved):

dhoehna:
Yes -- done in 331a48a, which is on #1054.

`loopbackPorts` was not a real field.  `ingress.default` and `ingress.hostLoopback` are the complete GA ingress surface, and both the 0.8 and 0.9 schemas close the ingress object to exactly those two properties, so anyone following that comment would have gone looking for a contract gap that is not there.  The header now names what is actually missing -- cross-namespace plumbing in both directions, since the container's 127.0.0.1 is not the host's, plus a runtime port-mapping contract the allow/deny policy does not express.

I narrowed the sentence above it too.  It called all three settings inbound bind/listen permission, and `hostLoopback` governs both directions.

PR #1050, opened 2026-08-26
Bring C# SDK up to parity with Rust SDK.

shschaefer:
## 📖 Description
Hardens and expands the .NET SDK API.

### Security and policy correctness

- Migrates the obsolete root `SandboxPolicy.CaptureDenials` value into `ProcessContainerContainment.CaptureDenials` when using `SandboxRequest`.
- Rejects conflicting capture-denial settings and incompatible containment instead of silently weakening policy.
- Marks the legacy property obsolete and documents the migration.
- Removes the system-drive read grant from PowerShell policy discovery.
- Revalidates mutable IsolationSession network options before serialization.
- Rejects malformed lifecycle sandbox IDs even when callers override the schema version.

### SDK API improvements

- Adds injectable `ISandboxRunner` and `ISandboxLifecycle` adapters while retaining the existing static convenience APIs.
- Adds `ISandboxProcess` so consumers can supply in-memory fakes for streaming execution.
- Passes `StateAwareExecOptions.TimeoutMs` into lifecycle processes, aligning timeout reporting with one-shot execution.
- Makes `MxcSandboxProcess.WaitBlocking()` private so public waits do not prevent concurrent `Kill()` or `Dispose()`.
- Caches streaming warnings and adds forward-compatible `Unknown` discovery values.
- Lets state-aware WSLC `exec` supply a proxy without restating `network.defaultPolicy`, which is fixed at provision.
- Adds a `stateAwareWslc` schema tier (`0.8.0-alpha`) so WSLC's state-aware default promotes independently.

### Compatibility

- Strengthens C#/Rust API parity checks across request fields, policies, backend mappings, capabilities, tiers, and lifecycle sandbox-ID prefixes.
- Makes the network proxy JSON converter discoverable through its type-level attribute.
- Enables the default .NET test suite on Linux and macOS CI.
- Replaces timing-dependent tests with explicit synchronization.
- Updates the .NET SDK documentation for capture-denial migration, dependency injection, lifecycle timeout behavior, and process waiting.

## 🔍 Validation
- Default .NET suite: 107 passed, 24 host-dependent tests skipped
- Optional-backend .NET suite: 107 passed, 24 host-dependent tests skipped
- Rust `mxc_ffi`: 53 passed, 4 ignored
- C#/Rust API parity passed
- Schema-version synchronization passed
- C# native-bindings code generation passed
- `git diff --check` passed

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1050)


sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -295,14 +499,10 @@ private SandboxWaitResult WaitCore(CancellationToken cancellationToken)
...
-    /// <summary>
-    /// Block until the child exits using the native blocking wait, which reports
-    /// a policy timeout distinctly (<see cref="SandboxWaitResult.TimedOut"/>).
-    /// Unlike <see cref="Wait"/> this cannot be interrupted by a concurrent
-    /// <see cref="Kill"/> — it holds the control lock for the whole wait — so use
-    /// it only when you will not race a kill.
-    /// </summary>
-    public SandboxWaitResult WaitBlocking()
+    // Called only after try_wait reports exit or the managed deadline expires.
+    // It holds the control lock across the native wait and must not be public:
```
SohamDas2021: This generalizes the TimeoutMs = 0 issue already flagged on MxcLifecycle.cs:200 — it applies to every timeout, not just zero. WaitCore reaches WaitBlocking() only after the managed deadline has expired, but neither native wait derives its deadline from spawn time: Windows is WaitForSingleObject(process, self.timeout_ms), Unix is Instant::now() + d. Both start a fresh, full-length window at call time, so a timed-out sandbox runs up to 2 × timeoutMs with _controlLock held for that entire second window. The comment at 485-489 asserts the opposite ("we are already at/past the deadline, so it returns promptly")?

Fix: have WaitCore call Kill() when its stopwatch expires before delegating, or record the spawn instant natively?

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -295,14 +499,10 @@ private SandboxWaitResult WaitCore(CancellationToken cancellationToken)
...
+    // callers need Wait/WaitAsync so concurrent Kill and Dispose stay responsive.
+    private SandboxWaitResult WaitBlocking()
```
SohamDas2021, line 531: Separate from the lifecycle signature breaks already noted on MxcLifecycle.cs:59 : WaitBlocking() was public at the base commit and is now private — another source + binary break. And Microsoft.Mxc.Sdk.csproj:11 is still  <Version>0.8.0</Version> for all of them. 

Please either bump to 0.9.0 across  src/Cargo.toml,  sdk/node/package.json, and the csproj, or keep [Obsolete] forwarders?

sdk/dotnet/Microsoft.Mxc.Sdk/StateAwareTypes.cs
```diff
@@ -28,24 +32,28 @@ public enum StateAwareNetworkDefault
...
-/// vocabulary the lifecycle phases use, a different layer from the one-shot
-/// <see cref="NetworkPolicy"/>.
+/// Network posture sent on a state-aware lifecycle request. Omitted values are
+/// resolved by the native backend using its fail-closed defaults.
 /// </summary>
 public sealed class StateAwareNetworkPolicy
 {
     /// <summary>The default action for outbound traffic.</summary>
-    public StateAwareNetworkDefault DefaultPolicy { get; set; }
+    public StateAwareNetworkDefault? DefaultPolicy { get; set; }
```
SohamDas2021, line 41: This and AllowLocalNetwork (line 44) becoming nullable is a CLR signature change on shipped properties, so compiled consumers break. The JSON also changes shape: defaultPolicy is now omitted where block was previously emitted, and several backends treat an absent network policy differently from an explicit deny (IsolationSession rejects an absent one outright). Same version-bump question as above.

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcLifecycleE2ETests.cs
```diff
@@ -171,6 +170,26 @@ public void Exec_PropagatesANonZeroExitCode()
         }
     }
 
+    /// <summary>A lifecycle exec timeout must be reported by the streaming
+    /// process wrapper just like a one-shot policy timeout.</summary>
+    [Fact]
+    public void Exec_ReportsConfiguredTimeout()
+    {
+        Assert.SkipUnless(HostRunsIsolationSession, "no isolation-session host available");
```
SohamDas2021, line 178: Exec_ReportsConfiguredTimeout is the only coverage for lifecycle exec TimeoutMs, and this gate means no CI lane runs it — consistent with the default and optional-backend suites both reporting an identical 107 passed / 24 skipped. That's why the WaitBlocking double-timeout is invisible to CI. 

Could we get a prepared-host lane, or an internal clock/native seam so the timeout path is exercisable without a live sandbox?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1532,7 +1539,18 @@ fn convert_wire_state_aware(
     }
 
     let require_process = phase == Phase::Exec;
-    let mut request = convert_wire_config(cfg, logger, require_process, allow_missing_command)?;
+    let state_aware_wslc_exec = phase == Phase::Exec
```
SohamDas2021, line 1542: This is a behavior change to the shared config parser — WSLC exec may now supply a proxy without restating defaultPolicy — landing inside a 6284-line C#-SDK-parity PR with no mention in the description. Reviewers scoped to the .NET surface won't look here, and it changes what every WSLC exec request may declare. 

Please call it out explicitly alongside the stateAwareWslc schema tier, or split it into its own bugfix PR?

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicyDiscovery.cs
```diff
@@ -0,0 +1,314 @@
...
+                normalizedParent + Path.DirectorySeparatorChar,
+                comparison);
+    }
+
+    private static void AddPowerShellPolicy(
+        IEnumerable<string> pathDirectories,
+        IReadOnlyDictionary<string, string?> environment,
+        ICollection<string> readwritePaths)
+    {
+        if (!OperatingSystem.IsWindows()
```
SohamDas2021, line 295: [Non-blocking] Rust and C# now grant different filesystem access for the same logical policy. This grants only PSReadLine, but Rust still grants the whole system-drive root read-only (mxc_engine/src/policy.rs:275), and its doc-comment at 248-255 claims it "mirrors the SDK". Rust is the more permissive side. 

If the narrower behavior is intended, narrow powershell_policy in Rust, update its tests and comment, and check getPowerShellPolicy in the TypeScript SDK for the same drift.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
```diff
@@ -29,11 +31,33 @@ public sealed class SandboxPolicy
...
+    /// <see cref="MxcSandbox.Spawn(SandboxPolicy, string)"/> overloads.
+    /// New code should set
+    /// <see cref="ProcessContainerContainment.CaptureDenials"/> explicitly.
     /// </summary>
+    /// <remarks>
+    /// Request-based execution migrates this value to ProcessContainer
+    /// containment. It rejects incompatible containment and conflicting
+    /// <see cref="ProcessContainerContainment.CaptureDenials"/> values.
+    /// </remarks>
+    [Obsolete("Set ProcessContainerContainment.CaptureDenials instead.")]
```
SohamDas2021: [Non-blocking] Two things: (a) the [JsonIgnore] on line 46 changes the documented SandboxPolicy  serialization contract — anyone deserializing, persisting, or forwarding legacy policy JSON silently loses captureDenials (MxcSandboxTests.cs:630 codifies it). 

Cleaner to keep it serializable and strip it only from the clone in SerializeRequest. 

(b) This attribute has no removal version and no DiagnosticId, so consumers can't tell how long the shim lasts or suppress just this warning — suggest [Obsolete("… Removed in 1.0.", DiagnosticId = "MXC0001")] plus the same note in the README.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandbox.cs
```diff
@@ -165,7 +239,146 @@ private static byte[] ToNullTerminatedUtf8(string value)
...
+        return JsonSerializer.Serialize(policy, JsonOptions);
+    }
+
+    internal static string SerializeRequest(SandboxRequest request)
+    {
+        ArgumentNullException.ThrowIfNull(request);
+        return JsonSerializer.Serialize(PrepareRequest(request), JsonOptions);
+    }
+
+    private static SandboxRequest PrepareRequest(SandboxRequest request)
```
SohamDas2021, line 251: [Non-blocking] The entire CaptureDenials legacy migration lives in the C# client, while Rust just fails closed on the old shape (request.rs:415). Every future language binding will re-implement this. 

Moving the normalization into the Rust deserializer would centralize it — and would make the [JsonIgnore] round-trip problem unnecessary.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxAdapters.cs
```diff
@@ -0,0 +1,247 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Mxc.Sdk;
+
+/// <summary>
+/// Injectable one-shot sandbox operations. Use <see cref="MxcSandboxRunner.Default"/>
+/// in production and substitute a fake or mock in unit tests.
+/// </summary>
+public interface ISandboxRunner
```
SohamDas2021, line 10: [Non-blocking] These interfaces are 9–13 members each with no default implementations and no abstract base class, so the first added lifecycle phase or RunAsync overload breaks every consumer mock. 

Related:  SandboxStreamCloser being sealed with an internal ctor (MxcSandboxProcess.cs:33) means external fakes can't be written today either. DIMs are supported on net8.0 and are nearly free before the first release.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
```diff
@@ -29,11 +31,33 @@ public sealed class SandboxPolicy
...
+    [JsonIgnore]
+    public CaptureDenialsPolicy? CaptureDenials
```
SohamDas2021: [Non-blocking] The policy/config POCOs are fully mutable ( { get; set; }  + mutable  List<T> ) but are validated and then serialized, so mutating after validation bypasses it silently. init + IReadOnlyList<T> fixes it — but that's itself source-breaking for the mutate-after-construction pattern the README uses, so it belongs in the same version bump as the other breaking changes, not a later one.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxAdapters.cs
```diff
@@ -0,0 +1,247 @@
...
+    ISandboxProcess Spawn(SandboxPolicy policy, string command);
+
+    /// <summary>Spawn a complete request with live standard streams.</summary>
+    ISandboxProcess Spawn(SandboxRequest request);
+}
+
+/// <summary>
+/// Stateless <see cref="ISandboxRunner"/> adapter over <see cref="MxcSandbox"/>.
+/// </summary>
+public sealed class MxcSandboxRunner : ISandboxRunner
```
SohamDas2021, line 48: [Non-blocking] This delegates to the static facade rather than the reverse, and MxcSandbox.Run still P/Invokes  mxc_run_request directly (MxcSandbox.cs:113-149). So the injection seam only exists for consumers who restructure around the interfaces; existing static-API code still needs native bits to test. 

Consider moving the native calls into the adapter and having the statics delegate through a default instance.

sdk/dotnet/README.md
```diff
@@ -72,28 +72,339 @@ when `permissiveLearningMode` disabled deny-by-default. MXC never writes these
 to the host's stderr, so inspecting `Warnings` is the only way to learn that
 containment was relaxed.
 
-### Denial capture (Windows)
+Streaming callers receive the same warnings immediately from
+`MxcSandboxProcess.Warnings`. They do not need to wait for the process to exit.
+
+### Dependency injection and testing
+
+Applications that use constructor injection can depend on `ISandboxRunner` for
```
SohamDas2021, line 83: [Non-blocking] Keeping the static MxcSandbox / MxcLifecycle methods alongside the new interfaces forks the execution surface — every future feature needs implementing and documenting twice, with no clear canonical entry point. Suggest exposing MxcSandboxRunner.Default  /  MxcSandboxLifecycle.Default as the non-DI path and deprecating the static forwarders. (Anchored in the README, but the fix is code — pairs with #9.)

src/ffi/mxc_ffi/src/request.rs
```diff
@@ -0,0 +1,534 @@
...
+};
+use mxc_sdk::{
+    build_request_with_containment, Containment, Error, ErrorCode, SandboxPolicy, SandboxRequest,
+    WslcSection,
+};
+use serde_json::Value;
+
+#[derive(serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+struct RequestSpec {
```
SohamDas2021, line 20: [Non-blocking] [Follow-up] The request/policy/containment model is now hand-maintained in three places — C# POCOs, these Rust structs, and the golden JSON — held together only by the parity script, which has already picked up three separate defects in this PR alone (CodeQL at  :98 / :130 , the non-advancing loop at  :155 ). Adding one field is a synchronized edit across three ecosystems. Worth treating the canonical JSON schema as the source and generating all three, consistent with  mxc_schema_gen  / csbindgen elsewhere in the repo.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -315,20 +512,20 @@ private static void SetBackendConfig(
         phaseConfig[key] = value;
     }
 
-    // Run an envelope phase via mxc_state_aware and return the parsed `result`
-    // object (may be an empty object). Throws MxcException on failure.
-    private static JsonObject? RunEnvelopePhase(JsonObject envelope)
+    private static JsonObject? RunEnvelopePhase(JsonObject envelope, bool dryRun)
```
SohamDas2021, line 521: [Non-blocking] [Follow-up]  MxcLifecycle is now 450+ lines mixing the public facade, JSON envelope construction ( SetCrossCuttingPolicies  at 462, the  Build*Envelope  helpers), validation, and raw P/Invoke marshalling. 

Extracting the envelope building into a pure  LifecycleRequestBuilder  would make the JSON shaping unit-testable without native code — currently impossible.

sdk/dotnet/golden/request-directional-network.json
```diff
@@ -0,0 +1,41 @@
+{
+  "policy": {
+    "version": "0.8.0-alpha",
+    "network": {
+      "allowOutbound": false,
+      "allowLocalNetwork": false,
+      "allowedHosts": [],
+      "blockedHosts": [],
```
bbonaby: note: for 0.8.0 networking you should omit 

`
     "allowOutbound": false,
      "allowLocalNetwork": false,
      "allowedHosts": [],
      "blockedHosts": [],
`
if you mix it with the new ergress and ingress we'll throw an error

sdk/dotnet/golden/request-process-container.json
```diff
@@ -0,0 +1,60 @@
...
+      "allowInputInjection": false
+    },
+    "timeoutMs": 30000
+  },
+  "command": "echo parity",
+  "containment": {
+    "type": "processContainer",
+    "leastPrivilege": true,
+    "learningMode": false,
+    "capabilities": ["internetClient"],
```
bbonaby: note: since this config is requesting to use a proxy configuration proxy, the OS will return an error if you pass in internetClient since the container shouldn't have direct internet access. The change for this file would be to remove it from the capabilities list.

sdk/dotnet/golden/request-process-container.json
```diff
@@ -0,0 +1,60 @@
...
+      "readwritePaths": ["C:\\work"],
+      "readonlyPaths": ["C:\\input"],
+      "deniedPaths": ["C:\\secret"],
+      "clearPolicyOnExit": true
+    },
+    "network": {
+      "allowOutbound": false,
+      "allowLocalNetwork": false,
+      "allowedHosts": [],
+      "blockedHosts": [],
```
bbonaby: note: same comment like above

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
adpa-ms: [AgentAssisted] Native-side fact for the forwarders-vs-bump question (also open on `MxcLifecycle.cs:59`) — I think it rules forwarders out.

This PR breaks the C ABI as well as the CLR surface: `mxc_sandbox_try_wait` goes from three out-parameters to four (`src/ffi/mxc_ffi/src/streaming.rs:678`), managed caller updated in lockstep.

An assembly compiled against 0.8.0 still P/Invokes the three-parameter form. Point it at a newer `mxc_ffi.dll` — reachable via `MXC_FFI_DIR` or an assembly-directory override — and the callee reads a fourth pointer that was never passed. The `out_timed_out.is_null()` guard only helps when that stale argument happens to be null; otherwise it writes an `i32` through garbage.

That makes the options asymmetric. Today such a consumer fails loudly with `MissingMethodException` before any P/Invoke; with forwarders it binds, runs, then corrupts — in exactly the configuration the shims exist to serve, restoring source compatibility while advertising a binary guarantee the native layer no longer honours.

Both ship in one package and are co-versioned (the crate docs call the C ABI "not a stable external contract"), so the bump looks like the honest fix: move off the released 0.8.0 and let the break be explicit.

.github/workflows/SDK.Dotnet.Test.Job.yml
```diff
@@ -65,3 +61,14 @@ jobs:
...
+      # Exercise the opt-in backends as one complete native unit. The managed
+      # tests dry-run both backends and verify WSLC's daemon + SDK DLL were
+      # propagated into the application output.
+      - name: Build .NET SDK with optional backends
+        if: matrix.os_label == 'windows'
+        run: dotnet build Microsoft.Mxc.Sdk.slnx -p:MxcWithWslc=true -p:MxcWithIsolationSession=true
+
+      - name: Test optional backend availability
+        if: matrix.os_label == 'windows'
+        run: dotnet test Microsoft.Mxc.Sdk.Tests/Microsoft.Mxc.Sdk.Tests.csproj --no-build --no-restore -p:MxcWithWslc=true -p:MxcWithIsolationSession=true
```
adpa-ms: [AgentAssisted] This is the only CI coverage for the `isolation_session` and `wslc` feature builds, and the only step here without the skip guard.

The default step above captures the log and parses `succeeded: (\d+)` because, as its own comment says, "a run in which everything skipped also exits 0". Nearly every test behind these flags is host-gated with `Assert.SkipUnless`, so that applies here with more force.

It isn't vacuous today — the two `*BuildSwitch_MatchesNativeAvailability*` tests assert on both branches unconditionally. But nothing pins that, and the failure mode is silent: if they ever became gated, this lane would keep reporting green while compiling the features and running nothing. The PR's own validation numbers show both suites at an identical 107 passed / 24 skipped, so the summary doesn't currently distinguish a feature-on run from a feature-off one.

Suggest the same capture-and-parse guard, ideally asserting a higher count than the default lane so the added coverage is what's verified.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -259,43 +325,174 @@ public static async Task<RunResult> ExecInSandboxAsync(
...
+        }
+
+        return id.Value[..separator] switch
+        {
+            "iso" => StateAwareContainment.IsolationSession,
+            "wsb" => StateAwareContainment.WindowsSandbox,
+            "wslc" => StateAwareContainment.Wslc,
+            _ => throw new MxcException(
+                ErrorCode.UnsupportedContainment,
+                $"no state-aware backend is registered for sandbox id '{id.Value}'"),
```
adpa-ms: [AgentAssisted] Two ids that are *missing* a prefix are misclassified here — `":payload"` (empty) and `default(SandboxId)` (null `Value`). An unrecognized non-empty prefix is correctly `UnsupportedContainment`.

`new SandboxId(":payload")` is accepted (the ctor rejects only null/empty), then passes the `separator < 0` guard because `IndexOf` returns 0, slices to `""`, and reaches `UnsupportedContainment`. But `""` isn't a registrable name — it's structural, which line 447 already routes to `MalformedId` for the missing-colon sibling. Native pins the same split: `parse_prefix_rejects_empty_prefix` asserts `":no-prefix"` → `MalformedId`.

`default(SandboxId)` is legal and leaves `Value` null, so line 443 throws `NullReferenceException` instead of a typed `MxcException`.

Neither is urgent — both fail loudly — but one fix covers both: treat null/empty `Value`, missing colon, and `separator == 0` as `MalformedId` before slicing. The added tests cover only the missing-colon form.

src/ffi/mxc_ffi/Cargo.toml
```diff
@@ -14,9 +14,13 @@ crate-type = ["cdylib", "staticlib", "lib"]
 
 [dependencies]
 mxc-sdk = { workspace = true }
+serde = { workspace = true }
 serde_json = { workspace = true }
 
 [features]
+# WSL Container backend (Windows only, experimental). The managed package must
+# also stage wxc-wslc-daemon.exe and wslcsdk.dll beside its application host.
+wslc = ["mxc-sdk/wslc"]
```
adpa-ms: [AgentAssisted] This says "application host", but the change in this PR moved sidecar resolution off exactly that.

`WslcSdk::load` resolves `wslcsdk.dll` via `module_path_for_address(Self::load as *const ())` — the directory of the loaded `mxc_ffi` module, not the executable. `module_path_for_address`'s own doc gives the reason: a framework-dependent .NET app loads `mxc_ffi.dll` from its NuGet runtime-asset directory while `current_exe` points at `dotnet.exe` or the apphost.

Following this comment would stage both files next to the host and `load()` wouldn't find them. Suggest "beside the loaded `mxc_ffi` native library (its `runtimes/<rid>/native` asset)".

src/ffi/mxc_ffi/src/request.rs
shschaefer: I moved the golden JSON.   The parity script should be replaced with a generator, but currently there lacks a clear definition of what to generate.  csbindgen is a mechanical translation of the FFI.  mxc_schema_gen covers the "wire" format which is the wrong shape.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
shschaefer: The envelope building is already separated and already unit-tested without native code —  Build{Provision,Start,Exec,Stop,Deprovision}Envelope  are internal static and return JsonObject with no P/Invoke (all native calls go through the single private RunEnvelopePhase  at :515). Microsoft.Mxc.Sdk.csproj:76  grants InternalsVisibleTo to the test assembly, and  MxcLifecycleTests.cs exercises the builders directly at 10 call sites, including the validation failure paths.

So the design goal here is met — just via an access modifier rather than a separate type. Promoting them to a public LifecycleRequestBuilder would expand semver-bound public API surface (gated by  check-dotnet-api-parity.js ) with no coverage gain, and an internal builder would be equivalent to what's there. Declining as a no-op refactor; happy to revisit if the file grows another backend.

sdk/dotnet/README.md
shschaefer: MxcSandboxRunner.Default  and  MxcSandboxLifecycle.Default  already exist, so the non-DI entry point is in place.
On deprecating the statics: the direction is the reverse of "forwarders" —  MxcSandbox / MxcLifecycle  are the implementations and the adapters forward to them. Marking them  [Obsolete]  would make the adapters' own call sites CS0618 requiring pragma suppression across a large number of members.
It would also invert this PR's premise: the Rust SDK's canonical surface is free functions ( mxc_sdk::run ,  spawn_sandbox ,  exec_attached  —  lib.rs:184/166/256 ) with no runner type. The static facade is the parity surface; the interfaces are a C#-idiomatic DI layer on top. This matches  TimeProvider  in .NET 8, which didn't deprecate  DateTime.UtcNow .

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxAdapters.cs
shschaefer: This is inverted again.  The point is to create a seam for consumers that use it.  MxcSandboxRunner.Default  is get only so statics delegating through it are still pinned to one concrete native-backed instance. Getting real substitutability requires making Default settable, i.e. mutable static ambient state, which pollutes across xUnit's parallel collections and is order-dependent.  It is moving code around for no benefit.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
shschaefer: This diverges from  mxc_engine::policy::SandboxPolicy which is a mutable struct with #[serde(default)] pub fields.  Semantic validation happens natively on every call, after serialization, so a post-construction mutation is always re-checked. Its an implementation pattern shared by the C#, Rust, and TS SDKs.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxAdapters.cs
shschaefer: External fakes are already writable.  ISandboxProcess  exposes  ISandboxStreamCloser?, not the sealed type and ISandboxStreamCloser is public. SandboxStreamCloser is sealed because it wraps MxcStreamCloserHandle, an unmanaged resource; the interface is the seam.

On DIMs: these interfaces are for test doubles, so a default body would have to be  throw new NotSupportedException()  — turning a compile error in a test project into a runtime failure, with a fake that silently claims to support a phase it doesn't.  Also, at  <Version>0.8.0</Version>  we're pre-1.0, so breaking changes are still on the table — that's the window where DIM insurance buys least. And interface growth isn't accidental: InjectableContractsMirrorStaticFacadeMethods fails the build if a public static lands without a matching interface member.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandbox.cs
shschaefer: captureDenials  on the policy doesn't exist outside C# — zero occurrences in the TS SDK's  types.ts , and  mxc_engine::policy  only has it on  Containment / ProcessContainer . It's back-compat debt from the C# SDK's own prior shape, so a future binding has nothing to re-implement: it starts at the canonical  containment.captureDenials  and never had the alias.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
shschaefer: Agree.  Fixed.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicyDiscovery.cs
shschaefer: I did this intentionally.  There is churn in the definition of this policy.  Rust, TS and C# will need to be updated to be consistent.

src/core/wxc_common/src/config_parser.rs
shschaefer: Fair call on disclosure — adding both to the description (the parser fix and the  stateAwareWslc  tier), since a .NET-scoped reviewer wouldn't look at config_parser.rs

sdk/dotnet/Microsoft.Mxc.Sdk/StateAwareTypes.cs
shschaefer: Yes, its a breaking change.  IsolationSession semantics are changing and should already not require a network policy in post provision phases.  I can document the breaking change, but this is the correct semantic for MXC - default deny.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
shschaefer: The premise that JSON shaping is "currently impossible" to unit-test without native code isn't accurate — there are already two independent seams.

1. The builders are already pure and already directly tested.
2. There's also a public dry-run API for every phase.

Maintaining parity across SDKs remains the priority.

.github/workflows/SDK.Dotnet.Test.Job.yml
shschaefer: Fixed:  Two steps. First, the same capture-and-parse guard as the default lane, for the reason you gave. Second, a by-name pin that's strictly stronger than a count:

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
shschaefer: Fixed and verified. Both defects were real and both are genuine parity divergences from native.

src/ffi/mxc_ffi/Cargo.toml
shschaefer: Fixed

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcLifecycleE2ETests.cs
shschaefer: Fixing the double-timeout issue.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
shschaefer: Fixing the double timeout issue.

bbonaby (approved):

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxPolicy.cs
```diff
@@ -29,9 +29,27 @@ public sealed class SandboxPolicy
...
+    /// This property stays serializable so legacy policy JSON round-trips
+    /// without silently losing the value. It is stripped from the clone sent
+    /// natively once it has been migrated onto the containment, so the native
+    /// layer never sees the deprecated shape.
+    /// </para>
+    /// </remarks>
+    [Obsolete(
+        "Set ProcessContainerContainment.CaptureDenials instead. Removed in 1.0.",
+        DiagnosticId = "MXC0001",
+        UrlFormat = "https://github.com/microsoft/mxc/blob/main/sdk/dotnet/README.md#{0}")]
```
bbonaby, line 52: note: I don't think anyone was actually using this yet so probably not needed.

SohamDas2021 (approved):

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -295,14 +504,31 @@ private SandboxWaitResult WaitCore(CancellationToken cancellationToken)
...
+        // A child that exited in the window between the poll above and here ran
+        // to completion, so surface its real status rather than a timeout.
+        if (TryGetTerminalStatus(out _, out _))
+        {
+            return WaitBlocking();
+        }
+
+        // Safe to call after an exit that races this check: every backend
+        // no-ops a kill once the child has been reaped.
+        Kill();
```
adpa-ms, line 522: [AgentAssisted] The comment at 520 doesn't hold for the state-aware exec path — the IsolationSession one.

`ExecSandboxProcess::kill` no-ops only when `self.exit.is_some()`, and `self.exit` is set solely by `join_waiter`. A process finishing after the poll at 515 is exited but not yet reaped, so the guard misses and the terminator runs against a corpse. The repo already treats that as an error: `kill_after_a_completed_wait_does_not_run_the_terminator` pins it with a terminator returning `Err("no such process")`.

`Wait()` then throws `MxcException` instead of returning `TimedOut`, and latches `kill_refused`, so every later `Kill()` claims "the sandbox refused an earlier request to terminate this exec" — untrue for an already-dead process.

Likelihood is low — one-shot backends are unaffected, and `WaitCore` checks the terminal branch first while the native waiter starts earlier, so 522 is usually unreachable. Worth closing since the comment states the opposite.

Suggest native `kill` join a finished waiter before terminating, or re-poll after a failed kill and propagate only if still running.

sdk/dotnet/CHANGELOG.md
```diff
@@ -0,0 +1,68 @@
...
+  MxcLifecycle.ProvisionSandbox(
+      StateAwareContainment.IsolationSession,
+      new ProvisionSandboxOptions { … });
+  ```
+
+- `StartSandbox`, `StopSandbox`, `DeprovisionSandbox`, and `ExecInSandbox`
+  accept an optional trailing options argument. Source-compatible; existing
+  call sites keep compiling. Because adding a parameter changes the emitted
+  method signature, assemblies compiled against 0.8.0 must be recompiled
+  rather than dropped in place.
```
adpa-ms, line 47: [AgentAssisted] The nullable `StateAwareNetworkPolicy` change was acknowledged as a break on the `StateAwareTypes.cs:41` thread but isn't covered here.

`DefaultPolicy` went from `StateAwareNetworkDefault` to `StateAwareNetworkDefault?`, and `AllowLocalNetwork` from `bool` to `bool?` — a CLR signature change on properties shipped in 0.8.0, so this bullet's recompile note applies to them too.

The JSON shape also changed, which this bullet's class doesn't cover: with `WhenWritingNull`, `defaultPolicy` is now omitted where `"block"` was emitted, and absent is not equivalent to explicit deny — IsolationSession refuses an absent policy at provision.

Worth a bullet here so the 0.9.0 upgrade note is complete.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcSandboxProcess.cs
```diff
@@ -8,6 +8,62 @@
...
+
+    /// <summary>Check whether the child has exited without blocking.</summary>
+    bool TryGetExitCode(out int exitCode);
+
+    /// <summary>Wait while capturing standard output and error.</summary>
+    Task<(SandboxWaitResult Result, byte[] Stdout, byte[] Stderr)> WaitForExitWithOutputAsync(
+        CancellationToken cancellationToken = default);
+
+    /// <summary>Kill the child and its process tree.</summary>
+    void Kill();
```
adpa-ms, line 64: [AgentAssisted] This promises more than IsolationSession can deliver, and it's the contract new consumers bind to since `ExecInSandbox` hands back `ISandboxProcess`.

For state-aware exec, `mxc-sdk` states the opposite: "`Sandbox::kill` reaches only the foreground process here; a descendant the workload backgrounded is reclaimed when the sandbox is stopped and deprovisioned." A caller trusting a tree-kill may skip `StopSandbox`/`DeprovisionSandbox` and leave descendants running.

`Id` on the concrete class already handles this exact case — it documents returning 0 for a process from `ExecInSandbox` because a state-aware exec is driven behind the backend's own waiter/terminator. The same caveat fits here.

Suggest qualifying the reach as backend-dependent, noting IsolationSession descendants survive until stop/deprovision.

PR #1051, opened 2026-08-26
[Bubblewrap] Reuse the validated egress plan and pin IPv6/ICMP rendering

### LINKED ISSUE #991: [Bubblewrap] Directional network follow-ups from PR #979 review
### Description of the task

Four non-blocking follow-ups raised in review of #979 (GA directional network schema for Bubblewrap). None affect correctness or enforcement — all four fail closed today — so they are deferred rather than blocking the PR. The blocking item from the same batch (bounding cumulative peer expansion in `lower_peers`) was fixed in #979 directly.

**1. Don't lower the same directional policy twice per spawn** (performance)

`EgressPlan::for_request` is computed at `src/backends/bubblewrap/common/src/bwrap_runner.rs:213` (`validate`) and recomputed at `:412` (the firewall spawn path). That is duplicate CIDR subtraction and rule expansion on every spawn. Reusing a prepared/validated plan would remove the duplicate work and keep validation and installation tied to the same derived result.

**2. Centralize the directional-shape predicate** (maintainability)

The `network_egress.is_some() || network_ingress.is_some()` discriminator (and its negation) appears at three production sites:

- `src/backends/bubblewrap/common/src/bwrap_command.rs:182` — inside `impl ResolvedNetworkMode`
- `src/backends/bubblewrap/common/src/bwrap_command.rs:450` — inside `external_proxy_host_rules_rejection`
- `src/backends/bubblewrap/common/src/network_rules.rs:795` — inside `EgressPlan::for_request`

Correction to the review text: `directional_network_rejection` (`bwrap_command.rs:308`) does **not** contain the two-field test, so it is not a fourth site; the review named it in place of `ResolvedNetworkMode`.

Checking only one section can silently route a policy into the wrong enforcement path, so a shared `is_directional(&ContainerPolicy)` helper would make this fail-closed invariant harder to drift.

Note when doing this: `wxc_common::validator.rs:66` already has a *related but different* helper, `directional_posture_supplied`, which also folds in `network_mode_specified` and the proxy. The two must not be conflated.

**3. Add a live directional config exercising IPv6 CIDR and ICMP-to-ICMPv6 rendering** (test)

None of the seven `tests/configs/bubblewrap_network_directional_*.json` configs contain an IPv6 CIDR or an ICMP peer, so no directional E2E sends those rules through the real `ip6tables-restore`. The string-level unit tests pin the rendered payload, but a token or transaction incompatibility with the real tool would not be caught by them.

Connectivity is not required — asserting successful installation/debug output is enough.

Related: this is the same family as item 6 on #955 (the live inbound test asserts only the IPv4 ingress chain), and the two should probably land together.

**4. The proxy parity test hard-depends on `https://api.github.com/zen`** (test)

`tests/configs/bubblewrap_network_directional_proxy.json` exits 1 with `PARITY_PROXIED_FETCH_FAILED` when that endpoint is unreachable, and `run_parity` (`tests/scripts/run_bwrap_directional_test.sh:344`) turns any nonzero exit into a FAIL. The same script pre-probes its other external anchors (`1.1.1.1:443`, `9.9.9.9:443`) via `probe_anchor` and `exit 77`s when they are down, so the treatment is internally inconsistent.

An unrelated outage, rate limit, or restricted CI network therefore surfaces as a product failure. Either pre-probe and skip consistently, or use a deterministic local upstream fixture that still proves end-to-end HTTP forwarding through `unix-test-proxy`.

Note that five already-merged configs (`bubblewrap_network_proxy_builtin`, `_allowlist`, `_blocklist`, `_host_rules`, `_egress_denied`) depend on the same endpoint, so a fix should cover them rather than only the new parity config.

### Additional context

- PR: #979
- Related: #955 (Bubblewrap networking follow-ups), #990 (LXC/Bubblewrap firewall consolidation)

SohamDas2021:
## 📖 Description

-----

## Summary

Three non-blocking follow-ups from the #979 review. No behavior change on the enforced posture; the only new coverage is a test config.

**Build the egress plan once per spawn.** `validate` computed an `EgressPlan` purely to reject bad rule addresses and then discarded it, so `spawn` rebuilt the identical plan moments later. `validate`'s body moves into `validate_prepared`, which returns the plan it already built; the trait method stays a thin delegate for callers that only want the check, and `spawn` passes the plan down to `spawn_bwrap`. `spawn_bwrap` keeps a recompute fallback so the change is provably behavior-preserving: the normalization between the two call sites touches only filesystem fields and carries network policy through unchanged.

**Share the directional-shape check.** The `network_egress.is_some() || network_ingress.is_some()` test was spelled out at
three sites, where a partial edit would route an ingress-only policy down the legacy path and silently discard its posture. It is now `bwrap_command::is_directional`, which lives there rather than in `network_rules` because that module is Linux-gated while `bwrap_command` compiles everywhere.

**Cover IPv6 and ICMP rendering with a live test.** No directional config carried a v6 peer or an ICMP rule, so `ip6tables-restore` never saw a v6 address or the `icmp` → `icmpv6` token swap the v6 table requires. `bubblewrap_network_directional_ipv6_icmp.json` exercises a `/128`, a `/48` with an `except` carve-out, an ICMP rule to a v6 peer, and a `to`-less ICMP rule that fans into both families. Connectivity is deliberately **not** asserted — the namespace has no IPv6 route (#955) — only that the rules install, which a rejected token would prevent by failing namespace setup.

The fourth item in the issue (a flaky test reaching `api.github.com`) was already fixed on main: every Bubblewrap proxy config now uses the local `mxc-test.invalid` fixture. No work was needed.

## Validation

- `cargo clippy --target x86_64-unknown-linux-gnu -p bwrap_common --all-targets -- -D warnings` — clean
- `cargo test -p bwrap_common` — 84 passed (includes a new table test pinning all four `is_directional` shape combinations)
- `node scripts/versioning/validate-configs.js` — 268 configs OK
- `bash -n tests/scripts/run_bwrap_directional_test.sh` — clean

The Linux-only `network_rules`/`bwrap_runner` unit tests and the new E2E config
require a Linux host with `bwrap`, so they are covered by CI rather than locally.

Closes #991

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1051)

### DIFF
--- a/src/backends/bubblewrap/common/src/bwrap_runner.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_runner.rs
@@ -92,6 +92,87 @@ impl SandboxBackend for BubblewrapScriptRunner {
     }
 
     fn validate(&self, request: &ExecutionRequest) -> Result<(), ScriptResponse> {
+        self.validate_prepared(request).map(|_| ())
+    }
+
+    fn spawn(
+        &mut self,
+        request: &ExecutionRequest,
+        logger: &mut Logger,
+        stdio: StdioMode,
+    ) -> Result<Box<dyn SandboxProcess>, ScriptResponse> {
+        validate_common(request)?;
+        // Keep the plan validation derived so the firewall path installs
+        // exactly what was accepted instead of deriving it again.
+        let egress_plan = self.validate_prepared(request)?;
+        // Object-based FS-policy normalization (D6): tighten aliases of the same
+        // host object to the strictest intent (deny > ro > rw). Done here, close
+        // to mount — config_parser stays string-only and the TOCTOU window
+        // between check and mount is minimized. Only clone the request when an
+        // aliasing conflict actually needs tightening (the common case is none);
+        // an unresolvable path with deniedPaths present fails closed.
+        let normalized;
+        let request = match wxc_common::filesystem_object::normalize_object_conflicts(
+            &request.policy,
+            logger,
+        ) {
+            Ok(Some(policy)) => {
+                normalized = ExecutionRequest {
+                    policy,
+                    ..request.clone()
+                };
+                &normalized
+            }
+            Ok(None) => request,
+            Err(msg) => return Err(ScriptResponse::error(&msg)),
+        };
+        // Delegation check (D3): reject any policy path the invoking user cannot
+        // access, so the sandbox never gains access the caller lacks. Runs AFTER
+        // object normalization so it is evaluated against the already-tightened
+        // intents (a path moved rw -> denied must not then require write access).
+        if let Err(msg) = wxc_common::filesystem_access::check_delegation(&request.policy) {
+            return Err(ScriptResponse::error(&msg));
+        }
+        // Resolve denied paths that traverse a symlink to their real host path
+        // and classify each as a file/dir mask (see [`resolve_denied_paths`]).
+        // Only clones the request when a path needs rewriting (common case:
+        // none). See docs/bwrap-support/bubblewrap-backend.md.
+        let plan = match resolve_denied_paths(&request.policy, logger) {
+            Ok(plan) => plan,
+            Err(msg) => return Err(ScriptResponse::error(&msg)),
+        };
+        let resolved;
+        let request = match plan.paths {
+            Some(denied_paths) => {
+                let mut policy = request.policy.clone();
+                policy.denied_paths = denied_paths;
+                resolved = ExecutionRequest {
+                    policy,
+                    ..request.clone()
+                };
+                &resolved
+            }
+            None => request,
+        };
+        // The masks are built from the list above, not from the one the caller
+        // wrote, so the pin conflict has to be judged against it too.
+        if let Err(msg) = check_pin_against_denied_hosts(request) {
+            return Err(ScriptResponse::error(&msg));
+        }
+        let child = self.spawn_bwrap(request, &plan.files, egress_plan, logger, stdio)?;
+        Ok(Box::new(BubblewrapSandboxProcess::new(child)))
+    }
+}
+
+impl BubblewrapScriptRunner {
+    /// `validate`, plus the egress plan it derived.
+    ///
+    /// Returning the plan lets `spawn` install exactly what validation
+    /// accepted instead of deriving it a second time on every spawn.
+    fn validate_prepared(
+        &self,
+        request: &ExecutionRequest,
+    ) -> Result<Option<network_rules::EgressPlan>, ScriptResponse> {
         validate_network_policy_support(request, self.network_policy_support())?;
         // User-input validation runs before the environmental `bwrap`
         // probe so config errors are reported deterministically even on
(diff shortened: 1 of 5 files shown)


src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -92,6 +92,87 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+        &mut self,
+        request: &ExecutionRequest,
+        logger: &mut Logger,
+        stdio: StdioMode,
+    ) -> Result<Box<dyn SandboxProcess>, ScriptResponse> {
+        validate_common(request)?;
+        // Keep the plan validation derived so the firewall path installs
+        // exactly what was accepted instead of deriving it again.
+        let egress_plan = self.validate_prepared(request)?;
+        // Object-based FS-policy normalization (D6): tighten aliases of the same
```
bbonaby: note: we can probably remove the D6. I still have no clue why copilot thinks they need to mention these in the comments.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -92,6 +92,87 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+                normalized = ExecutionRequest {
+                    policy,
+                    ..request.clone()
+                };
+                &normalized
+            }
+            Ok(None) => request,
+            Err(msg) => return Err(ScriptResponse::error(&msg)),
+        };
+        // Delegation check (D3): reject any policy path the invoking user cannot
```
bbonaby: note: same here. We can probably have copilot decrease the comment size for these to just the first sentences really.

tests/configs/bubblewrap_network_directional_ipv6_icmp.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "network": {
+    "egress": {
+      "default": "deny",
+      "allow": [
+        { "to": [{ "cidr": "2001:4860:4860::8888/128" }] },
+        { "to": [{ "cidr": "2001:db8:1::/48", "except": ["2001:db8:1:dead::/64"] }] },
+        { "to": [{ "cidr": "2606:4700:4700::1111/128" }], "ports": [{ "protocol": "icmp" }] },
+        { "ports": [{ "protocol": "icmp" }] }
+      ]
```
bbonaby, line 16: question: Does the test go out to these addresses? or do we use this config as like a dry run kind of thing?

SohamDas2021: No, nothing connects, this is a rendering/acceptance test, not a connectivity test. The `commandLine` only reads `/proc/self/ns/net`, checks `CapEff`, and echoes sentinels.

bbonaby (approved):

PR #1052, opened 2026-08-26
[Bubblewrap] Reject proxy with firewall enforcement, warn on unreachable IPv6 allows, and assert the v6 ingress chain

### LINKED ISSUE #955: [Bubblewrap] Network policy follow-ups: incomplete parser/validate() pairs, unusable IPv6 rules, and misleading diagnostics

SohamDas2021:
## 📖 Description

-----

## Summary

Addresses items 2, 3 and 6 of #955.

**Item 2 — runner twin for the proxy + firewall-enforcement rejection.** The parser rejects `network.proxy` combined with `network.enforcementMode` of `firewall` or `both`, but `validate()` had no twin — and could not reach one as written, since `ResolvedNetworkMode::from_request` tests `proxy_active` first and returns `ProxyOnly` before `enforcementMode` is ever consulted. Without a twin, an in-tree caller holding an `ExecutionRequest` has that explicitly-set, security-relevant field silently discarded. Every *public* entry point does reach the runner through the parser — `build_request*` maps a policy to wire JSON and parses it, and `SandboxRequest`'s inner request is `pub(crate)` — so this is layer parity and defense in depth rather than a live bypass. Adds `proxy_with_firewall_rejection` alongside the existing runner checks.

Per the issue's ask that the two paths agree "verbatim" and "cannot drift", the runner's copy lives in `bwrap_command` as `BWRAP_PROXY_WITH_FIREWALL`, carrying the parser's wording character-for-character. It is hand-duplicated exactly as the neighbouring `BWRAP_EXTERNAL_PROXY_HOST_RULES` already is, rather than hoisted into a shared constant, because that would mean editing `config_parser.rs` — an architect-owned file this PR deliberately leaves untouched. `config_parser.rs` is therefore unchanged.

Duplication normally invites drift, so it is pinned by a test: `the_parser_and_the_runner_reject_with_the_same_message` drives the real parser and asserts the `ConfigParse` payload it emits *is* the constant. Editing either literal alone fails that test, which was verified by drifting the parser's string and confirming the failure.

The gate is deliberately not schema-gated, matching the parser, and only fires when a caller explicitly asks for `firewall`/`both` — the default enforcement mode is `Capabilities`.

**Item 3 — warn on IPv6 allows that cannot carry traffic.** An IPv6 entry in `allowedHosts` validates, renders, and installs, but `slirp4netns` runs without `--enable-ipv6`, so the destination is unreachable regardless of the rule. This does not take either remedy the issue proposes: enabling IPv6 needs the RFC 4890 ICMPv6 carve-outs and v6-capable CI, and rejecting IPv6 rule addresses would break a config surface the schema accepts today. Since the posture already fails closed, the defect worth fixing is the *silence* — `spawn` now warns, naming the affected destinations. It lives in the runner so both the JSON and programmatic paths hear it from one site, and denies are excluded, since an unreachable deny is the posture the caller asked for. See the issue discussion for the proposal to close item 3 on this basis.

**Item 6 — assert the IPv6 ingress chain.** `run_bwrap_inbound_deny_test.sh` inspected only `iptables`, so nothing pinned the v6 chain's content against a live kernel. The per-family assertions are factored into `assert_ingress_chain <tool>` and run for both `iptables` and `ip6tables`, including the schema-0.7 absence check. Adds an `ip6tables` dependency probe so the test skips rather than fails where it is unavailable.

Items 4 and 5 are already fixed on `main` — `PrivateNetworkUse::remedy` gives firewall-mode callers mode-appropriate advice, and `rule_install_timeout` now adds `RULE_INSTALL_HEADROOM`, caps at `RULE_INSTALL_CEILING`, and clamps to the
request's own timeout. Item 1 is not addressed here — see the issue comment.

#1051, which extracted the `validate_prepared` helper that item 2 inserts into, has since merged; this branch is rebased onto it and is no longer stacked.

## Validation

- `cargo fmt --all -- --check`; `cargo clippy --workspace --all-targets -D warnings`   on both the host target and `x86_64-unknown-linux-gnu`
- `cargo test -p bwrap_common -p wxc_common`
- On a real kernel (WSL2 / Ubuntu 24.04): `run_bwrap_network_proxy_test.sh` (14/14), `run_bwrap_directional_test.sh` (9/9), and
  `sudo tests/scripts/run_bwrap_inbound_deny_test.sh` all pass, with the new `ip6tables MXC_INGRESS` assertion green — confirming `ip6tables -S` normalizes match output identically to `iptables -S`.

Note that item 2's runner gate is not reachable from a JSON config: the parser rejects first, and every public programmatic entry point parses too. It is defense in depth for an in-tree `ExecutionRequest` holder, so it is covered by the unit test `a_proxy_with_firewall_enforcement_is_rejected` rather than by an end-to-end config.

Closes #955

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1052)

### DIFF
--- a/docs/bwrap-support/bubblewrap-backend.md
+++ b/docs/bwrap-support/bubblewrap-backend.md
@@ -251,9 +251,10 @@ disagreed with the one behind the rules would hand the workload an address the
 chain never authorized. An IPv6 rule programs `ip6tables`, but the sandbox's
 namespace has no IPv6 connectivity today — slirp4netns is launched without
 `--enable-ipv6` — so an allowed IPv6 destination stays unreachable regardless
-of the rule (see #955). The terminal verdict of the unmatched family still
-follows `defaultPolicy`, so a v4-only allowlist under `block` does not leave
-IPv6 open.
+of the rule (see #955). MXC emits a warning naming those destinations rather
+than refusing them, since the posture fails closed. The terminal verdict of the
+unmatched family still follows `defaultPolicy`, so a v4-only allowlist under
+`block` does not leave IPv6 open.
 
 An IPv4-mapped address such as `::ffff:203.0.113.5` is programmed as IPv4:
 Linux puts a genuine IPv4 packet on the wire for one, so an `ip6tables` rule
--- a/docs/bwrap-support/bubblewrap-backend.md
+++ b/docs/bwrap-support/bubblewrap-backend.md
@@ -248,13 +248,18 @@ Rule addresses must be **IP literals or CIDR blocks**; a DNS name is rejected
 at validation time rather than resolved on the caller's behalf. The backend
 does not resolve, because the sandbox resolves names itself and a lookup that
 disagreed with the one behind the rules would hand the workload an address the
-chain never authorized. An IPv6 rule programs `ip6tables`, but the sandbox's
-namespace has no IPv6 connectivity today — slirp4netns is launched without
-`--enable-ipv6` — so an allowed IPv6 destination stays unreachable regardless
-of the rule (see #955). MXC emits a warning naming those destinations rather
-than refusing them, since the posture fails closed. The terminal verdict of the
-unmatched family still follows `defaultPolicy`, so a v4-only allowlist under
-`block` does not leave IPv6 open.
+chain never authorized.
+
+**IPv6 rules are filtered, but IPv6 traffic has nowhere to go.** The two are
+separate concerns and only the second is missing. Filtering works: an IPv6 rule
+programs `ip6tables`, and the terminal verdict of the unmatched family follows
+`defaultPolicy`, so a v4-only allowlist under `block` does not leave IPv6 open.
+What the sandbox lacks is IPv6 *connectivity* — slirp4netns is launched without
+`--enable-ipv6`, so the namespace has no IPv6 route at all (see #955). The
+consequence is one-sided: an IPv6 **block** is already satisfied, while an IPv6
+**allow** grants nothing in practice, because the destination stays unreachable
+regardless of the rule. MXC emits a warning naming those allowed destinations
+rather than refusing them, since the posture fails closed.
 
 An IPv4-mapped address such as `::ffff:203.0.113.5` is programmed as IPv4:
 Linux puts a genuine IPv4 packet on the wire for one, so an `ip6tables` rule
--- a/src/backends/bubblewrap/common/src/bwrap_command.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_command.rs
@@ -462,6 +462,21 @@ pub fn external_proxy_host_rules_rejection(request: &ExecutionRequest) -> Option
     conflicts.then_some(BWRAP_EXTERNAL_PROXY_HOST_RULES)
 }
 
+/// Validate-time twin of the parser's proxy + firewall-enforcement rejection.
+///
+/// [`ResolvedNetworkMode::from_request`] tests the proxy first and returns
+/// `ProxyOnly`, so `enforcementMode` is never consulted on that path and a
+/// programmatic caller's explicit field is discarded in silence. Not
+/// schema-gated, matching the parser.
+pub fn proxy_with_firewall_rejection(request: &ExecutionRequest) -> Option<&'static str> {
+    let conflicts = request.policy.network_proxy.is_enabled()
+        && matches!(
+            request.policy.network_enforcement_mode,
+            NetworkEnforcementMode::Firewall | NetworkEnforcementMode::Both
+        );
+    conflicts.then_some(wxc_common::error::BWRAP_PROXY_WITH_FIREWALL_MSG)
+}
+
 /// Validate-time twin of [`local_network_diagnostic`]: the same mismatch, but
 /// only for schemas that reject it rather than warn.
 ///
@@ -1172,6 +1187,42 @@ mod tests {
         r.policy.network_proxy.builtin_test_server = true;
         r.policy.allowed_hosts = vec!["10.0.0.1".into()];
         assert!(external_proxy_host_rules_rejection(&r).is_none());
+    }
+
+    /// The parser refuses this pairing; the runner did not, and could not:
+    /// `from_request` returns `ProxyOnly` before `enforcementMode` is read, so
+    /// a programmatic caller's explicit field was dropped without a word.
+    #[test]
+    fn a_proxy_with_firewall_enforcement_is_rejected() {
+        let mut r = base_request();
+        r.policy.network_proxy.address = Some(ProxyAddress::new("127.0.0.1".into(), 3128));
+
+        for mode in [
+            NetworkEnforcementMode::Firewall,
+            NetworkEnforcementMode::Both,
+        ] {
+            r.policy.network_enforcement_mode = mode.clone();
+            assert_eq!(
+                proxy_with_firewall_rejection(&r),
+                Some(wxc_common::error::BWRAP_PROXY_WITH_FIREWALL_MSG),
+                "{mode:?} must be refused with the parser's own message"
+            );
+        }
+
+        // `is_enabled()` covers both proxy flavors, so the builtin refuses too.
+        let mut builtin = base_request();
+        builtin.policy.network_proxy.builtin_test_server = true;
+        builtin.policy.network_enforcement_mode = NetworkEnforcementMode::Firewall;
+        assert!(proxy_with_firewall_rejection(&builtin).is_some());
+
+        // Negative controls: the default mode never invokes iptables, and the
+        // gate must key on the proxy rather than on the mode alone.
+        r.policy.network_enforcement_mode = NetworkEnforcementMode::Capabilities;
+        assert!(proxy_with_firewall_rejection(&r).is_none());
+
+        r.policy.network_enforcement_mode = NetworkEnforcementMode::Firewall;
+        r.policy.network_proxy.address = None;
+        assert!(proxy_with_firewall_rejection(&r).is_none());
     }
 
     /// A directional runtime proxy must not trip the legacy host-list guard.
--- a/src/backends/bubblewrap/common/src/bwrap_command.rs
+++ b/src/backends/bubblewrap/common/src/bwrap_command.rs
@@ -395,6 +395,17 @@ pub const BWRAP_EXTERNAL_PROXY_HOST_RULES: &str =
      'network.proxy.builtinTestServer: true' (testing only) for MXC-enforced host filtering, \
      or remove the host policy.";
 
+/// Rejection text for a proxy combined with firewall enforcement.
+///
+/// Duplicated (verbatim) from the parser's pre-existing check of the same
+/// combination, exactly as [`BWRAP_EXTERNAL_PROXY_HOST_RULES`] is. The parser
+/// keeps its copy for JSON configs; this one covers callers who hand a runner
+/// an `ExecutionRequest` directly. Keep the two strings identical.
+pub const BWRAP_PROXY_WITH_FIREWALL: &str = "Bubblewrap: network.proxy cannot be combined with \
+     network.enforcementMode='firewall' or 'both'. The cooperative \
+     env-var proxy enforces hosts at the proxy layer; iptables-based \
+     enforcement requires privilege and is mutually exclusive.";
+
 /// Rejection text for a proxy combined with a directional egress rule set.
 ///
 /// Distinct from [`BWRAP_EXTERNAL_PROXY_HOST_RULES`] because the two describe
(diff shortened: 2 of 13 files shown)


src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -1158,6 +1173,42 @@ mod tests {
         assert!(external_proxy_host_rules_rejection(&r).is_none());
     }
 
+    /// The parser refuses this pairing; the runner did not, and could not:
```
dhoehna: ` /// The parser refuses this pairing; the runner did not, and could not:` if the parser rejects the combo why have a bubblewrap test?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -345,6 +357,28 @@ fn check_pin_against_denied_hosts(request: &ExecutionRequest) -> Result<(), Stri
     }
 }
 
+/// Warn that an IPv6 allow rule installs but cannot carry traffic.
+///
+/// It fails closed, so this warns rather than rejects: refusing would break
+/// configs the schema accepts today (see #955). Emitted from `spawn`, which has
+/// a logger and which every caller reaches, so the JSON and programmatic paths
+/// both hear it from one site.
+fn warn_unreachable_v6_targets(plan: &network_rules::EgressPlan, logger: &mut Logger) {
```
dhoehna, line 388: Should this be a warning, or an error?

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -244,9 +244,10 @@ disagreed with the one behind the rules would hand the workload an address the
 chain never authorized. An IPv6 rule programs `ip6tables`, but the sandbox's
 namespace has no IPv6 connectivity today — slirp4netns is launched without
 `--enable-ipv6` — so an allowed IPv6 destination stays unreachable regardless
-of the rule (see #955). The terminal verdict of the unmatched family still
-follows `defaultPolicy`, so a v4-only allowlist under `block` does not leave
-IPv6 open.
+of the rule (see #955). MXC emits a warning naming those destinations rather
+than refusing them, since the posture fails closed. The terminal verdict of the
+unmatched family still follows `defaultPolicy`, so a v4-only allowlist under
+`block` does not leave IPv6 open.
```
bbonaby: question (non-blocking): can't tell but is this saying that even with slirp4netns bubblewrap can't do ipv6 filtering?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -123,9 +129,10 @@ impl SandboxBackend for BubblewrapScriptRunner {
             Ok(None) => request,
             Err(msg) => return Err(ScriptResponse::error(&msg)),
         };
-        // Reject any policy path the invoking user cannot access, so the sandbox
-        // never gains access the caller lacks. Runs after object normalization
-        // so it sees the already-tightened intents.
+        // Delegation check (D3): reject any policy path the invoking user cannot
```
bbonaby: note: this one still says "D3" in the comments.

src/backends/bubblewrap/common/src/bwrap_command.rs
SohamDas2021: `build_request` funnels through `config_parser` too (`policy.rs:751`) and `SandboxRequest.inner` is `pub(crate)`, so no caller bypasses the parser today. Keeping the gate as layer parity + defense-in-depth (same as the equally parser-shadowed `external_proxy_host_rules_rejection`), I've corrected the comment and test doc to say exactly that.

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: Warning is okay, it fails closed, and erroring would break configs the schema accepts.

docs/bwrap-support/bubblewrap-backend.md
SohamDas2021: Thanks, the wording conflated two things. Filtering *does* work, v6 rules program `ip6tables` and the unmatched family's verdict follows `defaultPolicy`. What's missing is v6 *connectivity* (`slirp4netns` runs without `--enable-ipv6`), so a v6 block is already satisfied while a v6 allow grants nothing in practice. I have split the paragraph to state both plainly.

src/backends/bubblewrap/common/src/bwrap_runner.rs
SohamDas2021: Rebased onto `main`, which drops both the `D3` and `D6` tags here.

bbonaby (approved):

Issue #1049, opened 2026-08-26
[WSLC] Promote WSLC, move it out of experimental


SohamDas2021:
### Description of the task

 Move WSLC out of experimental

SohamDas2021:
https://github.com/microsoft/mxc/pull/1187

PR #1058, opened 2026-08-27
[WSLC] Reject remaining policy fields the backend cannot honor

SohamDas2021:
## 📖 Description

-----

## Summary

WSLc accepted five policy fields, carried them into the runner, and never read them. A caller asking for a UI lockdown, a firewall enforcement mode, inbound local-network access, policy preservation, or a container that outlives the run got a container that silently did not have the posture they asked for. Each is now an explicit `policy_validation` rejection.

| Field | Before | Now |
|---|---|---|
| `ui` | accepted + ignored, every phase | rejected on every phase, both surfaces |
| `network.allowLocalNetwork` | ignored at state-aware provision | rejected at provision |
| `network.enforcementMode` | accepted + ignored | `firewall`/`both` rejected; `capabilities` accepted |
| `lifecycle.preservePolicy` | accepted + ignored on one-shot | rejected on one-shot |
| `lifecycle.destroyOnExit` | `false` accepted, behaved as `true` | `false` rejected on one-shot; `true` accepted |

**Rejection means abort, not teardown.** All three entry surfaces validate before anything is created: `ScriptRunner::run` ahead of `execute`, `SandboxBackend::spawn` ahead of `start_container`, and the state-aware dispatcher ahead of each phase body — and `connect_daemon()` lives inside `provision()`, so a refused provision never even spawns the daemon.

`ui` is rejected by **presence, not value**. The parser materializes a full-lockdown `ui` block into *every* request, so an explicitly supplied lockdown `ui` is indistinguishable by value from an absent one — a value-based check would reject every WSLc config in the repo while letting the most restrictive request a caller can write through unenforced. This uses the parse-derived `ContainerPolicy::ui_specified` flag, mirroring IsolationSession.

`enforcementMode`, `preservePolicy`, and `destroyOnExit` are rejected by **value** instead, because their defaults honestly describe WSLc's behavior — `destroyOnExit` defaults to `true`, so bare configs are untouched.

`destroyOnExit: false` corrects a claim this PR originally made. Selecting `WSLC_CONTAINER_FLAG_AUTO_REMOVE` is not the same as honoring the field: `StartedContainer` owns the `WslcSessionGuard`, whose `Drop` terminates the session — and with it the session-scoped container — regardless of the flag, and the WSLc SDK has no cross-process re-attach. `true` and `false` therefore produce an identical observable outcome, which the repo's own fixture had already admitted in a comment. Accepting `false` promised a container that was already gone, so it is now refused and that fixture converted into a rejection test.

No wire, schema, or parser-gating changes — this is a domain-model behavior change only, so it lands while WSLc is still experimental.

## Validation

`wslc_common` 211 tests (with and without `link-wslcsdk`), `wxc_common` 983, clean `fmt` and clippy on `wslc_common` / `wxc_common` / `wxc --features wslc`. All 43 `tests/configs/wslc*.json` pushed through the real parser + validator: 26 accept, 17 reject. Of the rejections, 16 are pre-existing (13 `{{SANDBOX_ID}}` placeholders, 3 named `*_rejected_*`) and 1 is the new fixture this PR adds. **No config regressed into an unintended rejection.**

The full WSLc one-shot E2E suite was run on a WSL2 host and is green, including both lifecycle fixtures.

**On E2E coverage, precisely.** One rejection is now asserted positively end-to-end: `wslc_destroy_on_exit_false_rejected.json` runs in the nightly WSLc suite and requires exit `-1` plus the rejection message, so the payload provably never runs. The other three are not — no fixture supplies a `ui` block, a `firewall` enforcement mode, or `preservePolicy`. What the config corpus covers for those is the inverse and, for this change, the higher risk: that none of the existing configs regressed into an over-rejection. The positive path is covered by unit tests instead, including the near-miss acceptances (`capabilities`, `allowLocalNetwork: false`, absent `ui`) that a gate flipping between value- and presence-based would break. Positive E2E fixtures for the remaining three are better added alongside the promotion in PR 3, where these fields move to the top-level surface and the configs have to be rewritten anyway.

## Related Issues
- #1049
- 
## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1058)


src/backends/wslc/common/src/policy.rs
```diff
@@ -119,6 +161,61 @@ fn reject_host_filtering(request: &ExecutionRequest) -> Result<(), MxcError> {
...
+/// Value-based, unlike [`reject_ui_policy`]: the default `capabilities` is an
+/// honest description of what WSLc does (the container's network is
+/// all-or-nothing, with nothing per-host to enforce), so an explicit
+/// `capabilities` is accepted. `firewall` and `both` ask for per-rule
+/// enforcement the container cannot perform — it has no `CAP_NET_ADMIN` — so
+/// accepting either would assert a guarantee that does not exist.
+///
+/// Shared by every WSLc phase and by the one-shot / streaming
+/// `validate_runner`, for the same reason as the UI check: it is a property of
+/// the backend, not of a phase.
```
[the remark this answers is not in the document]
SohamDas2021: Good catch - the doc was wrong. Fixed in cc7007a8: it now names the actual call sites (`validate_provision_policy` and the one-shot / streaming `validate_runner`) and explains that post-provision and exec deliberately do not call it, because they reject the whole network mode by presence via `reject_post_provision_network_mode` - which is broader, and routing them here would wrongly accept `enforcementMode: capabilities` after provision.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -688,6 +704,50 @@ impl ScriptRunner for WSLContainerRunner {
...
+///   paths become container volume mounts and the network posture is a
+///   container networking mode, both of which are properties of the container
+///   object itself and cannot be retained independently of it. There is
+///   nothing to preserve, so the request is refused rather than silently
+///   dropped.
+///
+/// The state-aware surface needs no counterpart: the parser rejects the whole
+/// one-shot `lifecycle` section on state-aware requests.
+fn reject_unsupported_lifecycle(request: &ExecutionRequest) -> Result<(), ScriptResponse> {
+    if request.lifecycle.preserve_policy {
```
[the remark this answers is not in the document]
SohamDas2021: Correct, and fixed in `fa97faa9`. Selecting `WSLC_CONTAINER_FLAG_AUTO_REMOVE` was not the same as honoring the field: `StartedContainer` owns the `WslcSessionGuard`, whose `Drop` terminates the session (and with it the session-scoped container) regardless of the flag, so `true` and `false` were observably identical. `false` is now rejected, `tests/configs/wslc_destroy_on_exit_false.json` became a rejection fixture asserted in the nightly suite, and the doc comment, PR description, and `copilot-instructions.md` were all corrected.

docs/wsl/wsl-container-getting-started.md
```diff
@@ -401,6 +410,30 @@ Paths in `filesystem.readwritePaths` and `filesystem.readonlyPaths` are mounted
...
+The check is on **presence, not value**. `ui`'s defaults are full lockdown, so
+an explicitly supplied lockdown `ui` is indistinguishable *by value* from an
+absent one — a value-based check would let the single most restrictive request
+you can write through unenforced. Omit the section entirely.
+
+### `lifecycle`: `destroyOnExit` is honored, `preservePolicy` is not
+
+`lifecycle.destroyOnExit` **is** honored: it selects the SDK's
+`WSLC_CONTAINER_FLAG_AUTO_REMOVE`, so both `true` (the default) and `false`
+behave as documented.
```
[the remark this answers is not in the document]
SohamDas2021: Correct, and fixed in `fa97faa9`. `destroyOnExit: false` is now **rejected** on the one-shot surface rather than documented as honored, and this section explains why one-shot structurally cannot deliver it.

.github/copilot-instructions.md
```diff
@@ -186,7 +186,7 @@ The Rust workspace (`src/`) implements multiple sandboxing backends behind the `
...
...at provision. exec honors `network.proxy` **url-form only** (injected as `HTTP_PROXY`/`HTTPS_PROXY`); start/stop/deprovision reject all policy. Rejects, on **both** surfaces and every phase, the fields it cannot honor: any supplied `ui` (presence-based via `ContainerPolicy::ui_specified`, since a WSLc container runs Linux while `ui` maps to Windows `JOB_OBJECT_UILIMIT_*` — an omitted `ui` is accepted and applies no restriction, so the schema's default-deny reading does not hold here) and `network.enforcementMode` other than `capabilities` (no `CAP_NET_ADMIN` for in-container rules). Provision/one-shot additionally reject `network.allowLocalNetwork=true` (all-or-nothing networking; the one-shot message points at `experimental.wslc.portMappings`, which the state-aware surface does not have), and one-shot rejects `lifecycle.preservePolicy=true` while genuinely honoring both `lifecycle.destroyOnExit` values (it selects `WSLC_CONTAINER_FLAG_AUTO_REMOVE`). Every rejection aborts before any container is created. ID prefix `wslc` (`wslc:<32-hex>`). Idle-timeout is env-overridable via `MXC_WSLC_DAEMON_IDLE_TIMEOUT_SECS`/`MXC_WSLC_DAEMON_IDLE_POLL_SECS`. See `docs/wsl/wslc-state-aware.md`. |
```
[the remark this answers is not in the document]
SohamDas2021: Correct, and fixed in `fa97faa9`. The WSLc row now states that one-shot rejects `lifecycle.destroyOnExit=false`.

.github/copilot-instructions.md
```diff
@@ -186,7 +186,7 @@ The Rust workspace (`src/`) implements multiple sandboxing backends behind the `
...
...`HTTPS_PROXY`); start/stop/deprovision reject all policy. Rejects, on **both** surfaces and every phase, the fields it cannot honor: any supplied `ui` (presence-based via `ContainerPolicy::ui_specified`, since a WSLc container runs Linux while `ui` maps to Windows `JOB_OBJECT_UILIMIT_*` — an omitted `ui` is accepted and applies no restriction, so the schema's default-deny reading does not hold here) and `network.enforcementMode` other than `capabilities` (no `CAP_NET_ADMIN` for in-container rules). Provision/one-shot additionally reject `network.allowLocalNetwork=true` (all-or-nothing networking; the one-shot message points at `experimental.wslc.portMappings`, which the state-aware surface does not have), and one-shot rejects `lifecycle.preservePolicy=true` and `lifecycle.destroyOnExit=false` (the container is session-scoped and the session dies with the one-shot process, so `false` cannot be honored; only the default `true` matches actual behavior). Every rejection aborts before any container is created. ID prefix `wslc` (`wslc:<32-hex>`). Idle-timeout is env-overridable via `MXC_WSLC_DAEMON_IDLE_TIMEOUT_SECS`/`MXC_WSLC_DAEMON_IDLE_POLL_SECS`. See `docs/wsl/wslc-state-aware.md`. |
```
bbonaby: note: UI is meant to be cross platform. On Windows it might map to `JOB_OBJECT_UILIMIT_` but on Linux we still want to see if there is something we can map it to.

docs/wsl/wsl-container-getting-started.md
```diff
@@ -401,6 +410,36 @@ Paths in `filesystem.readwritePaths` and `filesystem.readonlyPaths` are mounted
 into the container. Host path `C:\workspace` becomes `/mnt/c/workspace` inside
 the container.
 
+### `ui` is not supported
+
+A `ui` section is **rejected**. The section maps to Windows job-object UI limits
+(`JOB_OBJECT_UILIMIT_*`); a WSLC container runs Linux inside the WSL2 VM, so
+those limits have no analogue and nothing in the backend could apply them.
+
```
bbonaby, line 417: note: see my comment above about what UI is suppose to be for. It's ok for WSLC to not support it if it can't but I don't think we want to imply it's only for Windows when it's meant to be cross platform. We can probably remove that sentence from here actually.

docs/wsl/wsl-container-getting-started.md
```diff
@@ -401,6 +410,36 @@ Paths in `filesystem.readwritePaths` and `filesystem.readonlyPaths` are mounted
...
+`lifecycle.destroyOnExit: true` (the default) is honored: it selects the SDK's
+`WSLC_CONTAINER_FLAG_AUTO_REMOVE`, and teardown stops and deletes the container.
+
+`lifecycle.destroyOnExit: false` is **rejected**. It asks for the container to
+outlive the run, which a one-shot invocation cannot deliver: the container is
+scoped to a session this process owns, terminating that session at the end of
+the run reaps the container regardless of the AutoRemove flag, and the WSLC SDK
+has no cross-process re-attach. Use the state-aware lifecycle if you need a
+container to persist — its daemon holds the session open across phase processes.
```
bbonaby, line 433: quick question here if two apps use MXC and both you WSLC  state-aware are the two daemons operating in separate processes and if so do they know anything about the others WSL containers? Just want to confirm they won't stamp on each others toes is all.

docs/wsl/wslc-state-aware.md
```diff
@@ -85,7 +85,23 @@ rules do not apply — see the empirical finding in the plan history). Proxy is
...
+| `network.allowLocalNetwork` | rejected if `true` — WSLc networking is all-or-nothing, and the state-aware surface has no port-mapping escape hatch | rejected (any explicit network-mode field) | rejected (any explicit network-mode field) |
+| `network.enforcementMode` | rejected unless `capabilities` — `firewall` / `both` ask for per-rule enforcement the container cannot perform | rejected (any explicit network-mode field) | rejected (any explicit network-mode field) |
 | `network.proxy` | rejected | rejected | honored — **`url` form only** (`localhost` / `builtinTestServer` forms → `policy_validation`); injected as `HTTP_PROXY` / `HTTPS_PROXY` env vars |
+| `ui` | rejected | rejected | rejected |
+| `process.timeout` | n/a | n/a | honored → `ExecConfig.timeout_ms` |
+| `lifecycle` | rejected (whole section, at parse) | rejected | rejected |
+
+`ui` is rejected by **presence, not value**, on every phase. A WSLc container runs Linux, so the
+section's Windows job-object UI limits (`JOB_OBJECT_UILIMIT_*`) have no analogue inside it and no
+phase could honor it. Presence is the only workable test because `UiPolicy`'s defaults are full
```
bbonaby: note: see comment on UI above.

docs/schema.md
```diff
@@ -304,11 +304,15 @@ backend (via job-object UI restrictions plus the Win32k mitigation — see
...
-any supplied `ui` at every phase on both surfaces** — no `ui` posture is truthful
+whether it applies, rejects, or ignores the section. **IsolationSession and WSLc
+refuse any supplied `ui` at every phase on both surfaces**, and each accepts an
+omitted one without applying any UI restriction — so the section's default-deny
+reading does not hold on either. The reasons differ: no `ui` posture is truthful
 for a session-isolated sandbox (see
-[`isolation-session/state-aware-rust.md`](isolation-session/state-aware-rust.md)) —
-and accepts an omitted one without applying any UI restriction. The Windows
+[`isolation-session/state-aware-rust.md`](isolation-session/state-aware-rust.md)),
+while a WSLc container runs Linux and has no analogue of the Windows job-object
```
bbonaby: thought: "WSLc container", we probably want to start saying "WSL container" right?

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -688,6 +704,66 @@ impl ScriptRunner for WSLContainerRunner {
...
+/// * `preservePolicy: true` asks for filesystem and network policy to outlive
+///   the run. WSLc installs no persistent host-side enforcement: `rw`/`ro`
+///   paths become container volume mounts and the network posture is a
+///   container networking mode, both of which are properties of the container
+///   object itself and cannot be retained independently of it. There is
+///   nothing to preserve, so the request is refused rather than silently
+///   dropped.
+///
+/// The state-aware surface needs no counterpart: the parser rejects the whole
+/// one-shot `lifecycle` section on state-aware requests.
```
bbonaby: thought: Do we need these comments? In this case the code kind of speaks for itself really.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -2431,6 +2554,177 @@ mod tests {
...
+    // from `validate_runner`, which both `ScriptRunner::run` and
+    // `SandboxBackend::spawn` call *before* any container exists.
+
+    /// A WSLc container runs Linux; `ui` maps to Windows job-object UI limits
+    /// (`JOB_OBJECT_UILIMIT_*`) with no analogue inside it.
+    ///
+    /// Presence-based: `UiPolicy::default()` is full lockdown, so an
+    /// explicitly-supplied lockdown `ui` is indistinguishable by value from an
+    /// absent one. A value-based check would let the single most restrictive
+    /// request a caller can write through unenforced.
```
bbonaby: note: same thing for these comments on tests, the test seems like it explains themselves. The docs you have at the top already explain the reasoning.

src/core/wxc_common/src/models.rs
```diff
@@ -751,12 +751,13 @@ pub struct ContainerPolicy {
...
-    /// accept and ignore it, so this flag being set does not mean a UI policy
-    /// was honored anywhere; it means only that the caller supplied one.
+    /// Consumed by IsolationSession and WSLc today, neither of which has a
+    /// UI-restriction primitive: both refuse a supplied UI policy rather than
+    /// accepting and dropping it. The other backends that do not enforce
+    /// `policy.ui` — LXC and Bubblewrap on Linux, Seatbelt on macOS, Windows
+    /// Sandbox — still accept and ignore it, so this flag being set does not
+    /// mean a UI policy was honored anywhere; it means only that the caller
+    /// supplied one.
     #[serde(skip)]
```
bbonaby: note: probably don't need the full explanation, but as an aside, I'm starting to think we should have made these `*_supplied` booleans bit flags. Not for this PR but just for later.

tests/configs/wslc_destroy_on_exit_false_rejected.json
```diff
@@ -0,0 +1,20 @@
+{
+  "_comment": "Rejection fixture: the one-shot WSLc surface refuses lifecycle.destroyOnExit=false. The container is scoped to a session this process owns, and terminating that session reaps the container regardless of the AutoRemove flag, so `false` cannot be honored. Expects a rejection before any container is created -- the payload must never run.",
+  "version": "0.6.0-alpha",
```
bbonaby, line 3: question (non-blocking): I guess the version doesn't really matter here any more, as the behavior will be for all of them + stable after promotion.

src/backends/wslc/common/src/policy.rs
```diff
@@ -119,6 +161,65 @@ fn reject_host_filtering(request: &ExecutionRequest) -> Result<(), MxcError> {
...
+/// Reject any supplied UI policy. Presence-based, not value-based: the domain
+/// `UiPolicy::default()` is full lockdown, so an explicitly-supplied lockdown
+/// `ui` is indistinguishable from an absent one by value — the same blind spot
+/// `network_specified` closes for the network policy.
+///
+/// Shared by every WSLc phase and by the one-shot / streaming
+/// `validate_runner`: the reason is the container's OS, not the lifecycle
+/// phase, so there is no phase on either surface where a `ui` section could be
+/// honoured. Runs after the filesystem check so a filesystem rejection keeps
+/// precedence, and before the network checks.
```
bbonaby: issue: I think the comments in this file can probably be reduced, if the code is clear enough (in this case simply rejecting a specified policy + tests) we don't need to say what its doing. I think we can have copilot do an inventory to slim all these in the PR down or remove them to be honest.

docs/wsl/wsl-container-getting-started.md
SohamDas2021: **If both apps run for the same user**, they would share one daemon. Not one per app. Discovery is a single `daemon.json` under `%TEMP%\mxc-wslc\state-aware`, and `%TEMP%` is per-user, so the daemon is a **per-user singleton**. Both apps connect to the same process and the same shared `WslcSession`- one WSL2 utility VM, booted lazily on first provision and amortized across every sandbox.

**If they run for different users**, then they are fully isolated. The pipe DACL is current user + SYSTEM only matching the owner-only DACL on the record.

**Can app A see app B's containers? No.** `DaemonRequest` has no enumeration — only Provision / Start / Exec / Stop / Deprovision / Ping. Sandbox IDs are random UUIDs and aren't listable.

docs/wsl/wsl-container-getting-started.md
```diff
@@ -372,6 +372,15 @@ filtering at the proxy layer, or remove the host lists. The bare
 `defaultPolicy` forms with **no** host lists remain supported: `"block"` is a
 full network cutoff and `"allow"` is full outbound (NAT).
 
+### `enforcementMode` must be `capabilities`
+
+For the same reason, `network.enforcementMode: "firewall"` (or `"both"`) is
```
dhoehna: `For some reason`?  What is the reason?

src/backends/wslc/common/src/wsl_container_runner.rs
dhoehna: Gotta love Copilot.  I believe the reason for the long comments is two fold
1. It does not know what people assume and documents everything.
2. Documents any and all decisions made.

dhoehna (commented):
Look through some of the bigger comment sections and reduce them down.

tests/configs/wslc_destroy_on_exit_false_rejected.json
SohamDas2021: yes, that is correct, it just mirrors wslc_destroy_on_exit_true.json

src/backends/wslc/common/src/policy.rs
```diff
@@ -119,12 +148,49 @@ fn reject_host_filtering(request: &ExecutionRequest) -> Result<(), MxcError> {
...
-/// Reject any network *mode* field supplied after provision. The network
-/// posture (`defaultPolicy` / `enforcementMode` / `allowLocalNetwork` / host
-/// lists) is bound to the provision phase; presence — not value — is checked so
-/// an explicit `defaultPolicy: "block"` (indistinguishable from an omitted
-/// block by value) is rejected too. The cooperative proxy is a separate
-/// exec-time concern handled by the callers.
+/// Reject any supplied UI policy: WSLc cannot enforce UI restrictions.
+///
+/// Presence-based, not value-based — `UiPolicy::default()` is full lockdown, so
+/// a supplied lockdown `ui` is indistinguishable from an absent one by value.
```
bbonaby: note (non-blocking): we can probably do without the extra comments for these functions.

src/backends/wslc/common/src/sandbox.rs
```diff
@@ -393,6 +393,48 @@ impl Drop for WslcSandboxProcess {
...
+
+    /// Rejections must abort, not tear down a container after building one.
+    ///
+    /// The message alone cannot prove that: on a host that *has* `wslcsdk.dll`,
+    /// a guard moved after `start_container` would leak a container and still
+    /// return this message. The banner is what makes the check host-independent
+    /// — `start_container` writes it first, so an untouched buffer proves it was
+    /// never entered.
+    #[test]
+    fn spawn_rejects_policy_before_touching_the_sdk() {
```
bbonaby: note (non-blocking): do you mean the wslc sdk? If not then I think by the time the code flow gets to a runner we're already inside the SDK, so this test probably needs a name update.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -655,12 +657,18 @@ impl ScriptRunner for WSLContainerRunner {
     /// Mirrors the config parser so requests reaching the engine directly
     /// (an already-built `ExecutionRequest`, bypassing the parser) fail here
     /// instead of late in `execute` on the broken in-container iptables path.
+    ///
+    /// The single hook for both one-shot surfaces: `ScriptRunner::run` calls it
+    /// before `execute`, and `SandboxBackend::spawn` (via `validate`) before
+    /// `start_container`. Checks run lifecycle → ui → network.
```
bbonaby: note: probably don't need this comment here, I believe the code flow that calls the runners auto invokes this validate_runners function.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -688,6 +697,54 @@ impl ScriptRunner for WSLContainerRunner {
     }
 }
 
+/// Retag a shared [`policy`] rejection so it reaches SDK callers as
+/// `policy_validation` with the same message the state-aware surface emits.
+fn as_rejection(err: MxcError) -> ScriptResponse {
```
bbonaby: thought (non-blocking): `as_wslc_rejection`?

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -688,6 +697,54 @@ impl ScriptRunner for WSLContainerRunner {
...
+/// `destroyOnExit: true` genuinely matches the behaviour — it selects
+/// `WSLC_CONTAINER_FLAG_AUTO_REMOVE`. `false` is refused because
+/// [`StartedContainer`] owns the [`WslcSessionGuard`], whose `Drop` terminates
+/// the session — and with it the session-scoped container — regardless of the
+/// flag, and the SDK has no cross-process re-attach. `preservePolicy: true` is
+/// refused because mounts and the networking mode belong to the container
+/// object, so no host-side enforcement could outlive it.
+///
+/// The state-aware surface needs no counterpart: the parser rejects the whole
+/// one-shot `lifecycle` section on state-aware requests.
```
bbonaby: note: still think these comments can be removed tbh.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -2431,6 +2533,163 @@ mod tests {
         }
     }
 
+    // -- Accept-but-ignore closures --------------------------------------
+    //
+    // Fields that used to be parsed, carried into the runner, then never read.
+    // Now refused from `validate_runner`, before any container exists.
+
+    /// Presence-based: `UiPolicy::default()` is full lockdown, so a value-based
+    /// check would let the most restrictive request through unenforced.
```
bbonaby: note: these can probably be removed now, since they talk about things that use to be a thing, but now you have tests that make sure they aren't anymore.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -2431,6 +2533,163 @@ mod tests {
...
+            );
+            assert_eq!(
+                err.failure_phase,
+                wxc_common::models::FailurePhase::Rejected
+            );
+        }
+    }
+
+    /// Value-based, unlike `ui`: `firewall` / `both` need a `CAP_NET_ADMIN` the
+    /// container lacks, while an explicit `capabilities` is accepted.
```
bbonaby: note: same comment as above about whether we need the comment.

bbonaby (commented):
mostly still about comment clean up, other than that we should be ok

bbonaby (dismissed):

bbonaby (approved):

PR #1064, opened 2026-08-28
[WSLC] Promote from experimental to the stable config surface

### LINKED ISSUE #1049: [WSLC] Promote WSLC, move it out of experimental

SohamDas2021:
## 📖 Description

Closes #1049

**111 files changed — 58 test, 53 non-test** (42 source/config, 11 docs).

WSLc is now configured through the top-level `wslc` section instead of `experimental.wslc`, and selecting `"containment": "wslc"` no longer requires `--experimental` / `{ experimental: true }` / `SandboxRequest::set_experimental(true)`. This mirrors the Seatbelt promotion precedent.

Wire + schema
- Move `wslc` from `wire::Experimental` to the `MxcConfig` root, and from `experimental.rs` to `stable.rs` in the 0.9.0-alpha closed contract.
- Keep `experimental.wslc` as a rejection alias so a pre-promotion config fails with an actionable migration message rather than being silently ignored.
- Regenerate the two dev schemas and the two TypeScript wire oracles.

Parser + dispatch
- Generalize the state-aware dispatcher with `SectionRoot` (`Experimental` | `Stable`); `StatefulSandboxBackend::SECTION_ROOT` defaults to `Experimental` and only `WslcStateAwareRunner` overrides it. `ParsedStateAwareRequest` gains `stable_raw` for stable-rooted backends.
- Drop the experimental gate for WSLc on both the one-shot and state-aware surfaces.

TypeScript SDK
- Move `wslc` out of `ContainerConfig.experimental` and off `ExperimentalBackends`; add `BACKEND_SECTION_ROOT`, the TS twin of the Rust `SectionRoot`, so envelopes are built at the right root.
- Add a `wslcAvailable` probe fact end to end (Rust probe -> engine -> `wxc-exec --probe` -> SDK). Without it, removing the experimental gate unmasked the fact that `wslc` was never in `availableMethods`, which would have made the backend unreachable through the SDK.

Rust SDK, corpus, harnesses and docs
- Drop the `set_experimental(true)` requirement from `mxc-sdk`.
- Hoist `experimental.wslc` to top-level `wslc` in 31 test configs.
- Remove `--experimental` from the WSLc test harnesses.
- Update the WSLc docs, schema reference, SDK READMEs, setup script and `.github/copilot-instructions.md`.

## 🔍 Validation
Unit / static:
- `cargo clippy --workspace --all-targets --features wxc/wslc -- -D warnings`
- `cargo test -p wxc_common -p wslc_common -p mxc_engine -p wxc -p mxc-sdk --features wxc/wslc,mxc-sdk/wslc` — 1,346 tests
- `cargo test -p mxc_config_contract --features schema-gen` — 707 tests
- `cd sdk/node && npm test` — 283 tests, 0 fail
- All gates: `check-schema-versions`, `check-schema-codegen`, `check-sdk-types-codegen`, `check-contract-codegen`, `validate-configs` (267 configs), `check-version-sync`, `check-rust-toolchain-sync`, `check-psec-codegen`, `check-dotnet-errorcode-parity`

Live WSLc E2E on a WSL2 host (`run_wslc_all_tests.ps1`, which chains the per-scenario scripts and delegates to the state-aware suite): **78/79**. The 3 reported assertion failures are all `network.proxy` post-provision error-code mismatches that are **pre-existing on the base commit** — verified by rebuilding at the merge-base and reproducing byte-identical output — and are fixed independently by #1042.

Also verified end to end against a real `wxc-exec.exe` build: a promoted top-level `wslc` config parses and runs a container to completion, the pre-promotion `experimental.wslc` shape produces the migration error, and `--probe` reports `wslcAvailable: true`.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1064)


docs/wsl/wsl-container-support-plan.md
```diff
@@ -550,11 +539,11 @@ msiexec /i wsl.2.9.9.0.x64.msi
 # Verify WSLC is available
 wslc container run hello-world
 
-# Run a Linux command via MXC (requires --experimental)
-wxc-exec.exe --experimental --debug wslc-config.json
+# Run a Linux command via MXC (needs a build with the `wslc` Cargo feature)
```
dhoehna, line 542: Cargo has a `wslc` feature?

dhoehna (dismissed):

schemas/dev/mxc-config.schema.0.9.0-alpha.json
```diff
@@ -1568,6 +1522,52 @@
...
+          "description": "Container image reference.",
+          "type": "string"
+        },
+        "imageTarPath": {
+          "description": "Path to a local image tarball to import.",
+          "type": "string"
+        },
+        "memoryMb": {
+          "description": "Requested memory limit in megabytes.",
+          "maximum": 18446744073709551615,
```
bbonaby, line 1549: note: noticed this just now, do we cap this anywhere? I assume there is a limit WSLc  has under the hood

schemas/dev/mxc-config.schema.0.9.0-dev.json
```diff
@@ -254,7 +254,7 @@
               "type": "null"
             }
           ],
-          "description": "WSL container backend config."
+          "description": "WSL container backend config (pre-promotion alias). Promoted to the top-level `wslc` section; still parsed here so the parser can reject it with a migration message instead of silently ignoring it."
```
bbonaby: question: I'm now realizing that we have an alpha and a dev and now I'm wondering if the promotion process is now that you do the same thing you did in the alpha, in the dev?

scripts/check-dotnet-api-parity.js
```diff
@@ -443,49 +443,40 @@ if (!tierMatch) {
...
+  "dotnet",
+  "Microsoft.Mxc.Sdk",
+  "StateAwareTypes.cs"
+);
+compare(
+  "state-aware containment enum",
+  enumVariants(managedStateAware, "StateAwareContainment", "csharp"),
+  rustBackends
+);
+
```
bbonaby, line 473: note: currently confused  with the content of this file tbh. That said, did it need to change for the promotion of wslc? It looks almost the same as what it was prior, if I take into account the deletion + additions

sdk/dotnet/Microsoft.Mxc.Sdk.Tests/MxcLifecycleTests.cs
```diff
@@ -424,8 +424,13 @@ public void BuildProvisionEnvelope_WslcUsesV08AndNestsImageOptions()
         Assert.Equal(
             "allow",
             root.GetProperty("network").GetProperty("defaultPolicy").GetString());
-        var provision = root.GetProperty("experimental")
-            .GetProperty("wslc")
+        // WSLc is promoted to the stable surface, so its phase section is a
+        // closed top-level `wslc.<phase>` object rather than nested under
+        // `experimental`. The parser rejects the old shape outright.
```
bbonaby: issue: comment not needed here, also do we need the check for experimental if we're removing it in this PR?

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -493,22 +493,50 @@ private static void SetOptionalBackendConfig(
...
+    /// </summary>
+    private static bool IsExperimentalSection(string backend) => backend switch
+    {
+        IsolationSessionContainment => true,
+        WindowsSandboxContainment => true,
+        WslcContainment => false,
+        _ => throw new MxcException(
+            ErrorCode.UnsupportedContainment,
+            $"no wire section root declared for state-aware backend '{backend}'"),
+    };
```
bbonaby, line 512: note: still wondering if this is needed since wslcContainment is already out of the experimental section with this PR. The other things might be nice to have but just trying to keep it scoped.

sdk/dotnet/README.md
```diff
@@ -274,8 +273,8 @@ var request = new SandboxRequest(
 ```
 
 The image must already be cached unless `ImageTarPath` is supplied. The image
-store wins over the tar when both identify an already-cached image. WSLC is
-experimental, so `Experimental` is required; the native unit must also be built
+store wins over the tar when both identify an already-cached image. WSLC is a
+stable backend and needs no `Experimental` opt-in; the native unit must be built
 with WSLC support or execution returns `BackendUnavailable`.
```
bbonaby: note: I believe we can probably remove this section saying it's now stable since it doesn't really provide the user with new info. How I see it, if it doesn't say it's experimental it's stable as there are no other states..

sdk/node/src/generated/wire.ts
```diff
@@ -96,7 +96,7 @@ export interface Experimental {
    */
   windows_sandbox?: WindowsSandbox | null;
   /**
-   * WSL container backend config.
+   * WSL container backend config (pre-promotion alias). Promoted to the top-level `wslc` section; still parsed here so the parser can reject it with a migration message instead of silently ignoring it.
    */
```
bbonaby: note: see comment I had about this in the schema. Just calling it out, it's probably coming from the rust wire type I guess.

sdk/node/src/state-aware-helper.ts
```diff
@@ -61,6 +61,18 @@ export const PREFIX_TO_BACKEND: Record<string, StateAwareContainmentBackend> = O
...
+// `StatefulSandboxBackend::SECTION_ROOT` const: a backend that is still
+// experimental nests its phases under `experimental.<backend>.<phase>`, while
+// a promoted backend carries a closed top-level `<backend>.<phase>` section.
+// Typed exhaustively so adding a backend without declaring its root is a
+// compile error.
+export const BACKEND_SECTION_ROOT: Record<StateAwareContainmentBackend, 'experimental' | 'stable'> = {
+  isolation_session: 'experimental',
+  windows_sandbox: 'experimental',
+  wslc: 'stable',
+};
```
bbonaby, line 73: note: not for this PR but I assume we'd probably be changing how this works soon. That said, until then @adpa-ms, raising to your attention as an FYI in case you see something in state aware that might break sessions, since sessions will also be stable one day.

sdk/node/tests/unit/sandbox.test.ts
```diff
@@ -1126,21 +1126,15 @@ describe('createConfigFromPolicy', () => {
...
     });
 
-    it('should throw from spawnSandbox when experimental backend is used via config', () => {
+    it('should no longer require experimental mode for a wslc config', () => {
       const config = createConfigFromPolicy({ version: '0.6.0-alpha' }, 'wslc');
       config.process!.commandLine = 'echo hello';
+      // WSLc is promoted, so the experimental gate no longer fires. A host
+      // without the WSLc runtime still rejects the request, but on
+      // availability — never on `--experimental`.
       assert.throws(
```
bbonaby: thought: I'm thinking we don't have a test for this, since once it's stable it'll never go back to experimental so I feel like this test probably doesn't add much value here.

sdk/node/tests/unit/sandbox.test.ts
```diff
@@ -1281,13 +1275,23 @@ describe('resolveExecutableAndArgs (containment validation)', { skip: platformSk
...
-  it('should still require experimental mode for experimental backends like wslc', () => {
+  it('should still require experimental mode for experimental backends like windows_sandbox', () => {
     assert.throws(
-      () => resolveExecutableAndArgs(makeConfig('wslc'), { executablePath: fakeExe }),
+      () => resolveExecutableAndArgs(makeConfig('windows_sandbox'), { executablePath: fakeExe }),
       { message: /experimental mode/ },
     );
   });
 
+  it('should NOT require experimental mode for explicit wslc containment', () => {
```
bbonaby, line 1794: question: This just reminded me, wslc needs a dll to operate right?  for the inproc rust and C# SDKs, how do we plan on getting this dll in them? At least for C# in a build pipeline we can always add the dll into our nuget, but for rust you have a build.rs file but I'm not sure where the dll would get placed and how the sdk would find it.

sdk/node/README.md
```diff
@@ -138,11 +138,11 @@ for all three connectivity modes and backend-specific support.
 
 | Platform | Default backend | Other backends | Minimum build |
 | --- | --- | --- | --- |
-| Windows 11 24H2+ (verified on 25H2) | `processcontainer` | `windows_sandbox`, `wslc`, `microvm`, `isolation_session` | `processcontainer`: 26100 (24H2)<br>`isolation_session`: 26340.9212 ([Insider Preview](https://learn.microsoft.com/en-us/windows-insider/release-notes/experimental/preview-build-26340-9212)) |
+| Windows 11 24H2+ (verified on 25H2) | `processcontainer`, `wslc` | `windows_sandbox`, `microvm`, `isolation_session` | `processcontainer`: 26100 (24H2)<br>`isolation_session`: 26340.9212 ([Insider Preview](https://learn.microsoft.com/en-us/windows-insider/release-notes/experimental/preview-build-26340-9212)) |
```
bbonaby: issue: agree with the bot here, wslc should remain in the "other backends" column

sdk/node/README.md
```diff
@@ -299,7 +299,7 @@ capability names are reserved and must not be added directly to
 
 For long-lived sandboxes where you provision once, exec many times, and tear down at the end (e.g. agentic loops), use the state-aware lifecycle.
 
-> **Backend support:** the state-aware lifecycle is currently implemented for `isolation_session`, `windows_sandbox`, and `wslc` (all Windows-only; all still experimental, so every call must pass `{ experimental: true }`). The one-shot spawn APIs (`spawnSandbox` / `spawnSandboxFromConfig`) are the supported path for every other backend.
+> **Backend support:** the state-aware lifecycle is currently implemented for `isolation_session`, `windows_sandbox`, and `wslc` (all Windows-only; `isolation_session` and `windows_sandbox` are still experimental, so calls for those must pass `{ experimental: true }` — `wslc` is promoted and needs no opt-in). The one-shot spawn APIs (`spawnSandbox` / `spawnSandboxFromConfig`) are the supported path for every other backend.
```
bbonaby: issue: let's remove "`wslc` is promoted and needs no opt-in"

src/backends/appcontainer/common/src/probe.rs
```diff
@@ -70,6 +70,11 @@ pub struct ProbeFacts {
     /// the isolation-session backend; `wxc-exec --probe` overrides it when
     /// that backend is compiled in.
     pub isolation_session_available: bool,
+    /// Whether the host can actually run WSLc (WSL2 present and the WSLc
+    /// runtime loadable). Always `false` here — `appcontainer_common` has no
+    /// dependency on the WSLc backend; `wxc-exec --probe` overrides it when
+    /// that backend is compiled in.
+    pub wslc_available: bool,
```
bbonaby, line 77: note: we definitely need a refactoring (not for this PR), we have a probe for all the backends under the appcontainer common source :(.

src/backends/wslc/common/src/state_aware.rs
```diff
@@ -52,6 +53,9 @@ impl WslcStateAwareRunner {
 impl StatefulSandboxBackend for WslcStateAwareRunner {
     const ID_PREFIX: &'static str = "wslc";
     const BACKEND_KEY: &'static str = "wslc";
+    // WSLc is promoted to the stable surface: its per-phase config lives at the
+    // top-level `wslc` section, not under `experimental`.
```
bbonaby: issue: we can remove the comments here about promotion to stable.

src/core/mxc-sdk/src/lib.rs
```diff
@@ -50,9 +50,9 @@
 //! | Explicit ProcessContainer configuration | Windows | [`Containment::ProcessContainer`] |
 //! | WSLC (WSL Container) | Windows | [`Containment::Wslc`] |
 //!
-//! WSLC is **experimental**: build with the crate's `wslc` feature, and call
-//! [`SandboxRequest::set_experimental(true)`](SandboxRequest::set_experimental)
-//! on the request. Its container has no stdin (the WSLC SDK exposes no
+//! WSLC is opt-in at **build** time only: compile this crate with its `wslc`
+//! feature. It carries no `--experimental` gate — the section is part of the
+//! stable config surface. Its container has no stdin (the WSLC SDK exposes no
```
bbonaby: note: same as above we can delete comments through out that talk about wslc's promtion to stable.

src/core/mxc-sdk/README.md
```diff
@@ -397,16 +397,16 @@ Backends with no variant at all — Windows Sandbox, MicroVM, Hyperlight, LXC 
...
-memory, GPU, storage path, port forwards — are carried by the [`WslcSection`]
-inside [`Containment::Wslc`], mirroring the SDK's `experimental.wslc` block, and
-go through the same parser the executor uses — so a rejected value (e.g. a port
-mapping with a zero or duplicated host port) fails at build time, not at spawn.
+opt-in on one axis only: build this crate with its **`wslc` feature**. The
+backend is promoted, so it needs no `set_experimental(true)`. Its settings —
+image, vCPUs, memory, GPU, storage path, port forwards — are carried by the
+[`WslcSection`] inside [`Containment::Wslc`], mirroring the SDK's top-level
+`wslc` block, and go through the same parser the executor uses — so a rejected
+value (e.g. a port mapping with a zero or duplicated host port) fails at build
```
bbonaby: note: same as above

src/core/mxc_engine/src/policy.rs
```diff
@@ -1574,23 +1567,17 @@ mod tests {
...
     }
 
     #[test]
-    fn wslc_is_not_experimental_enabled_by_default() {
-        // Selecting the backend must not silently flip the experimental gate:
-        // the caller opts in explicitly, exactly like the SDK's
-        // `SandboxSpawnOptions.experimental`.
+    fn wslc_does_not_require_the_experimental_gate() {
+        // WSLc is promoted to the stable surface: selecting it must neither
+        // require nor silently flip the experimental gate.
```
bbonaby: note: test can probably be removed now as this is the PR that makes it stable.

src/core/mxc_engine/src/run.rs
```diff
@@ -175,18 +175,8 @@ fn resolve_runner_inner_windows(
...
-                }
-                let _ = writeln!(logger, "Using WSLContainer runner (--experimental)");
-                let wslc_config = request
-                    .experimental
-                    .wslc
-                    .as_ref()
-                    .cloned()
-                    .unwrap_or_default();
+                let _ = writeln!(logger, "Using WSLContainer runner");
+                let wslc_config = request.wslc.as_ref().cloned().unwrap_or_default();
```
bbonaby, line 179: issue: thinking about about the `unwrap_or_default();` line, shouldn't we actually return an error if we can't unwrap the users config rather than using a default?

src/core/wxc/src/main.rs
```diff
@@ -782,6 +782,16 @@ fn main() {
             output.probes.isolation_session_available = mxc_engine::isolation_session_available();
             output
         };
+        // Same story for WSLc: `appcontainer_common` cannot see that backend
+        // either. WSLc is promoted (non-experimental), so the SDK's
+        // availability check is the only gate on a `wslc` request — an
+        // un-overridden `false` here would make the backend unreachable.
```
bbonaby: issue: what's the same story? You mean the comment above with isolation session?  Also we can probably remove the promotion text.

Another thing for us to think about in the future is that these files are getting away from us. It's all getting pretty tangled. Nothing your PR can fix so it's not your PRs fault I'm gonna start thinking about what we can do in a refactoring effort.

src/core/wxc_common/src/state_aware_backend.rs
```diff
@@ -242,11 +243,17 @@ pub trait StatefulSandboxBackend {
...
+    /// Used by the dispatcher to navigate the per-phase config section in the
+    /// request envelope (see [`SECTION_ROOT`](Self::SECTION_ROOT)) and to
     /// resolve provision-phase requests to the right backend implementation.
     const BACKEND_KEY: &'static str;
 
+    /// Which envelope section holds this backend's per-phase config. Defaults
+    /// to the permissive `experimental.<BACKEND_KEY>` block; a backend promoted
+    /// to the stable surface overrides this to [`SectionRoot::Stable`] so its
+    /// config is read from the top-level `<BACKEND_KEY>` section instead.
+    const SECTION_ROOT: SectionRoot = SectionRoot::Experimental;
```
bbonaby, line 238: note: calling @adpa-ms to confirm sessions is still ok here, as he's also adding state-aware stuff at the same time and wanting to make sessions stable soon.

src/core/wxc_common/src/wire.rs
```diff
@@ -582,7 +586,9 @@ pub struct Experimental {
     pub test: Option<TestFeature>,
     /// Windows Sandbox backend config.
     pub windows_sandbox: Option<WindowsSandbox>,
-    /// WSL container backend config.
+    /// WSL container backend config (pre-promotion alias). Promoted to the
+    /// top-level `wslc` section; still parsed here so the parser can reject it
+    /// with a migration message instead of silently ignoring it.
     pub wslc: Option<Wslc>,
```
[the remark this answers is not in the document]
bbonaby: issue: @sohamdas2021, is this still true? also lets remove the promotion comments from here as well.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1490,6 +1497,49 @@ fn convert_wire_state_aware(
...
+        }
+        if !one_shot_only.is_empty() {
+            return Err(WxcError::ConfigParse(format!(
+                "State-aware lifecycle requests do not accept one-shot 'wslc' field(s): {}. \
+                 Use 'wslc.provision' for state-aware container configuration.",
+                one_shot_only.join(", ")
+            )));
+        }
+    }
+
```
bbonaby: note (non-blocking): I feel like this could all have been a helper/wslc specific function rathar than inside `convert_wire_state_aware` directly.

bbonaby (changes requested):

src/core/wxc_common/src/wire.rs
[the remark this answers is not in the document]
SohamDas2021: Done in `17d27c2f`. Fixing comments.

sdk/node/src/sandbox.ts
```diff
@@ -117,10 +117,8 @@ function buildWslcContainerConfig(
     config.containment = 'wslc';
     config.containerId = containerId;
 
-    config.experimental = {
-        wslc: {
-            image: 'alpine:latest',
-        },
+    config.wslc = {
+        image: 'alpine:latest',
```
adpa-ms, line 121: [AgentAssisted] This builder's output is always rejected by native WSLc: `createConfigFromPolicy` synthesizes `config.ui` unconditionally, `config_parser` sets `policy.ui_specified = cfg.ui.is_some()`, and `wslc_common::policy::reject_ui_policy` refuses any supplied `ui` before a container starts. Every `spawnSandbox` / `spawnSandboxFromConfig` WSLc call therefore fails, including one with no `ui` in the policy.

`mxc_engine::policy` already handles this — it emits `ui` only when the caller supplied one.

Missed by validation because `sdk/node/tests/integration/wslc-e2e.test.ts` is gated on `MXC_ENABLE_WSLC_TESTS=1` and no WSLc fixture sets `ui`.

Predates this PR — the rejection arrives with the base branch — so the fix may belong there. Raising it here because this is the PR that declares WSLc stable.

Suggested: drop `config.ui` here when `policy.ui` is absent, and keep an explicitly supplied one so native validation can still reject it. Test both cases.

src/core/wxc_common/src/wire.rs
```diff
@@ -648,16 +654,16 @@ pub struct Wslc {
...
     /// for UDP port mappings.
     pub port_mappings: Option<Vec<PortMapping>>,
-    /// State-aware provision-phase configuration
-    /// (`experimental.wslc.provision`). Carries the container-creation knobs
-    /// for the state-aware lifecycle; the flat sibling fields above remain the
-    /// one-shot surface. Absent on one-shot configs and non-provision phases.
+    /// State-aware provision-phase configuration (`wslc.provision`). Carries
+    /// the container-creation knobs for the state-aware lifecycle; the flat
+    /// sibling fields above remain the one-shot surface. Absent on one-shot
+    /// configs and non-provision phases.
```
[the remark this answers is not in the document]
adpa-ms: [AgentAssisted] Following up in case this was resolved in a batch — the behaviour looks unchanged at `17d27c2f`.

`wire::Wslc` carries `provision`, so `deny_unknown_fields` doesn't reject it on a one-shot config, and `convert_wire_config`'s WSLc block reads `targetOs`, `image`, `imageTarPath`, `cpuCount`, `memoryMb`, `gpu`, `storagePath`, `portMappings` — not `provision`. With `WslcConfig::default().image` = `alpine:latest`, a one-shot `{"wslc":{"provision":{"image":"my-image:latest"}}}` parses and runs Alpine.

The exact 0.9 contract already gets this right — one-shot `Wslc` in `stable.rs` has no `provision` field — so it's only reachable while the rolling parser is authoritative.

Happy to defer if there's a reason to leave it; otherwise rejecting `provision` in the one-shot converter closes it.

src/core/wxc_common/src/models.rs
```diff
@@ -75,7 +75,7 @@ impl ContainmentBackend {
             ContainmentBackend::ProcessContainer => Some("processContainer"),
             ContainmentBackend::Lxc => Some("lxc"),
             ContainmentBackend::WindowsSandbox => Some("experimental.windows_sandbox"),
-            ContainmentBackend::Wslc => Some("experimental.wslc"),
+            ContainmentBackend::Wslc => Some("wslc"),
```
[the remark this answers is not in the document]
adpa-ms: [AgentAssisted] Following up in case this was resolved in a batch — both directions look unchanged at `17d27c2f`.

`matching_key` is derived by `section_path().strip_prefix("experimental.")`. At the merge base `Wslc.section_path()` was `Some("experimental.wslc")`, so a foreign key was rejected; it's now `Some("wslc")`, so `matching_key` is `None` and the `None => return Ok(())` arm accepts a single foreign key, which `cfg.experimental = None` then drops. The reverse is thread 5: the `cfg.wslc` block in `convert_wire_state_aware` checks phase and one-shot-only fields but not the selected containment.

Neither direction mis-provisions anything — containment still wins and its policy is unaffected. The cost is that a malformed two-backend config is now accepted silently instead of rejected.

Worth noting this is per-promotion rather than WSLc-specific: `matching_key` is `None` for any containment whose `section_path()` has no `experimental.` prefix, so `isolation_session` will hit the same blind spot when it's promoted. Fixing the derivation rather than the instance would settle it once.

One-shot is unaffected — `present_backend_sections` / `validate_single_backend_section` were updated correctly.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1490,6 +1497,49 @@ fn convert_wire_state_aware(
...
+    // deserialize above already proved `json` is a JSON object.
+    let stable_raw = serde_json::from_str::<serde_json::Value>(json).ok();
+
+    // `wslc` is promoted to the stable surface but, unlike the other promoted
+    // sections, is not one-shot-only: the state-aware lifecycle reads its
+    // provision config from `wslc.provision`. So it is not a stray section —
+    // instead accept `provision` and reject the one-shot-only siblings, which
+    // the daemon-backed lifecycle does not honor, rather than silently
+    // dropping them.
+    if let Some(wslc) = cfg.wslc.as_ref() {
```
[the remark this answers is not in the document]
adpa-ms: [AgentAssisted] Following up in case this was resolved in a batch — unchanged at `17d27c2f`. Same root cause as the `section_path()` thread on `models.rs`; reasoning and suggested fix are there to keep it in one place.

src/core/wxc_common/src/state_aware_backend.rs
adpa-ms: [AgentAssisted] Sessions are fine as-is. `SECTION_ROOT` defaults to `Experimental` and `IsolationSessionRunner` doesn't override it, so iso dispatch and its `experimental.isolation_session.<phase>` path are unchanged by this PR.

Two things worth tightening before sessions are promoted:

1. The default means a backend promoted later that forgets to override it silently keeps the old root — and a mismatch fails quietly, since `deserialize_config` returns `Ok(None)` for a missing key and the backend then applies defaults. Dropping the default (one line in three impls, three stubs) makes it a compile error instead. Same shape as your `unwrap_or_default()` note on `run.rs`.

2. The root is declared in three places — here, `BACKEND_SECTION_ROOT`, and `IsExperimentalSection` — with no parity check; `check-dotnet-api-parity.js` covers the enum and id prefixes but not the root.

sdk/node/src/state-aware-helper.ts
adpa-ms: [AgentAssisted] Sessions are unaffected — `isolation_session` stays on the experimental root and this map is exhaustive, so adding a backend without declaring its root is a compile error. Changing an existing value isn't checked though; noted on the `SECTION_ROOT` thread along with the Rust-side default.

adpa-ms:
[AgentAssisted] Two documentation spots a comment sweep may not reach — both outside the current diff, so not attachable to a line.

`wire.rs`'s `Containment::Wslc` is still documented `/// WSL container (experimental).`, and that string is emitted as the `description` for the `wslc` containment value in the regenerated `0.9.0-dev` schema, so config authors see "experimental" in editor validation. The codegen gate will catch the regeneration once the doc comment is updated.

`docs/state-aware-lifecycle/mxc-state-aware-sandbox-api.md` still contains the literal `experimental.<backend>` 16 times, with no mention of `SectionRoot` / `SECTION_ROOT` / `stable_raw` / `BACKEND_SECTION_ROOT`. As the cross-backend contract doc it's now wrong for one of its three backends, and it's the doc the next promotion would follow.

src/core/wxc_common/src/config_parser.rs
SohamDas2021: yes, but not doing a refactor in here, this one is already too big.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
SohamDas2021: Needed. The switch isn't WSLC-specific routing — `IsolationSessionContainment` and `WindowsSandboxContainment` still return `true` and require `experimental.<backend>.<phase>`. Deleting it breaks both. The `WslcContainment => false` arm is also load-bearing: without it the `_ =>` arm throws `UnsupportedContainment` for wslc.

src/core/mxc_engine/src/run.rs
SohamDas2021: There's no failed unwrap to catch. `request.wslc` is `None` only when the section was omitted (in config_parser)- a malformed section already errors earlier. And when the section *is* present, the parser starts from `WslcConfig::default()` and overlays supplied fields. So `unwrap_or_default()` makes an absent section identical to `"wslc": {}`; erroring would let `{}` pass while omission fails.

schemas/dev/mxc-config.schema.0.9.0-alpha.json
SohamDas2021: That is a valid point.

Checked the SDK API reference (WslcSetSessionSettingsMemory), I don't think it caps anything. 
The only real ceiling is the ABI: the setter takes `uint32_t`, so `u32::MAX` MB. We already enforce exactly that in the runner before the call. So keeping as-is.

One thing worth knowing: `0` isn't "no memory" — the SDK treats it as a sentinel that resets to the default 2000 MB (`cpuCount: 0` likewise resets to 2). We land on the same behavior either way, since omitting the field skips the setter entirely.

docs/wsl/wsl-container-support-plan.md
SohamDas2021: Yes, declared in four crates - `wxc` (`wslc = ["dep:wslc_common", "wslc_common/link-wslcsdk", "mxc_engine/wslc"]`), `mxc_engine`, `mxc-sdk`, and `mxc_ffi` with `wslc_common`'s `link-wslcsdk` at the end of the chain. It's off by default; `build.bat --with-wslc` turns it on.

scripts/check-dotnet-api-parity.js
SohamDas2021: So, the old check derived the expected C# `StateAwareContainment` members from the `matches!` list in `state_aware.rs`, which is the *experimental opt-in* registry. Promoting WSLC removes `Wslc` from that list, so the old oracle would have demanded `Wslc` be deleted from the C# enum- even though it's still a dispatchable state-aware backend.

sdk/node/tests/unit/sandbox.test.ts
SohamDas2021: Yes, this is a gap currently. Filed https://github.com/microsoft/mxc/issues/1106 to track.

SohamDas2021: These are already handled.  

`build.rs` resolves the SDK- `WSLC_SDK_PATH` if set, else the pinned nupkg from the MxcDependencies feed, falling back to the vendored copy if that download fails. It then copies `wslcsdk.dll` to the cargo profile dir (`build.rs:112-119`) - beside the binary being built, including for downstream crates enabling the `wslc` feature.

**Discovery:** module-relative, not exe-relative — `wslc_bindings.rs:92-96` loads from `module_path_for_address(Self::load)`, so inside `mxc_ffi.dll` it resolves beside that DLL. No resolver work was needed for C#.

**NuGet:** `-p:MxcWithWslc=true` already stages `wslcsdk.dll` + `wxc-wslc-daemon.exe` into the native unit, SHA-verifies them, and packs to `runtimes/<rid>/native` (`sdk/dotnet/README.md:681-702`).

The gap is that none of this is documented for Rust SDK consumers, and `cargo install` drops the profile-dir copy. Filed as #1106.

sdk/node/src/generated/wire.ts
SohamDas2021: Right, this file is generated from the Rust wire type, it changed here because `wire.rs` changed

src/core/mxc_config_contract/src/dev/stable.rs
```diff
@@ -352,3 +353,57 @@ pub struct Seatbelt {
...
+}
+
+/// Windows WSL container backend settings.
+#[derive(Debug, serde::Deserialize)]
+#[cfg_attr(feature = "schema-gen", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct Wslc {
+    /// Target operating system inside the container.
+    #[serde(default)]
+    pub target_os: OptionalField<String>,
```
adpa-ms, line 387: [AgentAssisted] `targetOs` has no consumer — no read of `WslcConfig::target_os` anywhere in `src/backends/wslc/`, and the parser passes any string through, so `"windows"` or `""` silently runs Linux. Pre-existing and unchanged from `main`, so not a regression.

Flagging it here only because this block is the closed stable surface: removing the field is free today and a breaking change later. If it's reserved for planned multi-OS support, that's a good reason to keep it and no action is needed — otherwise dropping it, or validating `"linux"`, would close the gap.

SohamDas2021:
Done in #1187 - #1189 pr stack

PR #1070, opened 2026-08-31
Fix captureDenials probe parity

### LINKED ISSUE #785: wxc-exec --probe ignores captureDenials, reporting a launchable tier for runs that fail
### Relevant area(s)

Windows

`wxc-exec --probe` does not account for `processContainer.captureDenials`, so for a
config that uses it the probe can report a launchable tier for a run that will
actually fail. The probe is the supported way to ask "will this config run on this
host?", so a false positive here sends people to debug their config instead of
their OS build.

Two independent paths diverge between `run_probe` and the real dispatcher:

**1. Tier gating.** `dispatcher.rs:346-357` calls
`fallback_detector::detect_with_base_container_capabilities(...)` and then rejects
the request when capture is requested on a non-BaseContainer tier:

```rust
if request.policy.capture_denials.is_some() && decision.tier != IsolationTier::BaseContainer {
    return Err(DispatchError::CaptureDenialsUnsupported { tier: decision.tier });
}
```

`probe.rs:131` calls the plain detector and has no equivalent check:

```rust
match fallback_detector::detect(policy, /* prefer_base_container */ true) {
```

`probe.rs` contains no reference to `capture_denials`, `learning_mode`, or
`SecurityEnvironment` at all.

**2. V2 API availability.** Since #739, `captureDenials` requires the complete
official V2 Learning Mode and process security-environment export set, and a host
without it fails with `FailurePhase::BackendUnavailable`. `ProbeFacts`
(`probe.rs:118-130`) reports `base_container_api_present`, `bfscfg_present`,
`bfs_compiled_in`, `base_container_supports_deny_paths`, and `ui_capabilities` —
but nothing about the V2 export set, so the probe cannot surface this at all.

### Steps to reproduce

1. Take a Windows host where **either** the ProcessContainer tier resolves to
   AppContainer+DACL rather than BaseContainer, **or** the OS predates the complete
   V2 Learning Mode export set (validated in #739 against V1 `26657.1002`).
2. Author a config that sets `processContainer.captureDenials` (with
   `--experimental`).
3. Run `wxc-exec --probe` against it.
4. Run the same config for real.

### Expected behavior

`--probe` and the real run agree. For a `captureDenials` config the probe should
apply the same gating the dispatcher does, and report the request as unsupported on
a host that cannot service it — ideally with a `ProbeFacts` field naming the V2
export set so the reason is visible rather than inferred.

### Actual behavior

`--probe` reports a tier and no error, implying the config will launch. The real run
then fails — with `CaptureDenialsUnsupported` in the tier case, or
`FailurePhase::BackendUnavailable` in the V1-host case.

Note this is the *diagnostic* disagreeing with the runtime; the runtime itself fails
correctly and closed, so this is a triage-accuracy bug rather than a containment
issue.

### Additional context

Found while verifying review findings against `main` after #739 merged
(`18d73ecb`). Everything else raised on that PR landed — retryable trace delivery,
the lifecycle test seam, `OnceLock` export memoisation, the `wire.rs` /
generated-schema description, the OS build floor in
`docs/process-container/os-version-support.md`, and `.fbs` provenance with a pinned
`flatc`. This one was not picked up and nothing tracks it.

Suggested fix: have `run_probe` mirror the dispatcher — use
`detect_with_base_container_capabilities` with the same capability inputs, apply the
`capture_denials` tier check, and add a `learning_mode_v2_api_present` (or similar)
field to `ProbeFacts`. `run_probe` currently takes only a `&ContainerPolicy`, and
`capture_denials` lives on the policy, so the signature does not need to change.

huzaifa-d:
## 📖 Description
Align `--probe` capture-denial selection with runtime behavior and report native capture availability.

## 🔗 References
Resolves #785

## 🔍 Validation
`cargo test -p appcontainer_common`; targeted Clippy and rustfmt.

## ✅ Checklist
- [x] Linked to an issue

## 📋 Issue Type
- [x] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1070)


shschaefer (dismissed):

shschaefer (approved):

src/backends/appcontainer/common/src/probe.rs
```diff
@@ -119,31 +128,64 @@ impl From<EffectiveUiRestrictions> for UiCapabilitySupport {
...
+    probes: ProbeFacts,
+    base_container_usable: bool,
+    uses_native_capture: bool,
+    supports_deny_paths: bool,
+) -> ProbeOutput {
     match detect_request_tier(request, base_container_usable, supports_deny_paths) {
+        Ok(decision)
+            if request.policy.capture_denials.is_some()
+                && (decision.tier != fallback_detector::IsolationTier::BaseContainer
+                    || !uses_native_capture)
```
bbonaby, line 178: note: Would probably be best if we see if there was something we can do to combine these checks or consolidate them to they're easy to reason about at a glance.

bbonaby (approved):

src/backends/appcontainer/common/src/probe.rs
huzaifa-d: Noted

PR #1079, opened 2026-09-01
fix(seatbelt): Apply the cwd to the spawn_open (in-terminal) launch path.

theelliotm:
## 📖 Description

`spawn_open` never applied `process.cwd`. `open` hands the launch to Terminal,
which starts the helper in its own directory (normally `$HOME`), so the
resolved working directory was silently discarded — only `spawn_exec` set it.

The generated helper script now does what `spawn_exec` does before the sandbox
is applied: `cd` into `resolve_working_directory(request)` and export `PWD` to
it. The `cd` is guarded (`|| exit 1`) — a bare one would leave the workload
running in Terminal's directory. A directory that doesn't exist is rejected
before Terminal launches, matching the exec path's spawn failure, since
`open -W` reports its own status and not the helper's.

## 🔗 References

None.

## 🔍 Validation

- 5 new unit tests in `seatbelt_runner`: `cd` + `PWD` emitted, abort guard
  ordering, `PWD` assigned after caller env, single-quote escaping, and parity
  with the exec path's resolution.
- `cargo test -p seatbelt_common` — 106 passed.
- `cargo fmt --all -- --check`, `cargo clippy -p seatbelt_common -p mxc_darwin --all-targets -- -D warnings` — clean.
- Manual, macOS 26.6.2, against the real generated profile with `cwd` granted in `readwritePaths`:

  | probe | before | after |
  |---|---|---|
  | `pwd` | `/Users/agos` | `/tmp/mxc611_work` |
  | `cat f.txt` (relative) | `No such file or directory` | `hello` |

## ✅ Checklist

- [ ] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1079)


richiemsft (dismissed):

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -327,7 +327,26 @@ fn spawn_open(
...
+        Ok(cwd) => cwd,
+        Err(e) => {
+            let _ = fs::remove_file(&profile_path);
+            return Err(error_response(format!(
+                "failed to read the current directory to anchor the relative seatbelt working directory: {e}"
+            )));
+        }
+    };
+    if let Some(reason) = working_directory_error(&cwd) {
+        let _ = fs::remove_file(&profile_path);
```
bbonaby, line 345: note: fyi, might not have been considered in the past but looks like the remove_file usage here actually swallows the error https://doc.rust-lang.org/std/fs/fn.remove_file.html.

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -719,6 +732,88 @@ fn resolve_working_directory(request: &ExecutionRequest) -> String {
...
+    launch_dir: impl FnOnce() -> std::io::Result<PathBuf>,
+) -> std::io::Result<String> {
+    let candidate = Path::new(path);
+    if candidate.is_absolute() {
+        return Ok(path.to_string());
+    }
+    Ok(launch_dir()?.join(candidate).to_string_lossy().into_owned())
+}
+
+/// Why the helper's `cd` into `path` would fail, or `None` if it will succeed.
```
bbonaby: note: first sentence is a little clunky for a description of why this function exists.

theelliotm: Rewrote the comment to "Returns the error message for when the resolved cwd isn't usable".

src/backends/seatbelt/common/src/seatbelt_runner.rs
theelliotm: The return of fs::remove_file is ignored every time in the file. There are multiple attempts to delete the files again in cleanup_files, and if a file isn't removable than either (1) the file was never created, (2) the ownership of it changed due to the user, or (3) the file has already been removed prior. If in the future a bug is created that causes the file reference to be incorrect, I'll revisit this.

bbonaby (approved):

PR #1096, opened 2026-09-03
Add private exact contract parsing

MGudgin:
## 📖 Description

This PR adds private exact-contract parsing and test-only versioned policy builders while keeping the rolling parser authoritative for every production entry point.

Details
* Adds exact one-shot and state-aware version dispatch through the existing shared normalization seams.
* Adds explicit contract construction primitives, a hidden one-shot bridge, and private v0.6 through v0.9 test builders without implicit public conversion traits.

Tests
* `cargo clippy --workspace --all-targets -- -D warnings`
* `cargo test --workspace`
* `cargo check -p mxc_engine -p mxc-sdk --all-targets --target aarch64-apple-darwin`

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1096)


src/core/wxc_common/src/config_parser.rs
```diff
@@ -236,6 +237,176 @@ pub fn load_request_from_value(
...
+        ParseError::OneShot(WxcError::ConfigParse(message))
+    }
+}
+
+fn parse_exact_development(json: &str, logger: &mut Logger) -> Result<MxcRequest, ParseError> {
+    let state_aware = mxc_config_contract::dev::probe_phase(json)
+        .map_err(exact_phase_error)?
+        .is_some();
+    let request = mxc_config_contract::dev::parse_request(json)
+        .map_err(|error| exact_development_request_error(error, state_aware))?;
```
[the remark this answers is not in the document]
MGudgin: Fixed in b4f9fa52. All eight concrete development roots now deserialize through config_deserialize in wxc_common, retaining full paths, whole-document source locations, and shared escaping/redaction without introducing a runtime dependency into the contract crate. The new regressions cover all four one-shot versions, every development root, deep backend payload paths, and diagnostic escaping/secret-path handling. Four diagnostic tests failed against the previous tip and pass with the fix; the common, engine, and contract library suites also pass. The dependent differential matrix was expanded in #1097 (e26d5b29) to cover development one-shot and state-aware errors as well. This places the fix in this parser-staging PR rather than relying on the later cutover implementation.

shschaefer (dismissed):

src/core/mxc_engine/src/policy/exact/v0_9.rs
```diff
@@ -0,0 +1,414 @@
...
+        ime: contract::OptionalField::present(ui.ime),
+    }
+}
+
+fn map_legacy_enforcement(value: LegacyEnforcement) -> contract::NetworkEnforcementMode {
+    match value {
+        LegacyEnforcement::Capabilities => contract::NetworkEnforcementMode::Capabilities,
+        LegacyEnforcement::Firewall => contract::NetworkEnforcementMode::Firewall,
+        LegacyEnforcement::Both => contract::NetworkEnforcementMode::Both,
+    }
```
bbonaby, line 76: question: I was under the impression that with v0.9 we'd remove the legacy networking stuff. Or is that only for when we're moving to 1.0?

src/core/mxc-sdk/README.md
```diff
@@ -39,10 +39,10 @@ Ok(())
...
 
 [`build_request`] resolves the host's default containment backend (see
-[Supported backends](#supported-backends)), builds the wire config, and runs it
-through the shared parser. The command is supplied to [`build_request`], so the
-returned [`SandboxRequest`] is complete; optionally adjust its working directory
-or environment before spawning.
+[Supported backends](#supported-backends)), builds the rolling wire config, and
+runs it through the shared production parser. The command is supplied to
+[`build_request`], so the returned [`SandboxRequest`] is complete; optionally
+adjust its working directory or environment before spawning.
```
bbonaby, line 45: note: since this will eventually be viewed on crates.io we probably don't want to talk about the rolling wire or prod parser, especially since I think your plan is to remove the rolling stuff eventually.

bbonaby (dismissed):

src/core/mxc_engine/src/policy/exact/v0_9.rs
MGudgin: Phase 10

bbonaby (approved):

shschaefer (approved):

PR #1120, opened 2026-09-08
Add default environment behavior across SDKs

### LINKED ISSUE #483: Windows processcontainer: Bun exits 66 with no stderr; PATH env causes CreateProcessW error
## Summary

On Windows with `@microsoft/mxc-sdk@0.6.1`, `processcontainer` can successfully execute a simple PowerShell command, but attempting to execute Bun through the same SDK runner exits nonzero with no stderr. I also hit a separate env handling failure when passing the host `PATH` through the SDK env parameter.

This may be a Bun/runtime compatibility issue, but the current failure mode is hard to diagnose.

## Environment

- OS: Windows
- SDK: `@microsoft/mxc-sdk@0.6.1`
- Containment: `process` resolving to `processcontainer`
- Spawn mode: `spawnSandboxFromConfig(..., { usePty: false })`

`getPlatformSupport()` reported:

```json
{"isSupported":true}
```

The local capability wrapper resolved the backend as `processcontainer`.

## What worked

Running PowerShell inside the MXC process container worked:

```ts
await runner.exec(
  "C:/Windows/System32/WindowsPowerShell/v1.0/powershell.exe",
  ["-NoLogo", "-NoProfile", "-Command", "Write-Output mxc-smoke"],
  {
    cwd: tempDir,
    readonlyPaths: ["C:/Windows/System32"],
    readwritePaths: [tempDir],
    timeoutMs: 30_000,
  },
);
```

Result:

```json
{"stdout":"mxc-smoke\r\n","stderr":"","code":0}
```

## What failed

Running Bun through the same runner failed:

```ts
await runner.exec(
  process.execPath,
  ["-e", "console.log('mxc smoke')"],
  {
    cwd: tempDir,
    readonlyPaths: [process.execPath],
    readwritePaths: [tempDir],
    timeoutMs: 30_000,
  },
);
```

Result:

```json
{"stdout":"","stderr":"","code":66}
```

There was no stderr or diagnostic text explaining whether this is expected, unsupported, missing filesystem access, env-related, or a child runtime failure.

## Env-related failure

Before removing inherited `PATH`, passing `PATH` via the SDK env parameter produced:

```text
Process error: CreateProcessW failed: The system could not find the environment option that was entered. (0x800700CB)
{"error":{"code":"backend_error","message":"Process error: CreateProcessW failed: The system could not find the environment option that was entered. (0x800700CB)"}}
```

That may be expected if `PATH` contains entries that are invalid inside the container, but it would be useful for the SDK/backend to document or diagnose that case clearly.

## Expected behavior

Either:

- Bun works under Windows `processcontainer` when the executable path is readable and no broad env is passed, or
- MXC reports a clearer diagnostic explaining why that runtime cannot start in this containment mode.

For the env case, it would help to document whether callers should avoid passing host `PATH` and instead provide a container-safe env block.

shschaefer:
## 📖 Description
Fixes Windows ProcessContainer launches that use a caller-supplied environment without the variables required by Windows initialization.

This change preserves the distinction between:

- An omitted environment, which uses the backend’s default environment.
- An explicitly empty environment, which remains empty.
- A supplied environment, which is used verbatim.
- A supplied environment with `inheritDefaultEnv`, which is layered over the backend default.

The new `process.inheritDefaultEnv` option is exposed consistently through the Rust, Node.js, .NET, and FFI request surfaces. This option now builds the effective environment from the user profile block when inheritance is requested, while other backends retain their existing behavior.

The change also improves launch diagnostics for incomplete environment blocks, updates generated schemas and wire types, and documents the new behavior.

## 🔗 References
Resolves #483

## 🔍 Validation
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace --exclude wxc_e2e_tests`
- Built `wxc` with the `microvm` feature and ran the Windows E2E suite:
  - 6 passed
  - 14 host-dependent tests ignored
- Node.js SDK:
  - `npm run build`
  - `npm test`
- .NET SDK:
  - 118 passed
  - 24 backend-dependent tests skipped
- Schema-version consistency check
- SDK wire-type codegen check
- Development contract codegen check
- Package version consistency check

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type
- [x] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1120)

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -84,7 +84,7 @@ cannot mix both formats in one request.
 
 ```json
 {
-    "version": "0.6.0-alpha",              // Schema version (semver). Minimum supported: "0.6.0-alpha"; current stable: "0.8.0-alpha".
+    "version": "0.9.0-alpha",              // Schema version (semver). Minimum supported: "0.6.0-alpha"; current stable: "0.8.0-alpha".
     "containerId": "my-container",         // Externally assigned container ID
     "containment": "processcontainer",     // Backend (see table below)
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -100,7 +100,7 @@ cannot mix both formats in one request.
                                            //  than inheriting the launcher's — see
                                            //  "Working Directory" below)
         "env": ["MY_VAR=value"],           // Omitted: backend default; supplied: used verbatim
-        "inheritDefaultEnv": true,         // Layer env on the backend default (default false)
+        "inheritDefaultEnv": true,         // Layer env on the backend default (0.9.0-alpha+)
         "timeout": 30000                   // Timeout in ms (0 = no timeout)
     },
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -99,7 +99,8 @@ cannot mix both formats in one request.
                                            //  backend substitutes a granted directory rather
                                            //  than inheriting the launcher's — see
                                            //  "Working Directory" below)
-        "env": ["MY_VAR=value"],           // Environment variables as KEY=VALUE
+        "env": ["MY_VAR=value"],           // Omitted: backend default; supplied: used verbatim
+        "inheritDefaultEnv": true,         // Layer env on the backend default (default false)
         "timeout": 30000                   // Timeout in ms (0 = no timeout)
     },
 
--- a/sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
+++ b/sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
@@ -271,6 +271,10 @@ public static class MxcLifecycle
         {
             process["env"] = SerializeToNode(env);
         }
+        if (options?.InheritDefaultEnvironment is { } inheritDefaultEnv)
+        {
+            process["inheritDefaultEnv"] = inheritDefaultEnv;
+        }
         if (options?.TimeoutMs is { } timeout)
         {
             process["timeout"] = timeout;
--- a/sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
+++ b/sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
@@ -264,16 +264,9 @@ public static class MxcLifecycle
         ArgumentNullException.ThrowIfNull(command);
         ValidateExecOptions(id, options);
         var version = options?.Version;
-        if (options?.InheritDefaultEnvironment is not null)
+        if (options?.InheritDefaultEnvironment is not null && version is null)
         {
-            if (version is null)
-            {
-                version = InheritDefaultEnvironmentVersion;
-            }
-            else
-            {
-                ValidateInheritDefaultEnvironmentVersion(version);
-            }
+            version = InheritDefaultEnvironmentVersion;
         }
         var envelope = BuildIdEnvelope("exec", id, version);
         var process = new JsonObject { ["commandLine"] = command };
@@ -413,24 +406,6 @@ public static class MxcLifecycle
         containment == StateAwareContainment.Wslc
             ? WslcStateAwareVersion
             : StateAwareVersion;
-
-    private static void ValidateInheritDefaultEnvironmentVersion(string version)
-    {
-        var coreVersion = version.Split('-', 2)[0].Split('.');
-        if (coreVersion.Length < 2
-            || !uint.TryParse(coreVersion[0], out var major)
-            || !uint.TryParse(coreVersion[1], out var minor))
-        {
-            return;
-        }
-        if (major == 0 && minor < 9)
-        {
-            throw new MxcException(
-                ErrorCode.MalformedRequest,
-                $"process.inheritDefaultEnv requires schema version "
-                    + $"{InheritDefaultEnvironmentVersion} or later; got {version}");
-        }
-    }
 
     private static void ValidateProvisionOptions(
         StateAwareContainment containment,
--- a/sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
+++ b/sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
@@ -29,6 +29,8 @@ public static class MxcLifecycle
     /// <summary>Default state-aware schema for WSLC.</summary>
     public const string WslcStateAwareVersion = SchemaVersions.WslcStateAware;
 
+    private const string InheritDefaultEnvironmentVersion = "0.9.0-alpha";
+
     /// <summary>IsolationSession containment wire key.</summary>
     public const string IsolationSessionContainment = "isolation_session";
 
@@ -261,7 +263,19 @@ public static class MxcLifecycle
     {
         ArgumentNullException.ThrowIfNull(command);
         ValidateExecOptions(id, options);
-        var envelope = BuildIdEnvelope("exec", id, options?.Version);
+        var version = options?.Version;
+        if (options?.InheritDefaultEnvironment is not null)
+        {
+            if (version is null)
+            {
+                version = InheritDefaultEnvironmentVersion;
+            }
+            else
+            {
+                ValidateInheritDefaultEnvironmentVersion(version);
+            }
+        }
+        var envelope = BuildIdEnvelope("exec", id, version);
         var process = new JsonObject { ["commandLine"] = command };
         if (options?.WorkingDirectory is { } cwd)
         {
@@ -399,6 +413,24 @@ public static class MxcLifecycle
         containment == StateAwareContainment.Wslc
             ? WslcStateAwareVersion
             : StateAwareVersion;
+
+    private static void ValidateInheritDefaultEnvironmentVersion(string version)
+    {
+        var coreVersion = version.Split('-', 2)[0].Split('.');
+        if (coreVersion.Length < 2
+            || !uint.TryParse(coreVersion[0], out var major)
+            || !uint.TryParse(coreVersion[1], out var minor))
+        {
+            return;
+        }
+        if (major == 0 && minor < 9)
+        {
+            throw new MxcException(
+                ErrorCode.MalformedRequest,
+                $"process.inheritDefaultEnv requires schema version "
+                    + $"{InheritDefaultEnvironmentVersion} or later; got {version}");
+        }
+    }
 
     private static void ValidateProvisionOptions(
         StateAwareContainment containment,
--- a/sdk/dotnet/Microsoft.Mxc.Sdk/SandboxRequest.cs
+++ b/sdk/dotnet/Microsoft.Mxc.Sdk/SandboxRequest.cs
@@ -41,18 +41,19 @@ public sealed class SandboxRequest
     public string? WorkingDirectory { get; set; }
 
     /// <summary>
-    /// Environment variables supplied to the sandboxed process.
+    /// Optional environment variables supplied to the sandboxed process.
     /// </summary>
     /// <remarks>
-    /// When non-empty, this is the child's environment and is used verbatim —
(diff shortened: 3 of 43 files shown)


src/core/wxc_common/src/config_contract_adapters/v0_6.rs
```diff
@@ -31,6 +31,8 @@ fn convert_process(value: contract::Process) -> wire::Process {
         command_line: Some(command_line.into_inner()),
         cwd: cwd.into_option(),
         env: env.into_option(),
+        // The field postdates this released schema, so it is never set here.
+        inherit_default_env: None,
```
MGudgin, line 35: Yay! You had to do this because the new versioning stack is enforcing the 'make a decision about every field' rule :-)

src/backends/appcontainer/common/src/launch_diagnostics.rs
```diff
@@ -28,6 +28,78 @@ pub struct LaunchDiagnostic {
...
+
+/// Diagnose `ERROR_ENVVAR_NOT_FOUND` from a contained-process launch.
+///
+/// `supplied_env` is the caller's `process.env` — `None` when they supplied
+/// none, in which case MXC built the block itself and a missing variable is not
+/// the caller's doing, so no diagnostic is produced.
+///
+/// Returns `None` unless the error is 203 *and* a caller-supplied block is
+/// missing at least one of [`REQUIRED_CHILD_ENV_VARS`]; the generic
+/// [`diagnose_create_process_failure`] handles every other case.
```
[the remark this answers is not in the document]
shschaefer: This is incorrect.  I have verified that bun functions properly and the system returns reasonable diagnostics.

src/backends/appcontainer/common/src/launch_diagnostics.rs
```diff
@@ -144,13 +216,15 @@ fn check_exe_heuristics(
...
-    if exit_code == Some(STATUS_DLL_INIT_FAILED.0 as u32) && is_powershell(exe_path) {
+    if exit_code == Some(STATUS_DLL_INIT_FAILED.0 as u32) {
         return Some(LaunchDiagnostic {
             kind: "dll_init_failed_ui_required",
-            message: "PowerShell exited with STATUS_DLL_INIT_FAILED (0xC0000142). \
-                      This typically means the sandbox is blocking Win32k system calls \
-                      (UI subsystem access), which PowerShell requires to initialize. \
-                      Enable UI access in your sandbox policy (set `ui.allowWindows: true`)."
+            message: "The target executable exited with STATUS_DLL_INIT_FAILED (0xC0000142). \
+                      This often means the sandbox is blocking Win32k system calls \
```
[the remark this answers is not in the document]
shschaefer: As stated earlier, the PR fixes the source of the bun error and verifies that the reporter's script now functions properly.  The comment also states that this diagnostic is suggested, not guaranteed.  It makes no worse the assessment that launch of an arbitrary executable may result in this failure.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -396,6 +414,24 @@ private static string DefaultVersion(StateAwareContainment containment) =>
             ? WslcStateAwareVersion
             : StateAwareVersion;
 
+    private static void ValidateInheritDefaultEnvironmentVersion(string version)
+    {
+        var coreVersion = version.Split('-', 2)[0].Split('.');
+        if (coreVersion.Length < 2
+            || !uint.TryParse(coreVersion[0], out var major)
+            || !uint.TryParse(coreVersion[1], out var minor))
```
jsidewhite: This is no good - parsing the version (and the string version at that) in every front-end.

The engine beneath parses the version, and should perform this comparison/rejection.

sdk/node/src/sandbox.ts
```diff
@@ -481,6 +482,36 @@ export interface SandboxSpawnOptions {
...
+  /**
+   * State-aware lifecycle only: the correlation vector (MS-CV) returned by
+   * {@link provisionSandbox} as `correlationVector`. Relay it verbatim on every
+   * later phase (`start` / `exec` / `stop` / `deprovision`) so all phases of one
+   * lifecycle share a telemetry base prefix. The client relays it unchanged; the
+   * executor derives each phase's own vector from it (spinning a mutable base or
+   * reseeding a missing/malformed value). Ignored by one-shot spawns and by
+   * `provision` (which seeds its own).
+   */
+  correlationVector?: string;
```
jsidewhite: from Ramon??

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -86,6 +86,23 @@ use windows::Win32::System::Threading::{
...
+///   return `Ok(None)` so the API supplies that default itself.
+/// * `Some(entries)` — the caller's environment, used **verbatim**. MXC adds
+///   nothing to it, including when `entries` is empty: an explicitly empty
+///   environment produces an empty block, not the default one. Proxy variables
+///   are still injected, because the proxy is enforced policy rather than
+///   inherited environment.
+///
+/// When `inherit_default_env` is set, `entries` is layered on top of the
+/// default block instead of replacing it, which is the supported way to ask for
+/// "the user's profile block plus these".
```
jsidewhite, line 106: This seems off.

/// When `inherit_default_env` is set, `entries` is layered on top of 
/// ***this process's (e.g. wxc-exec.exe's env)*** block instead of replacing it, which is the supported way to ask for
/// "the ***current environment block*** plus these".

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxRequest.cs
```diff
@@ -40,9 +40,35 @@ public SandboxRequest(SandboxPolicy policy, string command)
...
+    /// Optional environment variables supplied to the sandboxed process.
+    /// </summary>
+    /// <remarks>
+    /// When non-null, this is the child's environment and is used verbatim —
+    /// including when the dictionary is empty. Nothing is merged into it, so
+    /// an environment missing what the platform requires fails the launch.
+    /// Set <see cref="InheritDefaultEnvironment"/> to layer these entries on
+    /// top of the default environment instead. Leave this property null to
+    /// give the child the backend's default environment (on Windows, the
+    /// user's profile block).
```
jsidewhite, line 53: I am wholly confused - `inheritDefaultEnv` is supposed to be to inherit from parent, not get a clean user profile block (i.e. the "login" block). #ByDesign

src/core/mxc_engine/src/policy.rs
```diff
@@ -245,6 +245,28 @@ fn env_or_process(env: Option<&[(String, String)]>) -> Cow<'_, [(String, String)
     }
 }
 
+fn environment_keys_equal(left: &str, right: &str) -> bool {
+    if cfg!(target_os = "windows") {
+        left.eq_ignore_ascii_case(right)
```
jsidewhite: >eq_ignore_ascii_case [](http://example.com/codeflow?start=13&length=20)

is that right for values?  isn't that only true for keys?

jsidewhite: i presume values are case sensitive because you don't know how someone will interpret them.  i.e. case must be preserved

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxRequest.cs
jsidewhite: Okay thinking for an additional 4 seconds, your field name does have the word *default* in it, but you can already inherit the default environment by saying `env = null`... so why do we need `InheritDefaultEnvironment`?

What we need is,

`InheritProcessEnvironment`

src/core/mxc_engine/src/policy.rs
```diff
@@ -647,21 +669,95 @@ impl SandboxRequest {
...
     }
 
+    /// Start from the *calling process's* environment and append `extra` on top.
+    ///
+    /// Note this is a different, generally larger and leakier set than
+    /// [`Self::inherit_default_env`]: it is whatever your process happens to be
+    /// running with, so anything you inherited — including secrets in the
+    /// ambient environment — is handed to the sandboxed child. Prefer
+    /// `inherit_default_env` unless you specifically need your own variables.
+    pub fn inherit_process_env<K, V>(
```
jsidewhite, line 739: >inherit_process_env [](http://example.com/codeflow?start=11&length=19)

This is cool, and what i thought needed to be exposed in the schema.  But nobody is calling this `fn`?

docs/schema.md
```diff
@@ -99,7 +99,8 @@ cannot mix both formats in one request.
                                            //  backend substitutes a granted directory rather
                                            //  than inheriting the launcher's — see
                                            //  "Working Directory" below)
-        "env": ["MY_VAR=value"],           // Environment variables as KEY=VALUE
+        "env": ["MY_VAR=value"],           // Omitted: backend default; supplied: used verbatim
+        "inheritDefaultEnv": true,         // Layer env on the backend default (0.9.0-alpha+)
```
jsidewhite, line 110: It's annoying to change the form now, but presumably you thought about and rejected stuffing this into the "env" node.

It would fit better there for clarity and avoid false assumptions.

..but too much nesting is gross too.

jsidewhite: instead of

    "inheritDefaultEnv": true

could do

    "envSeed": "user|process|null"

because the bool is confusing when considering "env" can be *null* and achieve the same result.  or specifying nothing at all achieves the same result.

jsidewhite (dismissed):
:shipit:

MGudgin (dismissed):

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
shschaefer: Fixed.

sdk/node/src/sandbox.ts
shschaefer: Yes, it was a rebase inheritance.  Will remove.

src/core/mxc_engine/src/policy.rs
shschaefer: Unix names are case-sensitive.  I changed "left/right" to "existing_key/override_key", but the values have to remain case-sensitive.

src/backends/appcontainer/common/src/base_container_runner.rs
shschaefer: As discussed, the semantics are correct.  If the caller wants "the current environment plus", they have to do it explicitly themselves.  This is considered a risky operation that we will not do automatically.

sdk/dotnet/Microsoft.Mxc.Sdk/SandboxRequest.cs
shschaefer: No.  There is no automatic inherit from parent.  Either no inheritance or inherit the default profile.  The caller has to be explicit if they want inherit process environment.

src/core/mxc_engine/src/policy.rs
shschaefer: We could expose this in the API, but for reasons described in the other comments we don't directly expose the behavior

docs/schema.md
shschaefer: This is possible.  I will wrap up this PR then consider this approach.

jsidewhite (dismissed):
:shipit:

jsidewhite (approved):
:shipit:

MGudgin (approved):

PR #1128, opened 2026-09-09
[Seatbelt] Grant the active developer directory in the baseline profile

### LINKED ISSUE #1110: macOS: Seatbelt baseline omits the active developer directory, breaking /usr/bin xcrun shims
### Relevant area(s)
macOS

The Seatbelt baseline does not grant the active developer directory, so the
`/usr/bin` `xcrun` shims (`python3`, `git`, ...) cannot start on a host where
`xcode-select` points at `Xcode.app`. The baseline grants `/Library`, which
covers a Command Line Tools install, but nothing covers `/Applications`.

This contradicts the documented baseline promise ("so the dynamic linker,
shells, and standard tools work") and breaks the shipped example
`tests/examples/21_mac_python_info.json` on a standard Xcode developer Mac.

### Steps to reproduce
1. `sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer`
2. `./src/target/release/mxc-exec-mac tests/examples/21_mac_python_info.json`

(Equivalent on any host: add `"deniedPaths": ["/Library/Developer"]` to a
config whose command runs `/usr/bin/python3`.)

### Expected behavior
`python3` runs, as the baseline documents for standard tools.

### Actual behavior
```
xcrun: error: unable to load libxcrun (dlopen(.../usr/lib/libxcrun.dylib):
  (file system sandbox blocked open()))
```

theelliotm:
## 📖 Description

Split out of #1125, where the seatbelt technical test suite found it.

`/usr/bin/python3`, `/usr/bin/git` and the other `/usr/bin` stubs get redirected
on Macs with Xcode installed. Instead of routing to the standard
"CommandLineTools" in /Library, they get routed to the installations of python
and git inside of the Xcode .app bundle.

The baseline now includes a read-only `subpath` grant for the active developer
directory, resolved **only** from the root-owned `xcode-select` symlinks
(`/private/var/db/xcode_select_link`, `/private/var/select/developer_dir`).
`DEVELOPER_DIR` is deliberately ignored: it is settable by any caller, so
honoring it would let an environment variable choose what the profile grants.

An Xcode developer directory (`<Xcode.app>/Contents/Developer`) widens to the
enclosing `.app` bundle, because the tools `xcrun` dispatches load sibling
frameworks from `Contents/SharedFrameworks`. Only that exact layout widens; a
Command Line Tools install is granted as-is. `deniedPaths` still overrides the
grant, and a host with no developer tools emits nothing.

## 🔗 References

Resolves #1110

Split out of #1125 — the seatbelt test suite, its CI wiring, and the profile
logging fix stay there.

## 🔍 Validation

- `cargo test -p seatbelt_common` — 111 tests pass
- `cargo fmt --all -- --check` and `cargo clippy -p seatbelt_common --all-targets -- -D warnings` clean
- Unit tests cover the emitted rule (read-only, quote-escaped), the bundle
  widening (beta / versioned / custom install locations, case-insensitive
  extension), the cases that must **not** widen (deeper paths inside a bundle,
  `..`), non-normalized paths, and the absent-tools no-op
- Tests in #1125 pass on both the macOS 15 and macOS 26 runners,
  confirming the grant works on a real Xcode host

## ✅ Checklist

- [x] Linked to an issue
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed) — N/A
- [ ] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes — N/A, `Cargo.lock` unchanged

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1128)

### DIFF
--- a/src/backends/seatbelt/common/src/profile_builder.rs
+++ b/src/backends/seatbelt/common/src/profile_builder.rs
@@ -155,18 +155,74 @@ const DEVELOPER_DIR_LINKS: [&str; 2] = [
 /// Command Line Tools install via its `/Library` grant, but an Xcode-selected
 /// host puts it under `/Applications`, where nothing in the baseline reaches.
 ///
-/// Resolved only from the root-owned `xcode-select` symlinks. The
-/// `DEVELOPER_DIR` override that `xcrun` honors is deliberately ignored: it
-/// is settable by any caller, so consulting it would let the environment
-/// choose what the profile grants.
+/// Resolved only from the `xcode-select` symlinks, and only when root controls
+/// them (see `link_is_root_controlled`). The `DEVELOPER_DIR` override that
+/// `xcrun` honors is deliberately ignored: it is settable by any caller, so
+/// consulting it would let the environment choose what the profile grants.
+///
+/// An untrusted candidate is skipped rather than fatal, so resolution
+/// continues to the legacy link.
 ///
 /// Returns `None` when no developer directory is installed
 fn active_developer_dir() -> Option<PathBuf> {
     DEVELOPER_DIR_LINKS
         .iter()
         .map(Path::new)
+        .filter(|link| link_is_root_controlled(link))
         .filter_map(|link| fs::read_link(link).ok())
         .find(|target| target.is_absolute() && target.is_dir())
+}
+
+/// Whether root alone decides what `link` resolves to.
+///
+/// This grant widens the baseline sandbox, so the ownership the doc comment
+/// above claims is enforced rather than assumed. Two independent facts are
+/// required: the entry is a symlink owned by root, and every directory leading
+/// to it is root-controlled. Neither implies the other — replacing a symlink is
+/// a directory operation, so a root-owned link inside a group-writable
+/// directory is still swappable by an unprivileged process, and a
+/// non-symlink at these paths is not something `xcode-select` produced.
+///
+/// The whole ancestor chain is checked, not just the parent: a writable
+/// directory anywhere along it lets its subtree be replaced wholesale.
+#[cfg(unix)]
+fn link_is_root_controlled(link: &Path) -> bool {
+    use std::os::unix::fs::{MetadataExt as _, PermissionsExt as _};
+
+    let Ok(link_meta) = fs::symlink_metadata(link) else {
+        return false;
+    };
+    if !link_meta.file_type().is_symlink() || link_meta.uid() != 0 {
+        return false;
+    }
+    // `ancestors()` starts at the link itself; the directories follow.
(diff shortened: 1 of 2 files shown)


src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -136,6 +139,87 @@ const SYSTEM_READ_ALLOW: &str = "\
...
+/// is settable by any caller, so consulting it would let the environment
+/// choose what the profile grants.
+///
+/// Returns `None` when no developer directory is installed
+fn active_developer_dir() -> Option<PathBuf> {
+    DEVELOPER_DIR_LINKS
+        .iter()
+        .map(Path::new)
+        .filter_map(|link| fs::read_link(link).ok())
+        .find(|target| target.is_absolute() && target.is_dir())
```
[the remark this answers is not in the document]
richiemsft: Agreed. The check should enforce the complete Unix trust boundary rather than only the link owner: use `symlink_metadata()` to require that the path is actually a symlink and is owned by UID 0, then require its containing directory to be owned by UID 0 with no group/world write bits (`mode & 0o022 == 0`) before calling `read_link()`. Symlink replacement is controlled by the parent directory, so checking only the symlink UID would enforce the documented ownership claim but would not establish that an unprivileged process cannot replace it. An untrusted candidate should be skipped so resolution can continue to the legacy link.

richiemsft (approved):

jsidewhite (approved):
:shipit:

src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -65,6 +67,7 @@ pub fn build_profile_with_proxy(
 
     // Filesystem — read-only system paths every process needs.
     out.push_str(SYSTEM_READ_ALLOW);
+    write_developer_dir_rule(&mut out);
```
jsidewhite, line 70: >write_developer_dir_rule(&mut out); [](http://example.com/codeflow?start=4&length=35)

why can't this be in the form of

`out.push_str(get_developer_dir_rule())`?

which feels better than a function with side_effects

theelliotm: If any of the security checks fail, or if there just isn't a developer directory, I wouldn't want to out.push_str(None). And giving the checks separate functions helps with readability imo

PR #1041, opened 2026-08-25
Fix various bugs CoPilot put into LXC.

dhoehna:
This PR branches from the traditional PR template because of hoe many times the reason for this PR has changed over the past 2 weeks.

tl;dr copilot added bugs, then did its best to work around those bugs.  Additionally the code quality and comment quality is also horrible.

In the Deny-All-Except-Proxy PR, this is when the story starts.  CoPilot, in its vanilla fashion, made the smallest, local change to get LXC to deny-al-except proxy.  The smallest change was "Get the interface, but then take off all capabilities".  This change turned the github pipeline red because MXC would not give the LXC container a network yet LXC needed the network to take away the capabilities.

CoPilots fix?  Add a set to the github pipeline to re-add br_netfilter.  This made the github pipeline green.  However, CoPilot did not change the ADO pipeline.  The ADO pipeline stayed red. Which is how this whole bug started.  Get the ADO pipeline green.

Since two weeks ago CoPIlot and I have substantially improved LXC in so many areas.

1. Accept a 0.7.0 network schema shape inside a 0.8.0 configuration.
2. Correct branch off of the network schema shape to determin is pre0.8.0 network rules will be applied or 0.8.0 network rules.
3. LXC correctly handles the no-network-no-proxy case
4. Substantially improved some E2E tests.
5. Added a new E2E test framework.
6. LXC now REJECTS configurations where the enforcementMode is either empty or capabilitiles.
    a. This breaks nodeSDK tests because the nodeSDK disallows users setting enforcmentMode and the SDK defaults to capabilities.
7. Removed br_netfilter and all code that assumes br_netfilter will be installed.

This is tested with using the full E2E suite that LXC offers.

PR #1146, opened 2026-09-11
Fix legacy LXC SDK enforcement mode

### LINKED ISSUE #1135: SandboxPolicy cannot express network.enforcementMode, so LXC rejects the pre-0.8 configs the SDK builds
## Problem

`SandboxPolicy` — the policy type a developer writes against — has no
`enforcementMode` field (`sdk/node/src/types.ts:413-476`).  The identifier
appears exactly once in that file, at `types.ts:204`, inside `NetworkConfig`,
which is the wire type rather than the caller's.

On schema 0.6 and 0.7 the SDK emits a network block regardless of what the
caller asked for.  When the policy names no network at all,
`createConfigFromPolicy` injects one (`sdk/node/src/sandbox.ts:386-388`):

```ts
config.network = {
    defaultPolicy: 'block',
};
```

It never writes `enforcementMode` — the branch that maps a caller-supplied
network omits it too (`sandbox.ts:378-384`), and the only writes in the file are
on the Windows path at `sandbox.ts:213` and `:215`.  The wire field is
`Option<NetworkEnforcement>` (`src/core/wxc_common/src/wire.rs:373`), so the
omission deserializes to the Rust default, `Capabilities`
(`src/core/wxc_common/src/models.rs:309-313`).  The 0.6 and 0.7 JSON schemas
document the same default (`schemas/stable/mxc-config.schema.0.6.0-alpha.json:123`,
`...0.7.0-alpha.json:133`).

Meanwhile the injected `defaultPolicy` alone is enough to set
`network_mode_specified` (`src/core/wxc_common/src/network_parser.rs:253-257`).

LXC then refuses the config.  That refusal is correct — `capabilities` means
Windows AppContainer capability SIDs, and LXC has no mechanism for them.

## Effect

Observed on #1041, [run 34532170422](https://github.com/microsoft/mxc/actions/runs/34532170422/job/103060722412):
`SDK Integration Tests (linux)` reports `fail 14` — 7 on schema 0.6.0-alpha and
7 on 0.7.0-alpha.  The same 7 each time:

- should execute hello world in LXC container
- should propagate exit code
- should report system info
- should select LXC backend (not Bubblewrap) on explicit containment="lxc"
- should mount readwrite filesystem path
- should mount readonly filesystem path
- should run multi-command pipeline

None of them asks for networking; the three tests that do are skipped in CI
(`linux-process-container.test.ts:117,151,166`).  These fail on an enforcement
mode they never requested.

Schema 0.8 is unaffected, but not because the field was removed —
`schemas/stable/mxc-config.schema.0.8.0-alpha.json:313` still declares it,
merely without a default.  0.8 escapes because a policy naming no network takes
the directional branch, which writes a network section only when `egress` or
`ingress` is present (`sandbox.ts:325-326`).  No network section means
`network_mode_specified` stays false and the guard never fires.

## The workaround is a layering violation

A caller *can* reach the field, by mutating the object `createConfigFromPolicy`
returns and spawning that instead — the function is documented as returning a
config "ready for modification" (`sandbox.ts:262`), and both it and
`spawnSandboxFromConfig` are public exports (`index.ts:69,72`).  The mutation
survives, since the config is serialized wholesale at `helper.ts:290`.

So this is not an impossibility.  It is a hole in the policy-level API that
forces callers past the abstraction the SDK exists to provide, to set a value
they have no way to know they need.

## The SDK already half-knows

`sdk/node/src/helper.ts:124-132` warns that `capabilities` "has no effect on
Linux" and advises "Use the default mode (auto-promotes to 'firewall' for
LXC ...)".

The default mode *is* `capabilities`, so the advice is circular.  The promotion
it describes runs only when the policy carries host rules — `hasHostRules &&
!hasProxy` at `helper.ts:136-138` — which none of these policies do.  And the
warning fires only on an explicit `capabilities`, so in this case nothing is
printed at all.

`sdk/node/src/types.ts:202` separately documents the default as `"both"`, which
matches neither the schema nor the SDK's behavior.

## What would close it

Give the policy surface a way to state firewall enforcement.  Either expose it
on `SandboxPolicy`, or have the Linux path select a mode the way the Windows
path already does explicitly (`sandbox.ts:211-217`).  Which of those is right is
an SDK API decision, not one the LXC backend can make.

Until then the lane stays red on #1041.

bbonaby:
## 📖 Description

Pre-0.8 Node SDK policies always emit a legacy network block. For explicit LXC containment, an omitted `enforcementMode` inherits the Windows-only `capabilities` default, which LXC cannot enforce and now rejects.

Set generated legacy LXC network policies to `firewall` while preserving Bubblewrap and schema 0.8 directional behavior. Add regression coverage for legacy deny-all and outbound policies, plus the schema 0.8 omission case.

## 🔗 References

Resolves #1135

Related to #1041

## 🔍 Validation

- `npm run build` from `sdk/node`
- `npm test` from `sdk/node`

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

---

for the steps.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1146)

### DIFF
--- a/sdk/node/src/helper.ts
+++ b/sdk/node/src/helper.ts
@@ -97,13 +97,14 @@ export function makeLogFilePath(dir: string): string {
  *      unprivileged enforcement path. The proxy applies the host policy
  *      for cooperating HTTP clients; raw-socket clients bypass it.
  *
- * Legacy LXC network blocks use `'firewall'` because capabilities are
- * Windows-only. Bubblewrap uses firewall only for host lists without a proxy.
+ * Legacy LXC network blocks with no explicit enforcement mode use `'firewall'`
+ * because capabilities are Windows-only. Bubblewrap uses firewall only for
+ * host lists without a proxy.
  *
  * If the caller explicitly passes `enforcementMode: 'capabilities'` we
  * warn: `'capabilities'` is a Windows/AppContainer concept (a token
- * capability mask) and has no Linux equivalent — the Linux runner will
- * not enforce anything and the field is silently dropped.
+ * capability mask) and has no Linux equivalent. The explicit value is left
+ * unchanged so native backend validation can report unsupported combinations.
  *
  * Shared between the explicit `'bubblewrap'` / `'lxc'` builders and the
  * abstract `'process'` branch on Linux (which resolves to Bubblewrap
@@ -129,11 +130,12 @@ export function applyLinuxNetworkPolicy(config: ContainerConfig): void {
       "for unprivileged Bubblewrap enforcement)."
     );
   }
-  // LXC cannot enforce Windows AppContainer capabilities, so legacy policies use firewall.
+  // LXC cannot enforce Windows AppContainer capabilities, so omitted legacy modes use firewall.
+  const isLegacyNetwork =
+    config.network.egress === undefined && config.network.ingress === undefined;
   if (
     config.containment === 'lxc' &&
-    config.network.egress === undefined &&
-    config.network.ingress === undefined &&
+    isLegacyNetwork &&
     config.network.enforcementMode === undefined
   ) {
     config.network.enforcementMode = 'firewall';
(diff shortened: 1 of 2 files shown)


jsidewhite (approved):
:shipit:

Issue #873, opened 2026-08-14
Relative process.cwd is resolved against the host process's working directory


caarlos0:
### Relevant area(s)

Linux, macOS, Windows

### Brief description of your issue

A relative `process.cwd` is never validated, so it is resolved against the **host process's** current directory. The sandboxed child then runs in a directory chosen by host state that the policy never granted — silently, with a successful exit.

Nothing in the config pipeline checks the shape of the value:

- `config_parser.rs:802` takes `process.cwd` verbatim (`process.cwd.unwrap_or_default()`).
- `validator.rs` has no rule for `working_directory` at all (it only appears in a test fixture).
- Backends hand it straight to the OS: Seatbelt via `Command::current_dir` (`seatbelt_runner.rs:226`), Bubblewrap via `--chdir` (`bwrap_command.rs:260`), Windows ProcessContainer via `CreateProcessW` (`working_directory.rs:71`).

For an *absolute* path this is correct and matches the documented contract. For a *relative* path, `current_dir` / `--chdir` resolve it against the launching process's cwd, so the effective sandbox cwd depends on where the host binary happened to be invoked from.

This defeats the reason the resolution logic exists in the first place (`core/wxc_common/src/models.rs:771`):

> Backends that must not let the child inherit the host process's cwd use this to fall back to a policy-granted path.

An explicit relative cwd re-introduces exactly that inheritance through the back door.

### Steps to reproduce

On macOS (same shape applies to Bubblewrap on Linux and ProcessContainer on Windows):

1. `mkdir -p /tmp/mxccwd/sub`
2. Build a policy granting only `/tmp/mxccwd` as `readwritePaths`, then:
   ```rust
   let mut request = build_request(&policy, None)?;
   request.set_script("pwd").set_working_directory("sub");
   let output = run(request)?;
   ```
3. Run the resulting binary **from `/tmp/mxccwd`**.

Equivalently via a JSON config with `"process": { "cwd": "sub" }`.

### Expected behavior

A relative `process.cwd` is rejected with a malformed-request / `policy_validation` error naming the value. The sandboxed child's working directory should never depend on the host process's cwd.

### Actual behavior

Observed output from the reproduction above:

```
host_cwd="/private/tmp/mxccwd" requested="sub" -> "/private/tmp/mxccwd/sub" (Exited(0))
```

The child ran in a host-relative directory, exit code 0, no error and no warning. Invoking the same binary from a different directory silently changes the sandbox's working directory.

---

[...]

Issue #872, opened 2026-08-14
WSLc one-shot silently ignores process.cwd when it is not a Windows drive path


caarlos0:
### Relevant area(s)

Windows

### Brief description of your issue

The WSLc **one-shot** runner silently discards `process.cwd` when the value is not a Windows drive path, so the sandboxed process runs in the container's default working directory with no error and no warning.

`wsl_container_runner.rs:1429` maps the cwd through `policy_mapping::windows_path_to_container_path` inside an `if let Some(...)` with **no `else` branch**:

```rust
let _cwd_cstr;
if !request.working_directory.is_empty() {
    if let Some(container_cwd) =
        policy_mapping::windows_path_to_container_path(&request.working_directory)
    {
        // ... WslcSetProcessSettingsWorkingDirectory ...
    }
    // no else: an untranslatable cwd is dropped on the floor
}
```

`windows_path_to_container_path` returns `None` for anything that is not `<drive>:` followed by `\`, `/`, or end-of-string (`backends/wslc/common/src/policy_mapping.rs:49`). An absolute in-container path such as `/workspace` therefore fails translation and is dropped.

This contradicts the documented contract in [`docs/schema.md`](../blob/main/docs/schema.md) ("Working Directory"):

> `process.cwd` is optional. When it is set, it is passed to the backend verbatim — an unusable value fails the launch rather than being silently replaced.

It also contradicts the WSLc **state-aware** surface, which handles the same field in the opposite direction and explicitly documents the behavior this code violates (`backends/wslc/common/src/container_steps.rs:422`):

```rust
// Working directory (an absolute in-container path, e.g. `/work`; empty =
// container default). It is passed straight to the SDK; a non-absolute
// value is rejected rather than silently ignored.
```

The two WSLc surfaces accept disjoint path shapes for the same config field:

| `process.cwd` | WSLc one-shot | WSLc state-aware |
|---|---|---|
| `/workspace` | silently ignored → wrong cwd | accepted |
| `C:\work` | translated to a container path | rejected (`must be an absolute in-container path`) |

Reachable from the Rust SDK (`mxc-sdk`), which supports WSLc streaming:

```rust
let mut req = mxc_sdk::build_request_with_containment(&policy, &Containment::Wslc, None)?;
req.set_experimental(true);
req.set_script("pwd").set_working_directory("/workspace");
let out = mxc_sdk::run(req)?;
```

[...]

PR #1147, opened 2026-09-12
fix(cwd): reject a relative process.cwd on schema 0.9 and stop WSLc silently dropping it

### LINKED ISSUE #902: [Cross-backend] process.cwd is silently discarded or resolved against host state
Both defects in how `process.cwd` is handled: one drops the value on WSLc one-shot, the other resolves a relative path against host state across every backend. Both fail silently.

This consolidates 2 separately filed issues so the backlog carries one entry per area instead of one per finding. Every original report is reproduced below in full, unedited. The originals are closed and point here.

| Folded from | Status | Title |
| --- | --- | --- |
| #872 | Open | WSLc one-shot silently ignores process.cwd when it is not a Windows drive path |
| #873 | Open | Relative process.cwd is resolved against the host process's working directory |

---

## WSLc one-shot silently ignores process.cwd when it is not a Windows drive path

*Originally #872.*

### Relevant area(s)

Windows

The WSLc **one-shot** runner silently discards `process.cwd` when the value is not a Windows drive path, so the sandboxed process runs in the container's default working directory with no error and no warning.

`wsl_container_runner.rs:1429` maps the cwd through `policy_mapping::windows_path_to_container_path` inside an `if let Some(...)` with **no `else` branch**:

```rust
let _cwd_cstr;
if !request.working_directory.is_empty() {
    if let Some(container_cwd) =
        policy_mapping::windows_path_to_container_path(&request.working_directory)
    {
        // ... WslcSetProcessSettingsWorkingDirectory ...
    }
    // no else: an untranslatable cwd is dropped on the floor
}
```

`windows_path_to_container_path` returns `None` for anything that is not `<drive>:` followed by `\`, `/`, or end-of-string (`backends/wslc/common/src/policy_mapping.rs:49`). An absolute in-container path such as `/workspace` therefore fails translation and is dropped.

This contradicts the documented contract in [`docs/schema.md`](../blob/main/docs/schema.md) ("Working Directory"):

> `process.cwd` is optional. When it is set, it is passed to the backend verbatim — an unusable value fails the launch rather than being silently replaced.

It also contradicts the WSLc **state-aware** surface, which handles the same field in the opposite direction and explicitly documents the behavior this code violates (`backends/wslc/common/src/container_steps.rs:422`):

```rust
// Working directory (an absolute in-container path, e.g. `/work`; empty =
// container default). It is passed straight to the SDK; a non-absolute
// value is rejected rather than silently ignored.
```

The two WSLc surfaces accept disjoint path shapes for the same config field:

| `process.cwd` | WSLc one-shot | WSLc state-aware |
|---|---|---|
| `/workspace` | silently ignored → wrong cwd | accepted |
| `C:\work` | translated to a container path | rejected (`must be an absolute in-container path`) |

Reachable from the Rust SDK (`mxc-sdk`), which supports WSLc streaming:

```rust
let mut req = mxc_sdk::build_request_with_containment(&policy, &Containment::Wslc, None)?;
req.set_experimental(true);
req.set_script("pwd").set_working_directory("/workspace");
let out = mxc_sdk::run(req)?;
```

...and equally from a JSON config with `"process": { "cwd": "/workspace" }` on the one-shot WSLc path.

### Steps to reproduce

1. On a Windows host with WSL2 + `wslcsdk.dll`, build with `--features wslc`.
2. Run a one-shot WSLc execution with `process.commandLine = "pwd"` and `process.cwd = "/workspace"` (via `wxc-exec --experimental` with a JSON config, or via `mxc_sdk::build_request_with_containment(..., &Containment::Wslc, ...)` + `set_working_directory("/workspace")`).
3. Observe the printed working directory.

### Expected behavior

Either the cwd is honored, or the launch fails with a `policy_validation` / malformed-request error naming the unusable value — matching the documented "an unusable value fails the launch rather than being silently replaced" and the state-aware surface's explicit "rejected rather than silently ignored".

### Actual behavior

The cwd is silently dropped and the process runs in the container's default working directory. No error, no warning, nothing in the logs.

---

Two things to decide when fixing:

1. The minimal fix is to turn the `if let` into a `match` with an `Err(...)` arm so an untranslatable cwd fails the launch.
2. Whether one-shot and state-aware WSLc should agree on the accepted path shape (Windows host path vs. absolute in-container path) is a separate design call — one-shot maps volumes from Windows paths, state-aware does not. Worth resolving deliberately rather than as part of the fail-loudly fix.

---

## Relative process.cwd is resolved against the host process's working directory

*Originally #873.*

### Relevant area(s)

Linux, macOS, Windows

A relative `process.cwd` is never validated, so it is resolved against the **host process's** current directory. The sandboxed child then runs in a directory chosen by host state that the policy never granted — silently, with a successful exit.

Nothing in the config pipeline checks the shape of the value:

- `config_parser.rs:802` takes `process.cwd` verbatim (`process.cwd.unwrap_or_default()`).
- `validator.rs` has no rule for `working_directory` at all (it only appears in a test fixture).
- Backends hand it straight to the OS: Seatbelt via `Command::current_dir` (`seatbelt_runner.rs:226`), Bubblewrap via `--chdir` (`bwrap_command.rs:260`), Windows ProcessContainer via `CreateProcessW` (`working_directory.rs:71`).

For an *absolute* path this is correct and matches the documented contract. For a *relative* path, `current_dir` / `--chdir` resolve it against the launching process's cwd, so the effective sandbox cwd depends on where the host binary happened to be invoked from.

This defeats the reason the resolution logic exists in the first place (`core/wxc_common/src/models.rs:771`):

> Backends that must not let the child inherit the host process's cwd use this to fall back to a policy-granted path.

An explicit relative cwd re-introduces exactly that inheritance through the back door.

### Steps to reproduce

On macOS (same shape applies to Bubblewrap on Linux and ProcessContainer on Windows):

1. `mkdir -p /tmp/mxccwd/sub`
2. Build a policy granting only `/tmp/mxccwd` as `readwritePaths`, then:
   ```rust
   let mut request = build_request(&policy, None)?;
   request.set_script("pwd").set_working_directory("sub");
   let output = run(request)?;
   ```
3. Run the resulting binary **from `/tmp/mxccwd`**.

Equivalently via a JSON config with `"process": { "cwd": "sub" }`.

### Expected behavior

A relative `process.cwd` is rejected with a malformed-request / `policy_validation` error naming the value. The sandboxed child's working directory should never depend on the host process's cwd.

### Actual behavior

Observed output from the reproduction above:

```
host_cwd="/private/tmp/mxccwd" requested="sub" -> "/private/tmp/mxccwd/sub" (Exited(0))
```

The child ran in a host-relative directory, exit code 0, no error and no warning. Invoking the same binary from a different directory silently changes the sandbox's working directory.

---

Note for whoever picks this up: the check is **not** a shared one-liner. "Absolute" is backend-relative — `C:\x` is not absolute per `Path::is_absolute` on Unix, and `/x` is a valid in-container path for WSLc but not a valid Windows host path. The rule likely belongs at each backend's resolution point (or as a per-backend predicate passed into the shared helper), not as a single check in `validator.rs`.

Related: #872 (WSLc one-shot silently drops an untranslatable `process.cwd`) — same field, same class of "wrong working directory, no diagnostic" failure, but a separate root cause and fix.

---

theelliotm:
## 📖 Description

A `process.cwd` the target cannot resolve fails silently today, in two ways:

- **A relative `cwd` resolves against the host process's working directory.** Nothing required it to be absolute, so `"cwd": "work"` lands wherever `wxc-exec` was started from — the result depends on the caller's shell, not on the policy.
- **WSLc one-shot dropped a `cwd` it could not translate**, falling through an `if let Some(..)` with no `else` and launching in the container's default directory.

**The WSLc fix is not version-gated.** Silently ignoring a requested `cwd` is wrong on every schema version, so an untranslatable path is now rejected on all of them. Only the relative-path rule below is gated on `0.9.0-alpha`.

Absoluteness belongs to whatever *consumes* the path, so `ContainmentBackend` declares a `WorkingDirectoryStyle` (`Windows`/`Unix`) per `WorkingDirectoryScope` (`OneShot`/`Exec`):

| Style | Backends | Accepts |
|----------|----------|----------|
| Windows | ProcessContainer, Windows Sandbox, IsolationSession, WSLc one-shot | drive-rooted and UNC/device paths       |
| Unix   | Seatbelt, LXC, Bubblewrap, WSLc exec, MicroVM, Hyperlight, VM| a leading / only|

Enforcement sits in `validate_common` and `validate_exec_common`, the choke points every backend passes through. Unlike a check in `config_parser`, they also catch a `cwd` injected after parsing via `SandboxRequest::set_working_directory` — the repro in #873.

Notes:
- The relative-path rule applies at schema >= 0.9; 0.6-0.8 and absent/malformed versions are unchanged.
- The exact supplied string is validated, since that is what the backends receive - " /tmp" is relative.
- Rejections carry FailurePhase::Rejected, so one-shot and state-aware callers both see policy_validation .
- Avoids Path::is_absolute, which answers for the compiled host. Windows rejects drive-relative paths (C:foo, \foo).
- No existing test or example config used a relative cwd.
## 🔗 References

Resolves #902 in schema version 0.9+ — both folded defects, #872 (WSLc drop) and #873 (host-relative resolution), are fixed here.

## 🔍 Validation

- `cargo test -p wxc_common --lib validator` — 23 passed, covering every backend, each style, cross-style rejection, the WSLc per-phase split, both sides of the version gate, an omitted `cwd`, and the state-aware `policy_validation` mapping.
- `wxc_common`, `mxc_engine`, `wslc_common`, `nanvix_runner` — no new failures; pre-existing sandbox-related failures confirmed identical against a stashed baseline.
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check` — clean.
- All four codegen/version gates and `validate-configs.js` (349 configs) pass.

The WSLc rejection is not unit-testable; it sits in the `unsafe` SDK-backed launch path.

Docs: the 0.9 rule lives in the `cwd` doc comments that generate the schema descriptions (`wire.rs`, `mxc_config_contract/src/dev/stable.rs`), with all four artifacts regenerated. The WSLc behavior change is documented in `docs/wsl/wsl-container-getting-started.md`.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1147)

### DIFF
--- a/src/backends/hyperlight/common/src/lib.rs
+++ b/src/backends/hyperlight/common/src/lib.rs
@@ -893,7 +893,7 @@ mod tests {
         let mut r = runner();
         let request = ExecutionRequest {
             script_code: "print('x')".to_string(),
-            working_directory: "C:/tmp".to_string(),
+            working_directory: "/tmp".to_string(),
             ..Default::default()
         };
         let mut logger = Logger::new(Mode::Buffer);
--- a/src/backends/wslc/common/src/wsl_container_runner.rs
+++ b/src/backends/wslc/common/src/wsl_container_runner.rs
@@ -681,6 +681,7 @@ impl ScriptRunner for WSLContainerRunner {
             .into_response());
         }
         policy::reject_unsupported_enforcement_mode(request).map_err(as_wslc_rejection)?;
+        reject_untranslatable_working_directory(request)?;
         // The shared validator returns an untagged response; retag it so its
         // rejections reach SDK callers as `policy_validation` like the checks above.
         validate_network_policy_support(request, NetworkPolicySupport::LEGACY)
@@ -697,6 +698,30 @@ impl ScriptRunner for WSLContainerRunner {
 /// `policy_validation` with the same message the state-aware surface emits.
 fn as_wslc_rejection(err: MxcError) -> ScriptResponse {
     WslcError::Rejected(err.message).into_response()
+}
+
+/// One-shot reads `process.cwd` as a Windows host path and maps it to its
+/// in-container mount point, so a value without a drive letter (a UNC path,
+/// say) has no equivalent inside the container.
+///
+/// Checked here so dry-run reports it and no SDK, session, or image work
+/// happens first; [`WSLContainerRunner::start_container`] repeats it because it
+/// owns the translated value.
+fn reject_untranslatable_working_directory(
+    request: &ExecutionRequest,
+) -> Result<(), ScriptResponse> {
+    if request.working_directory.is_empty()
+        || policy_mapping::windows_path_to_container_path(&request.working_directory).is_some()
+    {
+        return Ok(());
+    }
+
+    Err(WslcError::Rejected(format!(
+        "WSLc: process.cwd must be a Windows drive path that maps into the container \
+         (e.g. C:\\workspace -> /mnt/c/workspace), got {:?}",
+        request.working_directory
+    ))
+    .into_response())
 }
 
 /// The first line [`WSLContainerRunner::start_container`] writes, before any
@@ -1509,9 +1534,8 @@ impl WSLContainerRunner {
 
         let _cwd_cstr;
         if !request.working_directory.is_empty() {
-            // One-shot takes a Windows host path and translates it to its
-            // in-container mount point. An untranslatable value fails the
-            // launch rather than being dropped.
+            // Backstop for `reject_untranslatable_working_directory`, which
+            // already ran in `validate_runner`.
             let Some(container_cwd) =
                 policy_mapping::windows_path_to_container_path(&request.working_directory)
             else {
@@ -2554,6 +2578,41 @@ mod tests {
             err.failure_phase,
             wxc_common::models::FailurePhase::Rejected
         );
+    }
+
+    #[test]
+    fn validate_runner_rejects_a_cwd_that_does_not_map_into_the_container() {
+        let runner = WSLContainerRunner::new(&WslcConfig::default());
+        for cwd in ["\\\\server\\share", "/mnt/c/workspace", "relative"] {
+            let request = ExecutionRequest {
+                containment: wxc_common::models::ContainmentBackend::Wslc,
+                working_directory: cwd.to_string(),
+                ..Default::default()
+            };
+            let err = runner
+                .validate_runner(&request)
+                .expect_err(&format!("accepted '{cwd}'"));
+            assert!(
+                err.error_message.contains("maps into the container"),
+                "got: {}",
+                err.error_message
+            );
+            assert_eq!(
+                err.failure_phase,
+                wxc_common::models::FailurePhase::Rejected
+            );
+        }
+    }
+
+    #[test]
+    fn validate_runner_accepts_a_drive_rooted_cwd() {
+        let runner = WSLContainerRunner::new(&WslcConfig::default());
+        let request = ExecutionRequest {
+            containment: wxc_common::models::ContainmentBackend::Wslc,
+            working_directory: "C:\\workspace".to_string(),
+            ..Default::default()
+        };
+        assert!(runner.validate_runner(&request).is_ok());
     }
 
     #[test]
--- a/src/backends/wslc/common/src/wsl_container_runner.rs
+++ b/src/backends/wslc/common/src/wsl_container_runner.rs
@@ -1509,21 +1509,30 @@ impl WSLContainerRunner {
 
         let _cwd_cstr;
         if !request.working_directory.is_empty() {
-            if let Some(container_cwd) =
+            // One-shot takes a Windows host path and translates it to its
+            // in-container mount point. An untranslatable value fails the
+            // launch rather than being dropped.
+            let Some(container_cwd) =
                 policy_mapping::windows_path_to_container_path(&request.working_directory)
-            {
-                _cwd_cstr = format!("{}\0", container_cwd);
-                let hr = sdk.WslcSetProcessSettingsWorkingDirectory(
-                    &mut process_settings,
-                    _cwd_cstr.as_bytes().as_ptr() as PCSTR,
-                );
-                if hr != S_OK {
-                    return Err(sdk_error(
-                        "WslcSetProcessSettingsWorkingDirectory failed",
-                        hr,
-                        "",
-                    ));
-                }
+            else {
+                return Err(WslcError::Rejected(format!(
+                    "process.cwd must be a Windows drive path that maps into the container \
+                     (e.g. C:\\workspace -> /mnt/c/workspace), got {:?}",
+                    request.working_directory
+                ))
+                .into_response());
+            };
+            _cwd_cstr = format!("{}\0", container_cwd);
+            let hr = sdk.WslcSetProcessSettingsWorkingDirectory(
+                &mut process_settings,
+                _cwd_cstr.as_bytes().as_ptr() as PCSTR,
+            );
+            if hr != S_OK {
+                return Err(sdk_error(
+                    "WslcSetProcessSettingsWorkingDirectory failed",
+                    hr,
+                    "",
+                ));
             }
         }
 
--- a/src/core/wxc_common/src/models.rs
+++ b/src/core/wxc_common/src/models.rs
@@ -84,6 +84,93 @@ impl ContainmentBackend {
             | ContainmentBackend::Vm => None,
         }
     }
+
+    /// Path shape an explicit `process.cwd` must have for this backend on
+    /// `scope`.
+    ///
+    /// The shape follows the *target* the path is handed to, not the host MXC
+    /// runs on. Those coincide everywhere except one-shot vs. state-aware WSLc.
+    pub fn working_directory_style(&self, scope: WorkingDirectoryScope) -> WorkingDirectoryStyle {
+        match self {
+            ContainmentBackend::ProcessContainer
+            | ContainmentBackend::WindowsSandbox
+            | ContainmentBackend::IsolationSession => WorkingDirectoryStyle::Windows,
+            // NanVix and Hyperlight reject any working directory today; both
+            // run a POSIX-style guest, so a future implementation inherits the
+            // right shape rather than an exemption. `Vm` is non-Windows-only —
+            // a Windows host resolves that intent to `WindowsSandbox`.
+            ContainmentBackend::Lxc
+            | ContainmentBackend::Bubblewrap
+            | ContainmentBackend::Seatbelt
+            | ContainmentBackend::MicroVm
+            | ContainmentBackend::Hyperlight
+            | ContainmentBackend::Vm => WorkingDirectoryStyle::Unix,
+            // One-shot WSLc takes a Windows *host* path and translates it into
+            // the container; state-aware exec takes the in-container path.
+            ContainmentBackend::Wslc => match scope {
+                WorkingDirectoryScope::OneShot => WorkingDirectoryStyle::Windows,
+                WorkingDirectoryScope::Exec => WorkingDirectoryStyle::Unix,
+            },
+        }
+    }
+}
+
+/// The lifecycle surface a `process.cwd` is bound for. A backend can read the
+/// same field against a different target on each.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum WorkingDirectoryScope {
+    /// A one-shot run.
+    OneShot,
+    /// A state-aware `exec` against an already-provisioned sandbox.
+    Exec,
+}
+
+/// The path shape a backend treats as absolute for `process.cwd`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum WorkingDirectoryStyle {
+    /// Windows paths: `C:\dir`, `C:/dir`, or a UNC/device path.
+    Windows,
+    /// POSIX paths: `/dir`, plus the home-anchored `~` forms the Seatbelt
+    /// profile expands.
+    Unix,
+}
+
+impl WorkingDirectoryStyle {
+    /// Whether `path` is absolute in this style — i.e. it cannot resolve
+    /// against the launching process's working directory.
+    pub fn is_absolute(self, path: &str) -> bool {
+        match self {
+            WorkingDirectoryStyle::Windows => is_windows_absolute(path),
+            WorkingDirectoryStyle::Unix => is_unix_absolute(path),
+        }
+    }
+
+    /// An absolute path in this style, for error messages.
+    pub fn example(self) -> &'static str {
+        match self {
+            WorkingDirectoryStyle::Windows => "C:\\workspace",
+            WorkingDirectoryStyle::Unix => "/workspace",
+        }
+    }
+}
+
+/// These deliberately do not use `std::path::Path::is_absolute`, which answers
+/// for the *host* MXC was compiled for rather than for the target backend.
+fn is_windows_absolute(path: &str) -> bool {
+    let bytes = path.as_bytes();
+    let is_sep = |b: u8| b == b'\\' || b == b'/';
+    // UNC and device paths (`\\server\share`, `\\?\C:\dir`).
+    if bytes.len() >= 2 && is_sep(bytes[0]) && is_sep(bytes[1]) {
+        return true;
+    }
+    // A drive letter is only absolute when it is also rooted: `C:dir` is
+    // relative to that drive's current directory, and `\dir` to its current
+    // drive.
+    bytes.len() >= 3 && bytes[0].is_ascii_alphabetic() && bytes[1] == b':' && is_sep(bytes[2])
+}
+
+fn is_unix_absolute(path: &str) -> bool {
+    path.starts_with('/') || path == "~" || path.starts_with("~/")
 }
 
 impl From<crate::wire::Containment> for ContainmentBackend {
--- a/src/core/wxc_common/src/models.rs
+++ b/src/core/wxc_common/src/models.rs
@@ -101,10 +101,11 @@ impl ContainmentBackend {
             // a Windows host resolves that intent to `WindowsSandbox`.
             ContainmentBackend::Lxc
             | ContainmentBackend::Bubblewrap
-            | ContainmentBackend::Seatbelt
             | ContainmentBackend::MicroVm
             | ContainmentBackend::Hyperlight
             | ContainmentBackend::Vm => WorkingDirectoryStyle::Unix,
+            // Seatbelt is the only backend that expands `~` itself.
+            ContainmentBackend::Seatbelt => WorkingDirectoryStyle::MacOs,
             // One-shot WSLc takes a Windows *host* path and translates it into
             // the container; state-aware exec takes the in-container path.
             ContainmentBackend::Wslc => match scope {
@@ -130,9 +131,12 @@ pub enum WorkingDirectoryScope {
 pub enum WorkingDirectoryStyle {
     /// Windows paths: `C:\dir`, `C:/dir`, or a UNC/device path.
     Windows,
-    /// POSIX paths: `/dir`, plus the home-anchored `~` forms the Seatbelt
-    /// profile expands.
+    /// POSIX paths: `/dir` only. LXC and Bubblewrap hand the value to
+    /// `cd -- "$1"` and `--chdir` respectively, neither of which expands `~`.
     Unix,
+    /// POSIX paths plus the home-anchored `~` forms, which the Seatbelt
+    /// profile expands before use.
+    MacOs,
 }
 
 impl WorkingDirectoryStyle {
@@ -142,6 +146,9 @@ impl WorkingDirectoryStyle {
         match self {
             WorkingDirectoryStyle::Windows => is_windows_absolute(path),
             WorkingDirectoryStyle::Unix => is_unix_absolute(path),
+            WorkingDirectoryStyle::MacOs => {
+                is_unix_absolute(path) || path == "~" || path.starts_with("~/")
+            }
         }
     }
 
@@ -149,7 +156,7 @@ impl WorkingDirectoryStyle {
     pub fn example(self) -> &'static str {
         match self {
             WorkingDirectoryStyle::Windows => "C:\\workspace",
-            WorkingDirectoryStyle::Unix => "/workspace",
+            WorkingDirectoryStyle::Unix | WorkingDirectoryStyle::MacOs => "/workspace",
         }
     }
 }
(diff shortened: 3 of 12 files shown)


src/core/wxc_common/src/wire.rs
```diff
@@ -168,6 +168,14 @@ pub struct Process {
...
     /// fallback; LXC/WSL use the container root; NanVix and Hyperlight reject a
     /// working directory outright. See `docs/schema.md` ("Working Directory").
+    ///
+    /// From schema 0.9 on, a supplied value must be absolute for the target the
+    /// path reaches — `C:\workspace` or a UNC path for the Windows backends,
+    /// `/workspace` (or a `~` path on Seatbelt) for the Unix ones. WSL
+    /// Container reads it as a Windows host path one-shot and as an
+    /// in-container path on a state-aware `exec`. A relative path is rejected
+    /// because it would resolve against the launching process's working
+    /// directory.
```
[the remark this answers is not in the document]
theelliotm: Will only touch schema.md when 0.9 releases

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -699,6 +700,30 @@ fn as_wslc_rejection(err: MxcError) -> ScriptResponse {
...
+///
+/// Checked here so dry-run reports it and no SDK, session, or image work
+/// happens first; [`WSLContainerRunner::start_container`] repeats it because it
+/// owns the translated value.
+fn reject_untranslatable_working_directory(
+    request: &ExecutionRequest,
+) -> Result<(), ScriptResponse> {
+    if request.working_directory.is_empty()
+        || policy_mapping::windows_path_to_container_path(&request.working_directory).is_some()
+    {
```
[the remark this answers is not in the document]
theelliotm: out of scope

SohamDas2021:
This PR is marked *Resolves #902*, but the repro added for the issue is written against `0.8.0-alpha`, which the new gate exempts. It's designed to fail until #902 is fixed and after this change it still fails, because on the schema the issue reported the cwd is still resolved against host state. 

Shall we bump the repro to `0.9.0-alpha` so it flips to PASS and locks the fix in?

src/core/wxc_common/src/models.rs
```diff
@@ -84,6 +84,103 @@ impl ContainmentBackend {
...
+
+impl WorkingDirectoryStyle {
+    /// Whether `path` is absolute in this style — i.e. it cannot resolve
+    /// against the launching process's working directory.
+    pub fn is_absolute(self, path: &str) -> bool {
+        match self {
+            WorkingDirectoryStyle::Windows => is_windows_absolute(path),
+            WorkingDirectoryStyle::Unix => is_unix_absolute(path),
+            WorkingDirectoryStyle::MacOs => {
+                is_unix_absolute(path) || path == "~" || path.starts_with("~/")
```
SohamDas2021, line 153: Blessing `~` as absolute doesn't hold end-to-end. `seatbelt_runner.rs:747` is `expand_tilde(path).unwrap_or_else(|_| path.to_string())` - it swallows the error when `HOME` is unset and passes a literal `~` to `command.current_dir` (`:244`), which resolves against the launching process. 

Either drop `~` here or require expansion to a `/`-rooted path?

SohamDas2021:
The 0.9 absolute-`cwd` rule now ships in `wire.rs`, both 0.9 schema JSONs, both `wire.ts` files and `stable.rs`, but the doc they point at still says `cwd` is passed to the backend verbatim? 

Can you please do a scan of the docs and update if required? 
docs/schema.md, docs/seatbelt/seatbelt-backend.md, anywhere else?

src/core/wxc_common/src/validator.rs
```diff
@@ -314,6 +368,200 @@ mod tests {
...
+            script_code: "echo hi".to_string(),
+            schema_version: version.to_string(),
+            containment,
+            working_directory: cwd.to_string(),
+            ..Default::default()
+        }
+    }
+
+    #[test]
+    fn rejects_relative_cwd_on_schema_0_9_for_every_backend() {
```
SohamDas2021, line 386: This only covers values invalid under *all* styles. The positive matrix (`:456-473`) tests ProcessContainer, Seatbelt, Bubblewrap and WSLC. WindowsSandbox, IsolationSession, Lxc, MicroVm, Hyperlight and Vm never get a style-specific valid path, so a wrong entry in the new `working_directory_style` table wouldn't fail anything?

theelliotm: Made the test comprehensive.

src/core/wxc_common/src/models.rs
theelliotm: Mac no longer will accept ~

theelliotm:
> This PR is marked _Resolves #902_, but the repro added for the issue is written against `0.8.0-alpha`, which the new gate exempts. It's designed to fail until #902 is fixed and after this change it still fails, because on the schema the issue reported the cwd is still resolved against host state.
> 
> Shall we bump the repro to `0.9.0-alpha` so it flips to PASS and locks the fix in?

Tested both repros locally in 0.9 and they work flawlessly.

theelliotm:
> The 0.9 absolute-`cwd` rule now ships in `wire.rs`, both 0.9 schema JSONs, both `wire.ts` files and `stable.rs`, but the doc they point at still says `cwd` is passed to the backend verbatim?
> 
> Can you please do a scan of the docs and update if required? docs/schema.md, docs/seatbelt/seatbelt-backend.md, anywhere else?

Updated a ton of docs. I hope it covers any gaps.

SohamDas2021 (dismissed):

src/core/wxc_common/src/validator.rs
```diff
@@ -184,12 +187,66 @@ pub fn validate_state_aware_network_policy_support(
...
+    ))
+}
+
+/// A malformed version is left to the schema-version validator, which owns that
+/// diagnostic.
+fn schema_requires_absolute_cwd(version: &str) -> bool {
+    semver::Version::parse(version)
+        .is_ok_and(|parsed| (parsed.major, parsed.minor) >= ABSOLUTE_CWD_MIN_SCHEMA)
+}
+
```
jsidewhite, line 230: This isn't right.  validator.rs should not parse the version post-facto (whatever facto means) and then rebuild rules from the value.

The ExecutionRequest should already be correct.

Please have the 0.9.0 contract parser enforce the correct validation.

jsidewhite (dismissed):
:clock1:

src/core/wxc_common/src/validator.rs
theelliotm: I believe implementing cwd rejection in validator.rs is the best option. validator.rs is the only place that knows the schema version, the final selected backend, the `cwd` string, and is run by every entrypoint through `validate_common` or `validate_exec_common`. The 0.9 contract parser currently cannot enforce this, because the Rust SDK never routes a `cwd` through the contract at all. In the SDK, `cwd` is set to Default when parsed and changed via `set_working_directory` after parsing. I agree that reconstructing the version tuple isn't a great implementation, but 0.8's directional networking also does it in the same file.

jsidewhite (dismissed):
Will address the validator.rs changes in future.

src/core/wxc_common/src/models.rs
```diff
@@ -84,6 +84,90 @@ impl ContainmentBackend {
...
+
+impl WorkingDirectoryStyle {
+    /// Whether `path` is absolute in this style — i.e. it cannot resolve
+    /// against the launching process's working directory.
+    pub fn is_absolute(self, path: &str) -> bool {
+        match self {
+            WorkingDirectoryStyle::Windows => is_windows_absolute(path),
+            WorkingDirectoryStyle::Unix => path.starts_with('/'),
+        }
+    }
```
jsidewhite, line 146: Can't this be is_absolute()?

[https://doc.rust-lang.org/std/path/struct.Path.html#method.is_absolute](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_absolute)

src/core/wxc_common/src/models.rs
```diff
@@ -84,6 +84,90 @@ impl ContainmentBackend {
...
+    let is_sep = |b: u8| b == b'\\' || b == b'/';
+    // UNC and device paths (`\\server\share`, `\\?\C:\dir`).
+    if bytes.len() >= 2 && is_sep(bytes[0]) && is_sep(bytes[1]) {
+        return true;
+    }
+    // A drive letter is only absolute when it is also rooted: `C:dir` is
+    // relative to that drive's current directory, and `\dir` to its current
+    // drive.
+    bytes.len() >= 3 && bytes[0].is_ascii_alphabetic() && bytes[1] == b':' && is_sep(bytes[2])
 }
```
jsidewhite, line 171: can we delete?

jsidewhite (dismissed):
:shipit:

src/core/wxc_common/src/models.rs
theelliotm: WSLc wouldn't like it.

jsidewhite (dismissed):
:shipit:

jsidewhite (dismissed):
:shipit:

jsidewhite (dismissed):

src/core/wxc_common/src/validator.rs
```diff
@@ -184,12 +187,66 @@ pub fn validate_state_aware_network_policy_support(
...
     }
 
+    // `Rejected` so the SDK surfaces this as `policy_validation`, matching what
+    // state-aware `exec` returns for the same cwd. Only `error_message` is set:
+    // the CLI prints `standard_err` and then the error envelope, so populating
+    // both would report the same rejection twice.
+    validate_working_directory(request, WorkingDirectoryScope::OneShot).map_err(|message| {
+        ScriptResponse {
+            exit_code: -1,
+            failure_phase: FailurePhase::Rejected,
```
MGudgin, line 244: **Medium (correctness) - The one-shot error code does not match the PR's `policy_validation` claim.**

**Attribution:** `newly_exposed_by_change` — this new rejection sets `FailurePhase::Rejected`, but the byte-identical executor serializer in `script_runner.rs` ignores the phase and hardcodes `backend_error`.

Native streaming/state-aware dispatch maps `Rejected` to `policy_validation`, while `wxc-exec`/`lxc-exec`/`mxc-exec-mac` emit `backend_error`; executor-backed Node `spawnSandboxAsync` parses and forwards that code. The same caller-fixable rejection therefore has different public classifications.

**Fix:** Make `emit_backend_error_envelope` map `FailurePhase::Rejected` to `policy_validation`, and add CLI plus Node boundary tests for relative cwd and pre-0.9 WSLc translation rejection.

src/backends/wslc/common/src/wsl_container_runner.rs
```diff
@@ -1512,21 +1537,29 @@ impl WSLContainerRunner {
...
-                );
-                if hr != S_OK {
-                    return Err(sdk_error(
-                        "WslcSetProcessSettingsWorkingDirectory failed",
-                        hr,
-                        "",
-                    ));
-                }
+            else {
+                return Err(WslcError::Rejected(format!(
```
MGudgin: **Medium (testability/maintainability) - Consolidate the duplicated WSLc cwd translation.**

**Attribution:** `introduced_by_change` — this backstop and `reject_untranslatable_working_directory` both add the same conversion predicate and nearly identical rejection text.

Normal callers cannot reach this branch because validation rejects first, yet this copy runs only after COM/SDK initialization and session/image setup, making the changed defensive behavior impractical to test without the host SDK. The two copies can also drift in message or accepted forms.

**Fix:** Resolve cwd once in a pure helper before SDK initialization, then carry the prepared container path into `start_container`; test that helper and the pre-SDK rejection ordering.

docs/schema.md
```diff
@@ -269,7 +267,21 @@ that can be executed independently.
...
+replaced.
+
+**From schema `0.9.0-alpha` on, a set value must be absolute**, since a relative
+path would resolve against the launching process's working directory. Absolute
+means absolute for the target the path reaches, not for the host MXC runs on:
+
+| Backend | Absolute form |
+|---------|---------------|
+| Windows ProcessContainer / Windows Sandbox / IsolationSession | `C:\workspace`, `C:/workspace`, or a UNC path. `C:workspace` and `\workspace` are relative. |
+| WSL Container | One-shot: the Windows host path (`C:\workspace`), which the backend translates. State-aware `exec`: the in-container path (`/workspace`). |
```
MGudgin, line 282: **Medium (documentation) - This contradicts the section's “passed to the backend verbatim” statement.**

**Attribution:** `introduced_by_change` — the new row explicitly says WSLc translates the Windows host path, immediately after the section says a set cwd is passed verbatim.

**Fix:** Qualify the earlier sentence with the WSLc one-shot translation exception.

sdk/node/README.md
```diff
@@ -512,9 +512,9 @@ You must set `config.process!.commandLine = '…'` before calling `spawnSandboxF
 
 No `network` field → no network. No `readwritePaths` → process can't write `%TEMP%`. No `ui` → no GUI. Use the discovery helpers to compose a sensible baseline.
 
-### `process.cwd` doesn't grant filesystem access
+### `process.cwd`
 
-Setting `cwd` (or the `workingDirectory` argument) does **not** add that path to the policy. Add it to `readonlyPaths` / `readwritePaths` explicitly.
+Setting `cwd` (or the `workingDirectory` argument) does **not** add that path to the policy. Add it to `readonlyPaths` / `readwritePaths` explicitly. On `0.9.0-alpha` it must also be absolute — `C:\workspace` for the Windows backends, `/workspace` for the Unix ones.
```
MGudgin: **Medium (documentation) - The Windows/Unix shorthand omits WSLc's phase-dependent path style.**

**Attribution:** `introduced_by_change` — this new guidance can lead a Node user to supply `/workspace` to WSLc one-shot, which requires a Windows host path; state-aware exec is the surface that accepts `/workspace`.

**Fix:** Add the WSLc split explicitly: one-shot uses a Windows host path, while state-aware exec uses an in-container Unix path.

docs/schema.md
```diff
@@ -269,7 +267,21 @@ that can be executed independently.
...
+| Seatbelt (macOS) | `/workspace`. A `~` path is rejected: MXC would expand it from the launching host's `HOME`, and falls back to a literal `~` when that is unset. |
+| LXC / Bubblewrap | `/workspace`. `~` is not expanded. |
```
MGudgin: **Low (documentation) - State that `~` is rejected, not merely unexpanded.**

**Attribution:** `introduced_by_change` — the shared Unix-style validator rejects any value not starting with `/`, so LXC/Bubblewrap behave like the Seatbelt row above.

**Fix:** Use consistent wording such as “a `~` path is rejected; it is not expanded.”

src/backends/hyperlight/common/src/lib.rs
```diff
@@ -893,7 +893,7 @@ mod tests {
         let mut r = runner();
         let request = ExecutionRequest {
             script_code: "print('x')".to_string(),
-            working_directory: "C:/tmp".to_string(),
+            working_directory: "/tmp".to_string(),
```
MGudgin: **Low (scope) - This fixture change is unrelated to the new schema gate.**

**Attribution:** `introduced_by_change` — this request keeps the default empty schema version, so the new schema-0.9 cwd validator returns before checking path style; Hyperlight already rejects every non-empty cwd.

**Fix:** Revert this edit, or explain it as an intentional consistency-only cleanup rather than a change required by the fix.

MGudgin (commented):
## Summary

Reviewed the current PR head `481495b` against base `6e12d6d`; the merge-base matches the PR base and the locally captured diff matches `gh pr diff` apart from line-ending size differences.

The target-aware path predicate itself came back clean: the one-shot, streaming, dry-run, and state-aware routes reach the new validation; diagnostic input is escaped; Linux fallback, macOS override, Windows path forms, and the WSLc one-shot/exec split are consistent. I found no security, performance, or cross-platform-parity defects.

### Findings outside the diff

**Medium (test coverage) - State-aware dispatcher wiring is not regression-tested.** `src/core/wxc_common/src/state_aware_dispatch.rs:90,123` — `newly_exposed_by_change`: this file is byte-identical base versus head, but the PR adds a new invariant inside `validate_exec_common` and relies on these dispatcher calls to enforce it. The added tests call the validator directly; there is no sibling to the existing empty-command-line dispatcher test that passes a relative cwd through streaming or relayed exec and proves rejection occurs before backend validation/execution.

**Fix:** Add dispatcher-boundary tests for both exec entry points using the call-counting backend, asserting `policy_validation` and zero backend validation/exec calls.

**Medium (proportionality/claim) - The #902 closure and test-coverage claims conflict with the registered regression.** `PR description` — `claim_mismatch`: the description says “Resolves #902” and “No existing test or example config used a relative cwd,” but `tests/regression_tests/test_cases/Invoke-Issue902-RelativeWorkingDirectory.ps1` and its `TestCases.psd1` registration are byte-identical base versus head. That test uses `0.8.0-alpha` and passes only when relative cwd is rejected, while this PR intentionally leaves 0.8 unchanged. This is non-blocking as a claim mismatch, but the acceptance artifact and description need to agree.

**Fix:** Re-version the regression to `0.9.0-alpha` if that is the intended acceptance contract, or explicitly document that 0.6-0.8 retain the old behavior and correct the PR description.

None posted as findings. The unchanged files above are retained only because the PR newly depends on their dispatcher/error behavior or makes a claim about them; they are not charged as standalone pre-existing defects.

src/core/wxc_common/src/script_runner.rs
```diff
@@ -86,9 +87,16 @@ pub fn emit_backend_error_envelope(response: &ScriptResponse) {
...
+    // message that does not end in a newline would leave the envelope glued to
+    // its tail, where the line-oriented parsers the SDKs use cannot find it.
+    if !response.standard_err.is_empty() && !response.standard_err.ends_with('\n') {
+        eprintln!();
+    }
+
     let mut envelope = serde_json::json!({
         "error": {
-            "code": "backend_error",
+            "code": envelope_error_code(&response.failure_phase).as_str(),
```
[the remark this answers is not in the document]
theelliotm: Done

src/backends/hyperlight/common/src/lib.rs
theelliotm: Reverted

docs/schema.md
theelliotm: Done

sdk/node/README.md
theelliotm: Done

docs/schema.md
theelliotm: Done

src/backends/wslc/common/src/wsl_container_runner.rs
theelliotm: Done

src/core/wxc_common/src/validator.rs
theelliotm: Made this change.

PR #1158, opened 2026-09-14
Regression tests [process containment]: Issue 1130 - Environment block needs SystemRoot + LOCALAPPDATA

### LINKED ISSUE #1130: Windows SBOX: launch error 203 and explicit-environment fallback observability (SDK 0.8.0)
## Context

- `@microsoft/mxc-sdk` 0.8.0, schema 0.8.0-alpha; Windows 11 x64 build 26200.
- Official tag v0.8.0, commit `7dac1a952f0c9ad13f0a4cb089c4e0e8b3e0013a`.
  Packaged executor metadata reports the same commit; Authenticode signature
  is valid Microsoft. No local native rebuild.
- Unelevated host-owned Node worker with sanitized environment, SDK
  `usePty: false`. Two separately authorized benign invocations failed; neither
  established that the candidate command executed.

## Sanitized request shape

This describes the actual request; it is not an executed reduced reproducer.
`containment: processcontainer`; leastPrivilege true, capabilities empty,
learningMode false. Filesystem grants: one copied Node executable readonly,
new workspace and scratch readwrite, deniedPaths empty, clearPolicyOnExit true,
allowDaclMutation true after explicit authorization. Ingress/egress and host
loopback deny; windows/input disabled, clipboard none; no proxy/denial capture.
Explicit command line and workspace cwd, timeout 5000 ms; explicit child env
only CI, NO_COLOR, private TEMP/TMP and SystemRoot. The benign command writes
and reads a workspace marker and tests a synthetic outside canary. No real
private files or credentials are involved.

## Failure and source evidence

Both exited unsigned 4294967295 (-1). The final diagnostic identifies:

```text
CreateProcessInSandbox failed with error code 203 (0x000000CB).
Experimental_CreateProcessInSandbox failed: WIN32_ERROR(203)
```

Raw output was not retained. Offline reconstruction of the official
standard_err + compact JSON envelope templates with the recorded cwd,
UTF-8 and LF matched both original capture SHA-256 values exactly: 403 and
380 bytes. This identifies the reported launch failure without invoking MXC
again. Private cwd values and reconstructed raw envelopes are omitted here.
Templates: base_container_runner.rs:1744/1779, launch_diagnostics.rs:67,
working_directory.rs:60, wxc_common/src/script_runner.rs:95/112 and
wxc/src/main.rs:1299/1316, all in the pinned source.

The source selects tiers by capability probes, not just build number.
LPAC excludes PSEC (base_container_runner.rs:814), while SBOX remains eligible.
The exact probe result is not in our receipt. The API pointer is resolved from
System32 processmodel.dll and GetLastError is captured immediately after failure
(lines147/181). No intervening logging/cleanup was found. Error 203 maps to
ERROR_ENVVAR_NOT_FOUND, but the variable or internal OS failure remains unknown.
There is no T3 dispatch after this launch failure, and no production T3 selector
was found; MXC_FORCE_TIER is cfg(test).

A separate concern: launch_with_environment_fallback:147/205 retries SBOX with
NULL environment after error 50 (ERROR_NOT_SUPPORTED) with an explicit block.
Final 203 does not exclude an earlier 50 followed by retry; the warning remains
in the internal logger and our receipt does not observe that branch. The
worker/executor remains sanitized, so credential inheritance is not established,
but strict child-env enforcement is unproven. Dependent workloads remain blocked.

## Questions

1. Which host/child environment prerequisites does SBOX require on build 26200
   for LPAC, and how should error 203 be diagnosed? Is a supported structured
   diagnostic available without logging command lines or environment values?
2. Can callers fail closed instead of abandoning the explicit environment on
   error 50? Can SDK output expose the selected launch contract and retry branch?
3. Is there a supported AppContainer/DACL selection mechanism when SBOX probes
   succeed but launch fails, or should this remain blocked pending an upstream
   fix? No test seams or host feature settings were changed.

No retry, native modification, broader grants or permissive policy is proposed.
This is not a successful containment test or a demonstrated escape.

jsidewhite:
Emits friendly error message if env is empty for process container.  We require SystemRoot + LOCALAPPDATA.

Adds regression test.

Closes #1130


tests/regression_tests/test_cases/Invoke-Issue1130-SboxSparseEnvironment.ps1
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+# Purpose: Test that we reject launching process containment sandbox if the env is totally empty (require some vars: LOCALAPPDATA / SystemRoot)
+
```
bbonaby, line 5: note: we'll probably want to add these to Elliots validation testing when you're all done with them. Thanks for adding these btw

bbonaby (approved):

PR #1160, opened 2026-09-14
feat(bwrap): add filesystem preflight diagnostics

### LINKED ISSUE #507: feat(bwrap): Improve error reporting for deny by default file system semantics

Follow-up from review of #482, which switched the Bubblewrap backend to a deny-by-default filesystem baseline.

With the new baseline, a few situations fail at `bwrap` exec time with low-level/opaque errors that we can detect beforehand and surface as clear, actionable messages:

- **`working_directory` outside the baseline / policy paths** — e.g. a `cwd` of `~/project` with no matching `readonlyPaths` entry. We know the baseline mounts at build time, so we could validate the cwd up front and return an error naming the missing path and the fix (add it to `readonlyPaths`/`readwritePaths`).
- **DNS on hosts where `/etc/resolv.conf` symlinks outside the bound `/run/...` dirs** — name resolution silently fails. We could detect the symlink target and either bind it or warn that it needs to be listed in `readonlyPaths`.

These are not bugs in #482 (the deny-by-default posture is intentional), but better error messages would make the new baseline much easier to adopt.

In the Bubblewrap runner, before exec:
- Resolve `working_directory` and check it is covered by the baseline or a policy path; if not, return a descriptive `MxcError` instead of relying on `bwrap` to fail.
- Optionally resolve the `/etc/resolv.conf` symlink target and, if it falls outside the bound DNS dirs, emit a diagnostic so DNS failures are explained.

MGudgin:
This PR adds actionable Bubblewrap diagnostics for paths hidden by the deny-by-
default filesystem baseline. Launch-time cwd preflight rejects only paths that
are provably uncovered or denied and defers ambiguous namespace cases to
Bubblewrap with retained warnings.

Fixes #507

### Details

* Distinguish uncovered, denied, and inconclusive cwd diagnostics while keeping
  host-dependent checks out of the stable validate surface.
* Model synthetic `/dev`, `/proc`, and `/var/run -> /run` topology, including
  policy mounts that shadow the compatibility link.
* Preserve caller `deniedPaths` spellings through symlink, hard-link, and
  filesystem-object alias normalization without probing unchanged spellings.
* Preserve `lxc-exec` failure output in shell E2Es and accurately identify
  resolver-config inspection failures.
* Add unit, parser-corpus, documentation, and real Bubblewrap E2E coverage.

### Tests

* `cargo check -p bwrap_common --target x86_64-unknown-linux-gnu`
* `cargo clippy -p bwrap_common --target x86_64-unknown-linux-gnu -- -D warnings`
* `cargo test -p bwrap_common` in WSL (379 passed)
* Exact `wxc_common` differential repository corpus test
* `node scripts/versioning/validate-configs.js` (373 configs, 11 exempt)
* Rebuilt `lxc-exec`, basic Bubblewrap E2E, and cwd-preflight E2E


src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+fn warn_if_resolver_target_is_hidden(
+    request: &ExecutionRequest,
+    resolv_conf: &Path,
+    logger: &mut Logger,
+) {
+    let target = match resolver_symlink_target(resolv_conf) {
+        Ok(Some(target)) => target,
+        Ok(None) => return,
+        Err(error) => {
+            let _ = writeln!(
```
SohamDas2021, line 283: `writeln!(logger, …)` calls write to the logger's console/buffer stream, not the retained warnings channel. `mxc_engine::spawn` only hands `logger.take_warnings()` to the streaming handle, so SDK and FFI callers never receive this diagnostic at all and in CLI debug mode it lands on stdout ahead of workload output. 

Should be `logger.warning_line(...)`?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+            );
+            return;
+        }
+    };
+
+    if path_is_readable_in_sandbox(request, &target) {
+        return;
+    }
+
+    let _ = writeln!(
```
SohamDas2021, line 298: Same comment as above

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+        .iter()
+        .any(|path| cwd == Path::new(path))
+        || VIRTUAL_FILESYSTEM_PATHS
+            .iter()
+            .any(|path| cwd.starts_with(path));
+
+    if !visible {
+        for path in bwrap_command::BASELINE_RO_BIND_PATHS {
+            let mount = Path::new(path);
+            if mount.exists() && paths_overlap(&cwd, mount) {
```
SohamDas2021, line 223: `mount.exists()` is evaluated before the cheap prefix test, so any cwd outside the baseline stats all 18 `BASELINE_RO_BIND_PATHS` entries and `validate_working_directory` runs twice per launch, so ~36–54 stats per execution. One of those entries is `/mnt/wsl/resolv.conf`; on a host with autofs or a stale NFS mount under `/mnt`, that can block the launch path. Swapping the operands to `paths_overlap(&cwd, mount) && mount.exists()` drops it to zero stats for non-overlapping paths.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+        return false;
+    };
+
+    let readable = VIRTUAL_FILESYSTEM_PATHS
+        .iter()
+        .any(|mount| path.starts_with(mount))
+        || bwrap_command::BASELINE_RO_BIND_PATHS
+            .iter()
+            .map(Path::new)
+            .filter(|mount| mount.exists())
```
SohamDas2021, line 354: same comment as 223

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -51,6 +52,10 @@ use crate::{
     bwrap_version, network_rules, proxy_network,
 };
 
+const RESOLV_CONF_PATH: &str = "/etc/resolv.conf";
+const SYNTHETIC_DIRECTORY_PATHS: &[&str] = &["/", "/tmp", "/var"];
```
SohamDas2021, line 56: These constants re-state mount topology `bwrap_command.rs` already owns (`--tmpfs /tmp`, `--dev /dev`, `--proc /proc`, `--symlink /run /var/run`). Nothing couples them, so adding or changing a mount in the command builder silently desyncs the preflight and since this feeds a hard `validate()` rejection, drift turns into refusing valid configs. Can we derive both from one mount plan, or at minimum add a test asserting every mount `build_args` emits is recognized here?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+        }
+    }
+
+    if !visible {
+        visible = request
+            .policy
+            .readonly_paths
+            .iter()
+            .chain(&request.policy.readwrite_paths)
+            .filter_map(|path| normalize_absolute_path(Path::new(path)))
```
SohamDas2021, line 236: The `/var/run` -> `/run` translation is applied to the cwd but not to policy paths here. So `readonlyPaths: ["/var/run/custom"]` with `cwd: "/var/run/custom"` compares `/run/custom` against `/var/run/custom`, overlaps nowhere, and gets rejected with an error telling the user to add the path they already added?`deniedPaths: ["/var/run/secret"]` fails to mask for the same reason. 

Note the new test uses `/run/custom` for the policy path and `/var/run/custom` for the cwd- the one spelling combination that works. Translation should apply to both sides.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+
+    apply_policy_visibility(request, &path, readable)
+}
+
+fn apply_policy_visibility(request: &ExecutionRequest, path: &Path, mut visible: bool) -> bool {
+    for mount in wxc_common::filesystem_resolve::resolve_mount_order(&request.policy) {
+        let Some(mount_path) = normalize_absolute_path(Path::new(&mount.path)) else {
+            continue;
+        };
+        if path.starts_with(mount_path) {
```
SohamDas2021, line 365: This only matches when the mount is an ancestor of `path`, so a re-allowed descendant can't undo a denied ancestor. With `deniedPaths: ["/var"]`, `readonlyPaths: ["/var/cache/apt"]`, `cwd: "/var/cache"` this returns false and we reject, but bwrap creates `/var/cache` as a synthetic parent for the allowed mount and the config runs fine (verified by running it). 

Visibility needs to account for synthetic parents recreated beneath a denied directory?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+}
+
+/// Whether `cwd` will exist in the namespace assembled by the bwrap command.
+///
+/// A bind mount makes both its source-backed subtree and any synthetic parent
+/// directories needed for the mount destination available. Empty synthetic
+/// filesystems such as `/tmp` guarantee only their root until policy mounts are
+/// layered beneath them.
+fn working_directory_is_visible(request: &ExecutionRequest, cwd: &Path) -> bool {
+    let Some(cwd) = normalize_absolute_path(cwd) else {
```
SohamDas2021, line 208: `normalize_absolute_path` folds `..` lexically at 208, before the `/var/run` translation on 211. `cwd: "/var/run/../etc"` therefore becomes `/var/etc` and is rejected- but in the sandbox `/var/run` is a symlink to `/run`, so the kernel resolves that cwd to baseline-mounted `/etc` and it works (verified). Lexical `..` folding doesn't match kernel path resolution whenever a component is a symlink.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
         if let Err(msg) = check_pin_against_denied_hosts(request) {
             return Err(ScriptResponse::error(&msg));
         }
+        // Re-check against the final policy: object normalization and denied
```
SohamDas2021, line 166: The comment says this re-check exists because normalization and denied-symlink resolution can change which mount wins at the cwd. But there's no test where `validate()` passes and this call then rejects, so the scenario justifying the duplicate call is unexercised?

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +163,213 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    ))
+}
+
+/// Whether `cwd` will exist in the namespace assembled by the bwrap command.
+///
+/// A bind mount makes both its source-backed subtree and any synthetic parent
+/// directories needed for the mount destination available. Empty synthetic
+/// filesystems such as `/tmp` guarantee only their root until policy mounts are
+/// layered beneath them.
+fn working_directory_is_visible(request: &ExecutionRequest, cwd: &Path) -> bool {
```
SohamDas2021, line 207: Nothing tests that this model agrees with what bwrap actually mounts- the only cwd E2E uses `/`, a trivial synthetic root. Since this now hard-fails `validate()`, a modelling error becomes a user-facing regression CI won't catch. 

Could we add E2Es with cwd under a readonly mount, a readwrite mount, a denied subtree, and a baseline path, asserting both the preflight verdict and the child's real `pwd`?

docs/bwrap-support/bubblewrap-backend.md
```diff
@@ -153,16 +153,19 @@ Common consequences of this default:
   not readable from the sandbox.
 - `/opt` and `/usr/local` tooling is not on PATH; list either path under
   `readonlyPaths` if the script depends on it.
-- `working_directory` must live under the baseline or a policy path — a
-  `cwd` of `~/project` without a matching `readonlyPaths` entry will fail.
+- `process.cwd` must live under the baseline or a policy path. The runner
```
SohamDas2021, line 156: This covers the visibility check, but `validate_working_directory` also rejects any non-absolute `process.cwd` outright with a different message (`must be an absolute path`) - a new user-visible failure mode documented neither here nor in `docs/schema.md`. Worth adding, along with the other rules this preflight enforces: synthetic dirs `/`, `/tmp`, `/var` match only at their exact root, `/var/run` is translated to `/run`, and `deniedPaths` precedence can reject an otherwise-covered cwd.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -158,11 +162,397 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    if working_directory_is_visible(request, cwd) {
+        return Ok(());
+    }
+
+    Err(format!(
+        "Bubblewrap: process.cwd '{}' is outside the deny-by-default filesystem baseline and \
+         configured filesystem policy paths. Add the working directory (or an ancestor) to \
+         filesystem.readonlyPaths or filesystem.readwritePaths.",
+        request.working_directory
+    ))
```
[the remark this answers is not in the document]
SohamDas2021: Confirming this, and the new `bubblewrap_cwd_denied.json` fixture is a clean demonstration of it. That config already lists the ancestor `/tmp/mxc_bwrap_cwd/denied` in `readwritePaths`, so a caller who follows the error's advice has already done the thing it asks for and the request still fails: the exact deny at `/tmp/mxc_bwrap_cwd/denied/secret` wins in `path_is_visible` (exact match sets `visible = !mount.is_denied()`), and a shallower allow cannot override a deeper deny.

Suggest branching the message: when the mount that loses at the resolved cwd is `FsIntent::Denied`, name the covering `deniedPaths` entry and tell the caller to remove or narrow it; keep the current `readonlyPaths`/`readwritePaths` wording for the genuinely uncovered case. `run_bwrap_cwd_preflight_test.sh` could then assert the denied fixture produces the `deniedPaths` wording instead of `filesystem.readonlyPaths`, which would also stop that assertion from passing for the wrong reason.

Leaving this open since it is not addressed in d0adb328.

SohamDas2021 (dismissed):

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -163,6 +184,274 @@ impl SandboxBackend for BubblewrapScriptRunner {
     }
 }
 
+fn preflight_working_directory(
```
jsidewhite, line 186: >preflight_working_directory [](http://example.com/codeflow?start=3&length=27)

I'm confused why this only applies to the bubblewrap backend?  Couldn't any container backend have this problem?  e.g. process_container

jsidewhite (changes requested):
:clock1:

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -163,6 +184,274 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    }
+
+    let Some(cwd) = normalize_absolute_path(cwd) else {
+        return PathVisibility::Indeterminate("it could not be normalized");
+    };
+    let mut visible = cwd == Path::new("/")
+        || cwd == Path::new(bwrap_command::TMP_PATH)
+        || cwd == Path::new("/var")
+        || cwd.starts_with(bwrap_command::DEV_PATH)
+        || cwd.starts_with(bwrap_command::PROC_PATH);
```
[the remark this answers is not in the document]
MGudgin: Fixed in 804b131f. The preflight now treats only the `/dev` and `/proc` roots as definitely present. Unmodeled descendants are indeterminate unless a later policy mount makes them definitively visible or denied, so cwd validation is deferred to bwrap and resolver diagnostics are retained. I added unit coverage for both virtual roots, policy overrides, denied descendants, and resolver warnings, plus an E2E using `/proc/sys` that verifies both the advisory and the child's real cwd.

src/backends/bubblewrap/common/src/bwrap_runner.rs
MGudgin: Bubblewrap has a backend-specific version of this problem because it assembles a sparse mount namespace: a cwd can exist on the host but be absent from the namespace built by `bwrap_command`. The verdict therefore depends on Bubblewrap's ordered baseline, virtual-filesystem, and policy mounts.

The other backends have different cwd contracts and checks. ProcessContainer keeps the host filesystem namespace, selects a concrete cwd through `appcontainer_common::working_directory`, and includes that cwd in launch failures; Seatbelt checks existence/searchability; WSLC validates its in-container path through the SDK. A shared visibility model would therefore be misleading, although a follow-up audit for consistent actionable cwd diagnostics across backend-specific implementations makes sense. I kept this PR scoped to the Bubblewrap deny-by-default regression tracked by #507.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -4717,7 +4717,7 @@ mod tests {
             observed_counts, expected_counts,
             "explicit divergence inventory and observed category totals differ"
         );
-        let expected_inventory = (353, 329, 14);
+        let expected_inventory = (363, 339, 14);
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit 5affc3b6. The corpus expectation is now `(364, 340, 14)`. The exact failing differential corpus test passes locally.

docs/version-specific-parser-migration-inventory.md
```diff
@@ -9,6 +9,8 @@ over newer configuration fixtures, primarily the Seatbelt validation corpus
 from #1125. The rebase added 73 JSON documents and removed one, producing 67
 additional equivalent accepts and five additional shared rejections without
 changing the seven classified exact-stricter results.
+The corpus baseline was refreshed again on 2026-09-15 for #1160, whose ten
+Bubblewrap cwd-preflight fixtures are equivalent accepts under both parsers.
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit 5affc3b6. Both references now say eleven fixtures, and the documented equivalent-accept count is 340.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -163,6 +184,282 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    };
+    let virtual_filesystem_descendant = [bwrap_command::DEV_PATH, bwrap_command::PROC_PATH]
+        .iter()
+        .map(Path::new)
+        .any(|root| cwd != root && cwd.starts_with(root));
+    let mut visible = cwd == Path::new("/")
+        || cwd == Path::new(bwrap_command::TMP_PATH)
+        || cwd == Path::new("/var")
+        || cwd == Path::new(bwrap_command::DEV_PATH)
+        || cwd == Path::new(bwrap_command::PROC_PATH);
```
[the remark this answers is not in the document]
MGudgin: Fixed in 19f71d8d after rebasing onto `origin/main` at e7940f7c. Cwd and policy paths under `/var/run` are now translated to their `/run` namespace equivalents before host-symlink and baseline evaluation, while policy mounts still apply in their original resolved order and denied diagnostics retain the caller spelling. The regression test injects the exact `/run/NetworkManager` baseline case, so it is deterministic even when the host has a real `/var/run` directory.

tests/scripts/run_bwrap_basic_test.sh
```diff
@@ -16,5 +16,10 @@ if [ ! -f "$LXC_EXEC" ]; then
 fi
 
 echo "Running basic Bubblewrap test..."
-"$LXC_EXEC" --experimental "$REPO_DIR/tests/configs/bubblewrap_basic.json"
+OUTPUT=$("$LXC_EXEC" --experimental "$REPO_DIR/tests/configs/bubblewrap_basic.json" 2>&1)
+echo "$OUTPUT"
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit 4b16412d. The basic test now guards the command substitution, prints captured output, and exits with a specific failure message when `lxc-exec` fails.

tests/scripts/run_bwrap_cwd_preflight_test.sh
```diff
@@ -0,0 +1,151 @@
...
+echo "Running Bubblewrap root cwd test..."
+OUTPUT=$("$LXC_EXEC" --experimental \
+    "$REPO_DIR/tests/configs/bubblewrap_cwd_root.json" 2>&1)
+echo "$OUTPUT"
+echo "$OUTPUT" | grep -q "CWD_ROOT_OK" \
+    || fail "the sandbox root was not honored as the child's cwd."
+
+echo "Running Bubblewrap virtual-filesystem descendant cwd test..."
+OUTPUT=$("$LXC_EXEC" --experimental \
+    "$REPO_DIR/tests/configs/bubblewrap_cwd_virtual_descendant.json" 2>&1)
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit 4b16412d. A shared `run_expect_success` helper now guards every expected-success invocation in the cwd suite and prints captured output before failing. Both affected E2Es pass.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -163,6 +184,299 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    logger: &mut Logger,
+    originals: &HashMap<String, String>,
+) {
+    let target = match resolver_symlink_target(resolv_conf) {
+        Ok(Some(target)) => target,
+        Ok(None) => return,
+        Err(error) => {
+            logger.warning_line(&format!(
+                "WARNING: Bubblewrap: could not inspect symlinked '{}': {}; DNS may fail inside \
+                 the sandbox. Repair the host resolver path before retrying.",
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit 4b16412d. The warning now says `could not inspect resolver config`, covering missing, unreadable, and broken-symlink cases accurately; both unit assertions were updated and 375 tests pass.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -163,6 +184,299 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    }
+    let safe_cwd = wxc_common::diagnostic_text::escape(&request.working_directory);
+    match working_directory_visibility(request, cwd) {
+        PathVisibility::Visible => Ok(()),
+        PathVisibility::Denied(denied_path) => Err(denied_working_directory_error(
+            request,
+            originals
+                .get(&denied_path)
+                .map(String::as_str)
+                .unwrap_or(&denied_path),
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit d64dbfe7. Original denied spellings are now recovered by filesystem-object identity, not only canonical path equality, so hard-link and bind/object aliases retain the caller `deniedPaths` entry. A distinct alias-conflict diagnostic remains as the fail-safe if provenance cannot be recovered. Added a hard-link regression test; 377 unit tests pass.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -163,6 +184,299 @@ impl SandboxBackend for BubblewrapScriptRunner {
...
+    cwd: &Path,
+    baseline_path_exists: impl Fn(&Path) -> bool,
+) -> PathVisibility {
+    if cwd
+        .components()
+        .any(|component| component == Component::ParentDir)
+    {
+        return PathVisibility::Indeterminate("it contains parent-directory traversal");
+    }
+    let cwd = resolve_var_run_compat_path(cwd);
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit d64dbfe7. `/var/run` translation is now disabled when a policy mount at `/var/run` or an ancestor shadows the synthetic link. The preflight then evaluates the host-backed topology and preserves ordered allow/deny verdicts; deterministic tests cover both read-only and denied `/var` mounts.

src/backends/bubblewrap/common/src/bwrap_runner.rs
```diff
@@ -973,14 +1343,45 @@ fn resolve_denied_paths(
...
+    original_denied_paths
+        .iter()
+        .find(|original| {
+            original.as_str() == path
+                || wxc_common::filesystem_object::compare_existing_filesystem_objects(
+                    Path::new(path),
+                    Path::new(original),
+                ) == wxc_common::filesystem_object::ExistingObjectComparison::Same
+        })
+        .map(String::as_str)
```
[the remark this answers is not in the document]
MGudgin: Fixed in amended commit ebe9fe40. The lookup now performs a complete exact-string pass before any object-identity fallback, so unchanged denied paths perform zero metadata comparisons. Added an injected-comparator test proving zero fallback calls for an exact match and retained explicit coverage that true aliases still use the identity comparison. Format, Linux check/clippy, 379 unit tests, and the cwd E2E pass.

MGudgin:
Closing this PR because the implementation is not converging.

The original goal in #507 remains valid, but this approach has expanded into a second model of Bubblewrap namespace and path-resolution semantics. Repeated review rounds continue to uncover additional correctness gaps across mount precedence, symlinks, policy aliases, resolver visibility, and version-specific cwd behavior. Continuing to patch this branch would increase risk without establishing a stable design boundary.

#507 remains open. A replacement should start with a smaller design: separate the resolver diagnostic from cwd handling, and make any cwd preflight advisory-only or derive it from a shared structured mount plan consumed by both argument generation and diagnostics.

The branch is retained for reference.

PR #1163, opened 2026-09-15
Feature [process containment]: Enumerate FS config (Issue 1162)

### LINKED ISSUE #1162: Add BaseContainer support for PSEC fs_enumerate

Add BaseContainer support for the PSEC `fs_enumerate` filesystem policy. MXC currently builds the PSEC specification with `fs_read_only` and `fs_read_write`, but cannot express an enumeration-only filesystem grant.

This is needed so BaseContainer can request directory-enumeration access through the native process security environment contract without requiring a broader filesystem grant.

- Update the generated PSEC bindings to a specification version that includes `fs_enumerate`.
- Plumb the corresponding filesystem paths through the BaseContainer PSEC specification builder.
- Add capability/version handling so unsupported hosts fail closed or use an appropriate compatible fallback rather than silently dropping the policy.
- Add serialization, policy-mapping, and BaseContainer execution coverage for `fs_enumerate`.

jsidewhite:
Adds support for fs_enumerate for applications that call `FindFirstFile`/`FindNextFile`.

Closes #1162 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1163)


src/core/mxc-sdk/README.md
```diff
@@ -191,6 +191,8 @@ The reported `tier` is a **ceiling** — the strongest isolation the host can
 reach for that backend; a policy can still force a weaker tier at dispatch.
 `capabilities` lists optional features supported by that tier.
 `FilesystemDeniedPaths` covers native `filesystem.deniedPaths`.
+`FilesystemEnumeratePaths` covers PSEC 1.1
+`filesystem.enumeratePaths` without granting file-content reads.
```
bbonaby: thought (non-blocking): Should we remove PSEC 1.1 here? #Resolved

src/backends/appcontainer/common/src/probe.rs
```diff
@@ -227,6 +230,10 @@ fn format_fallback_error(e: &FallbackError) -> String {
         FallbackError::SystemRootUnresolved { reason } => {
             format!("Could not resolve Windows system directory: {reason}")
         }
+        FallbackError::EnumeratePathsUnsupported => {
+            "filesystem.enumeratePaths requires BaseContainer PSEC 1.1 with fs_enumerate support"
```
bbonaby: question (non-blocking): Should the error state that the users version of Windows doesn't support enumerate paths rather than calling out PSEC 1.1 directly? 

I guess the question is the same for the error message in src/backends/appcontainer/common/src/dispatcher.rs as well. 

Just wondering if folks see PSEC 1.1 if they'll know what we mean is all. That said, we can probably get away with just making sure the docs say what we mean. #Resolved

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -772,25 +802,61 @@ impl BaseContainerRunner {
...
-        })
+        if !request.policy.enumerate_paths.is_empty()
+            && !api
+                .supports_enumerate_paths()
+                .map_err(|error| ScriptResponse {
+                    failure_phase: FailurePhase::BackendUnavailable,
+                    ..ScriptResponse::error(&format!(
+                        "failed to query Process Security Environment filesystem enumeration support: {error}"
+                    ))
+                })?
```
bbonaby: note (non-blocking): For this if statement and the one below, it'd probably be preferrable to have the part after the `&&` go in a separate variable above it just so it's all readable/maintainable. #Resolved

src/backends/appcontainer/common/src/base_container_runner.rs
```diff
@@ -340,6 +345,9 @@ trait CaptureSessionFactory: Send + Sync {
 trait CapturePlatformSupport: Send + Sync {
     fn check_apis(&self, require_learning_mode: bool) -> Result<(), String>;
     fn supports_deny_paths(&self) -> Result<bool, String>;
+    fn supports_enumerate_paths(&self) -> Result<bool, String> {
```
bbonaby: question: is this one used?

src/backends/process_container/common/src/base_container_runner.rs
```diff
@@ -382,6 +390,19 @@ impl CapturePlatformSupport for RealCapturePlatformSupport {
                 format!("could not query process security-environment support: {error}")
             })
     }
+
+    fn supports_enumerate_paths(&self) -> Result<bool, String> {
```
bbonaby, line 422: note (non-blocking): perhaps this will be removed eventually with the refactor once secenv.rs is moved out of learning mode.  The `RealCapturePlatformSupport` just seems to add duplication. #WontFix

src/backends/wslc/common/src/policy.rs
```diff
@@ -72,6 +73,9 @@ const ERR_ENFORCEMENT_MODE: &str =
 /// mode; both are honoured here. Everything else in the module table is
 /// rejected.
 pub(crate) fn validate_provision_policy(request: &ExecutionRequest) -> Result<(), MxcError> {
+    if !request.policy.enumerate_paths.is_empty() {
```
bbonaby, line 136: question: do we need the same for bubblewrap and LXC? #WontFix

src/backends/appcontainer/common/src/probe.rs
jsidewhite: Yeah that's probably better.

src/backends/process_container/common/src/base_container_runner.rs
jsidewhite: Yeah I plan to do that right after this! :)

jsidewhite: For this PR, i wanted it to follow the same pattern as deny paths, for clarity.

jsidewhite: Agreed. `CapturePlatformSupport` is currently retained as the injectable test seam for native-capture eligibility. Consolidating it with `SecurityEnvironmentApi` fits the planned refactor that moves `secenv.rs` out of Learning Mode, so I would prefer to keep that broader cleanup out of this feature PR.

src/backends/wslc/common/src/policy.rs
jsidewhite: Bubblewrap and LXC do not need backend-local duplicates. Their one-shot `ScriptRunner`/`SandboxBackend` paths run `wxc_common::validator::validate_common`, which rejects non-empty `enumeratePaths` for every backend except ProcessContainer before backend execution. The WSLC check here is for the separate state-aware provision path, which does not use the one-shot runner validation path. I expanded the central validator test to explicitly cover Bubblewrap and LXC.

jsidewhite: Great callout, and maybe we should make it clearer which paths are validated centrally... or fix code so it's super clear.

Copilot says it's not a concern because all paths are covered...

> Bubblewrap and LXC do **not** need duplicate backend-local checks. Their one-shot execution paths 
always call `wxc_common::validator::validate_common`, which already rejects `enumeratePaths` for every backend except 
Windows ProcessContainer. The WSLC check is specifically needed for its state-aware provision path, which uses separate 
phase validation.

..but I agree with you the code should speak for itself.

src/core/mxc-sdk/README.md
jsidewhite: Agreed. I removed the PSEC version from the public Rust SDK capability description. It now describes the observable behavior: native support for `filesystem.enumeratePaths` without granting file-content reads.

src/backends/appcontainer/common/src/base_container_runner.rs
jsidewhite: Agreed. I extracted named requirement/support booleans for enumeration and unrestricted host loopback, while keeping each capability query conditional so an unused capability query cannot reject an unrelated request. I also corrected the nearby PSEC 1.1 comment.

src/backends/appcontainer/common/src/probe.rs
jsidewhite: Agreed. I changed the probe, dispatcher, fallback, and direct AppContainer messages to lead with “not supported by this version of Windows” and explain that enumeration-only access requires native ProcessContainer support and cannot fall back to AppContainer. PSEC details remain in internal capability diagnostics and backend documentation.

bbonaby (dismissed):

src/backends/process_container/common/src/base_container_helpers.rs
```diff
@@ -34,7 +34,9 @@ pub(super) enum PsecContract {
 
 impl PsecContract {
     pub(super) fn for_request(request: &ExecutionRequest) -> Self {
-        if unrestricted_host_loopback_allowed(&request.policy) {
+        if unrestricted_host_loopback_allowed(&request.policy)
+            || !request.policy.enumerate_paths.is_empty()
```
MGudgin, line 38: **High (correctness/security) - Enumeration can serialize an ingress table the host did not advertise.**

**Attribution:** `introduced_by_change` — this added condition makes an enumeration-only request select PSEC 1.1.

`build_psec_network_policy` treats every 1.1 contract as ingress-capable, while the support flags are independently advertised. The parser also creates a default `network_ingress` policy when `network` is omitted. Because `resolve_psec_contract` queries ingress support only for explicit host-loopback allow, a host with PSEC 1.1 + `PSE_SUPPORT_FS_ENUMERATE` but without `PSE_SUPPORT_NETWORK_INGRESS` receives an ingress table anyway.

**Fix:** Separate contract version from negotiated capabilities and serialize ingress only after a positive ingress-support result. Add the enumeration-true/ingress-false regression case.

sdk/dotnet/Microsoft.Mxc.Sdk/PlatformDiscovery.cs
```diff
@@ -69,6 +69,11 @@ public enum BackendCapability
     /// </summary>
     FilesystemDeniedPaths,
 
+    /// <summary>
+    /// Native filesystem enumeration-only access at the reported tier.
+    /// </summary>
+    FilesystemEnumeratePaths,
```
MGudgin: **High (correctness) - This insertion changes an existing public enum value.**

**Attribution:** `introduced_by_change` — the new member is inserted before the implicitly numbered `IngressHostLoopbackAllow`.

At the base, `IngressHostLoopbackAllow` is numeric value 4; at the head it becomes 5 and value 4 now means `FilesystemEnumeratePaths`. Already-compiled consumers embed the old constant, so they can mistake enumeration support for loopback support and fail to recognize actual loopback support.

**Fix:** Append the new member or assign explicit numeric values preserving every existing identity. Add a compatibility test for the released values.

src/backends/process_container/common/src/base_container_runner.rs
```diff
@@ -766,31 +796,77 @@ impl BaseContainerRunner {
...
         }
-        Err(ScriptResponse {
-            failure_phase: FailurePhase::Rejected,
-            ..ScriptResponse::error(
-                "network.ingress.hostLoopback='allow' requires Process Security Environment \
-                 contract version 1.1 with ingress support",
-            )
-        })
+
+        let requires_enumerate_paths = !request.policy.enumerate_paths.is_empty();
```
MGudgin: **Medium (test coverage/testability) - The production contract-resolution matrix has no deterministic test seam.**

**Attribution:** `introduced_by_change` — the rewritten resolver directly queries three OS capabilities, but no test can force its version/flag/error combinations.

Base and head each contain only the resolver definition and its production call; the added tests cover FlatBuffer serialization and pure compatibility predicates. That gap allowed the independent ingress-capability case to go untested.

**Fix:** Inject a capability snapshot/probe and test version unsupported, enumeration unsupported, enumeration supported without ingress, ingress-required unsupported, query errors, and full success.

sdk/node/src/sandbox.ts
```diff
@@ -428,6 +437,13 @@ export function createConfigFromPolicy(
         readonlyPaths: [...(policy.filesystem?.readonlyPaths ?? [])],
         deniedPaths: [...(policy.filesystem?.deniedPaths ?? [])],
     };
+    if (policy.processContainer?.filesystem?.enumeratePaths?.length) {
```
MGudgin: **Medium (reliability/cross-platform parity) - The portable builder emits a Windows-only backend section before resolving the target.**

**Attribution:** `introduced_by_change` — this new assignment runs before Linux `process` resolves to Bubblewrap, macOS resolves to Seatbelt, or explicit WSLC returns.

Those paths retain the new `processContainer` section and native parsing later rejects the mixed-backend configuration. Pre-0.9 policies also pass SDK validation and fail only after a native child is spawned.

**Fix:** Before assigning this section, require schema `0.9.0-alpha` and an effective Windows ProcessContainer target. Add Linux, macOS, WSLC, and old-version negative tests.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1120,6 +1153,10 @@ fn normalize_filesystem_paths(policy: &mut ContainerPolicy, logger: &mut Logger)
     for (paths, list_name) in [
         (&policy.readwrite_paths, "readwritePaths"),
         (&policy.readonly_paths, "readonlyPaths"),
+        (
+            &policy.enumerate_paths,
```
MGudgin: **Low (performance) - Enumeration-only paths now add serial host filesystem I/O during parsing.**

**Attribution:** `introduced_by_change` — this added list participates in the existing synchronous `fs::metadata` loop.

An enumeration-only request can now block on every configured local, removable, or UNC path before backend selection. This is downgraded to Low because the pre-existing path lists already had the same behavior.

**Fix:** Skip the advisory existence probe for enumeration-only paths, or make it bounded and testable.

src/backends/process_container/common/src/base_container_runner.rs
```diff
@@ -1023,13 +1128,21 @@ impl BaseContainerRunner {
...
+                request.policy.enumerate_paths.is_empty()
+                    || Self::query_psec_enumerate_support().unwrap_or(false),
             )
         {
             return true;
         }
         crate::fallback_detector::base_container_supports_deny_paths()
     }
 
+    pub(crate) fn supports_enumerate_paths_for_request(request: &ExecutionRequest) -> bool {
```
MGudgin: **Low (maintainability) - The new capability helper is not the single owner of its rule.**

**Attribution:** `introduced_by_change` — the PR adds this helper while new sibling paths independently re-derive the same `enumerate_paths.is_empty() || query` condition and thread another positional boolean through selection/probing.

The formulas already differ in shape, and boolean transposition remains type-correct.

**Fix:** Resolve a named capability snapshot once and reuse it across selection, probing, validation, and reporting.

tests/regression_tests/TestCases.psd1
```diff
@@ -120,4 +120,12 @@
         HarnessArguments = "Default"
         Destructive = $false
     }
+    "1162" = @{
+        FriendlyName = "BaseContainer Filesystem Enumeration"
+        Script = "test_cases\Invoke-Issue1162-ProcessContainerFsEnumerate.ps1"
+        ExpectedTierSupport = @("base-container-psec")
+        Prerequisites = @("wxc-exec", "PSEC fs_enumerate-capable host")
```
MGudgin, line 127: **Medium (testability) - This capable-host prerequisite has no execution effect.**

**Attribution:** `newly_exposed_by_change` — the regression harness is byte-identical base-to-head and consumes `Dependencies`, not `Prerequisites`; this new test is the code that newly relies on the inert field.

The all-regressions runner executes every non-destructive test, while this script turns unsupported-host probe rejection into failure. Broad runs therefore fail on hosts that cannot satisfy the declared prerequisite.

**Fix:** Add an executable capability preflight/skip to the harness, preserving failure when a host advertises support but behavior is wrong.

src/backends/isolation_session/common/src/policy.rs
```diff
@@ -88,6 +88,7 @@ pub(super) fn validate_post_provision_policy(
 fn reject_filesystem_policy(request: &ExecutionRequest) -> Result<(), IsolationSessionError> {
     if !request.policy.readwrite_paths.is_empty()
         || !request.policy.readonly_paths.is_empty()
+        || !request.policy.enumerate_paths.is_empty()
```
MGudgin, line 91: **Low (test coverage) - New backend-local rejection guards are untested.**

**Attribution:** `introduced_by_change` — equivalent new guards are added here, in Windows Sandbox, and in WSLC, while those files retain the same test counts as the base (32, 22, and 29 respectively).

These state-aware/provision paths do not all rely on the shared one-shot validator.

**Fix:** Add local tests in all three backend policy modules with populated `enumerate_paths`, asserting the intended backend-specific errors.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -1111,6 +1123,27 @@ fn normalize_filesystem_paths(policy: &mut ContainerPolicy, logger: &mut Logger)
...
+                 applying most-restrictive intent (enumerate)",
+                config_deserialize::escape_diagnostic_text(p)
+            ));
+            false
+        } else {
+            true
+        }
+    });
+    policy.enumerate_paths.retain(|p| {
+        if denied.contains(p) {
```
MGudgin, line 1139: **Low (test coverage) - Two new same-path precedence decisions have no regression test.**

**Attribution:** `introduced_by_change` — the PR adds readwrite-vs-enumerate, readonly-vs-enumerate, and enumerate-vs-denied branches, but only one `same_path_in_*enumerate*` test exists at the head.

**Fix:** Add same-path tests for readwrite-versus-enumerate and enumerate-versus-denied, including the normalized policy and diagnostic.

MGudgin (commented):
## Summary

Attribution was re-verified against
`ae1e60a54779f0f32d9eeec7edae9d65a4083b54...ed7c3db9b4df8b7f6e886f5318951ee35c78417e`.
The merge base exactly equals the PR base, and the local/GitHub diff content is
equivalent after ignoring blob-hash abbreviation and hunk-label differences.

Two High findings survive: PSEC 1.1 enumeration can serialize an unadvertised
ingress table, and the new C# enum member changes an existing public numeric
value. The review event is `COMMENT`; the High findings should still be fixed
before merge.

- Unsupported non-ProcessContainer backends received fail-closed rejection
  guards; no silent policy broadening was found.
- The PR adds independent support-bit coverage showing
  `PSE_SUPPORT_FS_ENUMERATE` and `PSE_SUPPORT_NETWORK_INGRESS` can differ.
- Generated wire/schema artifacts were treated as generated output rather than
  reviewed for hand-written implementation defects.
- The central resolver still has only its definition and production call in
  both base and head; the two added Rust tests cover serialization/predicates,
  not the resolver matrix.

### Findings outside the diff

**Low (documentation drift) - Capability design document is now stale.**
`docs/backend-support-probe-api-plan.md:83` —
`newly_exposed_by_change`: the file is byte-identical base versus head, but its
exhaustive capability list omits the new `filesystemEnumeratePaths` value added
by this PR. Add the new capability to the list.

**Low (documentation drift) - C# discovery documentation omits the new public
capability.** `sdk/dotnet/README.md:115-144` —
`newly_exposed_by_change`: the README is modified by the PR, but the capability
discovery example and prose are unchanged while `BackendCapability` gains
`FilesystemEnumeratePaths`. Add it to the example and explain the mapped policy
field.

No unrelated pre-existing finding is being charged to this PR. The unchanged
regression harness and design document are cited only where the PR's new
behavior newly depends on or invalidates them.

### Withdrawn

- `src/backends/process_container/common/src/appcontainer_runner.rs:1734-1740`
  — `invalid`: the added check is redundant in normal dispatch, but redundancy
  alone does not establish a defect and can be intentional defense in depth.

docs/schema.md
```diff
@@ -293,6 +296,12 @@ The `filesystem` section defines path access policy shared across backends:
 | `readonlyPaths` | string[] | `[]` | Paths the process can read but not write. |
 | `deniedPaths` | string[] | `[]` | Paths the process cannot access at all. |
 
+The ProcessContainer-only `processContainer.filesystem` section contains:
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `enumeratePaths` | string[] | `[]` | Paths the process can query or enumerate without reading file contents. Requires schema `0.9.0-alpha` and a Windows BaseContainer host with PSEC 1.1 `fs_enumerate` support. |
```
MGudgin: **Medium (documentation drift) - The new field's `leastPrivilege` incompatibility is omitted.**

**Attribution:** `introduced_by_change` — this added row states the schema/host requirements but not the policy-shape restriction.

PSEC selection rejects `least_privilege_mode`, while any non-empty `enumeratePaths` makes the legacy SBOX contract incompatible. The combination therefore fails even on a host advertising PSEC 1.1 and filesystem enumeration.

**Fix:** State here and in the OS support guide that `enumeratePaths` is incompatible with `processContainer.leastPrivilege` and fails rather than falling back.

src/backends/process_container/common/src/base_container_runner.rs
```diff
@@ -766,31 +796,77 @@ impl BaseContainerRunner {
...
+            failure_phase: FailurePhase::BackendUnavailable,
+            ..ScriptResponse::error(&format!(
+                "failed to query Process Security Environment contract support: {error}"
+            ))
+        })?;
+        if !version_supported {
+            return Err(ScriptResponse {
+                failure_phase: FailurePhase::Rejected,
+                ..ScriptResponse::error(
+                    "the request requires Process Security Environment contract version 1.1, \
```
MGudgin: **Medium (error behavior) - The common version check regresses the existing host-loopback diagnostic.**

**Attribution:** `introduced_by_change` — the base resolver returned the field-specific host-loopback/ingress error whenever PSEC 1.1 was unavailable; this new generic error intercepts that path.

It also bypasses the new enumeration-specific message when the host lacks 1.1 entirely, so neither trigger identifies the field that caused rejection.

**Fix:** Preserve trigger-specific diagnostics after the common version query: enumerate message for enumeration, ingress message for host-loopback allow.

src/backends/process_container/common/src/secenv.rs
```diff
@@ -491,6 +492,11 @@ impl SecurityEnvironmentApi {
         self.query_support(PSE_SUPPORT_FS_DENY)
     }
 
+    /// Whether the official PSEC API supports enumeration-only filesystem paths.
+    pub fn supports_enumerate_paths(&self) -> Result<bool, LearningModeError> {
```
MGudgin, line 480: **Low (performance) - The new immutable host capability is not cached.**

**Attribution:** `introduced_by_change` — this new method queries the support API directly, while the adjacent fixed deny-path capability memoizes results for the cacheable production API.

Selection, probing, validation, and contract resolution can repeat this query for one request.

**Fix:** Apply the same cacheability and typed-error memoization pattern used by `supports_deny_paths`, with a query-count test.

src/backends/process_container/common/src/probe.rs
```diff
@@ -227,6 +241,11 @@ fn format_fallback_error(e: &FallbackError) -> String {
         FallbackError::SystemRootUnresolved { reason } => {
             format!("Could not resolve Windows system directory: {reason}")
         }
+        FallbackError::EnumeratePathsUnsupported => {
+            "processContainer.filesystem.enumeratePaths is not supported by this version of Windows; \
```
MGudgin: **Low (maintainability) - The same unsupported condition now has four separately maintained messages.**

**Attribution:** `introduced_by_change` — probe, fallback, dispatcher, and AppContainer validation each add their own enumeration wording.

The strings already differ slightly, so future guidance can drift depending on the surfaced path.

**Fix:** Reuse the `FallbackError::EnumeratePathsUnsupported` display text and wrap it with path-specific context.

sdk/node/src/sandbox.ts
```diff
@@ -356,6 +359,12 @@ function buildMicroVmConfig(
             deniedPaths: policy.filesystem?.deniedPaths,
         };
     }
+    if (policy.processContainer?.filesystem?.enumeratePaths?.length) {
```
MGudgin, line 362: **Low (test coverage) - This new microvm rejection has no matching negative unit test.**

**Attribution:** `introduced_by_change` — the Node tests add one positive ProcessContainer serialization case, but no `assert.throws` case for this branch.

**Fix:** Add a microvm + `enumeratePaths` rejection test alongside the existing microvm policy-rejection tests.

src/backends/process_container/common/src/base_container_helpers.rs
jsidewhite: Fixed in f2638888. PSEC contract version and negotiated ingress capability are now separate. Enumeration-only PSEC 1.1 omits the ingress table when PSE_SUPPORT_NETWORK_INGRESS is absent, with a regression test for enumerate=true/ingress=false.

sdk/dotnet/Microsoft.Mxc.Sdk/PlatformDiscovery.cs
jsidewhite: Fixed in f2638888. BackendCapability now has explicit numeric values preserving IngressHostLoopbackAllow=4, with FilesystemEnumeratePaths appended as 5. A compatibility test locks all released values.

src/backends/process_container/common/src/base_container_runner.rs
jsidewhite: Fixed in f2638888. Added an injectable PsecCapabilityProbe and deterministic resolver tests covering version absence, missing enumerate/ingress flags, enumerate without ingress, each query failure, and full success.

sdk/node/src/sandbox.ts
jsidewhite: Fixed in f2638888. The Node builder now requires schema 0.9.0-alpha and an effective Windows ProcessContainer target before emitting the section. Added old-version, Linux process, macOS process, and WSLC rejection tests.

src/core/wxc_common/src/config_parser.rs
jsidewhite: Fixed in f2638888. Enumeration-only paths no longer participate in the synchronous advisory metadata/existence loop.

src/backends/process_container/common/src/base_container_runner.rs
jsidewhite: Fixed in f2638888. Added a named BaseContainerRequestCapabilities snapshot and reused it across dispatcher selection, probing/reporting, and fallback decisions instead of threading positional booleans.

tests/regression_tests/TestCases.psd1
jsidewhite: Fixed in f2638888. #1162 declares CapabilityPreflight; the harness invokes -CheckPrerequisites and treats exit 77 as skipped. The test skips only when support is not advertised and still fails when advertised support cannot probe successfully.

src/backends/isolation_session/common/src/policy.rs
jsidewhite: Fixed in f2638888. Added backend-local enumerate_paths rejection tests for IsolationSession provision, Windows Sandbox policy planning, and WSLC provision.

src/core/wxc_common/src/config_parser.rs
jsidewhite: Fixed in f2638888. Added readwrite-vs-enumerate and enumerate-vs-denied same-path tests, asserting both the normalized policy and the most-restrictive diagnostic.

docs/schema.md
jsidewhite: Fixed in f2638888. The schema reference and Windows support guide now state that enumeratePaths is incompatible with processContainer.leastPrivilege and fails rather than falling back.

src/backends/process_container/common/src/base_container_runner.rs
jsidewhite: Fixed in f2638888. The common version query now preserves trigger-specific rejection messages: enumeration identifies enumeratePaths, while unrestricted host loopback identifies ingress support.

src/backends/process_container/common/src/secenv.rs
jsidewhite: Fixed in f2638888. SecurityEnvironmentApi now caches the complete immutable support-flag result, including typed errors, and derives deny/enumerate/ingress from it. The query-count test verifies all three methods issue one native query.

src/backends/process_container/common/src/probe.rs
jsidewhite: Fixed in f2638888. Probe, dispatcher, and AppContainer rejection paths now reuse FallbackError::EnumeratePathsUnsupported; direct PSEC negotiation retains its more specific contract/capability diagnostic.

sdk/node/src/sandbox.ts
jsidewhite: Fixed in f2638888. Added the requested Node microVM + enumeratePaths negative unit test alongside the existing microVM policy rejection coverage.

MGudgin:
Re-verified the review against updated head `d2db390e` and the new PR base
`b773325c`.

**Disposition:** all 14 inline findings are addressed. The two original High
findings are fixed: PSEC ingress capability is resolved independently from the
1.1 contract version, and the C# capability enum preserves its released
numeric values.

Verification completed:

* 29 focused Rust tests passed, including the resolver matrix, enumeration
  without ingress, path precedence, backend rejection, and capability-cache
  tests.
* Node SDK unit tests: 432 passed, 20 skipped, 0 failed.
* Focused .NET capability tests: 6 passed, 0 failed.

The capable-host #1162 executor E2E was not run locally. CI's Windows x64/ARM64
build jobs and Rust CodeQL were still running when checked; the other reported
build, SDK, lint, versioning, and integration checks were passing.

Two **Low, non-blocking documentation items from the review body remain**:

* `docs/backend-support-probe-api-plan.md:83` still omits
  `filesystemEnumeratePaths` from its capability list.
* `sdk/dotnet/README.md:115-144` still omits
  `BackendCapability.FilesystemEnumeratePaths` from the discovery example and
  prose.

No blocking review finding remains open.

MGudgin (approved):

PR #1166, opened 2026-09-15
fix(seatbelt): let a sandbox listen without surrendering proxy confinement

caarlos0:
## 📖 Description

A sandbox under a runtime proxy could not run a listener on macOS.
`python3 -m http.server 8080 --bind 127.0.0.1` failed with `EPERM`, and no
setting fixed it, because the two things a caller needs were welded together.

`bind()`/`listen()` require `(allow network-inbound (local ip))`, which only
`ingress.default: "allow"` emits, and `validate` refused **any** `hostLoopback`
differing from `ingress.default`. So asking for a listener also forced
`hostLoopback: "allow"`, which under a denied egress default emits
`(allow network-outbound (remote ip "localhost:*"))` — every port on the host. A
client that ignores `HTTP_PROXY` then reaches host-local services directly,
which is precisely the confinement a proxy exists to provide. Callers had to
give one of the two up.

Only one of the two divergent ingress pairs is actually inexpressible, so
`validate` now says that instead of rejecting both.

- **`hostLoopback: "allow"` under `default: "deny"` stays rejected.** It promises
  inbound host-loopback connectivity, and the only rule that could carry it is
  the blanket inbound grant that `default: "deny"` withholds. Accepting it would
  advertise a grant the profile never makes.
- **`hostLoopback: "deny"` under `default: "allow"` is now accepted:**

```lisp
(deny default)
(allow network-outbound (remote ip "localhost:9000"))   ;; the proxy, only
(allow network-inbound  (local ip))                     ;; bind + listen
```

`write_host_loopback_rules` needed no change — it already emitted nothing when
loopback is denied under a denied egress default (the `(deny default)` baseline
carries it) and the explicit `localhost:*` deny when egress is allowed. Both
newly-reachable cells gained tests instead.

### Limitations, stated up front

The container→host half of `hostLoopback: "deny"` is fully enforced. **The
host→container half is not**: `(local ip)` cannot exclude a loopback peer, so a
host process can connect to the sandbox's listeners, and a workload that binds
`0.0.0.0` rather than `127.0.0.1` is reachable on that port from the LAN.
Seatbelt cannot express a local-address filter — a literal
`(local ip "127.0.0.1:8080")` is a profile syntax error, `host must be * or
localhost`, exactly as for `remote`. `--bind 127.0.0.1` is a convention the
workload follows, not one the sandbox can enforce.

This is documented rather than implied, in both the backend doc and the 0.8
networking spec. It is strictly narrower than the alternative it replaces:
reaching a listener through `hostLoopback: "allow"` surrenders the
container→host direction as well.

### Documentation

Also fixes the fact that made this look unfixable. Under a denied egress
default, `bind()` comes from `ingress.default`, not from the egress arm:
`(allow network-inbound (local ip))` grants `bind()` **and** `listen()`, while
`(allow network-bind (local ip))` grants `bind()` but **not** `listen()`. The
network policy table named `network-bind` only under `egress.default: "allow"`,
which reads as "a denied egress default denies `bind()`" and invites widening
the deny posture to "fix" it.

## 🔗 References

Related to #1032 (not resolved here), and reviewers should know about one
interaction with it. This config was **rejected** before and is **accepted**
now:

```json
{ "network": { "egress": { "default": "deny" },
               "ingress": { "default": "allow" } } }
```

An omitted `hostLoopback` parses as `deny`, so it is now a valid posture and
loopback is silently closed. Previously the symmetric check rejected it with a
message that said so explicitly ("an omitted `hostLoopback` is `deny`, not an
inherit of `default`"), which incidentally served as a diagnostic for the
ambiguity #1032 describes. That diagnostic is gone for this shape. Resolving
#1032 (inherit vs. absolute deny) is the right place to decide whether an
omitted `hostLoopback` under `default: "allow"` should instead be refused as
under-specified.

## 🔍 Validation

Automated, on macOS 26.6.2 (aarch64):

| Command | Result |
|---|---|
| `bash tests/scripts/run_seatbelt_network_test.sh` | 21 assertions passed (6 new) |
| `cargo test --locked --release --target aarch64-apple-darwin -p mxc_darwin -p seatbelt_common -p wxc_common -p wxc_e2e_tests -p mxc_engine` | green (133 / 995 / 110 + e2e) |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --target aarch64-apple-darwin --locked --release -p mxc_darwin -p seatbelt_common -- -D warnings` | clean |
| `shellcheck -S warning tests/scripts/run_seatbelt_network_test.sh` | clean |

New coverage: two `profile_builder` tests for the newly-reachable cells (denied
and allowed egress default), two `seatbelt_runner` tests (the rejected pair
keeps its message, the accepted pair validates), and two live-suite configs.
The pre-existing `rejects_mismatched_ingress_default_and_host_loopback` was
rewritten rather than deleted, so it now pins the narrower rejection.

The live suite also gains the `egress deny` + `ingress allow` cell, asserting
both the behavior and that **no** standalone `network-bind` rule is emitted —
without the second assertion, a change that re-granted `bind()` from the egress
side would keep the behavioral assertions green while widening the deny posture.

Manual, against real Seatbelt, with a loopback proxy and a separate host-local
origin listener, under `egress: deny` + `ingress: {default: allow, hostLoopback:
deny}`:

```
BIND_LISTEN_8080     = OK
PROXY_PORT           = REACHABLE(PROXYOK)
HOST_ORIGIN_DIRECT   = BLOCKED(EPERM)      <- no bypass
EXTERNAL             = BLOCKED(EPERM)

python3 -m http.server 8080 --bind 127.0.0.1  ->  host curl: HTTP 200
```

Review-feedback commits on top of the original change, each re-validated with
`cargo fmt --all -- --check`, `cargo test -p seatbelt_common` (133 passed),
`cargo check -p wxc_e2e_tests --tests`, and `shellcheck -S warning`:

| Commit | Change |
|---|---|
| `40130b19` | The rule-order assertion compared `Option<usize>` positions without first asserting the broad `(allow network-outbound)` exists, so `Some(deny) > None` kept it green if that rule disappeared. |
| `3df4fa0a` | The backend-guide rejection table and `schema-updates.md` still said every differing `hostLoopback`/`ingress.default` pair is rejected. |
| `6374c499` | Same stale claim in `local_network_allowed`'s rustdoc, the rejections-suite case label, the e2e characterization comment, and the Copilot instructions. The rejection fixture itself was already the correct pair; only its label was wrong. |

### Schema `0.9.0-alpha`

Since #1145 completed the directional cutover, both decisions were re-verified
against the exact v0.9 contract, end to end with a release `mxc-exec-mac`:

| v0.9 config | Result |
|---|---|
| `ingress: {default: allow, hostLoopback: deny}` | runs; `LISTEN_OK` |
| `ingress: {default: deny, hostLoopback: allow}` | rejected, `backend_error`, exit 255 |

With a host listener on `127.0.0.1:8099` (reachable from the host, `HTTP 200`),
the accepted pair under `egress.default: "deny"` gives
`HOSTLOOPBACK_BLOCKED` / `EXTERNAL_BLOCKED` / `LISTEN_OK` — identical to 0.8, so
the container-to-host confinement holds under v0.9.

Two `seatbelt_runner` tests now pin this through the real parser rather than a
hand-built `ContainerPolicy`. The rejection test is the guard: an adapter that
dropped `ingress` would make `validate` return `Ok` and fail it, while every
policy-level test in that module stayed green.

Not verified: only macOS 26.6.2 arm64 was exercised. `hostLoopback` validation
here is Seatbelt-specific, so other backends are untouched.

## ✅ Checklist

- [ ] Linked to an issue — no issue tracks this; #1032 is related but not resolved
- [x] Updated [Copilot instructions](.github/copilot-instructions.md) — no build, architecture, or convention change, but the Seatbelt row stated the old symmetric `hostLoopback` rejection and is corrected
- [ ] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes — `Cargo.lock` is unchanged

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1166)


richiemsft:
@theelliotm, @bbonaby

src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -1188,6 +1191,51 @@ mod tests {
...
+            default: NetworkAction::Allow,
+            ..Default::default()
+        });
+        r.policy.network_ingress = Some(wxc_common::models::NetworkIngressPolicy {
+            default: NetworkAction::Allow,
+            host_loopback: NetworkAction::Deny,
+        });
+        let p = build_profile(&r).unwrap();
+        assert!(p.contains("(allow network-inbound (local ip))"));
+        assert!(
```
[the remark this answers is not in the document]
caarlos0: Correct and fixed in 40130b19 — `assert!(p.contains("(allow network-outbound)"))` now runs before the position comparison, matching the sibling test.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -238,18 +277,44 @@ Two more things to know about `hostLoopback`:
...
-down.
+So MXC refuses the pair it cannot express and accepts the one it can.
+
+**`hostLoopback: "allow"` under `ingress.default: "deny"` is rejected.** It
+promises inbound host-loopback connectivity, and the only rule that could carry
+it is the blanket `(allow network-inbound (local ip))` that `default: "deny"`
+withholds. Accepting it would advertise a grant the profile never makes.
+
+**`hostLoopback: "deny"` under `ingress.default: "allow"` is accepted**, and is
+how you run a listener without opening host loopback:
```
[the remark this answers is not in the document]
caarlos0: Correct and fixed in 3df4fa0a. The rejection table row now names only the rejected pair (`hostLoopback: "allow"` under `default: "deny"`), and `schema-updates.md` states both halves: that pair rejected, the reverse accepted with container-to-host enforcement only, pointing at the backend guide for the limitation.

src/backends/seatbelt/common/src/seatbelt_policy.rs
```diff
@@ -160,27 +160,40 @@ pub fn validate_seatbelt_network_policy(policy: &ContainerPolicy) -> Result<(),
...
-            return Err(format!(
-                "macOS Seatbelt cannot enforce a network.ingress.hostLoopback \
-                 posture ('{}') that differs from network.ingress.default \
-                 ('{}'): the inbound half is not expressible in a Seatbelt \
-                 profile. Set both to the same value. Note that an omitted \
-                 'hostLoopback' is 'deny', not an inherit of 'default'.",
-                name(ingress.host_loopback),
-                name(ingress.default),
-            ));
+        if ingress.host_loopback == NetworkAction::Allow && ingress.default == NetworkAction::Deny {
```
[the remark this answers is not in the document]
caarlos0: Rejecting this one, with the contract text as the reason.

The mandate in the shared contract is directional, and names `allow` only: *"A backend that cannot enforce both directions must reject `hostLoopback: allow` rather than accept it with partial enforcement"* (`networking.md:112-113`). That sentence is upstream text this PR does not touch, and the branch keeps rejecting exactly that pair. There is no corresponding sentence for a partially-enforced `deny`; the general rule immediately above it provides for this case instead: *"Backend-specific docs identify temporary compatibility behaviors whose enforcement is intentionally coarser. Outside those documented exceptions, a backend rejects configurations it cannot enforce"* (`networking.md:266-269`). The exception is documented in both the backend guide and the Seatbelt section of the contract.

On the security direction: retaining the rejection does not keep the host-to-container half closed. It forces a caller who needs a listener to set `hostLoopback: "allow"`, which opens the host-to-container half *and* surrenders container-to-host egress to every port on the host — the confinement a model-2 proxy exists to provide. The accepted pair is strictly narrower than the configuration it replaces, not a weakening of it.

The alternative proposed here — a new asymmetric policy surface and capability bit — is a public contract addition for one backend's inexpressible half. If reviewers want the asymmetry surfaced in the capability model, that belongs in the #1032 discussion, which already owns the inherit-vs-absolute-deny decision for this field.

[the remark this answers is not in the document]
caarlos0: All four confirmed and fixed in 6374c499:

- `seatbelt_policy.rs` — `local_network_allowed`'s rustdoc now says `hostLoopback` is enforced separately on the container-to-host direction, instead of claiming it must match `default`.
- `run_seatbelt_rejections_test.sh` — the case label is now `hostLoopback=allow under ingress.default=deny is refused`. The fixture itself was already the still-rejected pair (`egress: allow`, `ingress: {default: deny, hostLoopback: allow}`), so only the label was wrong; the assertion is unchanged.
- `e2e_seatbelt_characterization.rs` — dropped the trailing "Seatbelt requires `hostLoopback` to equal `default`" sentence. The rest of that comment still holds: the allow variant states `ingress` explicitly because an omitted `hostLoopback` is `deny`.
- `.github/copilot-instructions.md` — the Seatbelt row now names the one rejected pair and the accepted reverse pair with its container-to-host-only enforcement.

I also grepped the repository for the remaining spellings (`diverge`, `must match`, `must equal`) against `hostLoopback`/`ingress` and found no other stale reference.

Validation: `cargo fmt --all -- --check` clean, `cargo test -p seatbelt_common` 133 passed, `cargo check -p wxc_e2e_tests --tests` clean, `shellcheck -S warning` clean.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -4685,7 +4685,14 @@ mod tests {
...
             "explicit divergence inventory and observed category totals differ"
         );
-        let expected_inventory = (354, 330, 14);
+        // +2 files / +2 equivalent accepts over the upstream inventory: the two
+        // Seatbelt live-suite configs that pin how `bind()`/`listen()` are
+        // granted -- `seatbelt_net_egress_deny_ingress_allow.json` and
+        // `seatbelt_net_ingress_allow_loopback_deny.json`. Both differ from
+        // their `seatbelt_net_egress_deny.json` sibling only by an added
+        // `ingress` block, so both land in `equivalent_accepts` and shift no
+        // rejection count.
```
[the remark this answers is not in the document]
caarlos0: Correct on the fact — the fixtures do change their probe commands (`seatbelt_net_egress_deny_ingress_allow.json` adds a separate `BIND_OK` probe, and `seatbelt_net_ingress_allow_loopback_deny.json` uses `HOSTLOOPBACK_*` / `LISTEN_BLOCKED_<errno>` markers), so "differ only by an added `ingress` block" overstated the relationship.

Reworded in d28ed798 to the fact the assertion actually rests on: both parsers accept each fixture and build identical runtime models, so both land in `equivalent_accepts` and shift no rejection count.

Comment-only change: `cargo fmt --all -- --check` clean and `differential_repository_corpus_has_no_unclassified_or_exact_looser_results` still passes.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -245,18 +285,44 @@ Two more things to know about `hostLoopback`:
...
```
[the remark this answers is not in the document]
caarlos0: Agreed, and fixed in 5a9ba388.

The backend guide now reads: "how you run a listener while keeping the container-to-host direction closed (the host-to-container direction stays open — see the limitation below)".

Two code comments used the same loose phrase and are now direction-explicit too: `seatbelt_policy.rs` (the `validate` rationale) and the `accepts_host_loopback_deny_under_an_ingress_default_allow` test doc.

On the second location: line 614 is the rejection-table row, which already names the direction ("the promised host-to-container grant could not be made"), so it needed no change. I grepped the repository for the remaining ambiguous spellings — `networking.md:297` ("without surrendering the container-to-host direction") and the Copilot-instructions row ("enforces the container-to-host half only") are already explicit.

Comment/doc-only change: `cargo fmt --all -- --check` clean and `seatbelt_common` 135 tests pass.

src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -1188,6 +1190,52 @@ mod tests {
         assert!(!p.contains("network-inbound"));
     }
 
+    /// `ingress.default: "allow"` with `hostLoopback: "deny"` is the posture a
+    /// runtime proxy needs: a listener is permitted, and the sandbox still
+    /// cannot originate a connection to any host-local service. Under a denied
+    /// egress default the `(deny default)` baseline carries the outbound half,
+    /// so the profile must emit the inbound rule and nothing else.
```
theelliotm: Remove this comment to keep it consistent with the other tests.

src/backends/seatbelt/common/src/profile_builder.rs
```diff
@@ -1188,6 +1190,52 @@ mod tests {
...
+        assert!(p.contains("(allow network-outbound (remote ip \"localhost:9091\"))"));
+        // The bypass this posture exists to avoid: every other host port.
+        assert!(!p.contains("(allow network-outbound (remote ip \"localhost:*\"))"));
+        assert!(!p.contains("(allow network-outbound)"));
+    }
+
+    /// Under an *allowed* egress default the baseline no longer carries the
+    /// outbound half, so `hostLoopback: "deny"` must emit the explicit
+    /// `localhost:*` deny — and it must still follow the blanket allow, since
+    /// Seatbelt takes the last matching rule.
```
theelliotm: Remove this comment to keep it consistent with the other tests.

src/backends/seatbelt/common/src/seatbelt_policy.rs
```diff
@@ -160,27 +160,40 @@ pub fn validate_seatbelt_network_policy(policy: &ContainerPolicy) -> Result<(),
...
+    // `hostLoopback: "deny"` under `default: "allow"` is permitted, and is the
+    // only way to ask for a listener while keeping the container-to-host
+    // direction closed. Its container-to-host half is fully enforced: under a denied egress default
+    // by the `(deny default)` baseline, and under an allowed one by the
+    // explicit `localhost:*` deny. Its host-to-container half is not, because
+    // `default: "allow"` emits `(allow network-inbound (local ip))` and that
+    // rule cannot exclude a loopback peer. That residual grant is narrower than
+    // the alternative it replaces: reaching a listener through
+    // `hostLoopback: "allow"` gives up the container-to-host direction too,
+    // which is what dissolves a runtime proxy's egress confinement.
```
theelliotm: Too verbose - the first paragraph might be enough.

src/core/wxc_common/src/config_parser.rs
```diff
@@ -4685,7 +4685,13 @@ mod tests {
             observed_counts, expected_counts,
             "explicit divergence inventory and observed category totals differ"
         );
-        let expected_inventory = (354, 330, 14);
+        // +2 files / +2 equivalent accepts: the two Seatbelt live-suite configs
+        // that pin how `bind()`/`listen()` are granted --
+        // `seatbelt_net_egress_deny_ingress_allow.json` and
+        // `seatbelt_net_ingress_allow_loopback_deny.json`. Both parsers accept
+        // each of them and build identical runtime models, so both land in
+        // `equivalent_accepts` and shift no rejection count.
```
theelliotm: Remove this comment.

tests/scripts/run_seatbelt_rejections_test.sh
```diff
@@ -38,7 +38,7 @@ expect_rejected "allowedHosts under defaultPolicy=block is refused" \
 
 # --- Network: the inbound half of hostLoopback is not expressible -----------
 
-expect_rejected "a hostLoopback that diverges from ingress.default is refused" \
+expect_rejected "hostLoopback=allow under ingress.default=deny is refused" \
     "seatbelt_reject_hostloopback_mismatch.json" \
     "cannot enforce a network.ingress.hostLoopback" "$S"
```
theelliotm:
```suggestion
    "cannot enforce network.ingress.hostLoopback" "$S"
```
This test currently errors. Remove the 'a' from the test's marker to reflect the updated label.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -197,12 +198,51 @@ This is the cross-backend
...
+
+Pinned by the `egress deny + ingress allow` cell in
+`tests/scripts/run_seatbelt_network_test.sh`, which asserts the behavior *and*
+that no standalone `network-bind` rule is emitted.
+
+> Binding is not an egress capability. A bound socket under a denied egress
+> default can neither `connect()` nor `sendto()`, and cannot receive inbound
+> datagrams without `network-inbound` — so it opens no data path. The worst
+> case is squatting on a port nothing is permitted to reach.
+
```
theelliotm: I would drop this section. Verbose and not need for this minor change.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -245,18 +285,45 @@ Two more things to know about `hostLoopback`:
 - **`deny` also cuts the sandbox off from its own loopback listeners**, since a
   Seatbelt sandbox shares the host's network stack.
 
-#### Why it must equal `ingress.default`
+#### Why only one divergent pair is allowed
```
theelliotm, line 249: I would make this section shorter and only mention the reasoning for why ingress=allow, hostLoopback=deny is OK and the opposite isn't. And I would mention elsewhere that you can't contain ingress with a proxy.

theelliotm (changes requested):
Add a test for the proxy usecase you mention as your reason for this change. Also, add a warning message (using the existing function) when a config has a proxy + ingress: allow, hostLoopback: deny so users know that ingress is unrestricted and only egress is being confined.

src/backends/seatbelt/common/src/profile_builder.rs
caarlos0: Removed in e8a4b561 — the test now matches the others in the module.

src/backends/seatbelt/common/src/profile_builder.rs
caarlos0: Removed in e8a4b561, same as the one above.

src/backends/seatbelt/common/src/seatbelt_policy.rs
caarlos0: Trimmed in e8a4b561 down to the first paragraph, plus one sentence naming which pair is inexpressible and why. The per-pair detail it used to repeat now lives only in the backend guide.

src/core/wxc_common/src/config_parser.rs
caarlos0: Removed in e8a4b561. The rebase onto #1136 moved the inventory to (353, 329, 14), so the line now reads (355, 331, 14) — +2 files / +2 equivalent accepts for the two new Seatbelt fixtures — with no comment.

tests/scripts/run_seatbelt_rejections_test.sh
caarlos0: Good catch — that was a real failure, applied in e8a4b561.

I changed the error message in this PR (it no longer has the article: `cannot enforce network.ingress.hostLoopback='allow' alongside ...`) but left the expected marker on the old wording, so the case failed with `refused, but not for '...'`. PR CI never showed it because this suite runs from the nightly validation matrix, not on pull requests.

Verified both directions against a release `mxc-exec-mac`:

- with the old marker restored: `FAIL: hostLoopback=allow under ingress.default=deny is refused (refused, but not for 'cannot enforce a network.ingress.hostLoopback')`
- with the fix: `Seatbelt rejections: 13 assertion(s) passed`

`run_seatbelt_network_test.sh` also passes, 21 assertions.

docs/seatbelt/seatbelt-backend.md
caarlos0: Dropped in e8a4b561 — the whole `Who grants bind()` section is gone.

The one fact it existed to correct survives where a reader actually looks, in the field table: the `ingress.default` row now reads "This single rule is what permits **both `bind()` and `listen()`**; `network-bind` alone grants `bind()` but not `listen()`", and the `egress.default` row notes that only the first of its three rules is egress. The behavior itself stays pinned by the `egress deny + ingress allow` cell in `run_seatbelt_network_test.sh`, so nothing is unverified — it is just no longer narrated.

docs/seatbelt/seatbelt-backend.md
caarlos0: Shortened in e8a4b561. The section is now the peer-scoping limitation, then one short paragraph per pair — why `hostLoopback: "allow"` under `default: "deny"` cannot be expressed, and why the reverse can — plus the existing address-scoping warning. The proxy-confinement argument and the rule listing are gone.

The ingress point moved to the proxy section as you suggested, as a new row in "what is and isn't enforced":

| Can the proxy contain *inbound* traffic? | No — a proxy confines egress only. Inbound is governed solely by `ingress.default`, and Seatbelt cannot scope that grant by peer or by address. |

src/core/wxc_common/src/config_parser.rs
```diff
@@ -4717,7 +4717,7 @@ mod tests {
             observed_counts, expected_counts,
             "explicit divergence inventory and observed category totals differ"
         );
-        let expected_inventory = (353, 329, 14);
+        let expected_inventory = (355, 331, 14);
```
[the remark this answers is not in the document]
caarlos0: Right — the report drifted from the new baseline, fixed in 99aaed95.

`docs/version-specific-parser-migration-inventory.md:60` now reads 331 and states the delta explicitly: the count dropped by one when the 0.9 removal of `seatbelt.launchMethod` retired `27_mac_terminal_sandboxed.json`, and rose by two when `seatbelt_net_egress_deny_ingress_allow.json` and `seatbelt_net_ingress_allow_loopback_deny.json` joined the corpus. Both parsers accept each of those two and build identical runtime models, so neither adds a shared rejection — 14 is unchanged.

I grepped the report for other count claims; line 60 was the only stale one. The file total is not stated there, so it needed no edit.

tests/scripts/run_seatbelt_network_test.sh
```diff
@@ -83,4 +108,19 @@ DENY_LINE=$(grep -n "deny network-outbound.*localhost" <<<"$OUT" | head -1 | cut
 [ "$DENY_LINE" -gt "$ALLOW_LINE" ] || fail "the localhost deny must follow the broad allow (last match wins)" "$OUT"
 pass "the generated profile denies localhost after the broad outbound allow"
 
-summary "Seatbelt directional network"
+# `ingress.default: allow` with `hostLoopback: deny` — a listener without host
+# loopback. The backend permits only this divergent pair; the inverse
+# (`hostLoopback: allow` under `default: deny`) promises inbound host loopback
+# that the profile cannot carry and is still refused.
```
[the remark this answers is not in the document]
caarlos0: Agreed, fixed in 99aaed95. The comment now reads "a listener with the container-to-host direction closed (host-to-container stays open; Seatbelt cannot scope the inbound grant by peer)".

The assertions below it were already direction-specific (`HOSTLOOPBACK_BLOCKED` is the container-to-host probe); only the prose overstated the guarantee.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -270,6 +285,7 @@ This distinction matters, and it's easy to get backwards.
 | Can the sandbox reach anything *other than* the proxy? | **No — kernel-enforced**, provided `ingress.hostLoopback` stays `"deny"` (see the caveat below). |
 | Will a client actually *speak to* the proxy? | Not enforced — cooperative. |
 | Is traffic transparently redirected into the proxy? | No. |
+| Can the proxy contain *inbound* traffic? | No — a proxy confines egress only. Inbound is governed solely by `ingress.default`, and Seatbelt cannot scope that grant by peer or by address. |
```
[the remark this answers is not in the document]
caarlos0: Correct, and it made the PR's own use case look unsupported. Fixed in fd795b7d.

Verified the claim before changing the text: `ingress.default: "allow"` emits only `(allow network-inbound (local ip))` (`profile_builder.rs::write_local_network_rules`) — inbound, no outbound rule. Only `hostLoopback: "allow"` emits `(allow network-outbound (remote ip "localhost:*"))`. The live suite already pins this with the `ingress.default=allow leaves the egress deny intact` assertion.

The warning now requires only `hostLoopback: "deny"` for proxy-only egress, and a second paragraph describes the inbound tradeoff separately: `{"default": "allow", "hostLoopback": "deny"}` keeps proxy-only egress while permitting a listener, at the cost of an inbound grant that cannot be scoped by peer or address.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -559,7 +575,7 @@ the rule applies to both.
 | Config | Why it's rejected | Do this instead |
 |---|---|---|
 | `egress.allow` / `egress.deny` (non-empty) | No CIDR/port/protocol filtering primitive | Use `egress.default` alone |
-| `ingress.hostLoopback` ≠ `ingress.default` | Only the outbound half is expressible; see [the trap](#the-hostloopback-trap) | Set both to the same value |
+| `ingress.hostLoopback: "allow"` + `ingress.default: "deny"` | The inbound half is not expressible, so the promised host-to-container grant could not be made; see [the trap](#the-hostloopback-trap) | Set both to `"allow"` |
```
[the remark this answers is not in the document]
caarlos0: Fair — fixed in fd795b7d. The introduction now names the exception before claiming the table is complete:

> One documented exception: `ingress.hostLoopback: "deny"` under `ingress.default: "allow"` is accepted with the container-to-host half enforced and the host-to-container half left open, because Seatbelt cannot scope an inbound grant by peer. See [the trap](#the-hostloopback-trap).

While checking the anchor I also found a dangling link this PR had introduced: renaming the section to "Why only one divergent pair is allowed" left `[above](#why-it-must-equal-ingressdefault)` pointing at nothing. That is fixed too, and I verified every in-file anchor now resolves against an existing heading.

theelliotm (approved):

bbonaby (approved):

PR #1177, opened 2026-09-16
Fix PSEC proxy loopback peer serialization

caarlos0:
## Summary

When `runtimeConfig.networkProxy` is combined with explicit PSEC 1.1 ingress, MXC serializes `allowed_appcontainer_peer` directly from `processContainer.network.allowedProxyPeer`. For unrestricted host loopback that field is intentionally absent, so the generated process-security-environment policy contains no peer identity and `CreateProcessSecurityEnvironment` rejects it with `E_INVALIDARG`.

This change uses the existing `allowed_appcontainer_peer` helper for PSEC 1.1 ingress as well as PSEC 1.0. It emits MXC's reserved `MXC-Loopback` sentinel for unrestricted host loopback while preserving an explicitly configured proxy peer. Runtime-proxy denial capture therefore remains on native PSEC rather than falling through to a backend that cannot enforce the network policy.

## Reproduction

Save this as `proxy-capture.json` on a Windows host with native capture support:

```json
{
  "version": "0.8.0-alpha",
  "containment": "processcontainer",
  "process": {
    "commandLine": "cmd.exe /c echo sandbox-started"
  },
  "network": {
    "egress": { "default": "deny" },
    "ingress": {
      "default": "allow",
      "hostLoopback": "allow"
    }
  },
  "runtimeConfig": {
    "networkProxy": "http://127.0.0.1:8080"
  },
  "processContainer": {
    "captureDenials": { "mode": "block" }
  }
}
```

Run:

```powershell
wxc-exec.exe --config proxy-capture.json
```

Before this change, startup fails before the child launches:

```text
captureDenials: failed to start learning-mode capture: CreateProcessSecurityEnvironment failed (HRESULT = 0x80070057)
```

The same malformed PSEC network policy affects ordinary runtime-proxy launches with explicit ingress; capture exposes it while creating the learning-mode security environment.

## Validation

- Added a unit regression proving PSEC 1.1 unrestricted host-loopback ingress serializes `allowed_appcontainer_peer = "MXC-Loopback"`; it failed before the fix with `left: None` and passes after it.
- Added a native-capture selection regression for `runtimeConfig.networkProxy` plus `captureDenials`.
- Added an ignored Windows ProcessContainer E2E that asserts the proxied native-capture child launches. Live execution requires the brokered learning-mode API, PSEC 1.1, and a compatible host; the normal Windows suite compiles it.
- `cargo test -p appcontainer_common` (320 passed)
- Focused `wxc_e2e_tests` compilation passed.
- `cargo fmt --all -- --check` and `cargo clippy -p appcontainer_common --all-targets -- -D warnings` passed earlier on the branch with the pinned Rust 1.93 toolchain.

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1177)


src/backends/process_container/common/src/base_container_runner.rs
```diff
@@ -951,6 +951,7 @@ impl BaseContainerRunner {
 
         request.policy.capture_denials.is_some()
             && native_capture_usable
+            && !request.policy.network_proxy.is_enabled()
```
bbonaby: issue: wait, actually this should work. You should be able to set a proxy and be able to run capture denials.

I think the issue is on this line: https://github.com/microsoft/mxc/blob/7f5d19ced35757bd13b63ff0b314ec4ebb005d7b/src/backends/process_container/common/src/base_container_helpers.rs#L192. You see how it just has 

```rust
network.allowed_appcontainer_peer = policy.allowed_proxy_peer.clone();
```
it should be just like in the else case below and should be
```rust
network.allowed_appcontainer_peer = allowed_appcontainer_peer(policy);
```

instead.  allowed_appcontainer_peer isn't allowed to be empty so mxc passes in a sentinel value.

Can test this out to confirm?

bbonaby (changes requested):

src/backends/process_container/common/src/base_container_runner.rs
caarlos0: Confirmed — this is the actual failure. I added an assertion to the existing PSEC 1.1 ingress spec test before changing production code; it failed with `allowed_appcontainer_peer()` returning `None` instead of `Some("MXC-Loopback")`, matching the malformed input that reaches `CreateProcessSecurityEnvironment`.

I pushed 38a4e7e8, which uses `allowed_appcontainer_peer(policy)` in the ingress branch, restores runtime-proxy capture to native PSEC, and updates the E2E/docs accordingly. `cargo test -p appcontainer_common` passes all 320 tests and the Windows E2E compiles.

bbonaby (approved):

PR #1080, opened 2026-09-01
Enforce ProcessContainer ingress policy

bbonaby:
## 📖 Description

Connects ProcessContainer `network.ingress.default` and `network.ingress.hostLoopback` to the OS Process Security Environment (PSEC) ingress contract. MXC emits PSEC 1.1 only when `IsProcessSecurityEnvironmentVersionSupported` reports contract 1.1 or newer and `QueryProcessSecurityEnvironmentSupport` advertises `PSE_SUPPORT_NETWORK_INGRESS`. Otherwise, compatible policies retain PSEC 1.0 or legacy SBOX capability behavior.

The optional version export is resolved during the memoized PSEC API load so repeated SDK/FFI launches do not acquire additional `processmodel.dll` references. OS support-query failures remain retryable `backend_unavailable` errors. When PSEC 1.1 ingress is unavailable, `hostLoopback: allow` is rejected because older contracts cannot represent it. A final compatibility check prevents a late PSEC-to-SBOX fallback from dropping PSEC-only network policy.

## 🔗 References

- Related contract: #1076

## 🔍 Validation

- `cargo fmt --manifest-path src\Cargo.toml --all -- --check`
- `cargo test --manifest-path src\Cargo.toml -p learning_mode_windows --lib` (229 passed)
- `cargo test --manifest-path src\Cargo.toml -p appcontainer_common --lib` (293 passed)
- `cargo clippy --manifest-path src\Cargo.toml -p learning_mode_windows -p appcontainer_common --all-targets -- -D warnings`
- `cargo build --manifest-path src\Cargo.toml -p wxc`
- Real schema 0.8 ProcessContainer connectivity: container-to-host and host-to-container connections succeeded with `hostLoopback: "allow"` and were blocked with `hostLoopback: "deny"`

## ✅ Checklist

- [x] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [ ] Linked to an issue
- [x] Updated documentation (if applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed)
- [ ] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes (see [docs/pull-requests.md](https://github.com/microsoft/mxc/blob/main/docs/pull-requests.md))

## 📋 Issue Type

- [ ] Bug fix
- [x] Feature
- [ ] Task

PR #1180, opened 2026-09-16
fix(process-container): restore unprivileged host proxy connectivity

caarlos0:

-----

## Summary

Restore unprivileged client-to-host proxy connectivity for ProcessContainer policies using an unpackaged host proxy.

The peer fix in #1177 is necessary but insufficient on Windows build 26691.1002:

- Combining a native PSEC ingress table with a proxy makes `CreateProcessSecurityEnvironment` return `0x80070057`.
- Omitting the table allows creation but leaves the proxy unreachable without the explicit `networkLoopback` capability removed during #1080.

This change omits the native ingress table only in proxy mode and restores `networkLoopback` only when the policy allows host loopback. It retains the proxy endpoint, `MXC-Loopback` peer, private-network capability, and existing contract-version gate. Direct-egress ingress and identity-scoped proxy permissions remain unchanged.

The proxy may run in the caller's host process, outside the sandboxed workload. No elevation, firewall changes, or additional proxy process is required.

The branch contains one commit replayed onto current `main`.

## Known blocker

This PR is **not merge-ready**. `runtime_proxy_preserves_allowed_host_to_container_loopback` remains failing: the host cannot connect to a listener inside the proxy-configured sandbox. The same unprivileged full-ingress non-proxy control succeeds, and a PSEC 1.0 proxy probe behaves the same way.

Is host-to-container ingress supported in Windows proxy mode, and what native representation preserves it? Otherwise, the advertised bidirectional `hostLoopback: "allow"` contract needs an explicit backend limitation. The regression remains enabled rather than ignored or weakened.

## Related work

- Follow-up to #1177.
- Restores the proxy-specific capability removed in #1080.

## Validation

On Windows `processmodel.dll` 10.0.26691.1002:

- Added native creation and serialization regressions first; both failed before the fix, including the exact reported HRESULT `0x80070057`.
- Native creation, child execution, proxy TCP connectivity, alternate IPv4/IPv6 loopback denial, identity-scoped serialization, and native denial capture pass without elevation or firewall exceptions.
- The focused backend suite reports **7 passed, 1 failed, 0 ignored**; the single failure is the inbound blocker above.
- The IPv6 case checks bypass denial with an IPv4 proxy; it does not claim IPv6 proxy endpoints work.
- `cargo fmt --all -- --check` passes with the pinned Rust 1.93 toolchain.
- `cargo clippy --workspace --all-targets -- -D warnings` passes.

Current-history checks confirm the PR is one commit ahead of `main` and has no whitespace errors.


docs/process-container/networking.md
```diff
@@ -103,15 +103,26 @@ client policy and package identity are the same; the enforcement table below sep
...
+The public ingress policy remains required, and host-loopback allow retains the PSEC 1.1 host requirement.
+Requests that do not allow host loopback use the
 PSEC 1.0 capability mapping; `hostLoopback: "allow"` is rejected when the PSEC 1.1 ingress contract is unavailable.
 The legacy SBOX path remains eligible only when its capability mapping can preserve the request.
 
+**Unresolved ingress limitation:** on Windows build 26691.1002, this proxy mapping permits unprivileged connections
+to the configured proxy and blocks direct loopback bypasses, but host-to-container listener connections still time
+out. The same listener succeeds under a non-proxy PSEC ingress policy. The proxy mapping is therefore not yet a
+complete implementation of the shared bidirectional `hostLoopback: "allow"` contract.
+
```
bbonaby, line 125: note: I'll approve  but I think this paragraph will eventually need to get updated, as it is fixed on newer builds. Can't remember if Windows build 26691.1002 has the loopback fix or not that will go out for 9D. If not then those machines will evetually get it in the coming weeks at least.

To test, I tried it with curl.exe. I added a server loopback config to this comment. when you run that via wxc-exec it'll print out the port number it's listening on.

Then in a separate powershell window run:

```powershell
curl.exe --silent --show-error --max-time 10 --fail --header `
"Accept: application/json" --header "Content-Type: application/json" --data `
"{\`"jsonrpc\`":\`"2.0\`",\`"id\`":1,\`"method\`":\`"tools/call\`",\`"params\`":{\`"name\`":\`"fetch_example\`",\`"arguments\`":{}}}" `
http://127.0.0.1:52069/mcp

```
You should then see the following output in powershell `curl: (52) Empty reply from server`. You'll have to change the port number in the powershell script above based on what your sandbox prints out before running thre script.

[processcontainer_loopback_server.json](https://github.com/user-attachments/files/32310305/processcontainer_loopback_server.json)

bbonaby (approved):

PR #1182, opened 2026-09-16
BaseContainer: Remove legacy experimental CreateProcessInSandbox (SBOX)

jsidewhite:
(CreateProcessSecurityEnvironment (PSEC) completely usurps it.)

Closes #1183
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1182)


bbonaby (dismissed):
end of an era

bbonaby (dismissed):

bbonaby (approved):

src/core/wxc_common/src/models.rs
```diff
@@ -1185,7 +1185,7 @@ pub struct ScriptResponse {
     pub standard_err: String,
     pub error_message: String,
     /// Raw system/API error detail intended for developers and diagnostics
-    /// (e.g. "Experimental_CreateProcessInSandbox failed: WIN32_ERROR(1920)").
+    /// (e.g. "CreateProcessSecurityEnvironment failed: HRESULT(...)").
```
MGudgin, line 1188: **Low (maintainability) - The new error example is not an emitted format.**

**Attribution:** `introduced_by_change` — this line replaces the prior SBOX
example, but neither surviving PSEC failure path emits
`CreateProcessSecurityEnvironment failed: HRESULT(...)`.

The launch path reports
`CreateProcessW(PROC_THREAD_ATTRIBUTE_SECURITY_ENVIRONMENT) failed:
WIN32_ERROR(...)`; environment creation reports
`failed to create the process security environment: ...`. The illustrative
string will mislead anyone searching diagnostics for the documented format.

**Fix:** Use one of the actual emitted formats in this example.

tests/regression_tests/TestCases.psd1
```diff
@@ -99,24 +99,24 @@
...
         Script = "test_cases\Invoke-Issue1109-BaseContainerVolumeRoot.ps1"
-        ExpectedTierSupport = @("base-container-psec", "base-container-sbox")
+        ExpectedTierSupport = @("base-container-psec")
         Prerequisites = @("wxc-exec", "BaseContainer-capable host")
         HarnessArguments = "Default"
         Destructive = $false
     }
     "1130" = @{
-        FriendlyName = "SBOX Environment Fallback"
+        FriendlyName = "PSEC Sparse Environment Rejection"
```
MGudgin, line 116: **Low (maintainability/test coverage) - The rewritten PSEC regression retains
an SBOX-specific filename.**

**Attribution:** `newly_exposed_by_change` — this PR changes the test's
friendly name, purpose, output, and expected tier to PSEC-only behavior, but
the adjacent `Script` entry still points to
`Invoke-Issue1130-SboxSparseEnvironment.ps1`.

That stale name makes repository searches suggest a surviving SBOX regression
after the implementation and test intent have moved to PSEC.

**Fix:** Rename the script to
`Invoke-Issue1130-PsecSparseEnvironment.ps1` and update the catalog path.

src/backends/process_container/common/src/dispatcher.rs
```diff
@@ -481,13 +480,7 @@ fn select_backend_with_fallback(
...
             // when the OS enforces `fs_deny` natively, so there is nothing
             // for a host DACL to add.
-            let runner = if guarded_capture_required {
-                BaseContainerRunner::new().with_guarded_capture_factory(Arc::clone(
-                    capture_factory.expect("guarded capture factory checked above"),
-                ))
-            } else {
-                BaseContainerRunner::new()
-            };
+            let runner = BaseContainerRunner::new();
```
MGudgin, line 483: **Low (test coverage) - Pin the invariant that makes this plain
BaseContainer construction safe for `captureDenials`.**

**Attribution:** `newly_exposed_by_change` — the PR removes the guarded
capture factory from the BaseContainer arm. Correctness now depends on
BaseContainer usability and native-capture eligibility agreeing for every
`captureDenials` request.

The implementation currently enforces that relationship and runner validation
rechecks it, but no focused test documents that this arm cannot be selected
with `guarded_capture_required=true`.

**Fix:** Add a test or assertion that, for every `captureDenials` request,
BaseContainer usability equals native-capture eligibility and a non-native
result cannot select this arm.

MGudgin (commented):
## Summary

I reviewed the current PR head `baac8f3` against its current base `3fd44f9`
across security, reliability, performance, maintainability, correctness,
testability, and test coverage.

### Findings outside the inline diff

**Medium (correctness) - LPAC requests that prohibit DACL mutation now fail.**
`src/backends/process_container/common/src/base_container_runner.rs:575-579`
— `introduced_by_change`: PSEC still rejects
`processContainer.leastPrivilege`, while the deleted SBOX path accepted that
policy. The request now reaches Tier 3, where
`ensure_dacl_augmentation_allowed` requires
`fallback.allowDaclMutation=true` even with no filesystem policy. A previously
supported LPAC request that explicitly forbids host ACL changes therefore
fails before launch.

**Fix:** Preserve a native LPAC-capable route until PSEC supports LPAC, or make
the compatibility break explicit and versioned instead of requiring callers
to relax their no-mutation policy.

**Medium (correctness) - LPAC fallback can omit explicitly requested
filesystem grants.**
`src/backends/process_container/common/src/dispatcher.rs:298-303`
— `newly_exposed_by_change`: the redundant-grant filter is pre-existing, but
this PR newly routes LPAC requests through it. The filter counts access from
ALL APPLICATION PACKAGES, while LPAC startup explicitly opts out of that
group. A path whose access comes only from that group is removed from the
per-container ACE list, leaving the LPAC workload unable to use the requested
path.

**Fix:** Make the grantability precheck and redundant-grant filter LPAC-aware;
do not count ALL APPLICATION PACKAGES access when `leastPrivilege` is enabled.
Add a regression test with a path granted only to that group.

**Medium (security/performance) - SBOX-compatible requests silently move to
host-DACL-mutating Tier 3.**
`src/backends/process_container/common/src/fallback_detector.rs:365-413`
— `introduced_by_change`: legacy proxy policies, LPAC, and supported hosts
without usable PSEC previously had a native SBOX route. With SBOX removed and
Tier 2 disabled in shipping builds, they fall through to Tier 3. This expands
the host-side ACL mutation surface and adds synchronous per-path ACL setup,
state persistence, and teardown. The generic `BaseContainerUnavailable`
reason does not identify the policy incompatibility that caused the
degradation.

**Fix:** Emit a policy-specific warning for the Tier-1-to-Tier-3 transition,
add dispatcher coverage for LPAC and legacy proxy requests, and benchmark
launch/teardown against the base revision. If the posture or latency is
unacceptable, retain or enable a non-DACL path.

### Verified clean

- The current merge base exactly matches the PR base (`3fd44f9`), and the
  current diff remains 59 files, +446/-4980.
- The current head changed after the initial review by merging `main`, but the
  source files supporting these findings are unchanged from the reviewed head.
- The reliability and testability passes found no regressions in cleanup,
  fail-closed capture behavior, or deterministic test seams.
- Review passes reported `cargo test -p appcontainer_common --lib` passing all
  322 tests and `cargo check -p appcontainer_common` completing cleanly.

This is a `COMMENT` review rather than a blocking review because the surviving
findings are Medium or Low.

Issue #1165, opened 2026-09-15
inheritDefaultEnv is not honored on IsolationSession or WSLc


theelliotm:
### Relevant area(s)

Windows

### Brief description of your issue

#1153 gave LXC, Bubblewrap, and Seatbelt the schema-0.9 `process.inheritDefaultEnv` contract that #1120 shipped for the Windows ProcessContainer tiers. IsolationSession and WSLc were left out and still flatten the four states into two.

The contract the other backends now honor:

| `process.env` | `inheritDefaultEnv` | Result |
| --- | --- | --- |
| omitted | — | the backend default |
| `[]` | — | empty |
| `["FOO=bar"]` | `false` (default) | verbatim |
| `["FOO=bar"]` | `true` | layered over the default; a caller entry wins |

**IsolationSession** (`src/backends/isolation_session/common/src/process_options.rs:179-187`) fetches `proc_options.Environment()` and *inserts* into it, so it behaves as permanently `inheritDefaultEnv: true`. A caller who asks for a verbatim environment — which is the default — silently gets the session''s full default environment merged in. `IMap::Clear()` is available for the replace case.

**WSLc** (`src/backends/wslc/common/src/wsl_container_runner.rs:1490`, `src/backends/wslc/common/src/container_steps.rs:404`) guards the SDK call with `if !effective_env.is_empty()`, flattening omitted and `[]` together. Whether `WslcSetProcessSettingsEnvVariables` replaces or augments the image environment is undocumented — `wslcsdk.h:343` has no comment — so that needs to be established first. If it augments, `Explicit` may not be honorable and should be rejected rather than silently downgraded.

Both backends currently use `ExecutionRequest::env_entries()`, whose doc comment now names them as the remaining callers.

### Steps to reproduce

1. Build with `--features isolation_session`.
2. Run a `0.9.0-alpha` config against `containment: "isolationsession"` with `process.env` set to `["FOO=bar"]` and no `inheritDefaultEnv`.
3. Have the command print its full environment.

### Expected behavior

Only `FOO` is present — the environment is used verbatim, as it is on every other backend at 0.9.

### Actual behavior

`FOO` is present alongside the isolation session''s entire default environment.

PR #1195, opened 2026-09-17
0.9: Honor inheritDefaultEnv on LXC, Bubblewrap, and Seatbelt

### LINKED ISSUE #1153: Default PATH behavior is inconsistent across Linux and macOS backends
### Relevant area(s)

Linux, macOS

The non-Windows containment backends do not share a documented contract for the default `PATH` supplied to sandboxed workloads. This issue is split from #1102 after PR #1120 aligned the Windows ProcessContainer environment behavior.

Current behavior still differs by backend:

| Backend | `process.env` omitted | `process.env` set, no `PATH=` |
|---|---|---|
| Seatbelt (macOS) | `/usr/bin:/bin:/usr/sbin:/sbin` | same default |
| LXC (Linux) | inherits host `PATH` | liblxc default `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` |
| Bubblewrap (Linux) | no `PATH` supplied by MXC | no `PATH` supplied by MXC |

Bubblewrap clears the environment without restoring `PATH`, so command resolution falls back to the shell's distro-specific compiled default. This has produced different results between Debian/Ubuntu and RHEL. LXC also changes its `PATH` behavior when an unrelated environment variable is added because a non-empty `process.env` switches it from inherited environment behavior to liblxc's cleared-environment default. Seatbelt supplies a stable default, but that contract is not shared or documented across the non-Windows backends.

### Steps to reproduce

1. Build MXC on Linux or macOS.
2. Run a config that prints `PATH` and resolves a command normally found under `sbin`, with `process.env` omitted.
3. Add an unrelated variable such as `"env": ["FOO=bar"]` and run the same config again.
4. Compare Bubblewrap and LXC on Debian/Ubuntu and RHEL, and compare Seatbelt on macOS.

Example process configuration:

```json
"process": {
  "commandLine": "bash -c 'echo \"PATH=[${PATH-<unset>}]\"; command -v iptables || echo NO_IPTABLES'"
}
```

### Expected behavior

Linux and macOS backends follow a single documented default-environment contract. Adding an unrelated environment variable does not unexpectedly change command resolution.

### Actual behavior

Bubblewrap can leave `PATH` unset, LXC switches between the host path and liblxc's default based on whether `process.env` is non-empty, and Seatbelt applies a separate fixed default. Command resolution therefore varies by backend and host distribution.

Additional context: nightly run [33731889129](https://github.com/microsoft/mxc/actions/runs/33731889129) failed on RHEL 10 while the Debian/Ubuntu Bubblewrap jobs passed because their shell fallback paths included `sbin`.

theelliotm:
## 📖 Description

The Linux and MacOS backends had no standard default environment. This PR gives all three the same schema-0.9 contract the Windows process container got in #1120, implemented per-backend:

* An omitted `process.env` gets a default block of `PATH`, `HOME`, and `TERM`.
* An explicitly empty `process.env` stays empty, distinct from omitted.
* A supplied `process.env` is used verbatim.
* `inheritDefaultEnv` layers a supplied environment over the default, with a caller entry replacing the same-named default rather than duplicating it.

`PATH` is per-OS: Seatbelt keeps its existing value, Linux gets the `sbin` directories. `HOME` is the resolved working directory (else `/tmp`), since the child runs as the launching uid and that user's real home is not reachable under the sandbox's filesystem policy.

### Scope

Behavior below 0.9 is unchanged on all three. IsolationSession and WSLc are not covered and are tracked separately.

Seatbelt's resolution lives in a new host-agnostic `default_env` module, since `seatbelt_runner` is `target_os = "macos"` and would otherwise be untested off a Mac.

IsolationSession and WSLc are not covered — see #1165.

## 🔗 References
Resolves #1153

Related:
- #1120 — the Windows ProcessContainer contract this mirrors
- #1165 — `inheritDefaultEnv` on IsolationSession and WSLc, still open

## 🔍 Validation

* Unit tests per backend for all four states plus the sub-0.9 passthrough
* `run_bwrap_environment_test.sh` and `run_lxc_env_09_test.sh`, wired into the suite runners; `run_seatbelt_environment_test.sh` extended
* cargo fmt --all -- --check
* cargo clippy --workspace --all-targets -- -D warnings
* cargo clippy for x86_64-unknown-linux-gnu and aarch64-apple-darwin
* cargo test for the touched crates
* node scripts/versioning/validate-configs.js

`Cargo.lock` changes only because `lxc_common` and `seatbelt_common` now depend on `semver` to parse the schema version. No new crate enters the graph — `semver = "1"` is already a direct dependency of `wxc_common`, `bwrap_common`, `mxc_ffi`, and `isolation_session_bindings`.

## ✅ Checklist
- [x] Linked to an issue

## 📋 Issue Type
- [x] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1195)


src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -832,26 +833,24 @@ fn build_helper_script(
...
+/// [`resolve_environment`]). `PWD` is set separately alongside the cwd.
 fn apply_clean_environment(
     command: &mut Command,
     request: &ExecutionRequest,
     proxy_address: Option<&ProxyAddress>,
 ) {
     command.env_clear();
-    command.env("PATH", DEFAULT_SANDBOX_PATH);
+    // From 0.9 `resolved_env` carries `PATH`, and an explicitly empty
+    // `process.env` must stay empty rather than keep a floor under it.
```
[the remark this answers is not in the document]
theelliotm: Documented as an exception. Seatbelt needs it to be included to prevent "operation not permitted" errors when walking up the working directory tree. See PR: https://github.com/microsoft/mxc/pull/688.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -559,6 +559,90 @@ pub(crate) fn local_network_diagnostic_for_mode(
...
+
+/// The directory the child is actually started in, if any.
+///
+/// [`build_args_classified_with_mode`] emits `--chdir` for exactly this value
+/// and [`default_env`] points `HOME` at it, so the two cannot name different
+/// directories. A policy grant is deliberately *not* consulted: bwrap enters
+/// one only when `process.cwd` names it, so treating it as the start directory
+/// would put `HOME` somewhere the child never went.
+fn start_directory(request: &ExecutionRequest) -> Option<&str> {
+    Some(request.working_directory.as_str()).filter(|dir| !dir.is_empty())
```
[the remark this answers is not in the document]
theelliotm: Relative cwd get rejected in 0.9 PR: #1147

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -23,6 +23,91 @@ use crate::signal_cleanup;
...
+
+/// The directory the child is actually started in, if any.
+///
+/// [`LxcContainer::attach_run`] wraps the command in a `cd` for exactly this
+/// value and [`default_env`] points `HOME` at it, so the two cannot name
+/// different directories. A policy grant is deliberately *not* consulted: with
+/// `process.cwd` omitted the child starts at the container root, so treating a
+/// grant as the start directory would put `HOME` somewhere it never went.
+fn start_directory(request: &ExecutionRequest) -> Option<&str> {
+    Some(request.working_directory.as_str()).filter(|dir| !dir.is_empty())
```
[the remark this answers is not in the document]
theelliotm: Relative cwd get rejected in 0.9 PR: #1147

src/backends/seatbelt/common/src/seatbelt_runner.rs
```diff
@@ -234,13 +235,23 @@ fn spawn_exec(
     // Always start from a cleared environment so untrusted sandboxed code never
     // inherits the host's env. When a proxy is active its HTTP_PROXY/HTTPS_PROXY
     // vars are injected here (and caller-supplied proxy vars stripped).
-    apply_clean_environment(&mut command, request, proxy.address());
+    let resolved_cwd = resolved_working_directory_opt(request);
+    apply_clean_environment(
+        &mut command,
+        request,
+        proxy.address(),
+        resolved_cwd.as_deref(),
```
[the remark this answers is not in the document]
theelliotm: Relative cwd get rejected in 0.9 PR: https://github.com/microsoft/mxc/pull/1147

docs/lxc-support/lxc-backend.md
```diff
@@ -92,6 +92,25 @@ Any combination that lxc-create supports.
...
+
+By default, the backend supplies `PATH`
+(`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`), `HOME` (the
+directory the child is started in, else `/tmp`), and `TERM`
+(`xterm-256color`).
+
+| `process.env` | `inheritDefaultEnv` | Result |
+|---------------|---------------------|--------|
+| omitted | — | the default block |
+| `[]` | — | nothing |
```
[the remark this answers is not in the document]
theelliotm: Read the next paragraph.

src/backends/seatbelt/common/src/default_env.rs
```diff
@@ -0,0 +1,188 @@
...
+use wxc_common::models::ExecutionRequest;
+
+/// Baseline `PATH` for the sandboxed child. We always start from a cleared
+/// environment (so the host process's env — cloud creds, API tokens — never
+/// leaks into untrusted sandboxed code), which means we must supply a default
+/// `PATH` for the `/bin/sh` wrapper and common tools to resolve.
+pub const DEFAULT_SANDBOX_PATH: &str = "/usr/bin:/bin:/usr/sbin:/sbin";
+
+/// `HOME` when the request resolves no working directory.
+pub const FALLBACK_HOME: &str = "/tmp";
```
SohamDas2021: The generated profile is `deny default` and only grants `/private/tmp` under `guiAccess`, so without GUI access a workload that writes anything under `$HOME` gets `EPERM`, and *with* it `HOME` points at shared host `/private/tmp`, where another local user can preplant `.npmrc`/`.gitconfig`/`.ssh/config` that the sandbox then reads. 

Suggest a per-run mode-0700 directory granted explicitly in the profile, plus a non-GUI test that writes under `$HOME`.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -528,6 +528,90 @@ pub(crate) fn local_network_diagnostic_for_mode(
...
+    ]
+}
+
+/// The entries the child should get, as `KEY=VALUE` strings.
+///
+/// From schema 0.9 the four states of `process.env` stay distinct: omitted
+/// takes the default, `[]` is empty, a supplied environment is used verbatim,
+/// and `inheritDefaultEnv` layers a supplied environment over the default.
+/// Below 0.9 the caller's entries are passed through untouched.
+fn resolved_env(request: &ExecutionRequest) -> Vec<String> {
```
SohamDas2021, line 582: [Follow-up] This function is identical to `lxc_runner.rs:82` and `seatbelt/default_env.rs:60`. With Windows that's four copies of one contract.

Not in this PR, but file a follow-up ADO task to refactor and share? Do not file any more gh issues for now.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -717,6 +801,160 @@ mod tests {
...
+        }
+
+        fn value<'a>(entries: &'a [String], key: &str) -> Option<&'a str> {
+            entries
+                .iter()
+                .find_map(|kv| kv.strip_prefix(&format!("{key}=")))
+        }
+
+        #[test]
+        fn below_0_9_nothing_is_supplied() {
```
SohamDas2021, line 819: Worth adding `r.env = Some(vec!["FOO=bar".into()])` asserting exact passthrough?

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -23,6 +23,91 @@ use crate::signal_cleanup;
...
+}
+
+/// The default environment, from schema 0.9: `PATH`, `HOME`, and `TERM`.
+///
+/// `HOME` names the directory the child actually runs in, so it is a path the
+/// container has rather than a host path that was never mounted. With no start
+/// directory it is [`FALLBACK_HOME`], which every image provides writable.
+fn default_env(request: &ExecutionRequest) -> Vec<(String, String)> {
```
SohamDas2021, line 69: `process.cwd` may be relative, right? This copies the raw string into `HOME`. Same issue on Seatbelt runner. We should resolve to the absolute launch path

src/backends/bubblewrap/common/src/bwrap_command.rs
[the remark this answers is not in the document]
SohamDas2021: Do we know if 1147 will go in by 1.0?

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -528,6 +528,90 @@ pub(crate) fn local_network_diagnostic_for_mode(
...
+    }
+
+    let entries = match (&request.env, request.inherit_default_env) {
+        (None, _) => default_env(request),
+        (Some(supplied), false) => return supplied.clone(),
+        (Some(supplied), true) => {
+            let mut entries = default_env(request);
+            // A caller entry replaces the same-named default rather than
+            // being appended: `--setenv` would otherwise be passed twice.
+            for (key, value) in supplied.iter().filter_map(|kv| kv.split_once('=')) {
```
SohamDas2021, line 594: `filter_map(split_once('='))` silently drops an entry with no `=`, so `inheritDefaultEnv: true` with `["FEATURE_FLAG"]` loses it with no diagnostic and the path at `:594` passes that same entry straight through. The two modes disagree on identical input. Either reject malformed entries at validation with a clear message or handle them identically in both paths.

src/backends/seatbelt/common/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license.workspace = true
 # the ScriptRunner trait that seatbelt_runner implements. Same dep that
 # lxc_common uses.
 wxc_common = { workspace = true }
+semver = "1"
```
SohamDas2021: [Follow-up] `supports_default_env` is implemented identically in all three backends, `wxc_common` is already a dependency of all three and already depends on `semver` for gates like `supports_directional_network`. Refactor and share? Fold it into the refactor ADO item you filed earleir to follow up later.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -717,6 +801,160 @@ mod tests {
...
+        fn a_supplied_env_is_used_verbatim() {
+            let mut r = request("0.9.0-alpha");
+            r.env = Some(vec!["FOO=bar".into()]);
+            assert_eq!(resolved_env(&r), vec!["FOO=bar".to_string()]);
+        }
+
+        #[test]
+        fn inherit_default_env_layers_over_the_default_block() {
+            let mut r = request("0.9.0-alpha");
+            r.env = Some(vec!["FOO=bar".into(), "PATH=/only/mine".into()]);
```
SohamDas2021, line 876: The replacement test uses a single well-formed entry. The `split_once`/replace logic also has edges worth pinning: the same key supplied twice, a bare entry with no `=`, and an empty value like `PATH=` (which should still replace, not be skipped). Shall we add them for each backend?

tests/scripts/run_bwrap_environment_test.sh
```diff
@@ -0,0 +1,108 @@
...
+    echo "PASS: $1"
+    PASS_COUNT=$((PASS_COUNT + 1))
+}
+
+run_config() {
+    RC=0
+    OUT=$("$LXC_EXEC" --experimental "$CONFIG_DIR/$1" 2>&1) || RC=$?
+}
+
+expect_ok() {
```
SohamDas2021, line 46: Every assertion here is an expected-success case. Nothing checks that sub-0.9 `inheritDefaultEnv` is rejected with a nonzero exit and no workload sentinel. The parser unit test covers the error text but not the CLI exit contract. An `expect_rejected` helper plus one config would close it.

docs/seatbelt/seatbelt-backend.md
```diff
@@ -407,19 +407,45 @@ baseline `/Library` and `/System` allows.
...
+> `"inheritDefaultEnv": true` to get the old behavior, or supply `PATH`
+> yourself.
+
+Tools installed outside the default `PATH` need both an env entry **and** a
+`readonlyPaths` grant — e.g. Homebrew on Apple silicon needs
+`"PATH=/opt/homebrew/bin:…"` plus `readonlyPaths: ["/opt/homebrew"]`.
+
+### Before schema 0.9
+
+`PATH` defaults to `/usr/bin:/bin:/usr/sbin:/sbin` and each `process.env` entry
```
SohamDas2021, line 450: The old guide had a prominent warning that `$HOME` is unset inside the sandbox unless passed explicitly. This rewrite documents pre-0.9 `PATH` but drops that, so a reader on 0.8 no longer learns that `HOME` and `TERM` are not populated by MXC. Worth one sentence here?

src/backends/seatbelt/common/Cargo.toml
theelliotm: `semver` removed, now using `ContractVersion` enum in a shared `supplies_default_env` function in models.rs:1109

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: They now share properly.

src/backends/seatbelt/common/src/default_env.rs
theelliotm: Going to simply omit HOME when cwd doesn't resolve.

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: done

src/backends/lxc/common/src/lxc_runner.rs
theelliotm: Relative cwds are officially unsupported and about to be blocked. Worth waiting for Gudge?

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: Lifted to default_env.rs

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: Added

tests/scripts/run_bwrap_environment_test.sh
theelliotm: Added

docs/seatbelt/seatbelt-backend.md
theelliotm: Added

src/core/wxc_common/src/models.rs
```diff
@@ -972,33 +972,36 @@ pub struct ExecutionRequest {
...
     /// Three states, deliberately distinct:
     ///
     /// * `None` — the caller supplied no environment. Backends provide a
-    ///   default: on Windows, the user's profile block.
-    /// * `Some(vec![])` — the caller asked for an *empty* environment. This is
-    ///   not the same as `None`, and on the Windows process container it is
-    ///   expected to fail at process creation, because the OS requires certain
-    ///   names to be present (see `REQUIRED_CHILD_ENV_VARS`).
+    ///   default: on Windows, the user's profile block; on LXC, Bubblewrap, and
+    ///   Seatbelt, `PATH` + `HOME` + `TERM`.
```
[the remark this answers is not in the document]
theelliotm: Done

src/backends/lxc/common/src/lxc_runner.rs
theelliotm: Resolved

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -528,6 +528,85 @@ pub(crate) fn local_network_diagnostic_for_mode(
...
+}
+
+/// The default environment: `PATH`, `HOME`, and `TERM`.
+///
+/// `HOME` names the directory the child actually runs in, so it is a path the
+/// sandbox can reach rather than the launching user's real home, which the
+/// bind-mount policy would not have made visible. With no start directory it
+/// is [`FALLBACK_HOME`], which `build_args` always mounts as a fresh tmpfs.
+fn default_env(request: &ExecutionRequest) -> Vec<(String, String)> {
+    let home = start_directory(request)
```
MGudgin: **High (correctness) - Normalize relative `cwd` before assigning `HOME`.**

**Attribution:** `introduced_by_change` - this PR adds the default `HOME` assignment from the raw `process.cwd` value.

For `cwd: "work"`, the generated arguments contain both `--chdir work` and `HOME=work`. After the child changes into `work`, `HOME` is still relative, so a program resolving `$HOME/file` from the new current directory can target `work/work/file`. The strings are equal, but they no longer name the same directory, contradicting the invariant documented above.

The current `home_and_chdir_agree` test covers only `""` and `"/workspace"`, so it cannot expose this behavior.

**Fix:** Resolve `process.cwd` once to a normalized absolute sandbox path and use that exact value for both `--chdir` and `HOME`. Add cases for `work`, `./work`, and parent components, plus a live assertion that `$PWD == $HOME` when the default environment is selected.

MGudgin (commented):
## Review finding

Verified against current head `cc22dac`: the PR merge base equals its declared base, and the cited resolver and test coverage are added by this PR. The finding is therefore `introduced_by_change`; no pre-existing code is being attributed to this change.

The inline comment describes the relative-`cwd` failure mode and the regression coverage needed.

MGudgin:
**High — Pre-0.9 `process.env: []` regresses from fail-closed rejection to a populated Windows environment.**

`src/core/wxc_common/src/models.rs:1140-1152`

**Attribution:** `introduced_by_change` — this PR adds `supplied_env()`, which converts `Some([])` to `None` for `LegacyCompatible` requests, and changes both ProcessContainer runners and launch validation to consume that result.

This reverses behavior that was introduced and explicitly protected by two prior commits:

- `491ac954` (#1120) intentionally preserved omitted versus explicitly empty environments. It added the `explicitly_empty_env_yields_an_empty_block_not_the_default` regression test and documented that a caller-owned empty environment must not be upgraded to the default profile block.
- `519c8c0` (#1158) added `validate_required_child_env()` to both Windows runners, making an empty or incomplete caller-owned block fail closed with `FailurePhase::Rejected`.

With this PR, the same pre-0.9 request changes from **rejected before creating a child** to **successfully launching with the populated user-profile environment**. The released pre-0.9 contracts represent `env` as `OptionalField<Vec<String>>`, and their adapters preserve `None` versus `Some([])`, so this collapse is not required by the wire format.

**Fix:** Preserve `Some([])` for ProcessContainer at every schema version. If changing the established Windows contract is intentional, separate and version that compatibility change rather than folding it into the Linux/macOS default-environment work.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -528,6 +528,85 @@ pub(crate) fn local_network_diagnostic_for_mode(
...
+///
+/// `--clearenv` leaves the child with no `PATH`, so resolution fell through to
+/// the shell's compiled-in default. That value varies: it matches this one on
+/// Debian and Ubuntu, so nothing changes there, but on RHEL it omitted the
+/// `sbin` directories. Setting it explicitly removes the dependency.
+const DEFAULT_PATH: &str = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin";
+
+/// `HOME` when the request resolves no working directory. `build_args` always
+/// mounts a fresh tmpfs at `/tmp`, so this is always writable.
+const FALLBACK_HOME: &str = "/tmp";
```
MGudgin: **Medium (security/reliability) - Do not use policy-rebindable `/tmp` as the default `HOME`.**

**Attribution:** `introduced_by_change` - this PR adds `HOME=/tmp` when schema 0.9 selects the default environment and no `process.cwd` is supplied.

The comment says `/tmp` is always a fresh tmpfs, but policy mounts are deliberately emitted after `--tmpfs /tmp` so they win. With `readwritePaths: ["/tmp"]`, Bubblewrap effectively emits `--tmpfs /tmp` followed by `--bind /tmp /tmp`, replacing the private directory with the host's shared `/tmp` while still exporting it as `HOME`.

LXC has the analogous issue at `lxc_runner.rs:35-37`: policy grants bind a host path at the same container path, and reused containers can also retain files in their image `/tmp`. Tools may then consume pre-existing `.gitconfig`, `.curlrc`, `.npmrc`, `.config`, caches, or helper state from a shared or persistent HOME.

This requires the default environment, no explicit cwd, and a `/tmp` bind or applicable LXC reuse, so it is not a universal escape.

**Fix:** Create a private backend-owned home directory that policy cannot replace, with per-sandbox identity and restrictive permissions, or omit `HOME` when no safe private directory can be guaranteed.

src/backends/bubblewrap/common/src/bwrap_command.rs
```diff
@@ -528,6 +528,85 @@ pub(crate) fn local_network_diagnostic_for_mode(
...
+    }
+
+    let entries = match (&request.env, request.inherit_default_env) {
+        (None, _) => default_env(request),
+        (Some(supplied), false) => return supplied.clone(),
+        (Some(supplied), true) => {
+            let mut entries = default_env(request);
+            // A caller entry replaces the same-named default rather than
+            // being appended: `--setenv` would otherwise be passed twice.
+            for (key, value) in supplied.iter().filter_map(|kv| kv.split_once('=')) {
```
MGudgin: **Medium (maintainability, also performance) - Share the four-state environment resolver instead of implementing it independently in three backends.**

**Attribution:** `introduced_by_change` - this PR adds effectively identical version/state dispatch and override merging in Bubblewrap (`bwrap_command.rs:582-608`), LXC (`lxc_runner.rs:88-114`), and Seatbelt (`default_env.rs:49-75`).

Only construction of the platform defaults is backend-specific. Each copy otherwise handles legacy pass-through, omitted defaults, explicit empty/verbatim environments, inherited overlays, malformed entries, duplicate keys, and output formatting.

The copies have already diverged in a correctness-relevant way: LXC anchors a relative cwd to the container root and Seatbelt receives a resolved path, while Bubblewrap copies the relative value directly into `HOME`, producing the separate High finding.

The overlay loop also linearly scans the growing output vector for every caller entry, making large inherited environments O(n^2) in all three implementations.

**Fix:** Move the state dispatch and merge into a pure `wxc_common` helper that accepts backend-constructed defaults. Use a key-to-index map so default ordering, append order, and last-value-wins behavior remain stable without quadratic scans. Keep only platform-specific PATH/HOME construction and plumbing tests in each backend.

src/backends/process_container/common/src/base_container_runner.rs
```diff
@@ -3028,6 +3030,26 @@ mod tests {
         assert_eq!(environment, vec![0u16, 0u16]);
     }
 
+    #[test]
+    fn below_0_9_an_explicitly_empty_env_still_gets_the_default_block() {
```
MGudgin: **Medium (test coverage) - Add an end-to-end Windows case for the exact empty environment list changed here.**

**Attribution:** `newly_exposed_by_change` - the Windows lifecycle suite has meaningful environment coverage, but this PR changes `Some([])` specifically and no live test sends `"env": []`.

`run_processcontainer_lifecycle_test.ps1` already runs on both `base-container` and `appcontainer-dacl` and covers 0.9 inheritance, omitted defaults, and supplied verbatim environments. Its nominal “empty” case is actually a non-empty array containing one empty-valued entry (`MXC_LC_EMPTY=`), so it cannot exercise this branch or catch the fail-closed-to-success regression.

The omitted/verbatim cases also use the suite's default 0.8 schema rather than pinning the complete schema-0.9 four-state matrix. This PR adds the legacy-empty assertion only at the BaseContainer helper level; the changed AppContainer environment-block path has no equivalent targeted test.

**Fix:** Add an exact pre-0.9 fixture with `"env": []` and assert the established rejection. Add explicit 0.9 omitted, empty-list, verbatim, and inherited fixtures, preserve the empty array through fixture generation, and run the matrix on both Windows tiers.

docs/lxc-support/lxc-backend.md
MGudgin: **Medium (documentation) - Distinguish the environment MXC supplies from what the LXC child observes.**

**Attribution:** `introduced_by_change` - this PR adds a table saying `[]` produces “nothing” and a verbatim `FOO` environment produces “FOO only”, immediately followed by a paragraph saying liblxc always supplies a baseline `PATH`.

The implementation and `run_lxc_env_09_test.sh` confirm that the child always observes liblxc's baseline PATH unless the caller replaces PATH explicitly. Thus an empty MXC environment is not an empty observed environment, and `FOO`-only input is observed as baseline PATH plus FOO.

The current wording can make users expect `printenv` to be empty for `[]`, or contain only FOO for the verbatim case, unlike the actual LXC behavior.

**Fix:** Either label the table “Environment entries supplied by MXC”, keeping liblxc additions separate, or describe the child-observed result directly: `[]` means baseline PATH only; `FOO` means baseline PATH plus FOO; and a caller-supplied PATH replaces the baseline.

src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -24,6 +24,96 @@ use crate::signal_cleanup;
...
+    Some(request.working_directory.as_str()).filter(|dir| !dir.is_empty())
+}
+
+/// `lxc-attach` starts at the container root, so a relative start directory
+/// resolves against `/` and `HOME` must name the same absolute path.
+fn container_absolute(dir: &str) -> String {
+    if dir.starts_with('/') {
+        dir.to_string()
+    } else {
+        format!("/{}", dir.trim_start_matches("./"))
```
MGudgin: **Low (reliability) - Normalize the LXC start directory once before using it for both `cd` and `HOME`.**

**Attribution:** `introduced_by_change` - this PR adds `container_absolute()` and claims that HOME always names the directory where the child starts.

The helper only prefixes `/` and strips leading `./`; it preserves `..`. For ordinary paths such as `../etc`, both spellings normally resolve to `/etc`, so this is not a container-root escape. The edge case is a path containing a symlink followed by `..`: the shell's logical `cd` handling can resolve it differently from applications opening files beneath the absolute HOME string. Non-canonical HOME strings can also affect tools that compare or persist paths textually.

Current tests cover `work`, `./work`, and `/workspace`, but no parent-component or symlink cases.

**Fix:** Define one lexical root-anchoring/normalization rule, use the resulting absolute path for both `attach_run` and HOME, define whether attempts above `/` are clamped or rejected, and test parent plus symlink cases.

src/backends/seatbelt/common/src/default_env.rs
```diff
@@ -0,0 +1,238 @@
...
+/// runner resolved one — `HOME`.
+///
+/// `HOME` is left unset with no resolved directory: the profile is deny-default
+/// and grants `/private/tmp` only under `guiAccess`, so a `/tmp` fallback would
+/// be either unwritable or a shared preplant target.
+fn default_env(working_directory: Option<&str>) -> Vec<(String, String)> {
+    let mut entries = vec![("PATH".to_string(), DEFAULT_SANDBOX_PATH.to_string())];
+
+    if let Some(home) = working_directory {
+        entries.push(("HOME".to_string(), home.to_string()));
```
MGudgin: **Low (security) - Setting `HOME` to the workspace promotes repository files to user-level tool configuration.**

**Attribution:** `newly_exposed_by_change` - this PR newly sets HOME to the resolved working directory on Seatbelt, LXC, and Bubblewrap when the default environment is selected.

Tools can consequently consume workspace-controlled files such as `.gitconfig`, `.curlrc`, `.npmrc`, and `.config/*` as user-level configuration rather than merely project input. That gives untrusted workspace contents another way to influence the caller-selected command. The sandbox still limits the resulting process, and some tools already read project-local configuration, so this is a bounded Low-risk behavior rather than an escape.

**Fix:** Prefer a private backend-created HOME while leaving cwd on the workspace, or explicitly document and test the trust consequence if workspace-as-HOME is intentional.

src/core/wxc_common/src/models.rs
```diff
@@ -1067,6 +1091,16 @@ impl NetworkEnforcementCompatibility {
     }
 }
 
+impl DefaultEnvCompatibility {
+    /// Stable diagnostic spelling for tests.
+    pub const fn as_str(self) -> &'static str {
```
MGudgin: **Low (maintainability) - Remove the unused public `DefaultEnvCompatibility::as_str` helper.**

**Attribution:** `introduced_by_change` - this PR adds the method, but there are no callers in source, tests, SDKs, scripts, or documentation.

Unlike `NetworkEnforcementCompatibility::as_str`, this spelling is not consumed by policy identity or diagnostics. Keeping an unused public helper creates API surface and stable string spellings without a present requirement.

**Fix:** Remove the method until a concrete diagnostic or serialization consumer needs it, or use it in that consumer in this change.

MGudgin (changes requested):
## Summary

Changes requested for two High regressions verified against head `cc22dac`:

1. Bubblewrap copies a relative `process.cwd` into `HOME` while separately passing it to `--chdir`. After the directory change, the relative HOME no longer names the child's actual start directory. This is covered by the previously posted inline comment.
2. Pre-0.9 Windows `process.env: []` changes from fail-closed rejection to a successful launch with the populated user-profile environment. That reverses behavior deliberately introduced by `491ac954` (#1120) and enforced by `519c8c0` (#1158). This is covered by the previously posted PR comment.

The seven inline comments in this review cover the additional confirmed findings: policy-rebindable `/tmp` as HOME, triplicated and quadratic environment resolution, the narrowed Windows empty-list end-to-end coverage gap, contradictory LXC environment documentation, LXC parent-component normalization, workspace-as-HOME configuration influence, and an unused compatibility helper.

The original broad Windows coverage claim was rechecked and narrowed: both Windows tiers already run meaningful environment tests, but neither sends the exact empty environment list changed by this PR.

src/backends/bubblewrap/common/src/bwrap_command.rs
jsidewhite: Strongly agree, if it's possible.

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: Done. Moved to shared helper.

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: HOME is now omitted rather than falling back. Documentation is updated.

src/backends/bubblewrap/common/src/bwrap_command.rs
theelliotm: Done. Moved to common helper.

src/backends/process_container/common/src/base_container_runner.rs
theelliotm: Windows changes were reverted. Test should no longer be needed.

docs/lxc-support/lxc-backend.md
theelliotm: Done

src/core/wxc_common/src/models.rs
theelliotm: Removed

src/backends/lxc/common/src/lxc_runner.rs
theelliotm: LXC-only path no longer exists. The shared path does what you ask.

src/backends/seatbelt/common/src/default_env.rs
theelliotm: explicitly documented this.

MGudgin (approved):
Re-verified the updated implementation at 8a13c3af. The runtime and documentation findings are addressed, and the targeted regression tests pass. The remaining exact-empty Windows live-fixture request is non-blocking coverage hardening.

jsidewhite (approved):
:shipit:

PR #1203, opened 2026-09-17
[LXC] Reject runtimeConfig.networkProxy with a backend-specific error and document it

SohamDas2021:

-----

## Summary

LXC cannot reach a host loopback proxy: the container has its own network namespace, so the loopback endpoint `runtimeConfig.networkProxy` is required to name resolves to the container rather than the host. Proxied egress is not supported on this backend, so a request asking for it should say so plainly.

LXC already withheld the `RUNTIME_PROXY` support bit, so such a request was refused — but only by the shared support gate, whose message reads `runtimeConfig.networkProxy is not supported by the selected backend`. That names no backend, no reason, and no alternative. Nothing pinned the behavior either, so declaring the bit later would have silently accepted the config and left the workload failing at runtime against a loopback address with nothing listening.

This makes the refusal explicit and documents it.

- `LxcScriptRunner::validate_runner` now refuses `runtimeConfig.networkProxy` up front with an LXC-specific message giving the reason and pointing at `network.proxy.url` on schema 0.6-0.8. The shared gate stays as a backstop.
- Because the check keys on `runtime_network_proxy_specified` rather than on the schema version, it also covers SDK and FFI callers that build an `ExecutionRequest` directly and never touch the config parser.
- Two regression tests: one asserts the declared support set still withholds `RUNTIME_PROXY`, the other that the refusal names both the field and the backend.
- `docs/lxc-support/lxc-backend.md` gains a `### Proxy` section. Its lead-in previously pointed v0.9 readers at `runtimeConfig` for proxy values, which LXC rejects; that is corrected.

No behavior change for requests that do not carry the field. Schema 0.6 and 0.7 are unaffected — `runtimeConfig` is not part of those contracts, so they are rejected earlier as an unknown field; the new error applies from 0.8 onward.

## Validation

`cargo test -p lxc_common` (340 pass), `cargo fmt --check`, and `cargo clippy --all-targets -- -D warnings` all clean.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1203)


src/backends/lxc/common/src/lxc_runner.rs
```diff
@@ -650,6 +650,11 @@ pub const LXC_CAPABILITIES_MODE_UNSUPPORTED: &str =
      or state the policy in the 0.8 network.egress / network.ingress form, which carries no \
      enforcement mode.";
 
+pub const LXC_RUNTIME_PROXY_UNSUPPORTED: &str =
+    "LXC: runtimeConfig.networkProxy is not supported. It must name a loopback endpoint, which \
+     inside the container's own network namespace is the container rather than the host. On \
+     schema 0.6-0.8, use network.proxy.url with an address routable from inside the container.";
```
jsidewhite, line 656: >schema 0.6-0.8, [](http://example.com/codeflow?start=5&length=15)

unfortunate that we need to mention schema versions here.  what about 0.9?

jsidewhite:
Are we supposed to ensure probe works too or no?

jsidewhite (approved):

SohamDas2021:
> Are we supposed to ensure probe works too or no?

Not in this PR, will have to look into the probe story for linux backends, what is available and what needs work.

src/backends/lxc/common/src/lxc_runner.rs
SohamDas2021: Wanted to let the user know if they want use proxy (host routable), they can still fall back to an older version of the schema.

PR #1206, opened 2026-09-17
[WSLC] Document how an unsupported Windows version surfaces

SohamDas2021:
## 📖 Description

Documentation-only. It corrects the WSLC prerequisites and, more importantly, writes down **how an unsupported Windows version actually surfaces** — because MXC deliberately does not check for one.

This PR originally added a minimum-Windows-version gate to `GetPlatformSupport`. Review established that was the wrong mechanism, and the code has been reverted in full. The reasoning, now captured in the docs:

- **The existing probe already gates correctly.** `WslcGetMissingComponents()` — called from `is_available()`, which backs `platform_support()` / `GetPlatformSupport()` / `available_backends()`, and again from each backend preflight — reports `WslPackage` missing on a host too old for WSL 2, because the WSL runtime package cannot install below build 19041. `wslc` is omitted from the available backends either way.
- **WSL already owns and reports the floors.** `wsl --update` prints *"Windows version {} does not support the packaged version of Windows Subsystem for Linux"* with a KB link, and `wslcsdk.dll` statically links WSL's service-connect guard, which raises `WSL_E_OS_NOT_SUPPORTED` (`0x80040327`) — or `WSL_E_WSL_OPTIONAL_COMPONENT_REQUIRED` (`0x80040321`). Verified by disassembling the shipped DLL: an `lxss` service probe feeding a `cmovne` between those two codes, matching WSL's `svccomm.cpp` guard `!IsWindows11OrAbove() && !IsWslSupportInterfacePresent()`.
- **Hardcoded floors would duplicate a moving target.** WSL's requirements are architecture-dependent, include an update-build-revision component on x64, and change over time — this very work item exists because they changed. Duplicating them in MXC means tracking WSL's servicing.

### Changes

- `docs/wsl/wsl-container-getting-started.md` — replaces the incorrect "Windows 11 required" prerequisite with WSL 2's actual architecture-dependent floors, and adds a **How an unsupported Windows version is reported** section covering the probe, the `wsl --update` message, and both HRESULTs. Two troubleshooting rows replace the one describing the removed check.
- `docs/wsl/wsl-container-support-plan.md` — same prerequisite correction.
- `.github/copilot-instructions.md` — records that WSLC carries no version check of its own and how the verdict surfaces.

## 🔗 References

- ADO [63135991](https://microsoft.visualstudio.com/OS/_workitems/edit/63135991) — [WSLC] Down-level support for WSLC SDK nuget
- [WSL 2 system requirements](https://learn.microsoft.com/windows/wsl/install-manual#step-2---check-requirements-for-running-wsl-2)

## 🔍 Validation

Documentation only — no code, no `Cargo.lock` change. `src/` is byte-identical to the branch base.

The behaviour described was verified before the code was reverted: the full WSLC E2E suite (`run_wslc_all_tests.ps1`, including the delegated state-aware suite) ran on a live host — build `26100.9448`, x64, WSL `2.9.12.0` — at **76/77**, with the single failure (`D: exec (directional network change rejected structurally)`) reproducing identically on a clean tree and therefore pre-existing and unrelated.

## ✅ Checklist

- [x] Linked to an issue
- [x] If this PR changes `Cargo.lock`, the `dependency-feed-check` check passes — N/A, `Cargo.lock` is unchanged

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [x] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1206)


docs/wsl/wsl-container-getting-started.md
```diff
@@ -449,6 +449,7 @@ explicit `provision` / `deprovision` phases rather than by per-run flags.
 | Error | Cause | Fix |
 |---|---|---|
 | `WSLC backend not compiled` | Binary built without `--features wslc` | Rebuild with `build.bat --with-wslc` |
+| `WSLc requires Windows 10 ...; this host is build <n>` | The host is below WSL 2's minimum Windows version for its architecture | Upgrade Windows. Updating WSL cannot clear this, and `getPlatformSupport` / `--available-backends` omit `wslc` on such a host |
```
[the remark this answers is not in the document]
SohamDas2021: Good catch — verified and fixed in c643b21e.

`--available-backends` is defined only in `src/core/lxc/src/main.rs:55`, and `wxc-exec --help` on a Windows host lists `--probe` with no such flag. Since WSLC is Windows-only, that row was pointing users at a flag they could never run.

Replaced it with the APIs that actually report this on Windows: `platform_support()` (Rust) and `GetPlatformSupport()` (C#). I deliberately didn't keep the bare `getPlatformSupport` spelling either — the TypeScript SDK's `getPlatformSupport` never reports `wslc` on any host, so naming it unqualified would have been a second inaccuracy.

src/backends/wslc/common/src/host_requirements.rs
```diff
@@ -0,0 +1,250 @@
...
+fn meets_minimum(architecture: HostArchitecture, build: u32, update_build_revision: u32) -> bool {
+    match architecture {
+        HostArchitecture::Arm64 => build >= BUILD_2004,
+        _ => match build {
+            BUILD_1903 | BUILD_1909 => update_build_revision >= MIN_UBR_1903_1909,
+            _ => build > BUILD_1909,
+        },
+    }
+}
+
```
bbonaby: note: I know querying the OS builds are always tricky and perhaps maybe these minimum builds are ok but I think it might be better to attempt querying one of the wslc APIs to see if it exists on the build. Is that possible? Hoping there is some unsupported HRESULT or module not found error. Or even the presence of an API set would be most ideal.

SohamDas2021: We already do that probe one step later: `WslcGetMissingComponents()` runs right after this check in `probe_components()`, `init_and_load_sdk()`, and `load_sdk_checked()`, and it's what actually decides availability.

It can't replace the version check though: we ship `wslcsdk.dll` ourselves (`build.rs:118` copies it from the pinned NuGet, and `load()` reads from our own module dir), so its exports reflect our pin, not the host.

So, effectively, the WSL package can't install, so the component probe already omits `wslc` - the build numbers just change the message to "upgrade Windows" instead of an unhelpful "run `wsl --update`".

jsidewhite: why is "run wsl --update" unhelpful?  won't wsl.exe itself immediately then return the right error message with the windows build that is required?

(i agree with bbonaby that all this custom code to query the os version is the wrong way to tackle it)

jsidewhite (dismissed):
:shipit:

src/backends/wslc/common/src/host_requirements.rs
SohamDas2021: Okay, I validated from the wslc sdk code, that it does surface WSL_E_OS_NOT_SUPPORTED error. So, wsl --update will help see the error.

bbonaby (approved):

PR #1211, opened 2026-09-18
[LXC] [Test/docs] State the network chains' limits, and test that inbound default-deny drops packets

### LINKED ISSUE #897: [LXC] Inbound default-deny: install window, namespace identity, and workload escape
> ## Status — 2026-09-17
>
> Current snapshot, verified against `origin/main` @ `6e12d6da5`. This replaces
> every earlier scope note on this issue; read it as the present state, not as a
> change log.
>
> ### This issue covers egress as well as ingress
>
> PR #1041 (merged 2026-09-16, `0d4d6934`) moved LXC egress enforcement from a
> host-side `FORWARD` hook into the container's own network namespace on
> `OUTPUT` (`network_iptables.rs:1581`). Inbound default-deny has been in that
> namespace since #836. **Both chains now live where the workload can reach
> them**, so every finding here applies in both directions.
>
> The "Contrast with egress" paragraph in the folded #854 text below — egress
> chains live on the host, hooked into `FORWARD`, beyond the workload's reach —
> is **obsolete**. The asymmetry it describes no longer exists, and the lines it
> cites are gone.
>
> ### What is already fixed
>
> `confine_network_capabilities` (`lxc_bindings.rs:139`) drops `CAP_NET_ADMIN`
> from the bounding set of the `lxc-attach` child whenever chains are installed
> (`:411`), guarded by `e2e_lxc_network_capability.rs`. **The attached workload
> can no longer flush or delete either chain.** That is a real narrowing of
> #854, and it is the only part of this issue that has shipped.
>
> ### Three folded reports are dead
>
> **PR #849 was closed unmerged on 2026-09-16.** #859, #905, and the
> concurrent-exec addendum all describe `src/backends/lxc/common/src/state_aware.rs`,
> which does not exist on `main` and will not until LXC state-aware is re-cut.
> Verified: `mxc_engine::run_state_aware` has no LXC arm, and `signal_cleanup.rs`
> contains no `RollbackStep` or `SignalRollback` — the note under #859 below
> asserting otherwise describes branch-only code.
>
> They are **not reachable defects**. They survive as design constraints for a
> future re-cut, tracked in ADO, not as work remaining here.
>
> ### Work raised here, and where it now lives
>
> This issue is closed by the pull request that implements the two items marked
> below. **It is a record, not an owner** — every item, done or open, is tracked
> on an internal ADO work item. Those are not linkable from a public issue, so
> the status column says only that they are tracked there.
>
> The originals these came from (#850, #853, #854) are closed and point here, so
> this issue remains the public description of the open work even though the
> scheduling lives elsewhere.
>
> | Work | Status |
> |---|---|
> | Extend the `CAP_NET_ADMIN` drop past `lxc-attach` to **container init**. MXC writes no `lxc.cap.drop`/`lxc.cap.keep` anywhere — the only occurrence in the tree is a test fixture (`filesystem_mounts.rs:215`) — so LXC's defaults apply, PID 1 and its forks keep the capability, and a preserved rootfs (`destroyOnExit: false`) can carry an init unit that runs before or outside the attach path. From #854 | **Open**, tracked in ADO. Blocked on a design decision: dropping `net_admin` from container init also stops the in-container DHCP client configuring the interface, so `wait_for_network` never sees an address and the run aborts fail-closed |
> | Pin the namespace with a handle instead of a recyclable PID. Six `nsenter -t <pid> -n` sites re-resolve the number per command — five in `network_ingress.rs` (`build_nsenter_argv:208`), one in `network_iptables.rs` (`command_argv:1087`) — plus the watchdog slot held for the container's lifetime. Now carries a second catastrophic case: a terminal egress policy on the **host's** `OUTPUT`. From #853 | **Open**, tracked in ADO |
> | State the limitations in `docs/lxc-support/lxc-backend.md`. #1041 removed 569 lines including the containment-boundary and startup-window notes, and nothing was written in their place | **Implemented**, pending a pull request. Tracked in ADO |
> | Traffic-level validation on a live LXC host. `run_lxc_inbound_deny_test.sh` asserts rule programming and namespace containment and explicitly disclaims reachability, so nothing proves a packet is actually dropped. From #850 | **Implemented**, pending a pull request. Tracked in ADO |
>
> ### What is tracked elsewhere
>
> **Do not implement these here.**
>
> | Work | Status |
> |---|---|
> | Start-to-install fail-open window, now spanning both directions: start at `lxc_runner.rs:376`, up to 30 s of readiness polling, chains installed at `:428` | **Open**, in #896 as #764, which already carries a design. Tracked in ADO |
> | Atomic chain swap. `reset_family` (`network_ingress.rs:463`) unhooks a leftover chain before its replacement exists | **Open**, in #896 as #764's final step. Tracked in ADO |
> | Removing the `ESTABLISHED,RELATED` accept from the egress chain, which is what keeps a flow opened during the startup window usable afterwards | **Open**, tracked in ADO. Not in #896 — it was raised in review here rather than as its own issue, so it appears in no umbrella's fold list |
> | DHCP renewal cannot complete. The base chain opens `lo`, `ESTABLISHED,RELATED`, and port 53, with no port-67 allow; measured on Ubuntu 24.04 / LXC 5.0.3 / stock `lxc-net`, renewal at ~30 min is a NEW flow against a 30 s conntrack timeout and hits the DROP | **Open**, tracked in ADO alongside the item above. #764's carve-out covers the quarantine chain during startup, not the completed chain the container then lives under |
> | `CAP_NET_RAW` is retained so an explicit `protocol: "icmp"` allow works, so `AF_PACKET` frames bypass the egress chain entirely | **Open**, tracked in ADO |
> | #859, #905, and the concurrent-exec addendum below | **Not reachable defects**, tracked in ADO against the postponed state-aware lifecycle |
>
> ### Two things to know before starting
>
> **The `ESTABLISHED,RELATED` removal is egress-only.** The argument for removing
> it — destination allow rules carry no state restriction, replies to
> container-initiated flows arrive on `INPUT` rather than `OUTPUT`, and proxy mode
> already omits it deliberately (`network_iptables.rs:1489`) — holds for the
> **`OUTPUT`** chain (`network_iptables.rs:530`). On the **`INPUT`** chain
> (`network_ingress.rs:520`) the same rule is load-bearing: it is what lets
> replies to the container's own outbound connections back in. Removing it there
> would break all outbound traffic. Closing the ingress half of the window means
> flushing conntrack at install time, not deleting the rule.
>
> **Out of scope:** `CAP_NET_RAW` is deliberately retained so an explicit
> `protocol: "icmp"` allow stays functional, and `AF_PACKET` frames therefore
> bypass the chain. Tracked in ADO.
>
> This gates GA decisions **D1** (default-deny outbound), **D2**, and **D7**,
> which requires cooperative routing to sit above an enforcing layer.
>
> The folded reports below are the original filings, reproduced unedited. Where
> they disagree with this snapshot, this snapshot is correct.

---
The inbound (INPUT) default-deny chain: the window before it exists, how the target namespace is identified, and what the workload can do to it once inside.

This consolidates 5 separately filed issues so the backlog carries one entry per area instead of one per finding. Every original report is reproduced below in full, unedited. The originals are closed and point here.

| Folded from | Status | Title |
| --- | --- | --- |
| #850 | Open | LXC inbound: close the start-to-install fail-open window and make the ingress reset atomic |
| #853 | Open | LXC ingress: pin the container network namespace instead of re-resolving a recyclable PID |
| #854 | Open | LXC inbound default-deny is removable by the workload: CAP_NET_ADMIN is never dropped |
| #859 | Fixed in #849, closes on merge | LXC state-aware firewall mode never installs the inbound default-deny INPUT chain |
| #905 | Open | LXC start leaves a live unfiltered-inbound container if killed between start and ingress, and a retry is refused |

Rows marked *Fixed in #849* are already addressed by that pull request and are reproduced here only so the record stays complete. Read them as history once #849 merges, not as work remaining.

---

## LXC inbound: close the start-to-install fail-open window and make the ingress reset atomic

*Originally #850.*

Follow-up from #836.  Two reviewer findings on that PR were confirmed real and deliberately deferred, because both need a design change validated on a live LXC host rather than a late edit to a PR already at +2707.  They are contained in one another, so they should be fixed by the same change.

## 1. Inbound is unfiltered between container start and ingress install

`container.start()` runs at `src/backends/lxc/common/src/lxc_runner.rs:182-190`.  `wait_for_network` follows at :205 -- its return value is discarded, so a timeout falls through rather than aborting -- then veth discovery and the egress apply at :211-236, and only then the ingress install at :250-310.  For that whole interval the container's INPUT policy is unchanged, so a reused container's services, or a new container's init services, are reachable despite a requested default-deny.

Two things bound the severity today: the workload runs via `lxc-attach` only after the ingress install returns, so no sandboxed code executes during the interval, and the exposure is therefore to external traffic only.

The naive fix does not work.  Ingress cannot simply move ahead of `wait_for_network`, because the IPv6 classification needs the namespace to have come up.  The shape that does work is an early unconditional deny, with family-specific rules completed once the namespace has settled.

## 2. `reset_family` unhooks a leftover policy before the replacement exists

In `src/backends/lxc/common/src/network_ingress.rs`: resetting a leftover *hooked* chain from a crashed run removes every INPUT hook, then recreates and re-hooks.  On a still-running reused container that is a fail-open interval.  It does not affect a fresh install, where there is no prior hook to remove, and the doc comment has been corrected to say so rather than claiming no window exists at all.

The fix is an atomic swap: build the replacement chain, then repoint INPUT.  `iptables-restore` is the mechanism, and adopting it means rewriting the install path away from incremental commands -- which is also what the per-resource ownership tracking and its teardown depend on.

## Why they belong to one change

(2) is strictly contained inside (1).  An early-deny design that holds INPUT at deny across the whole startup window removes the exposure in (2) as well, so fixing them separately would build the same machinery twice.

## Validation

Neither is credible without a live host.  `tests/scripts/run_lxc_inbound_deny_test.sh` exists and asserts namespace containment, and it has now been executed against real LXC containers (Ubuntu 24.04 under WSL 2, LXC 5.0.3, iptables 1.8.10 nf_tables).  An earlier version of this issue said no LXC host was available and that WSL could not provide one; that was wrong, and it was never checked.  That run surfaced a separate defect in the merged code, fixed in #863.  It still does not prove *this* issue fixed.  The test that would actually prove this fixed is traffic-level: start listeners inside the container, probe them from the host over IPv4 and IPv6, both during and after the startup window.

Source: the review threads on #836 at `lxc_runner.rs:256` and `network_ingress.rs:630`.

---

## LXC ingress: pin the container network namespace instead of re-resolving a recyclable PID

*Originally #853.*

Raised in review of #836 (https://github.com/microsoft/mxc/pull/836#discussion_r3780037109).

## Problem

The LXC ingress path addresses the container network namespace by **numeric PID**, and re-resolves that PID for every privileged command instead of pinning the namespace once.

- `lxc_runner.rs` takes `container.init_pid()` and passes the raw `u32` to `IngressManager::new(&container_name, pid)`.
- `network_ingress.rs` builds every command as `nsenter -t <pid> -n <argv>`, so the PID is resolved again per command.
- `signal_cleanup.rs:93-98` stores the same raw PID in the watchdog slot, and `signal_cleanup.rs:214-215` uses it at signal time via `IngressManager::force_cleanup(&name, pid, ..)`.

A PID is a recyclable name for a process, not a stable handle on its namespace.  If container init exits and the PID is reallocated, `nsenter -t <pid> -n` enters whatever process now holds that number -- commonly a host process, and therefore the host network namespace.

## Severity, split by operation

These are not equally exposed, and the split matters for prioritization:

- **Install is the dangerous operation but has the narrow window.**  It creates an `MXCI-<container>` chain and hooks it into `INPUT` with a terminal DROP.  Doing that in the host namespace would firewall the host.  The window is only the multi-command install, and reuse requires PID wraparound -- `PID_MAX_LIMIT` is 2^22 (4194304) on 64-bit Linux, so this needs ~4M PIDs allocated within seconds.  Hosts with a lowered `/proc/sys/kernel/pid_max` shrink that requirement.
- **Cleanup has the wide window but is far less dangerous.**  The watchdog holds the PID for the container's whole lifetime, so reuse there is entirely plausible.  However it only issues targeted deletes against the uniquely-named `MXCI-<container>` chain, so entering a foreign namespace produces "no chain by that name" errors rather than damage.

So the catastrophic case is unlikely and the likely case is benign -- but the catastrophic case is host-wide inbound DROP, which is not a risk worth carrying on a probabilistic argument.

## Proposed fix

Pin the namespace with an open handle rather than a number: open `/proc/<pid>/ns/net` once at `IngressManager` construction and use that handle for install, cleanup, and watchdog state.  An open namespace file descriptor keeps referring to the namespace it was opened on, so PID reuse cannot redirect it.

Explicitly rejected as a substitute: re-checking the PID's identity immediately before each command.  That narrows the window without closing it -- the check and the `exec` are still separate -- and a partial guard here would read as a solved problem.

## Why not in #836

The change touches three subsystems (install, teardown, watchdog signal handling) and depends on file-descriptor inheritance semantics across the `nsenter` exec, which cannot be exercised on the Windows development host or in CI -- neither runs LXC.  It wants its own PR with a Linux host to validate against.

## Validation

Needs a live LXC host: start a container, capture the init PID, confirm the ingress chain lands in the container namespace and not the host's, and confirm that a handle opened at construction still resolves after the original PID is gone.

---

## LXC inbound default-deny is removable by the workload: CAP_NET_ADMIN is never dropped

*Originally #854.*

### Summary

The inbound default-deny chain added in #836 is installed **inside the container's own network namespace**, and the sandboxed workload retains `CAP_NET_ADMIN` in that namespace. A workload that wants to can run `iptables -F MXCI-<name>` (or delete the chain outright) and restore full inbound reachability.

This does not make the feature useless — it still closes external reachability for any workload that does not deliberately tear it down, including services the workload itself starts — but it does mean inbound default-deny is **not a containment boundary against the sandboxed code**, and it should not be relied on as one.

### Evidence

1. The chain is installed in the container netns, via `nsenter -t <init-pid> -n iptables ...` — `network_ingress.rs`, `nsenter_command`.
2. MXC creates containers with the stock download template and no capability configuration:
   - `lxc_bindings.rs:227-235` — `lxc-create -t download -- -d <dist> -r <release> -a <arch>`.
   - The only config key MXC ever writes is `lxc.mount.entry` (`filesystem_mounts.rs:239,250,304`). A repo-wide search for `lxc.cap` / `lxc.seccomp` / `lxc.apparmor` / `lxc.init.uid` returns only two path-injection **test fixtures** in `filesystem_mounts.rs:341-342`.
3. LXC's defaults therefore apply, and they leave `net_admin` in place:
   - `lxc/lxc`, `config/templates/common.conf.in`: `lxc.cap.drop = mac_admin mac_override sys_time sys_module sys_rawio` — `net_admin` is not listed.
   - `lxc/lxc`, `config/templates/userns.conf.in`: `# Start with a full set of capabilities in user namespaces.` followed by empty `lxc.cap.drop =` and `lxc.cap.keep =`.
4. The workload runs as container root: `attach_run` (`lxc_bindings.rs:337-368`) invokes `lxc-attach` with only `build_attach_args` (env/cwd/command layering). No `-u` / `-g`, no capability or credential flags.

### Contrast with egress

Egress is not exposed this way. Its chains live on the **host** and are hooked into `FORWARD` by the container's host-side veth (`network_iptables.rs:1142-1148`, `-I FORWARD -i <veth> -j <chain>`), which the workload cannot reach. The asymmetry is inherent to where each chain is installed, not an oversight in one of them.

### Why ingress was put in the container netns anyway

Host-side `FORWARD` filtering only sees traffic **routed through** the host to the container. Traffic originating on the host itself and addressed to the container does not traverse `FORWARD`, so a purely host-side inbound chain would miss it. Enforcing in the container netns catches every inbound path regardless of origin. That is a real reason for the current design, and any replacement has to keep that coverage.

### Options (not yet evaluated)

- Drop `net_admin` for the container (`lxc.cap.drop`), or attach the workload as a non-root uid. Cheapest, but may break workloads that legitimately configure their own networking.
- Move inbound enforcement to the host and add explicit coverage for host-origin traffic, so it matches the egress trust model.
- Keep the current mechanism and document it as best-effort exposure reduction rather than a boundary. This is what #836 does today.

### Scope

#836 documents the limitation truthfully. Choosing and implementing a tamper-proof mechanism is this issue.

Found in review of #836 by @copilot-pull-request-reviewer.

---

## LXC state-aware firewall mode never installs the inbound default-deny INPUT chain

*Originally #859.  Fixed in #849; closes on merge.*

Found while reviewing #849.  This is a fail-open gap: a state-aware container accepts inbound connections that the one-shot path would drop.

## The asymmetry

**One-shot installs ingress.**  `src/backends/lxc/common/src/lxc_runner.rs:20` imports `IngressManager`; `:247` computes `use_firewall`; `:265` constructs `IngressManager::new(&container_name, pid)`; `:296` fails closed when the PID is unavailable in firewall mode.  `network_ingress.rs:8` states inbound traffic is dropped by default, `:39` gates on `ContainerPolicy::allow_local_network`, and `:50-51` builds the container's own INPUT chain.

**State-aware never does.**  `state_aware.rs` contains no reference to `IngressManager`, `network_ingress`, or `Ingress` at all.  `apply_network_policy` installs only the host-side egress FORWARD chain via `NetworkIptablesManager` (`state_aware.rs:351`).

Since `allowLocalNetwork` defaults to false, a container started through the state-aware path with an accepted firewall policy can still listen for host and LAN connections.

## Note on the rollback step

`signal_cleanup.rs` has a `RollbackStep::RemoveIngress`, which might look like the state-aware path handles ingress.  It does not.  State-aware start uses `set_active_network_only` (`state_aware.rs:532`) which yields `SignalRollback::NetworkOnly`, and that plan is only `StopContainer` + `RemoveFirewall` (`signal_cleanup.rs:107-113`).  `RemoveIngress` appears only in the `DestroyContainer` plan (`:114-119`), reached from the one-shot path that does install ingress.  The rollback is correctly paired; the apply side is what is missing.

## Suggested direction

Install inbound rules with `IngressManager` after start on the state-aware path, and roll the start back if it fails -- matching the one-shot fail-closed posture.  Alternatively, reject firewall mode on the state-aware path until it is supported, rather than reporting it as enforced.

---

## LXC start leaves a live unfiltered-inbound container if killed between start and ingress, and a retry is refused

A kill between `container.start()` and `apply_ingress_policy` in `src/backends/lxc/common/src/state_aware.rs` leaves the container running with no inbound policy, and a retry is rejected as already started.

---
---

## Additional state-aware exposure: concurrent exec can enter the start-to-ingress window

The state-aware path has a second way to enter the same unfiltered interval described above: a normal concurrent `exec`, without requiring SIGKILL, OOM, or a process crash.

`start` holds `LifecycleLock`, calls `container.start()`, and only afterwards installs the inbound policy through `apply_ingress_policy` (`src/backends/lxc/common/src/state_aware.rs:963`). `exec` does not acquire that lock (`state_aware.rs:999-1054`); it only checks `is_defined()` and `is_running()` before calling `lxc-attach`. Because LXC transitional live states are treated as running, an exec racing start can attach after the container becomes live but before the INPUT hook is installed.

Failure sequence:

1. `start` installs egress and invokes `lxc-start`.
2. The container becomes live, but ingress installation has not completed.
3. A concurrent `exec` observes the container as running and enters through `lxc-attach`.
4. The workload executes while inbound traffic is still unrestricted.

This makes the exposure reachable during ordinary concurrent API use, not only abnormal termination. It also means the statement inherited from the one-shot analysis—that sandboxed code cannot execute during the startup window—does not hold for state-aware LXC.

### Acceptance criteria

- An exec submitted while start is in progress must either block until start has completed successfully or fail with a deterministic `busy` / `not_started` result.
- No exec may attach before ingress installation has completed and the start transition has released its lifecycle guard.
- Add a regression test with a barrier between `container.start()` and `apply_ingress_policy`; a concurrent exec must not enter the container while the barrier is held.
- Stop and deprovision must likewise be unable to race active execs unless the lifecycle contract explicitly defines and safely implements that behavior.

SohamDas2021:
## 📖 Description

Two things a reader of the LXC backend could not previously find out.

**The guide did not say what the network chains fail to protect against.** It advertised network isolation and listed no limitations at all — the only backend guide in the repo without a `## Limitations` section, and the one with the most caveats. A reader had no way to learn that a process running as root inside the container can flush the chains confining it, that raw sockets skip them entirely, that there is an interval at startup where no policy is in force, or that a long-lived container loses its address when its DHCP lease expires. The guide now states all of these, plus the two unsupported surfaces (proxied egress, state-aware lifecycle).

**Nothing proved the inbound chain actually drops a packet.** The existing inbound test checks that the chain is built correctly and lands in the container's namespace, and says in its own header that it does not check whether anyone can reach the container. Every other LXC test that puts traffic on the wire probes outbound. So a chain that was programmed perfectly and dropped nothing passed the suite. A new test sends a real connection at the container and requires both that it fails and that the chain's drop counters move — a refused connection on its own is indistinguishable from an address that was never reachable.

The test brings its own network interface into the container rather than going over the bridge. That is what makes the result independent of the host's network setup, which is the reason the existing test gave for not asserting reachability in the first place.

Two smaller corrections came out of writing the above: the guide claimed `preservePolicy` retained only the outbound chains when it retains both, and adding a configuration document moved a corpus total the parser tests assert on.

## 🔗 References

Resolves #897

## 🔍 Validation

Run on Ubuntu 24.04 (WSL 2), LXC 5.0.3, iptables 1.8.10 nf_tables:

- `tests/scripts/run_lxc_all_tests.sh` — **28 passed, 0 failed, 0 skipped, 1 disabled**. The disabled entry is the pre-existing `LXC Network` quarantine (needs container DNS + outbound HTTPS), untouched here.
- The new test standalone — inbound dropped on both IPv4 and IPv6, counters `0 -> 4` each.
- `cargo test -p wxc_common` — 1094 passed. `cargo test -p lxc_common` — 358 passed.
- `cargo fmt --all -- --check` clean; `node scripts/versioning/validate-configs.js` validates 363 documents.

The probe logic was also exercised against a synthetic namespace carrying the same chain shape, with the chain deliberately unhooked. It fails there, which is the property that matters: a deny test that still passes once enforcement is removed is not testing anything.

That exercise caught two defects before they shipped. Reading only the NEW-state drop rule, as the Bubblewrap inbound test does, works on IPv4 but reports zero on a host where IPv6 connection tracking is not loaded — the packet reaches the chain and lands on its terminal drop instead, so the test would have reported "the packet never arrived" and failed. And `bash` will not open a connection to a bracketed IPv6 literal. Both are fixed; the IPv6 half of this test would have been wrong without them.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [x] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1211)


docs/lxc-support/lxc-backend.md
```diff
@@ -181,4 +181,30 @@ pty.onExit((e) => console.log('Exit:', e.exitCode));
...
+  not deliberately tear it down, including services the workload itself starts.
+- **Raw sockets bypass egress filtering.** `CAP_NET_RAW` is retained so that an
+  explicit `protocol: "icmp"` allow works. It also permits `AF_PACKET` sockets,
+  which write link-layer frames straight to the interface without traversing the
+  filter chain.
+- **Policy is not in force while the container starts.** The chains are installed
+  after the container has started and its address has settled, so container init
+  and anything it starts run unfiltered in both directions for that interval. The
+  requested command is attached afterwards. A connection opened during that window
+  keeps working once the rules land, because the chains accept established flows.
```
bbonaby, line 203: thought (non-blocking): Wonder if this will be a problem one day.

docs/lxc-support/lxc-backend.md
```diff
@@ -181,4 +181,30 @@ pty.onExit((e) => console.log('Exit:', e.exitCode));
...
+- **A filtered container cannot renew a DHCP lease.** The chains permit loopback,
+  established flows, and DNS, with no carve-out for DHCP. A container that
+  outlives its lease loses its address; one that finishes within the lease period
```
bbonaby, line 206: thought (non-blocking): same with this one I guess, though i doubt folks will have long running processes and run into this issue since dhcp leases usually lasts multiple days.

docs/version-specific-parser-migration-inventory.md
```diff
@@ -65,7 +65,9 @@ when the Seatbelt directional-ingress fixtures
 `seatbelt_net_egress_deny_ingress_allow.json` and
 `seatbelt_net_ingress_allow_loopback_deny.json` joined the corpus. Both parsers
 accept each of those two and build identical runtime models, so they add no
-shared rejection. Both retain no
+shared rejection. It rose by one again when the LXC inbound-traffic fixture
+`lxc_inbound_traffic_deny.json` joined the corpus, which both parsers likewise
+accept identically. Both retain no
```
bbonaby: note: probably can remove this.

bbonaby (dismissed):

docs/lxc-support/lxc-backend.md
SohamDas2021: It already is, just bounded, the window runs from container start until the chains land, so only init and what it spawns are exposed, not the workload. Tracked in ADO

docs/lxc-support/lxc-backend.md
SohamDas2021: Not under stock `lxc-net`, dnsmasq's man page says "if not given, the default lease time is one hour for IPv4", giving T1 renewal ≈ 30 min.

bbonaby (approved):

PR #1216, opened 2026-09-18
[LXC] Subtract an egress 'except' from its own rule so a carve-out cannot shadow a later deny

### LINKED ISSUE #1010: LXC: CIDR except carve-outs shadow later explicit egress rules
> **Security notice:** When reporting BSODs or security issues, **DO NOT** attach memory dumps, logs, or traces to GitHub issues. Instead, send them to secure@microsoft.com referencing the GitHub issue. For application crashes, include a Feedback Hub link if possible (open with Win+F, choose "Share My Feedback" after submission).
### Relevant area(s)
Linux

LXC schema 0.8 lowers each CIDR `except` as an early rule carrying the direction default. Because the chain is first-match-wins, that carve-out shadows later independent rules. Under `egress.default: allow`, `deny 10.0.0.0/8 except 10.10.0.0/16` followed by `deny 10.10.1.0/24` permits the `/24`, violating explicit deny precedence. This was the blocking review comment on PR #983 (threads 2, 24, and 28).

### Steps to reproduce
1. Create a schema 0.8 LXC policy with `egress.default: allow`.
2. Add a deny for `10.0.0.0/8` with `except: [10.10.0.0/16]`.
3. Add a later deny for `10.10.1.0/24`.
4. Inspect the generated chain or attempt traffic to `10.10.1.1`.

### Expected behavior
The exception removes addresses only from its parent rule. Later independent rules remain eligible, and the explicit `/24` deny wins.

### Actual behavior
`lower_rule` emits an ACCEPT for the entire exception before the later DROP. The first ACCEPT wins and silently permits explicitly denied traffic. The documentation also describes the same incorrect fall-through-to-default behavior.

SohamDas2021:
## 📖 Description

On the LXC backend, a CIDR `except` exclusion on a schema-0.8 directional egress rule was emitted as its own iptables rule ahead of its parent CIDR. The chain is first-match-wins, so that carve-out escaped the rule that declared it and answered for every later rule covering the same addresses.

A caller could write:

```json
"egress": {
  "default": "allow",
  "deny": [
    { "to": [{ "cidr": "10.0.0.0/8", "except": ["10.10.0.0/16"] }] },
    { "to": [{ "cidr": "10.10.1.0/24" }] }
  ]
}
```

and the container still reached `10.10.1.1`, despite a rule denying it outright. The exclusion only ever turned a deny into an accept, so the failure was silent and always in the permissive direction.

An exclusion now narrows only the rule that declares it and states no verdict of its own. The exclusion is subtracted from its peer and the covering blocks that remain are programmed, mirroring what the Bubblewrap backend already does in `src/backends/bubblewrap/common/src/network_rules.rs`.

Subtraction splits the surrounding block once per prefix level, so a peer can expand into many blocks. Two ceilings bound that: a peer may expand into at most 256 blocks, and one egress policy may lower into at most 65,536 entries. A policy exceeding either is rejected, the container is torn down, and the script never runs — no partial policy is installed. Both limits are now stated in the 0.8 networking contract, which had documented neither.

**Correction.** An earlier version of this description said that rejection happens before the sandbox is created. It does not. The check runs during firewall setup, which is after the container has started, because the surrounding setup needs the container's network namespace. The observable outcome above is unchanged — nothing runs unenforced — but making it a genuine preflight is worth a follow-up.

`tests/configs/` gained two fixtures, so the corpus inventory pinned in `config_parser.rs` moves from `(368, 344, 14)` to `(370, 346, 14)` — two files, two equivalent accepts, no new divergence and no new rejection.

## 🔗 References

Resolves #1010

## 🔍 Validation

Every number below is from a local run.

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy -p lxc_common -p wxc_common --all-targets -- -D warnings` | exit 0 |
| `cargo test -p lxc_common --all-targets` | 314 + 3 + 28, 0 failed |
| `cargo test --workspace` | 101 suites, 4502 passed, 0 failed, 30 ignored |
| `tests/scripts/run_lxc_all_tests.sh` (WSL2 Ubuntu 24.04, root) | 28 passed, 0 failed, 0 skipped, 1 disabled |

The one disabled entry is the pre-existing quarantined "LXC Network" test. These counts match the baseline taken on `main` before the change.

**The tests were checked against the bug.** Reverting only `network_iptables.rs` and re-running:

- Unit: 312 passed, 2 failed — `an_exclusion_does_not_shadow_a_later_rule_that_names_it` and `a_peer_that_expands_past_the_block_ceiling_is_refused`.
- Integration: `run_lxc_network_ga_egress_test.sh` exits 1 on the new shadow case.

The egress script needed the new fixtures to see this at all. Its two existing `except` cases are both `default: deny` with an `allow` rule, a direction that was already correct, so the suite passed unchanged against a binary built without the fix. The shadow case supplies the missing direction, and the shadow-control case keeps it honest by requiring the same address to stay reachable once the second deny is removed — otherwise a rule set that blocked everything would pass.

## ✅ Checklist

- [x] Linked to an issue

## 📋 Issue Type

- [x] Bug fix
- [ ] Feature
- [ ] Task
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1216)


src/backends/lxc/common/src/network_iptables_ga_egress_spec.rs
```diff
@@ -152,6 +152,26 @@ fn new_connection_action<'a>(
         })
 }
 
+/// The verdict the whole chain gives a packet, including the closing default a
+/// destination no rule names falls through to.
+fn chain_verdict(
```
jsidewhite, line 157: >chain_verdict [](http://example.com/codeflow?start=3&length=13)

`#[test]`?

jsidewhite:
Does this logic really only apply to LXC?  It seems generic to all backends.

SohamDas2021:
> Does this logic really only apply to LXC? It seems generic to all backends.

The bug was LXC-only, Bubblewrap already subtracts, and WSLC/Seatbelt have no destination filtering.

src/backends/lxc/common/src/network_iptables_ga_egress_spec.rs
SohamDas2021: this is a test helper, not a test

sylvesterkaczmarek (commented):
Different-family except entries are still silently filtered out here. An IPv4 peer carrying an IPv6 exclusion, or vice versa, therefore applies the full parent rule instead of rejecting the malformed peer; for an allow rule under a deny default that widens access. Could this return an error for a family mismatch rather than dropping the exclusion, matching the fail-closed handling for an invalid prefix?

SohamDas2021:
> Different-family except entries are still silently filtered out here. An IPv4 peer carrying an IPv6 exclusion, or vice versa, therefore applies the full parent rule instead of rejecting the malformed peer; for an allow rule under a deny default that widens access. Could this return an error for a family mismatch rather than dropping the exclusion, matching the fail-closed handling for an invalid prefix?

Thanks for the catch, I have fixed it.

On erroring for a genuine cross-family mismatch: I kept the drop, since an IPv6 range can't lie inside an IPv4 peer and Bubblewrap deliberately discards these too.

sylvesterkaczmarek (changes requested):
Rechecked current 8d719929. The different-family except case I raised is still explicitly fail-open: peer_destinations() filters exclusions to the peer family, and the regression an_exclusion_that_cannot_overlap_its_peer_is_dropped_rather_than_applied now pins 10.0.0.0/8 except 2001:db8::/32 as an unrestricted allow of the IPv4 parent. For a security policy, a malformed/mismatched exclusion should not silently widen an allow rule. Please reject an except entry whose programmed address family does not match its peer, while still allowing equivalent IPv4-mapped notation that resolves to the same programmed family, and cover that with a fail-closed regression.

SohamDas2021:
> Rechecked current [8d71992](https://github.com/microsoft/mxc/commit/8d7199292d8a9f58f4f21e3d5883b8a2c2454a9f). The different-family except case I raised is still explicitly fail-open: peer_destinations() filters exclusions to the peer family, and the regression an_exclusion_that_cannot_overlap_its_peer_is_dropped_rather_than_applied now pins 10.0.0.0/8 except 2001:db8::/32 as an unrestricted allow of the IPv4 parent. For a security policy, a malformed/mismatched exclusion should not silently widen an allow rule. Please reject an except entry whose programmed address family does not match its peer, while still allowing equivalent IPv4-mapped notation that resolves to the same programmed family, and cover that with a fail-closed regression.

Fixed.

jsidewhite (approved):
:shipit:

PR #1219, opened 2026-09-18
[API Surface] State-aware: Add 'phase' APIs + remove 'phase' ops from schema

### LINKED ISSUE #1218: [API Surface] State-aware: Remove 'phase' operations from json schema
### Description of the task

State aware operations don't belong in json schema.  They should be APIs.

### Additional context

_No response_

jsidewhite:
1. State-aware is now functions (not "pass a config json")
2. Removed 'phase' and 'provision' (duplicated schema bits) from the 0.9.0-alpha json schema

**Rust mxc_sdk example**
```rust
let provisioned = sandbox::provision(
    r#"{"version":"0.9.0-alpha","containment":"isolation_session",
        "network":{"egress":{"default":"allow"},
          "ingress":{"default":"allow","hostLoopback":"allow"}}}"#,
    true,  // experimental
)?;
```

Closes #1218 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/mxc/pull/1219)


docs/isolation-session/state-aware-rust.md
```diff
@@ -293,7 +292,7 @@ meaning for this backend.
 | `containerId` | accepted, no effect | rejected | rejected | rejected | rejected | rejected |
 | `process.commandLine` | **honored** | rejected | rejected | **honored** | rejected | rejected |
 | `process.{cwd,env,timeout}` | **honored** | rejected | rejected | **honored** | rejected | rejected |
-| `experimental.isolation_session.provision.appId` | rejected | **honored** | n/a | n/a | n/a | n/a |
+| `experimental.isolation_session.appId` | rejected | **honored** | n/a | n/a | n/a | n/a |
 | `experimental.isolation_session.<another phase>.*` | rejected | rejected | rejected | rejected | rejected | rejected |
```
adpa-ms: Can get rid of this line since phase is dead

schemas/dev/mxc-config.schema.0.9.0-dev.json
```diff
@@ -300,27 +300,10 @@
...
-      },
-      "type": "object"
-    },
-    "IsolationSessionProvisionPhase": {
-      "description": "Provision-phase IsolationSession configuration (state-aware lifecycle).\n\nThe only phase that takes a per-phase payload, so it is its own type rather than a shared one: a shared type would advertise its fields on every phase in the generated schema. The domain configs and the SDK types are already split per phase; this keeps the wire model aligned with them.",
+      "description": "IsolationSession backend configuration.",
       "properties": {
         "appId": {
-          "description": "Optional identifier for the calling application.\n\n**A packaged application must supply its Package Family Name in the form `PFN:<packageFamilyName>`** (for example `PFN:Contoso.App_8wekyb3d8bbwe`). An unpackaged application may pass any string. Carried inside the `sandboxId` so later lifecycle phases can recover it without the caller re-supplying it.",
+          "description": "Lifecycle provision application identifier.",
```
adpa-ms: Might not want to lose *all* of the information from the previous comment here. The PFN format is important for the underlying OS API call.

sdk/node/src/generated/wire.ts
```diff
@@ -127,26 +127,11 @@ export interface Filesystem {
...
- * Provision-phase IsolationSession configuration (state-aware lifecycle).
- *
- * The only phase that takes a per-phase payload, so it is its own type rather than a shared one: a shared type would advertise its fields on every phase in the generated schema. The domain configs and the SDK types are already split per phase; this keeps the wire model aligned with them.
- */
-export interface IsolationSessionProvisionPhase {
-  /**
-   * Optional identifier for the calling application.
-   *
-   * **A packaged application must supply its Package Family Name in the form `PFN:<packageFamilyName>`** (for example `PFN:Contoso.App_8wekyb3d8bbwe`). An unpackaged application may pass any string. Carried inside the `sandboxId` so later lifecycle phases can recover it without the caller re-supplying it.
+   * Lifecycle provision application identifier.
```
adpa-ms, line 134: Similar comment to a previous one - please keep some of the context from the older comment about the format if passing in the PFN.

tests/configs/isolation_session_state_aware_deprovision.json
adpa-ms, line 1: Trying to understand, have I lost a bunch of my tests?

src/core/wxc_common/src/config_parser.rs
```diff
@@ -443,80 +418,42 @@ where
...
+    phase: Phase,
+    sandbox_id: Option<&str>,
+    logger: &mut Logger,
+) -> Result<ParsedStateAwareRequest, ParseError> {
+    let discriminator: RequestDiscriminator<'_> = config_deserialize::from_str(json)
+        .map_err(|error| ParseError::StateAware(MxcError::malformed_request(error.to_string())))?;
+    reject_legacy_telemetry_raw(discriminator.experimental.map(RawValue::get))
+        .map_err(|error| ParseError::StateAware(MxcError::malformed_request(error.to_string())))?;
+    let request: mxc_config_contract::dev::OneShotRequest =
+        deserialize_development_root(json, "lifecycle operation", true)?;
```
adpa-ms: [AgentAssisted] `development_network_migration` keys its backend-specific guidance on the contract labels `"IsolationSession provision"` and `"WSLC provision"`, but `deserialize_development_root` is now only ever called with `"lifecycle operation"` or `"one-shot"`, so 6 of its 10 arms are unreachable.

The effect is a wrong instruction: an `isolation_session` config using `network.allowedHosts` now gets `use network.egress.allow/deny CIDR rules`, which IsolationSession cannot satisfy — it accepts only the all-allow posture, which is what the previous message named.

`exact_development_network_migration_guidance_is_contract_aware` was updated to assert the message no longer contains `IsolationSession requires`, so the test now pins this under a name claiming the opposite.

Suggest keying the lookup on the selected containment so the backend-specific arms stay reachable.

src/core/wxc_common/src/config_contract_adapters/dev/state_aware.rs
```diff
@@ -1,310 +1,226 @@
...
+    };
+    Ok(StateAwareOperation::Provision(provision))
 }
 
-pub(super) fn deprovision_into_input(
-    request: contract::DeprovisionRequest,
+pub(super) fn operation_into_input(
+    request: contract::OneShotRequest,
+    phase: Phase,
+    sandbox_id: Option<&str>,
```
adpa-ms, line 207: [AgentAssisted] This function now owns all per-operation validation that the deleted per-phase contract roots enforced structurally, including IsolationSession provision. The `#[cfg(test)] #[path = "state_aware_tests/mod.rs"] mod tests;` declaration was removed here, but the 7 files (16 `#[test]`) remain on disk and are compiled by nothing; `one_shot.rs` keeps its equivalent declaration.

They also cannot just be re-linked — `state_aware_tests/common.rs` uses `adapt_request` and `AdaptedWireRequest::StateAware`, both removed in this change.

Either port them to `operation_into_input(request, phase, sandbox_id)` and restore the declaration, or delete them. As it stands the new validation has no unit coverage beside it.

src/core/wxc/src/main.rs
```diff
@@ -1208,11 +1241,45 @@ fn main() {
...
+                        ParseError::Decode(error)
+                        | ParseError::Version(error)
+                        | ParseError::OneShot(error)
+                        | ParseError::OneShotMalformed(error) => {
+                            MxcError::malformed_request(error.to_string())
+                        }
+                    };
+                    log_state_aware_dispatch_error(&mut logger, &error);
+                    print_error_envelope(&error);
+                    process::exit(1);
```
adpa-ms, line 1268: [AgentAssisted] This branch prints the diagnostic and envelope then exits without calling `log_request_parse_rejection`, so lifecycle parse rejections no longer emit the `mxc.ConfigRejected` audit record — `log_state_aware_dispatch_error` only writes a diagnostic line. One-shot parse failures still emit it via the `Err` arm below, and state-aware failures did too before this change, since both went through that path.

Suggest emitting the structured rejection before printing the envelope.

src/core/wxc_common/src/config_contract_adapters/dev/state_aware.rs
```diff
@@ -1,310 +1,226 @@
...
+        }
+        Phase::Deprovision => {
+            reject_non_exec_process(&common, phase)?;
+            reject_non_exec_policy(&common, phase)?;
+            StateAwareOperation::Deprovision {
+                sandbox_id: require_sandbox_id(phase, sandbox_id)?,
+            }
+        }
+    };
+    StateAwareInput::new(common, operation)
```
adpa-ms, line 255: [AgentAssisted] `take_experimental` runs only in `provision_operation`, so on exec/start/stop/deprovision `common.experimental` survives into `StateAwareInput::new` and is rejected as `State-aware common input must not contain operation or one-shot field(s): experimental`. That reads as though `experimental` is one-shot-only, but it is accepted on provision.

`"experimental":{}` was valid on all four of these operations before this change — the cases covering it were dropped from `public_preflight_duplicate_experimental_fields_follow_the_selected_request_kind` and from the `policy_identity` phase loop rather than re-asserted. If rejecting it is intended, `reject_non_exec_policy` is the natural place, with a message matching the others.

sdk/dotnet/Microsoft.Mxc.Sdk/MxcLifecycle.cs
```diff
@@ -681,25 +691,30 @@ private static void SetBackendConfig(
...
+    private static JsonObject? RunEnvelopePhase(
+        JsonObject envelope,
+        int operation,
+        SandboxId? sandboxId,
+        bool dryRun)
     {
         var requestBuf = ToNullTerminatedUtf8(envelope.ToJsonString());
+        var sandboxIdBuf = sandboxId is null
+            ? null
+            : ToNullTerminatedUtf8(sandboxId.Value.Value);
```
adpa-ms, line 706: [AgentAssisted] Sandbox IDs now cross the ABI as raw NUL-terminated strings and `ToNullTerminatedUtf8` does not reject an embedded NUL, so `CStr::from_ptr` truncates at it — `iso:<valid>\0suffix` addresses `iso:<valid>`. Previously the ID travelled inside the JSON envelope, where `\u0000` survived to reach validation and be rejected.

Low likelihood while IDs only originate from `ProvisionSandbox`, but the failure is a silent wrong target. Suggest rejecting embedded NUL in `SandboxId` or in the marshalling helper.

tests/scripts/run_isolation_session_state_aware_tests.ps1
```diff
@@ -207,12 +206,18 @@ function Invoke-StateAware {
...
+    if ($Operation -ne 'provision' -and -not $SandboxId) {
+        throw "Invoke-StateAware requires -SandboxId for the $Operation operation"
+    }
 
     $b64 = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($json))
 
     $argList = @()
     if ($Experimental.IsPresent) { $argList += '--experimental' }
     if ($DryRun.IsPresent) { $argList += '--dry-run' }
+    $argList += @('--operation', $Operation)
```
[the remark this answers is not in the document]
adpa-ms: [AgentAssisted] Confirming from the other direction: the equivalent assertions in `run_wslc_state_aware_tests.ps1` were migrated in this change (6 `unknown field` assertions at the base, 0 now) and `run_windows_sandbox_state_aware_tests.ps1` has none — so this is the only harness still asserting the removed structural diagnostics. At the current head it has 9 of them (6 `filesystem`, 2 `network`, 1 `ui`) alongside the `malformed_request` code expectations.

Worth noting this suite is host-gated, so CI cannot catch the failure.

docs/isolation-session/state-aware-rust.md
adpa-ms: [AgentAssisted] Two more statements in this file are now inaccurate for the same reason: the bullet saying `process` on non-exec phases is "structurally rejected" (it is now a semantic check in the operation adapter), and the error-table row saying "fields excluded by an exact request root fail earlier as `malformed_request`" (there is one shared root now).

Provision `filesystem` and `network` also reach IsolationSession policy validation and surface `policy_validation`, not `malformed_request`.

jsidewhite:
Gudge will be taking over this Issue.

